PR #6, opened 2025-05-20
Draft: attempting to add inc_mw_com tests

MarkSchmitt:
I wanted to try and include the bigdata test from inc_mw_com .. well, I think I'm missing a lot of dependencies, but still wanted to document the work done.

Maybe we can have someone with proper access to the dependencies try to extract the depenencies necessary and make the bigdata example (and possibly others?) compile and run on Linux? That'd be a huge help to get this rolling!

Please don't actually spend much time on reading the changes .. it's all just renaming, commenting out or plain bad guessing ...!


MarkSchmitt:
@darkwisebear  as discussed today, I tried, but there're some things like the @amp depencency missing that I can't resolve.

castler:
Hi @MarkSchmitt ,

we still have the BigData test to open source. This depends on an internal framework (SCTF) which we are right now removing dependencies to open source it.

In the meantime, we will try to maybe only open source the big data application - if this is of any help.

MarkSchmitt:
> Hi @MarkSchmitt ,
> 
> we still have the BigData test to open source. This depends on an internal framework (SCTF) which we are right now removing dependencies to open source it.
> 
> In the meantime, we will try to maybe only open source the big data application - if this is of any help.

that would be very helpful in itself, I didn't plan on using the test framework at this point!

LakmalBMW:
> > Hi @MarkSchmitt ,
> > we still have the BigData test to open source. This depends on an internal framework (SCTF) which we are right now removing dependencies to open source it.
> > In the meantime, we will try to maybe only open source the big data application - if this is of any help.
> 
> that would be very helpful in itself, I didn't plan on using the test framework at this point!

We started on this, will be tracked here. https://github.com/eclipse-score/communication/issues/19

PR #7, opened 2025-05-20
fix broken links of diagrams

phongt:
What changing:
- Add the `png` which convert from `uxf` diagrams  (java -jar umlet.jar  -action=convert -format=png -filename=*.uxf)
- Update the src img link of README.md


LittleHuba:
Hi @phongt,
thanks for your contribution.

Unfortunately, we can't merge any binary data into the repository, since this would lead to a huge git history very fast. Instead the diagrams should be ported, following: https://eclipse-score.github.io/score/main/contribute/documentation/docs-as-code.html#diagrams

Your contribution in that direction would be more than welcome!

phongt:
@LittleHuba thanks for your review. I will close this as now.

PR #33, opened 2025-07-23
Bazel formatting

PiotrKorkus:
Add bazel targets for starlark language sever and formatting. 
If save on exit doesn't work, try restarting VSC.

Closes: https://github.com/eclipse-score/communication/issues/37


FScholPer:
@PiotrKorkus Can you please add an issue with a short description about the change

LittleHuba:
Sorry that this PR takes so long to integrate. The problem is tooling on our side that messes with the formatting. So only adding formatting rules here would not fix the problem.
We are working on it though!

PR #39, opened 2025-07-29
Test that Rust and C++ ipc_bridge example work with each other

### LINKED ISSUE #38: Bug: Example C++ application crash when using --help
### Description

`ipc_bridge_cpp` crashes when `--help` is used. This is a valid use case and should yield a return code of 0.

### Error Occurrence  Rate

Reproducible

### How to reproduce

```shell
bazel build //...
./bazel-bin/score/mw/com/example/ipc_bridge/ipc_bridge_cpp --help
```

### Supporting Information

_No response_

### Classification

minor

### Affected Version

0.5

### Expected Closure Version

0.5

### Category

- [ ] Safety Related
- [ ] Security Related

lurtz:
CI should check that C++ and Rust code can communicate via mw::com. To simplify the tests the CLI interface of the Rust code has been more aligned with the C++ interface.

Fixes #38

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,9 +85,6 @@ Camera App ──► [Video Frame] ──┬──► Display App
 > This repository offers a [DevContainer](https://containers.dev/).
 > For setting this up and enabling code completion read [eclipse-score/devcontainer/README.md#inside-the-container](https://github.com/eclipse-score/devcontainer/blob/main/README.md#inside-the-container).
 
->**Note**:
-> If you are using Docker on Windows **without `WSL2`** in between, you have to select the alternative container `eclipse-s-core-docker-on-windows`.
-
 ### Building the Project
 
 ```bash
--- a/score/mw/com/example/ipc_bridge/BUILD
+++ b/score/mw/com/example/ipc_bridge/BUILD
@@ -95,3 +95,13 @@ cc_library(
         "//score/mw/com/impl/rust:bridge_macros",
     ],
 )
+
+# Rust code dumps core when run in parallel with unit tests
+sh_test(
+    name = "integration_test_ipc_bridge",
+    srcs = ["test_ipc_bridge.sh"],
+    data = [
+        ":ipc_bridge_cpp",
+        ":ipc_bridge_rs",
+    ],
+)
--- a/score/mw/com/example/ipc_bridge/ipc_bridge.rs
+++ b/score/mw/com/example/ipc_bridge/ipc_bridge.rs
@@ -16,7 +16,7 @@ use std::thread::sleep;
 use std::time::Duration;
 
 use clap::{Parser, ValueEnum};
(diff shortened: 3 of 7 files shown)


score/mw/com/impl/bindings/lola/service_discovery/client/BUILD
```diff
@@ -61,6 +61,7 @@ cc_gtest_unit_test(
         "//score/mw/com/impl/bindings/lola/service_discovery/test:service_discovery_client_test_resources",
         "//score/mw/com/impl/configuration/test:configuration_store",
     ],
+    tags = ["exclusive"],
```
darkwisebear, line 64: Why do you need this tag? Requiring the "exclusive" tag feels as if multiple tests use similar resources (in this case services). If we can avoid that, we might be able to leverage parallel testing again. Or is there another reason?

score/mw/com/impl/bindings/lola/test/proxy_component_test.cpp
```diff
@@ -64,9 +64,25 @@ using namespace ::testing;
 const filesystem::Path kTmpPath{"/tmp_discovery/mw_com_lola/service_discovery"};
 constexpr auto kServiceInstanceUsageMarkerFile =
     "/tmp_discovery/mw_com_lola/partial_restart/usage-0000000000052719-00016";
+const std::vector<char const *> files_for_cleanup {
```
darkwisebear, line 67: Such cleanup shouldn't be necessary unless a test explicitly "kills" the sender. In that case, the cleanup should be test specific.

score/mw/com/impl/bindings/lola/service_discovery/client/BUILD
lurtz: You can also sometimes see it fail in CI: https://github.com/eclipse-score/communication/actions/runs/16971034338/job/48107771406#step:6:225

But due to me rebasing and cleaning up commit, I cannot tell anymore to which git revision this run belongs to.

score/mw/com/impl/bindings/lola/test/proxy_component_test.cpp
lurtz: should this have been a response to https://github.com/eclipse-score/communication/pull/39#discussion_r2287460836 ?

FYI @LittleHuba is absent for some weeks.

darkwisebear: Maybe... I was referring to the instability due to parallelism, which is visible here, and also with the "exclusive" tag. All the tests run stable in our environment (RunsOnMyMachine(tm)).

LittleHuba (commented):
This test should be shifted to an integration test to simplify the logic and use our infrastructure.
@lurtz would you mind?

PR #63, opened 2025-09-10
LoLa benchmark implementation

piotrchodorowski:
Implementation of benchmark for LoLa that is benchmark-publish-subscribe.

### DIFF
--- a/score/mw/com/benchmark/benchmark.cpp
+++ b/score/mw/com/benchmark/benchmark.cpp
@@ -27,17 +27,7 @@ using namespace std::chrono_literals;
 using namespace score::mw::com;
 using namespace std::chrono;
 
-std::mutex cout_mutex{};
 score::cpp::latch benchmark_ab_start_point{3}, benchmark_ab_finish_point{3}, init_ab_sync_point{3}, deinit_ab_sync_point{2};
-score::cpp::latch benchmark_multi_start_point{1 + kThreadsMultiTotal}, benchmark_multi_finish_point{1 + kThreadsMultiTotal}, init_multi_sync_point{kThreadsMultiTotal}, deinit_multi_sync_point{kThreadsMultiTotal};
-const auto instance_specifier_skeleton_a_optional = InstanceSpecifier::Create("benchmark/SkeletonA");
-const auto instance_specifier_skeleton_b_optional = InstanceSpecifier::Create("benchmark/SkeletonB");
-
-score::cpp::optional<std::reference_wrapper<impl::ProxyEvent<DummyBenchmarkData>>> GetBenchmarkDataProxyEvent(
-    BenchmarkProxy& proxy)
-{
-    return proxy.dummy_benchmark_data_;
-}
 
 void SetupThread(int cpu)
 {
@@ -64,7 +54,7 @@ void SetupThread(int cpu)
     }
 }
 
-void Transmitter(int cpu, bool starter, impl::InstanceSpecifier skeleton_instance_specifier, impl::InstanceSpecifier proxy_instance_specifier)
+void Transceiver(int cpu, bool starter, impl::InstanceSpecifier skeleton_instance_specifier, impl::InstanceSpecifier proxy_instance_specifier)
 {
     SetupThread(cpu);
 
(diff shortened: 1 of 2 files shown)


crimson11 (commented):
1st: I really appreciate, that you are coming up with this effort to develop a basic benchmark test! We internally did have some iterations for benchmarking already - but it is often hard to decide, WHAT to exactly measure! 
We are typically differentiating between MACRO and MICRO benchmarks.
I guess, what you are trying to do here is a mix-up or rather some MACRO benchmark.

### Micro benchmark
So in case of MICRO benchmarks, we typically try to measure the runtime of a specific API. Sometimes it isn't even a public API, but some internal API/logic/algorithm. So executing a  rather small amount of code frequently and then averaging runtime.

### Macro benchmark
In a MACRO benchmark, we have a more complex application setup, which does a lot of `mw::com` communication for either a given amount of time or a given number of communication calls and then we look at throughput and CPU usage at the end of the test.

### Latency notion

So - your benchmark calculates a "Latency" according to the output.
From the code it is apparently the measurement between `benchmark_ab_start_point`/`benchmark_ab_finish_point,`
So it is the loop of the "inner sequence":
- proxy checks for a new sample
- skeleton allocates a new sample
- skeleton sends the new sample

So from this I guess you mean the "latency" in `score::mw::com` between a communication partner allocating/sending a sample until it is "accessible" by a subscriber?

Imho it is very hard to argue about how to measure "correctly". Let me explain:
Technically, the sample is "visible" to the consumer, after the provider has done the following steps:

1. Allocating memory for the sample
2. Writing the sample data into the allocated memory
3. Having successfully called `Send()`

After the last step, the sample data is residing in shared-memory (thus generally "visble" to the subscriber) AND the "meta-info" in shared-memory is updated, so that the subscriber is able to find the new sample within the shared-memory.

So at least the time those 3 steps take are part of the "latency".

Btw.: step (2) you aren't measuring at all in your test. Because you're not feeding data to the sample, which means, that it is either default constructed or contains the data from an older sample, which had the same memory location before!
But imho it is crucial to really "fill" the sample with valid data. Because these are memory writes, which take time and MUST be also measured! Because if you would compare the `score::mw::com` implementation (with shared-mem/zero-copy) with some other binding/middleware, which isn't zero-copy you would see BIG differences in latency, when the sample-data is large! So it only makes sense to measure/compare latency in conjunction with a specific sample data type/size and where you really write this data!

Since the consumer side call to `GetNewSamples()` is **surely** part of the "latency" (without this call "succeeding" no access to the sample is possible) I'm fine with also adding its runtime to the latency measurement. If I wanted to be nitpicky, the measurement/timer would need to get stopped already in the callback/lambda you hand over to `GetNewSamples()` because this is the exact point in time, where the consumer gets the access to the sample ;)

Now the most interessting part of the latency thing: When does the consumer side (calling `GetNewSamples()`) know, that there is a new sample and the call makes sense?
Answer: It only knows it in the so-called "event-driven" reception mode, where the consumer did register a so called `ReceiveHandler` at the event-instance before, which will fire as soon as the provider side has called `Send()`.
So in "real-world" scenarios, where provider/skeleton and consumer/proxy reside in different processes, you can measure "latency" in a meaningful way only in this "event-driven" mode and then there the whole runtime of this notification path (which in `score::mw::com` is done through a message-passing side-channel) becomes part of the latency! And here for sure context-switches are involved also!

So your decision, to have the consumer/proxy side `GetNewSamples()` call directly next to the `Send()` call by the provider/skeleton makes perfect sense in the way, that you don't need the additional `EventReceiveHandler` path ... which is good to estimate "latency" for users not using the "event-driven" mode with registered `EventReceiveHanlders` but just "poll" for new samples (which is a very common use case!).

### Further points to consider

- if you look at your "inner loop" in line 129ff, there is a risk, that you measure unintended retry-loops in failure cases. I would get rid of this! E.g. an error in either of these 3 steps is unexpected in your benchmark setup and should lead to termination of the test.
- having provider and consumer in the same process, while having some benefits in latency measurement (see above), on the other hand might be "unrealistic". In a real world setup with provider/consumer in different processes, maybe the risk that after the context switch to the provider took place, the caches are invalidated and therefore the access to the sample in the consumer takes longer/higher latency because the data from the sample in shared-memory must first be loaded into the cache again ...
- also on the consumer side I would then add "code", which really accesses/reads the sample to take such things into account.
- for having really tight/good measurement without OS overhead (using CPU specific performance counters/timers) a framework like google-benchmark is really helpful.

### Summary

Again - kudos to you, to get in touch with `score::mw::com` not only superficially ... but going deeper!

Georg Dadunashvili at our side is our benchmarking dude. I will get in touch with him after his vacation and inform him, that there is seemingly interest in the community about benchmarks/performance data! So we will see, how we can contribute here stuff, which we alread have.

PR #69, opened 2025-10-11
Tracing: switch BUILD deps to generic_trace_library interface targets

Abdelaziz177:
**Summary**

Update score/mw/com/impl/tracing/BUILD to depend on the public interface targets of generic_trace_library instead of the umbrella generic_trace_library target. This aligns the tracing code with the current layout in score-baselibs where public headers live under …/generic_trace_library/interface_types/*.

**Why**

While building the IPC example for a QNX image, the build broke in tracing with missing headers like:

fatal error: score/analysis/tracing/library/generic_trace_api/error_code/error_code.h: No such file or directory

The correct include path in latest baselibs is under …/generic_trace_library/interface_types/.... Depending on the umbrella target pulled in the wrong structure and caused those “file not found” errors

**Impact**

Fixes missing-header errors in tracing (and thus unblocks building the IPC example inside the QNX image).

Narrows linkage to interface-only targets (cleaner dependency graph).

**Testing**

Local build of tracing targets succeeded:

bazel build //score/mw/com/impl/tracing:all


LittleHuba:
Closing this PR since no feedback was received for two weeks. If you want to continue work on this PR, please reopen the PR or create a new one. Thank you!

PR #99, opened 2025-11-19
Fixed prefix of command line argument to use double dashes

### LINKED ISSUE #83: Service instance manifest command line argument needs two `-` prefixes to be compliant with the convention
### What

Currently the lola runtime factory function takes two arguments argc and argv

mw::com::Runtime::InitializeRuntime(argc, argv);
which expects a commandline flag "-service_instance_manifest" as can be seen [here](https://cc-github.bmwgroup.net/swh/safe-posix-platform/blob/bed79b686307e2a1b38fc9ae47341d8c6511e094/platform/aas/mw/com/impl/runtime.cpp#L34).
This is a wrong. It is customary in UNIX like systems to have two dashes before flags that are longer than a single letter.

### How

Since `-service_instance_manifest` is part of the public API, the best thing would be to introduce the proper command line flag with two dashes `--service_instance_manifest` as an additional option. And deprecate the single dash option.

A deprecation warning shall be issued when the deprecated flag is used (including a timeline 2 months for the final removal).
A ticket shall be created to clean up this deprecated flag.

### Estimates for realization

- Low effort
- Users will have to migrate to new flag. Migration path is provided.

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

mikehaller:
Fixes #83 

Added the `--service_instance_manifest` command line argument as an alias to the now deprecated `-service_instance_manifest`. The latter will issue a warning log message when used.

Migration path:
- Use `--service_instance_manifest` (with double dashes as prefix)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,22 +3,5 @@
     "image": "ghcr.io/eclipse-score/devcontainer:latest",
     "runArgs": [
         "--privileged"
-    ],
-    // Initialize bazel cache on host
-    "initializeCommand": "mkdir -p ${localEnv:HOME}/.cache/bazel",
-
-    // Mount ca-certificates from host
-    "mounts": [
-        {
-            "source": "/usr/local/share/ca-certificates/",
-            "target": "/usr/local/share/ca-certificates/",
-            "type": "bind"
-        }
-    ],
-    // Add your personal customizations
-    "onCreateCommand": {
-        "update certificates": "sudo apt update && sudo apt install -y --no-install-recommends ca-certificates-java openjdk-17-jre-headless && sudo update-ca-certificates",
-        "bazel use system trust store": "echo 'startup --host_jvm_args=-Djavax.net.ssl.trustStore=/etc/ssl/certs/java/cacerts --host_jvm_args=-Djavax.net.ssl.trustStorePassword=changeit' | sudo tee --append /etc/bazel.bazelrc",
-        "set wsl proxy": "echo 'startup --host_jvm_args=-Dhttps.proxyHost=wsl-proxy --host_jvm_args=-Dhttps.proxyPort=3129' | sudo tee --append /etc/bazel.bazelrc"
-    }
+    ]
 }


score/mw/com/runtime_configuration.cpp
```diff
@@ -79,6 +80,21 @@ std::optional<filesystem::Path> RuntimeConfiguration::ParseConfigurationPath(
...
         }
+        if (command_line_argument_key == kDeprecatedConfigurationPathCommandLineKey)
+        {
+            score::mw::log::LogWarn("lola")
+                    << "Command line argument" << kDeprecatedConfigurationPathCommandLineKey 
+                    << "is deprecated, please use" << kConfigurationPathCommandLineKey;
+            const auto index_of_configuration_path = arg_idx + 1U;
+            if (index_of_configuration_path >= num_args)
+            {
+                score::mw::log::LogFatal("lola")
```
castler: I think the positive branch of the `if-statement` is not covered.

I think we need an additional test similar to: `TerminatesIfCommandLineArgsContainPathKeyButNoPath`

castler (changes requested):
Thank you for your contribution.

I would have a nit, and a request for an additonal test case.

score/mw/com/runtime_configuration.cpp
mikehaller: Added additional test case to cover that branch and additional asserts for checking the termination reason is in log output.

score/mw/com/runtime_configuration_test.cpp
```diff
@@ -11,9 +11,12 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 #include "score/mw/com/runtime_configuration.h"
-
 #include "score/memory/string_literal.h"
 
+#include "score/mw/log/logging.h"
+#include "score/mw/log/recorder_mock.h"
+#include "score/mw/log/slot_handle.h"
```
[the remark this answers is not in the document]
mikehaller: Thanks for spotting that.
Removed all unnecessary includes, there were a couple more which were unnecessary.

Rahul-Sutariya:
**PR under integration — no further changes or pushes allowed.**

mikehaller:
@Rahul-Sutariya 

> **PR under integration — no further changes or pushes allowed.**

There were still changes requested by @castler

Starting integration before all code reviews are done is not allowed.

castler (commented):
Changes look good thanks.

maybe in future you could try to add commit reviews via ammends to the original commits - to not bloat git history.

Sorry, I told Rahul yesterday to import the stuff already.

PR #113, opened 2025-12-05
Defer EventReceiveHandler registration until service provider is available

### LINKED ISSUE #89: Register EventReceiveHandler only when service available
### What

Currently the registration of a EventReceiveHandler is done within NotSubscribedState::SubscribeEvent() independent of the state in state_machine_.provider_service_instance_is_available_.

I.e. even if provider_service_instance_is_available_ is FALSE the call to state_machine_.event_receive_handler_manager_.Register() is done!

But this Register() will finally lead to a register-message-sending via message-passing. But if the service instance is NOT available, this registration will most likely fail anyhow (provider not there, not listening on message-passing receiver)! This failing will only be detected by an error-log-message! There is no result/return-code in state_machine_.event_receive_handler_manager_.Register()!

So the call to register shall done only in case provider_service_instance_is_available_ == true, which minimizes the risk to do a message-passing send without success anyways.

----------------
This change/fix is an optimization/nice-to-have. In case the Register() call is done while service isn't reachable (and therefore the message-passing send is lost), as soon as the service gets available again a re-register will done and this should then succeed.

### How

_No response_

### Estimates for realization

- Effort medium
- Risk that the scope may extend if this requires adaptations in other places
- No impact on users expected

### Category

- [x] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

jorgecasal:
This pull request implements issue #89. It optimizes the event subscription workflow by deferring handler registration until the service provider becomes available, preventing unnecessary failed message-passing attempts that would only be detected via error logs.

When a client subscribes to an event while the service provider is unavailable, the handler is now kept pending and automatically registered when the provider becomes available via `ReOfferEvent()`. This eliminates wasted message-passing sends that would fail anyway since the provider is not listening.

resolves #89

### DIFF
--- a/score/mw/com/impl/bindings/lola/subscription_state_machine_methods_test.cpp
+++ b/score/mw/com/impl/bindings/lola/subscription_state_machine_methods_test.cpp
@@ -254,11 +254,9 @@ TEST_F(StateMachineMethodsNotSubscribedStateFixture, SubscribeWithUnavailablePro
                 RegisterEventNotification(QualityType::kASIL_QM, element_fq_id_, ::testing::_, ::testing::_))
         .Times(0);
 
-    // Make provider unavailable
     EnterSubscribed(max_num_slots_);
     state_machine_.StopOfferEvent();
-    state_machine_.UnsubscribeEvent();
-    EXPECT_EQ(state_machine_.GetCurrentState(), SubscriptionStateMachineState::NOT_SUBSCRIBED_STATE);
+    EXPECT_EQ(state_machine_.GetCurrentState(), SubscriptionStateMachineState::SUBSCRIPTION_PENDING_STATE);
 
     // When we set a handler and subscribe with unavailable provider
     state_machine_.SetReceiveHandler(CreateMockScopedEventReceiveHandler(receive_handler));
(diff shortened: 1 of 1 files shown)


score/mw/com/impl/bindings/lola/subscription_not_subscribed_states.cpp
```diff
@@ -81,9 +81,14 @@ ResultBlank NotSubscribedState::SubscribeEvent(const std::size_t max_sample_coun
...
     {
-        state_machine_.event_receive_handler_manager_.Register(
-            std::move(state_machine_.event_receiver_handler_.value()));
-        state_machine_.event_receiver_handler_.reset();
+        // Defer handler registration until provider is available to avoid failed registration attempts.
+        // When provider is unavailable, keep handler pending for deferred registration when provider re-appears.
+        if (state_machine_.provider_service_instance_is_available_)
+        {
+            state_machine_.event_receive_handler_manager_.Register(
+                std::move(state_machine_.event_receiver_handler_.value()));
```
crimson11: You could eventually do some "cleanup", while doing this change:
1. In the past the "handler" stored in `event_receiver_handler_` was a move-only callback ( `amp::callback` ). Since it has been changed in the meantime into a `weak_ptr<scoped_function<..>>` ... the `std::move` here is unnecessary and just "bloats" the code imho. So I would remove the std::move!
2. I just realized, that `event_receiver_handler_` is still an `amp::optional` ... we are step-by-step moving away from `amp::optional` to `std::optional` (since we are C++17 now).
So could you also in this PR (separate commit) change `event_receiver_handler_` to `std::optional` ?
If there comes up any issues with the change ... then leave it.

score/mw/com/impl/bindings/lola/subscription_state_machine_methods_test.cpp
```diff
@@ -245,6 +245,78 @@ TEST_F(StateMachineMethodsNotSubscribedStateFixture, CallingGetTransactionLogInd
...
+    state_machine_.UnsubscribeEvent();
+
+    // When we subscribe with handler while unavailable and then unsubscribe
+    state_machine_.SetReceiveHandler(CreateMockScopedEventReceiveHandler(receive_handler));
+    EXPECT_TRUE(state_machine_.SubscribeEvent(max_num_slots_).has_value());
+    EXPECT_EQ(state_machine_.GetCurrentState(), SubscriptionStateMachineState::SUBSCRIPTION_PENDING_STATE);
+
+    state_machine_.UnsubscribeEvent();
+
+    // Then the state returns to NOT_SUBSCRIBED and resources are properly cleaned up
```
crimson11: So the only difference to the test case above is, that an "unsubscribe" happens at the end?
What do you want to show with this "extended" test?
I do not see the 2nd part in your "then" expectation being "checked":
// ... and resources are properly cleaned up
which resources and how did you check?

PR #117, opened 2025-12-10
clippy: added clippy support

dcalavrezo-qorix:
added support for running clippy checks

Addresses: eclipse-score/score#2008

### DIFF
--- a/score/mw/com/example/ipc_bridge/BUILD
+++ b/score/mw/com/example/ipc_bridge/BUILD
@@ -67,7 +67,7 @@ rust_library(
     name = "ipc_bridge_gen_rs",
     srcs = ["ipc_bridge_gen.rs"],
     visibility = [
-        "//visibility:public",
+        "//:__subpackages__",
     ],
     deps = [
         ":ipc_bridge_gen_cpp",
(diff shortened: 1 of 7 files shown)


score/mw/com/example/ipc_bridge/BUILD
```diff
@@ -67,7 +67,7 @@ rust_library(
     name = "ipc_bridge_gen_rs",
     srcs = ["ipc_bridge_gen.rs"],
     visibility = [
-        "//score/mw/com:__subpackages__",
+        "//visibility:public",
```
pawelrutkaq: maybe the __subpackages__ visibility shall be changed to `public in module` only ? I guess most of them shall not be public as such for external consumers.

score/mw/com/impl/rust/com-api/com-api-concept/BUILD
```diff
@@ -28,6 +28,10 @@ rust_library(
 rust_test(
     name = "com-api-concept-test",
     crate = ":com-api-concept",
-    tags = ["manual"],
+    tags = [
+        "manual",
+        "no-clippy",
```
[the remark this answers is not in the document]
dcalavrezo-qorix: rust_test

score/mw/com/example/ipc_bridge/BUILD
dcalavrezo-qorix: good point

PR #130, opened 2026-01-12
Make message_passing public

NEOatNHNG:
Other infrastructure components (also external to S-CORE) such as the SOME/IP Gateway or the diagnostics stack may re-use it for their internal communication needs.

Apart from the reduced implementation and resource overhead for infrastructure components, it is also beneficial to use message_passing instead of trying to build on top of mw::com to lessen the amount of configuration the integrators have to write in order to make those infrastructure components work. Otherwise they have to configure the infrastructure component and in addition also provide an mw::com configuration for the service instances used internally in the infrastructure component. This is error prone and provides negligible additional value.

### DIFF
--- a/score/message_passing/BUILD
+++ b/score/message_passing/BUILD
@@ -199,7 +199,7 @@ cc_library(
         "additional_warnings",
     ],
     tags = ["FFI"],
-    visibility = ["//visibility:public"],
+    visibility = ["//score/mw/com/impl:__subpackages__"],
     deps = [
         ":message_passing_common",
         ":qnx_resource_path",
--- a/score/message_passing/BUILD
+++ b/score/message_passing/BUILD
@@ -57,11 +57,7 @@ cc_library(
         "additional_warnings",
     ],
     tags = ["FFI"],
-    visibility = [
-        "//score/mw/com/impl:__subpackages__",
-        "@score_logging//score/datarouter:__subpackages__",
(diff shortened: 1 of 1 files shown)


score/message_passing/BUILD
```diff
@@ -203,11 +199,7 @@ cc_library(
         "additional_warnings",
     ],
     tags = ["FFI"],
-    visibility = [
-        "//score/mw/com/impl:__subpackages__",
-        "@score_logging//score/datarouter:__subpackages__",
-        "@score_logging//score/mw/log/detail/data_router:__subpackages__",
-    ],
```
castler, line 210: A user should just depend on `//score/message_passing`, not on `//score/message_passing:message_passing_qnx_dispatch`

Could you remove this part of the changset please?

castler (commented):
So we are not ready yet to give full advice (in written statement) when to use message passing and when to use LoLa.

In short:
If you need 1:N communication - you should use mw::com.
If you transfer bigger data (then you need zero-copy) - you should use mw::com

If you have only 1:1 communication and small data -> use message_passing.

When you could not change the visibility of the QNX dispatch target, we could merge this.

score/message_passing/BUILD
NEOatNHNG: Thank you for the hint. Previously message_passing was an alias target that is why previously the extended visibility was also needed here. I have reduced it now.

castler (approved):

Issue #155, opened 2026-02-05
Improvement: COM-API CommData derive macro


bharatGoswami8:
### What

User need to implement manually CommData for each type

```
struct UserType{
a:i32,
}

impl CommData for UserType
{
 const ID: &'static str = "UserType";
}
```

### How

```
#[derive(CommData)]
struct UserType{
a:i32,
}
```

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [ ] Requirements / Architecture are not affected by this change?

pawelrutkaq:
CommData macro shall probably allow specifying another type identifier like CommData(ident="str"). Imagine
```
mod x {
#[derive(CommData)]
struct UserType{
   a:i32,
}
}

mod y {
#[derive(CommData)]
struct UserType{
   a:i32,
   b:i32;
}

}
```

This will end up with the same `ID` pretty easily. So either we record full qualified type name as `module::module::type` or allow the user to change it.

Second point is that ComData macro shall internally also use Reloc macro to bring all checks needed (just appending #[Derive(reloc)] into code from macro ie.)

PR #156, opened 2026-02-06
Rust::com Derive macro for CommData

bharatGoswami8:
* Implemented derive macro for CommData trait impl

https://github.com/eclipse-score/communication/issues/155

### DIFF
--- a/score/mw/com/impl/rust/com-api/com-api-concept-macros/lib.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-concept-macros/lib.rs
@@ -39,7 +39,7 @@ use syn::{parse_macro_input, parse_quote, Data, DeriveInput, Fields, Generics, M
 ///    const ID: &'static str;
 /// }
 /// use com_api_concept_macros::{CommData, Reloc};
-/// #[derive(Debug, Reloc, CommData)]
+/// #[derive(Reloc, CommData)]
 /// #[repr(C)]
 /// pub struct MyData {
 ///    value: u32,
@@ -52,7 +52,7 @@ use syn::{parse_macro_input, parse_quote, Data, DeriveInput, Fields, Generics, M
 ///   const ID: &'static str;
 /// }
 /// use com_api_concept_macros::{CommData, Reloc};
-/// #[derive(Debug, Reloc, CommData)]
+/// #[derive(Reloc, CommData)]
 /// #[repr(C)]
 /// #[comm_data(id = "CustomID_MyData")]
 /// pub struct MyData {
@@ -474,7 +474,7 @@ fn reloc_fails_on_generic_types_that_do_not_impl_reloc() {}
 ///    const ID: &'static str;
 /// }
 /// use com_api_concept_macros::{CommData, Reloc};
-/// #[derive(Debug, Reloc, CommData)]
+/// #[derive(Reloc, CommData)]
 /// pub struct MyData {
 ///     value: u32,
 /// }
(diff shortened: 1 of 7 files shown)


score/mw/com/impl/rust/com-api/com-api-concept-macros/lib.rs
```diff
@@ -12,9 +12,110 @@
...
+/// for structs and C-like enums, ensuring that the type is marked with `#[repr(C)]`.
+/// It internally also derives the `Reloc` trait for the type.
+/// It also supports an optional attribute to specify a custom ID string:
+/// ```
+/// #[comm_data(id = "CustomID_String")]
+/// ```
+/// If no custom ID is provided, the struct or enum name is used as the ID.
+/// Example usage:
+/// ```
+/// pub trait CommData {
```
pawelrutkaq: add  to doc of ComData trait that user shall not implement manually but use this macros.

pawelrutkaq (changes requested):

score/mw/com/impl/rust/com-api/com-api-concept-macros/lib.rs
```diff
@@ -12,9 +12,110 @@
...
+        return syn::Error::new_spanned(
+            &ident_name,
+            "The #[derive(CommData)] macro requires #[repr(C)] on the type",
+        )
+        .to_compile_error()
+        .into();
+    }
+
+    // Validate that the type is supported (struct, tuple struct, unit struct, or C-like enum)
+    if let Err(()) = collect_field_types(&input_args.data) {
```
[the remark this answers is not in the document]
bharatGoswami8: yes this check is part of reloc , it was added for early error detection but reloc call derive_reloc to generate Reloc implementation moved starting of CommData so validation can be done first.

score/mw/com/impl/rust/com-api/com-api-concept-macros/lib.rs
```diff
@@ -12,9 +12,110 @@
...
+        if !attr.path().is_ident("comm_data") {
+            continue;
+        }
+
+        let Meta::List(list) = &attr.meta else {
+            continue;
+        };
+
+        let Ok(Meta::NameValue(nv)) = list.parse_args::<Meta>() else {
+            continue;
```
pawelrutkaq: if its anything else then also error ?

pawelrutkaq: maybe add test cases for it

bharatGoswami8: Agreed, we should return error.

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
```diff
@@ -197,7 +198,16 @@ where
...
+/// # Important
+/// Users must NOT implement the `CommData` trait manually. Always use the derive macros instead.
+/// Since `Reloc` is a supertrait of `CommData`, both must be derived explicitly:
+/// `#[derive(Reloc, CommData)]` on the struct definition.
+/// This is to ensure that all necessary trait bounds and metadata are correctly applied to the communication data types.
+/// Also if user wants to specify a custom ID for the communication data type,
+/// they can use the `#[comm_data(id = "CustomID")]` attribute on the struct definition.
+/// The custom ID must be unique across all communication data types to avoid conflicts in the system.
+/// If no custom ID is provided, the struct name will be used as the default ID with module path as prefix to ensure uniqueness.
 pub trait CommData: Reloc + Send + Debug + 'static {
```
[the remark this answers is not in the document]
pawelrutkaq: I think `Debug` shall be bound to API itself and not here, even though you need to state it in many places later (agree with @darkwisebear ). Is `Unpin` here something really necessary? Thie pin history is future workound in Rust as i remeber and  not sure if here we really need to mention it ?

PR #176, opened 2026-03-11
mw/com/impl: Add get/set to field

crimson11:
Added Get/Set support to ProxyField.

Issue: SWP-249574


KrishaDeshkool:
These changes are already merged as part of different PR.

PR #183, opened 2026-03-12
Fix download_utils dependency for json_schema_validator_test

mariuswbr:
When score_communication is used in another module, the `download_utils` dependency is stripped away, as it is declared as a `dev_dependency` within score_communication. There are `use_repo_rule` calls upon this `dev_dependency` to get repo_rule_proxy(s) `download_archive` and `download_file` which are then used to create among others the `@json_schema_validator` repo. 

grafik

When using the `validate_json_schema_test` macro outside of score_communication, like in the score_someip_gateway for example, this fails during BUILD file loading, as the macro references the `@json_schema_validator` repo. When trying to create that repo, using the repo_rule, this fails in the score_someip_gateway context, as the necessary `download_utils` is marked as `dev_dependency` and therefore not available outside of score_communication.

Usage of the macro in score_someip_gateway:
grafik

Build error in score_someip_gateway:
grafik

Scope of the PR is to make this a "normal" dependency.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -184,7 +184,8 @@ bazel_dep(name = "score_crates", version = "0.0.7", repo_name = "score_communica
 bazel_dep(name = "boost.program_options", version = "1.87.0")
 bazel_dep(name = "boost.interprocess", version = "1.87.0")
 
-bazel_dep(name = "download_utils", version = "1.2.2", dev_dependency = True)
+bazel_dep(name = "download_utils", version = "1.2.2")
+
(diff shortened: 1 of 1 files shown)


mariuswbr:
> @mariuswbr please rebase your branch to newest main to get the newer QNX workflow to run for this PR.

@LittleHuba Thanks for the hint! Rebased to latest main.

castler:
@mariuswbrI think the best place would be eclipse-score-tooling, I am afraid that this will get to big at some point - but it also feels wrong to me right now to create a module / repository, JUST for this schema validation.

PR #207, opened 2026-03-18
fields: Introduce `Get` functionality

muhseth:
- introduces EnableGet flag on SkeletonField Side.
- implements Get function functionality.

### DIFF
--- a/score/mw/com/impl/skeleton_field.h
+++ b/score/mw/com/impl/skeleton_field.h
@@ -272,7 +272,8 @@ SkeletonField<SampleDataType, EnableSet, EnableNotifier, EnableGet>::SkeletonFie
                                                      SkeletonFieldBindingFactory<SampleDataType>::CreateEventBinding(
                                                          SkeletonBaseView{parent}.GetAssociatedInstanceIdentifier(),
                                                          parent,
-                                                         field_name),
+                                                         field_name,
+                                                         (EnableSet || EnableGet) ? 1U : 0U),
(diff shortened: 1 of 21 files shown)


score/mw/com/impl/traits.h
```diff
@@ -150,8 +150,8 @@ class SkeletonTrait
     template <typename SampleType>
     using Event = SkeletonEvent<SampleType>;
 
-    template <typename SampleType>
-    using Field = SkeletonField<SampleType>;
+    template <typename SampleType, bool EnableSet = false, bool EnableNotifier = false, bool EnableGet = false>
```
KrishaDeshkool, line 153: Note: on the proxy side this is set as
`    template <typename SampleType, bool EnableSet = false, bool EnableGet = false, bool EnableNotifier = false>
    using Field = ProxyField<SampleType>;`

we should stick to one convention since this same Field would be used a proxy and skeleton then this different bools would set different things.

score/mw/com/impl/skeleton_field.h
```diff
@@ -26,19 +27,75 @@
...
 #include <utility>
 
 namespace score::mw::com::impl
 {
 
-template <typename SampleDataType>
+namespace detail
+{
+/// Tag types for constructor overload disambiguation based on EnableGet/EnableSet template parameters
+struct EnableBothTag
```
KrishaDeshkool, line 41: Just a note:
This we can reuse between proxy and skeleton side, i have that on my PR in a new header called method_type.h that lets us disambiguate the method types.

PR #209, opened 2026-03-18
Suya lola field support implement skeleton field set handler

soldier-sky:
Implement RegisterSetHandler() method to allow users to register callbacks for handling remote Set requests on skeleton fields. The Set handler wraps user-provided validation/modification logic and automatically broadcasts accepted values to all field subscribers via Update().

Issue: SWP-249527

### DIFF
--- a/score/mw/com/impl/skeleton_field.h
+++ b/score/mw/com/impl/skeleton_field.h
@@ -137,20 +137,19 @@ class SkeletonField : public SkeletonFieldBase
                       "The argument initially holds the proxy-requested value and may be modified in-place.");
         set_handler_ = std::move(handler);
 
-        auto wrapped_callback = [this](const FieldType& new_value) -> FieldType {
+        auto wrapped_callback = [this](FieldType& new_value) -> FieldType {
             // Allow user to validate/modify the value in-place
-            FieldType value = new_value;
-            set_handler_(value);
+            set_handler_(new_value);
 
             // Store the (possibly modified) value as the latest field value
-            auto update_result = this->Update(value);
+            auto update_result = this->Update(new_value);
             if (!update_result.has_value())
             {
                 score::mw::log::LogError("lola") << "Set handler: failed to update field value.";
             }
 
             // Return the accepted value to the proxy
-            return value;
+            return new_value;
         };
 
         is_set_handler_registered_ = true;
--- a/score/mw/com/impl/skeleton_field_test.cpp
+++ b/score/mw/com/impl/skeleton_field_test.cpp
@@ -1354,6 +1354,69 @@ TEST(SkeletonFieldSetHandlerTest, UserCallbackCanModifyValueInPlace)
     capturing_binding_ref.captured_handler_(in_span, out_span);
 }
 
+// Handler wrapping: Update() failure inside the wrapped handler is logged, not propagated
+TEST(SkeletonFieldSetHandlerTest, WrappedHandlerLogsWhenUpdateFails)
+{
+    RecordProperty("Description",
+                   "When the event binding's Send() fails inside the wrapped set handler, the "
+                   "failure shall be logged and the handler shall complete normally without "
+                   "propagating the error to the proxy caller. The user callback is still "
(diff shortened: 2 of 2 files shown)


score/mw/com/impl/skeleton_field.h
```diff
@@ -124,22 +123,43 @@ class SkeletonField : public SkeletonFieldBase
     }
 
     // SFINAE-gated: only exists when EnableSet == true
-    template <bool ES = EnableSet, typename std::enable_if<ES, int>::type = 0>
-    ResultBlank RegisterSetHandler(SetHandlerType handler)
+    //
+    // \ brief Registers a handler that is invoked by the middleware whenever a proxy calls the field setter.
+    //
+    // \tparam CallableType Any callable (std::function, score::cpp::callback, lambda, ...) with the signature:
+    //         void(Result<FieldType>& result, const FieldType& new_value)
```
gdadunashvili: As we discussed this today, I think the callback should expect `void(FieldType& new_value)`

score/mw/com/impl/skeleton_field.h
```diff
@@ -92,6 +122,50 @@ class SkeletonField : public SkeletonFieldBase
...
+    {
+        static_assert(std::is_invocable_v<CallableType, Result<FieldType>&, const FieldType&>,
+                      "RegisterSetHandler: handler must be callable as "
+                      "void(Result<FieldType>&, const FieldType&). "
+                      "First arg is the result/return out-ref, second is the incoming value (read-only).");
+        set_handler_ = std::move(handler);
+
+        // Capture only self_ptr_ to stay within TypeErasedHandler capacity.
+        // The user callback is stored in set_handler_ on the SkeletonField itself.
+        auto wrapped_callback = [self = self_ptr_](FieldType new_value) -> Result<FieldType> {
```
crimson11: Also this signature isn't correct.
The `wrapped_callback` you are going to internally register as the method-handler in this case (Set) has the following signature:
`FieldType(FieldType&)` at least for now! When my ticket here gets done/fixed: Ticket-252582

then the signature shall look like `void(FieldType& result, const FieldType& input)`

score/mw/com/impl/skeleton_field.h
```diff
@@ -92,6 +122,40 @@ class SkeletonField : public SkeletonFieldBase
...
+    //   - new_value : the value requested by the proxy.
+    template <bool ES = EnableSet, typename std::enable_if<ES, int>::type = 0, typename CallableType>
+    ResultBlank RegisterSetHandler(CallableType&& handler)
+    {
+        static_assert(std::is_invocable_v<CallableType, FieldType&>,
+                      "RegisterSetHandler: handler must be callable as void(FieldType& value). "
+                      "The argument initially holds the proxy-requested value and may be modified in-place.");
+        set_handler_ = std::move(handler);
+
+        auto wrapped_callback = [this](FieldType new_value) -> Result<FieldType> {
```
crimson11: This signature is imho wrong. Right now - before Ticket-252582 gets implemented - the signature of the handler registered via `SkeletonMethod::RegisterHandler` looks like:
`ReturnType(const InArgType1&, const InArgType2&, ...)`
i.e. there is NO `Result<xxx>` and you should take the in-arg by reference!
see Ticket-251657

score/mw/com/impl/skeleton_field.h
```diff
@@ -92,6 +122,41 @@ class SkeletonField : public SkeletonFieldBase
...
+
+            // Store the (possibly modified) value as the latest field value
+            auto update_result = this->Update(value);
+            if (!update_result.has_value())
+            {
+                score::mw::log::LogError("lola") << "Set handler: failed to update field value.";
+            }
+
+            // Return the accepted value to the proxy
+            return value;
```
crimson11: Here you also return new_value (reference) ... currently this will lead to a copy, since your  `wrapped_callback` needs for now to return byVal. This will be changed/refactored later, when this ticket gets implemented: Ticket-252582

Issue #13, opened 2025-06-05
Expose SamplePtr age in C++ API


MarkSchmitt:
In order to sort Samples by age as was discussed as a necessity in workshop https://github.com/orgs/eclipse-score/discussions/1076 , the C++ API needs to expose the sample's age.

PR #258, opened 2026-03-30
Expose SamplePtr age in C++ API

Gfilinic:
I’m interested in contributing to this project and started by looking into Issue #13.

## Changes
- updated the constructor to carry the necessary context
- added a `GetTimestamp()` function as requested
- implemented a unit test to verify the behavior

## Questions
- Are there any specific commit message or PR guidelines I should follow?
- Is there a preferred way to assign or request reviewers?

## Design question
For the default/null constructors, I currently initialize the internal state (`event_data_ctrl_ = nullptr`, empty `slot_indicator_`, etc.) so that `GetTimestamp()` safely returns a default value.

Does this align with the intended design, or would you prefer a different approach for representing an "invalid" `SamplePtr`?

I was advised to reach out to @crimson11 for this part of the code, as he likely has deeper insight into the design.

### DIFF
--- a/score/mw/com/impl/bindings/lola/sample_ptr.h
+++ b/score/mw/com/impl/bindings/lola/sample_ptr.h
@@ -105,11 +105,12 @@ class SamplePtr final
 
     /// \brief Compares two SamplePtr instances based on their timestamp
     /// \param other SamplePtr to compare against
-    /// \return true if this instance is older than \p other, false otherwise or if any of the SamplePtr are invalid  
+    /// \return true if this instance is older than \p other, false otherwise or if any of the SamplePtr are invalid
     bool operator<(const SamplePtr& other) const noexcept
     {
-        if(!(*this) || !other) {
-             return false;
+        if (!(*this) || !other)
+        {
+            return false;
         }
 
         return timestamp_ < other.timestamp_;
@@ -120,7 +121,8 @@ class SamplePtr final
     /// \return true if this instance is newer than \p other, false otherwise or if any of the SamplePtr are invalid
     bool operator>(const SamplePtr& other) const noexcept
     {
-        if(!(*this) || !other) {
+        if (!(*this) || !other)
+        {
             return false;
         }
         return timestamp_ > other.timestamp_;
--- a/score/mw/com/impl/bindings/lola/sample_ptr.h
+++ b/score/mw/com/impl/bindings/lola/sample_ptr.h
@@ -50,6 +50,7 @@ class SamplePtr final
               TransactionLogIndex transaction_log_idx) noexcept
         : SamplePtr{ptr, std::make_optional<SlotDecrementer>(event_data_ctrl_local, slot_index, transaction_log_idx)}
     {
+        timestamp_ = (event_data_ctrl)[slot_indicator.GetIndex()].GetTimeStamp();
     }
 
     ~SamplePtr() noexcept = default;
@@ -102,6 +103,29 @@ class SamplePtr final
         return managed_object_;
     }
 
+    /// \brief Compares two SamplePtr instances based on their timestamp
+    /// \param other SamplePtr to compare against
+    /// \return true if this instance is older than \p other, false otherwise or if any of the SamplePtr are invalid  
+    bool operator<(const SamplePtr& other) const noexcept
+    {
+        if(!(*this) || !other) {
+             return false;
+        }
+
+        return timestamp_ < other.timestamp_;
+    }
+
+    /// \brief Compares two SamplePtr instances based on their timestamp
+    /// \param other SamplePtr to compare against
+    /// \return true if this instance is newer than \p other, false otherwise or if any of the SamplePtr are invalid
+    bool operator>(const SamplePtr& other) const noexcept
+    {
+        if(!(*this) || !other) {
+            return false;
+        }
+        return timestamp_ > other.timestamp_;
+    }
+
   private:
     explicit SamplePtr(pointer managed_object, std::optional<SlotDecrementer>&& slog_decrementer) noexcept
         : managed_object_{managed_object}, slot_decrementer_{std::move(slog_decrementer)}
@@ -110,6 +134,7 @@ class SamplePtr final
 
     pointer managed_object_;
     std::optional<SlotDecrementer> slot_decrementer_;
+    EventSlotStatus::EventTimeStamp timestamp_;
 };
 
 }  // namespace score::mw::com::impl::lola
--- a/score/mw/com/impl/bindings/lola/sample_ptr.h
+++ b/score/mw/com/impl/bindings/lola/sample_ptr.h
@@ -50,7 +50,7 @@ class SamplePtr final
               TransactionLogIndex transaction_log_idx) noexcept
         : SamplePtr{ptr, std::make_optional<SlotDecrementer>(event_data_ctrl_local, slot_index, transaction_log_idx)}
     {
-        timestamp_ = (event_data_ctrl)[slot_indicator.GetIndex()].GetTimeStamp();
+        timestamp_ = (event_data_ctrl_local)[slot_index].GetTimeStamp();
     }
 
     ~SamplePtr() noexcept = default;
--- a/score/mw/com/impl/bindings/lola/sample_ptr.h
+++ b/score/mw/com/impl/bindings/lola/sample_ptr.h
@@ -49,7 +49,7 @@ class SamplePtr final
               const SlotIndexType slot_index) noexcept
         : SamplePtr{ptr, std::make_optional<SlotDecrementer>(event_data_ctrl_local, slot_index)}
     {
-        timestamp_ = (event_data_ctrl)[slot_indicator.GetIndex()].GetTimeStamp();
+        timestamp_ = (event_data_ctrl_local)[slot_index].GetTimeStamp();
     }
 
     ~SamplePtr() noexcept = default;
--- a/score/mw/com/impl/bindings/lola/sample_ptr.h
+++ b/score/mw/com/impl/bindings/lola/sample_ptr.h
@@ -104,11 +104,12 @@ class SamplePtr final
 
     /// \brief Compares two SamplePtr instances based on their timestamp
     /// \param other SamplePtr to compare against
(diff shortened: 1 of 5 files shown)


castler (changes requested):
Thanks for your contribution.

I am the opinion that we should not expose this implementation detail - especially not the implementation data types.

We still want to give the opportuntity for a user to sort the samples.
Thus, we propose that the SamplePtr gets an `bool IsNewer(SamplePtr<T> other)` function instead.

score/mw/com/impl/bindings/lola/sample_ptr.h
```diff
@@ -49,17 +49,27 @@ class SamplePtr final
...
               TransactionLogSet::TransactionLogIndex transaction_log_idx) noexcept
-        : SamplePtr{ptr, std::make_optional<SlotDecrementer>(&event_data_ctrl, slot_indicator, transaction_log_idx)}
+        : SamplePtr{ptr,
+                    &event_data_ctrl,
+                    slot_indicator,
+                    std::make_optional<SlotDecrementer>(&event_data_ctrl, slot_indicator, transaction_log_idx)}
     {
     }
 
+    bool IsNewer(const SamplePtr& other) const noexcept
```
castler: How do we want to treat the error case? What if `other` is a nullptr - or if `this` is default constructed?

Do we let the user know via a command - or would it not be better to return a result?

score/mw/com/impl/bindings/lola/sample_ptr.h
```diff
@@ -49,17 +49,37 @@ class SamplePtr final
...
     {
     }
 
+    /// \brief checks whether this SamplePtr refers to a strictly newer sample than \p other.
+    /// \details Returns true only if both SamplePtrs refer to valid event slots and the timestamp of this sample is
+    /// greater than the timestamp of \p other. Returns false if either SamplePtr is invalid or if this sample is not
+    /// strictly newer.
+    /// \param other SamplePtr to compare against
+    /// \return true if this SamplePtr is strictly newer than \p other, false otherwise
+    bool IsNewer(const SamplePtr& other) const noexcept
```
crimson11: Maybe I'm missing the context ... but a "user" will never see a `lola::SamplePtr` directly .... he will see an `impl::SamplePtr` ... so do you plan to introduce this `IsNewer()` also to `impl::SamplePtr`? Otherwise I fail to see, how it is gonna be used.
 And a 2nd point ... in a short discussion with Jan some time ago, I said: "Yes, making SamplePtrs `comparable` regarding their age makes sense" ... but why not introduce an `bool operator<`?

score/mw/com/impl/bindings/lola/sample_ptr.h
```diff
@@ -49,17 +49,37 @@ class SamplePtr final
...
+    }
+
     ~SamplePtr() noexcept = default;
 
     /// \brief assign nullptr.
     /// \return reference to nullptr assigned (invalid) SamplePtr
     SamplePtr& operator=(std::nullptr_t) & noexcept
     {
         managed_object_ = nullptr;
+        event_data_ctrl_ = nullptr;
```
crimson11: Why are you not directly storing the `timestamp` value, which is an `uint32` -> 4Byte?
Instead you are storing now an additional pointer -> 8 Byte AND a `ControlSlotIndicator ` ... which is another 10 - 16 Byte (rather 16 Byte, because the type is poorly layouted :) ) ... so you are adding 24 Byte instead of 4 Byte to the `SamplePtr` ... from which we have flowing around hundreds in applications ...

... and the access to the timestamp like you are doing it now, will "kill" your performance ... I guess. So having the timestamp 4 Byte locally stored ... when you access the `SamplePtr` ... then you already have the timestamp "in your hand" -> it will be in the same cache-line.... with your approach to access it via the pointer plus index ... you are jumping into a complete separate memory-location ... in most/many cases suffering from cache-miss and therefore taking "ages".

... maybe I'm missing the point here completely and there was a strong reasoning for all these indirections?

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/sample_ptr.h
Gfilinic: Thanks for joining the discussion and for your feedback.

I’m still getting familiar with how SamplePtr is used across the different namespaces, so I may be missing some context here. My initial approach was to add this on the lola layer just to get the functionality in place.

You’re right that if users only interact with impl::SamplePtr, then exposing IsNewer() only on lola::SamplePtr would not be sufficient. I’ll revisit this and align the API so it’s available at the correct abstraction level.

Regarding the API design, I initially went with IsNewer() as a straightforward solution to get things moving. Introducing comparison operators felt like a broader design decision, so I preferred to start with something simpler.

Also, since @castler explicitly suggested adding a bool IsNewer(SamplePtr<T> other) function, I followed that direction and did not explore alternative approaches at that stage.

That said, I agree that comparison operators now seem like a more natural fit, and I’ll take that into account in the next iteration.

score/mw/com/impl/bindings/lola/sample_ptr.h
Gfilinic: I introduced the additional members because I initially wanted to keep the implementation flexible in case we later needed access to more than just the timestamp. At the time, that felt like the safer starting point.

That said, I agree with your point  regarding the memory overhead and potential cache impact. Given that, this is not the optimal approach, and storing the `uint32` directly is clearly simpler and more efficient.

I’ll rework this into a more optimized version and clean up the PR accordingly, unless you would prefer to keep the current state visible for discussion/history.

Gfilinic:
Thanks for the feedback and the discussion so far.

I revisited the implementation and reworked the approach based on the suggestions.

### Changes

- Replaced the initially proposed dedicated function (`IsNewer`) with standard comparison operators:
  - `operator<`
  - `operator>`

- Simplified the internal representation:
  - Instead of storing multiple internal members, the implementation now only keeps the timestamp (`EventTimeStamp`, underlying `uint32_t`)
  - Timestamp is captured once during construction

- Updated `impl::SamplePtr` accordingly to expose the same comparison operators at the abstraction level

### Notes on testing (impl::SamplePtr)

I did not introduce additional comparison tests for `impl::SamplePtr`.

The current plumbing tests rely on `mock_binding::SamplePtr`, which is defined as a `std::unique_ptr` alias and does not provide timestamp or comparison semantics.

Because of that:
- comparisons in the plumbing layer would always evaluate to `false`
- meaningful tests would require turning the mock binding into a full class with timestamp support

I intentionally did not extend the mock binding in this PR to keep the scope focused.

### Additional notes

I force-pushed the branch with only these changes to clean up the git history, but I do have a backup branch with the previous implementation in case that approach is preferred.

score/mw/com/impl/plumbing/sample_ptr.h
```diff
@@ -143,6 +143,43 @@ class SamplePtr final
...
+        {
+            return false;
+        }
+
+        return std::visit(
+            score::cpp::overload(
+                [](const lola::SamplePtr<SampleType>& lhs, const lola::SamplePtr<SampleType>& rhs) noexcept -> bool {
+                    return lhs < rhs;
+                },
+                [](const auto&, const auto&) noexcept -> bool {
```
crimson11, line 153: why no overload for the `mock_binding::SamplePtr`? Just add a correponding mockable </> operator to the `mock_binding::SamplePtr`

score/mw/com/impl/plumbing/sample_ptr.h
```diff
@@ -143,6 +143,43 @@ class SamplePtr final
         return get();
     }
 
+    bool operator<(const SamplePtr& other) const noexcept
+    {
+        if (!(*this) || !other)
```
crimson11: why this "redundant" check on this level? In case of `score::cpp::blank>` you already have the fall-through in the visitor ... and every other binding (your lola::SamplePtr) does this check internally ...

crimson11 (changes requested):

score/mw/com/impl/plumbing/sample_ptr.h
crimson11: Ah. OK - saw your comment ... and realized, that the mock_binding::SamplePtr isn't a gmock ... so you can ignore this comment

score/mw/com/impl/plumbing/sample_ptr.h
Gfilinic: Great catch, thank you for pointing this out!

I initially tried to be defensive and mirrored the behavior from the LoLa `SamplePtr`, essentially following the same pattern.

However, you’re right that this is already handled at this level, so the additional check is redundant. I’ve removed it to keep the implementation aligned with the existing design.

score/mw/com/impl/plumbing/sample_ptr.h
```diff
@@ -143,6 +143,33 @@ class SamplePtr final
         return get();
     }
 
+    bool operator<(const SamplePtr& other) const noexcept
```
crimson11, line 146: Now we only need some basic tests for `<` /  `>` on `impl::SamplePtr` level, which show, that these ops are correctly forwarded to the bindingspecific impl ...

crimson11 (changes requested):

score/mw/com/impl/plumbing/sample_ptr.h
Gfilinic: ## Update

Added test in impl by implementing actual sampleptr instead of using generic mock pointer.

Gfilinic:
# Update
Added test in impl by implementing actual sampleptr instead of using generic mock pointer.

Gfilinic:
Will there be more discussion in this PR or should I leave it be?

LittleHuba:
Hi @Gfilinic,
sorry for the extended silence. The colleague that did the reviews so far is on sick leave and this PR slipped through on the take over. I'll take the review up and will provide you feedback ASAP.

LittleHuba (dismissed):
Hi @Gfilinic,
your PR looks good now. It can be merged as it currently is.
Though, ideally I would like an immediate follow-up that provides the implementation of this API in the mock_binding. Otherwise we have an API that is not mockable...
Are you interested in working on that?

castler (approved):

Issue #91, opened 2025-11-12
Extract move assign / construction functionality in SkeletonBase into separate class


LittleHuba:
### What

SkeletonBase currently has a Cleanup method which is called both in the destructor and the move assignment operator. We should try to extract the functionality needed in the move constructor / assignment operator to a separate class so that SkeletonBase can have a default move constructor / move assignment operator if possible.

### Estimates for realization

- Effort medium
- No impact on users expected

### Category

- [x] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

JochenSatETAS:
@LittleHuba Maybe I'm looking into the wrong branch, but here: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/skeleton_base.h the SkeletonBase does not have a explicitly defined move-assignment operator.

bemerybmw:
Cleanup was previously removed from SkeletonBase. Cleanup on destruction is now initiated from SkeletonWrapperClass: https://github.com/eclipse-score/communication/blob/d4581a997d0536f72fa3fabb599246e4fa7825d9/score/mw/com/impl/traits.h#L246

PR #259, opened 2026-03-31
Feature/extraction skeleton service elements

rudresh-systream:
Ticket:
Ticket Extract move assign / construction functionality in SkeletonBase into separate class #91
91 is the issue
Issue link: https://github.com/eclipse-score/communication/issues/91

Context:
The original change introduced a new SkeletonServiceElements wrapper class and moved the events_, fields_, and methods_ maps from SkeletonBase into that wrapper. The intention was to separate the service-element map handling and remove duplicated UpdateSkeletonReference loops from the SkeletonBase move constructor and move assignment operator.

Review feedback summary:
After review, the SkeletonServiceElements wrapper was considered unnecessary because it did not provide real encapsulation. SkeletonBase still needs to directly operate on the registered events, fields, and methods for service offering, stop-offering, tracing callback setup, binding validation, and post-move reference fix-up.

Therefore, I updated the approach to keep the useful part of the refactor while removing the unnecessary wrapper abstraction.

Updated solution:
The updated implementation keeps events_, fields_, and methods_ directly inside SkeletonBase and removes the SkeletonServiceElements wrapper class.

The duplicated post-move reference update logic is now extracted into a private SkeletonBase helper:

    UpdateAllServiceElementReferences()

This method updates the back-reference of every registered event, field, and method to the new SkeletonBase instance after move construction or move assignment.

This keeps the move constructor and move assignment operator simple and avoids duplicated loops, while also avoiding the additional wrapper class that the review identified as unnecessary.

Changes made:
[ 1] Removed SkeletonServiceElements wrapper class.
[ 2] Removed skeleton_service_elements.h.
[ 3] Removed skeleton_service_elements.cpp.
[ 4] Removed skeleton_service_elements_test.cpp.
[ 5] Removed skeleton_service_elements Bazel library target.
[ 6] Removed skeleton_service_elements_test Bazel test target.
[ 7] Restored events_, fields_, and methods_ as direct members of SkeletonBase.
[ 8] Kept SkeletonBase::UpdateAllServiceElementReferences() as the common helper for post-move reference fix-up.
[ 9] Updated SkeletonBase move constructor to move events_, fields_, and methods_, then call UpdateAllServiceElementReferences().
[10] Updated SkeletonBase move assignment operator to move events_, fields_, and methods_, then call UpdateAllServiceElementReferences().
[11] Updated all internal SkeletonBase logic to use events_, fields_, and methods_ directly again.
[12] Made UpdateEvent, UpdateField, and UpdateMethod symmetric.
[13] Replaced the previous event-specific std::map::at() behavior with the same find() + SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE() + defensive std::terminate() pattern used for all three update methods.
[14] Removed the overdone logging from UpdateField and UpdateMethod.
[15] Removed the unnecessary test refactor because the public SkeletonBase / SkeletonBaseView API did not change.
[16] Kept the existing SkeletonBase-level test coverage for registration, move construction, move assignment, and back-reference fix-up.

Why this still solves the ticket:
The core issue was duplicated move construction / move assignment logic in SkeletonBase, specifically the repeated loops that update registered service elements after a SkeletonBase object is moved.

That duplication is now removed by UpdateAllServiceElementReferences().

So the ticket use case is still solved, but without introducing the SkeletonServiceElements wrapper because the reviewer pointed out that the wrapper did not add enough value.

Final design:
- SkeletonBase continues to own the service element maps directly.
- Move constructor and move assignment remain responsible for moving the members.
- UpdateAllServiceElementReferences() centralizes the post-move back-reference update.
- UpdateEvent, UpdateField, and UpdateMethod now follow the same consistent behavior.
- No downstream API breakage is expected because SkeletonBase::SkeletonEvents, SkeletonBase::SkeletonFields, and SkeletonBase::SkeletonMethods are still available as type aliases.

Validation:
- git diff --check passed locally in the generated patch environment.
- Bazel tests passed be in the repo environment with:

    bazel test //score/mw/com/impl:skeleton_base_test
    bazel test //...

The updated version solves the same ticket more directly:
it removes duplicated move-operation logic through UpdateAllServiceElementReferences(), keeps the maps inside SkeletonBase, and avoids unnecessary structural changes.
 
Summary :
a)The first version solved the issue by extracting the maps into a new wrapper class, but the review showed that this abstraction was not useful.
b)Forking: Created my own sandbox to work in.
c)Upstream: Linked my local repo to the main project to stay updated.
d)Feature Branch: Isolated my specific work so it doesn't get mixed up.
e)Bazel: Used the project's specific build tool to ensure high code quality.

### DIFF
--- a/score/mw/com/impl/skeleton_base.cpp
+++ b/score/mw/com/impl/skeleton_base.cpp
@@ -23,6 +23,7 @@
 
 #include "score/mw/log/logging.h"
 
+#include <score/assert.hpp>
 #include <score/utility.hpp>
 
 #include <algorithm>
@@ -104,22 +105,7 @@ SkeletonBase::SkeletonBase(SkeletonBase&& other) noexcept
       skeleton_mock_{std::move(other.skeleton_mock_)},
       service_offered_flag_{std::move(other.service_offered_flag_)}
 {
-    // Since the address of this skeleton has changed, we need update the address stored in each of the events and
-    // fields belonging to the skeleton.
-    for (auto& event : events_)
-    {
-        event.second.get().UpdateSkeletonReference(*this);
-    }
-
-    for (auto& field : fields_)
-    {
-        field.second.get().UpdateSkeletonReference(*this);
-    }
-
-    for (auto& method : methods_)
-    {
-        method.second.get().UpdateSkeletonReference(*this);
-    }
+    UpdateAllServiceElementReferences();
 }
 
 // Suppress "AUTOSAR C++14 A6-2-1" rule violation. The rule states "Move and copy assignment operators shall either move
@@ -142,21 +128,8 @@ SkeletonBase& SkeletonBase::operator=(SkeletonBase&& other) noexcept
     skeleton_mock_ = std::move(other.skeleton_mock_);
     service_offered_flag_ = std::move(other.service_offered_flag_);
 
-    // Since the address of this skeleton has changed, we need update the address stored in each of the events and
-    // fields belonging to the skeleton.
-    for (auto& event : events_)
-    {
-        event.second.get().UpdateSkeletonReference(*this);
-    }
+    UpdateAllServiceElementReferences();
 
-    for (auto& field : fields_)
-    {
-        field.second.get().UpdateSkeletonReference(*this);
-    }
-    for (auto& method : methods_)
-    {
-        method.second.get().UpdateSkeletonReference(*this);
-    }
     return *this;
 }
 
@@ -317,6 +290,105 @@ auto SkeletonBase::AreBindingsValid() const noexcept -> bool
         });
 
     return is_skeleton_binding_valid && are_service_element_bindings_valid;
+}
+
+void SkeletonBase::UpdateAllServiceElementReferences() noexcept
+{
+    for (auto& event : events_)
+    {
+        event.second.get().UpdateSkeletonReference(*this);
+    }
+    for (auto& field : fields_)
+    {
+        field.second.get().UpdateSkeletonReference(*this);
+    }
+    for (auto& method : methods_)
+    {
+        method.second.get().UpdateSkeletonReference(*this);
+    }
+}
+
+SkeletonBaseView::SkeletonBaseView(SkeletonBase& skeleton_base) : skeleton_base_{skeleton_base} {}
+
+InstanceIdentifier SkeletonBaseView::GetAssociatedInstanceIdentifier() const
+{
+    return skeleton_base_.instance_id_;
+}
+
+SkeletonBinding* SkeletonBaseView::GetBinding() const
+{
+    return skeleton_base_.binding_.get();
+}
+
+void SkeletonBaseView::RegisterEvent(const std::string_view event_name, SkeletonEventBase& event)
+{
+    const auto result = skeleton_base_.events_.emplace(event_name, event);
+    const bool was_event_inserted = result.second;
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(was_event_inserted, "Event cannot be registered as it already exists.");
+}
+
+void SkeletonBaseView::RegisterField(const std::string_view field_name, SkeletonFieldBase& field)
+{
+    const auto result = skeleton_base_.fields_.emplace(field_name, field);
+    const bool was_field_inserted = result.second;
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(was_field_inserted, "Field cannot be registered as it already exists.");
+}
+
+void SkeletonBaseView::RegisterMethod(const std::string_view method_name, SkeletonMethodBase& method)
+{
+    const auto result = skeleton_base_.methods_.emplace(method_name, method);
+    const bool was_method_inserted = result.second;
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(was_method_inserted, "Method cannot be registered as it already exists.");
+}
+
+void SkeletonBaseView::UpdateEvent(const std::string_view event_name, SkeletonEventBase& event) noexcept
+{
+    auto event_name_it = skeleton_base_.events_.find(event_name);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(
+        event_name_it != skeleton_base_.events_.cend(),
+        "SkeletonBaseView::UpdateEvent failed to update event because the requested event doesn't exist");
+
+    event_name_it->second = event;
+}
+
+void SkeletonBaseView::UpdateField(const std::string_view field_name, SkeletonFieldBase& field) noexcept
+{
+    auto field_name_it = skeleton_base_.fields_.find(field_name);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(
+        field_name_it != skeleton_base_.fields_.cend(),
+        "SkeletonBaseView::UpdateField failed to update field because the requested field doesn't exist");
+
+    field_name_it->second = field;
+}
+
+void SkeletonBaseView::UpdateMethod(const std::string_view method_name, SkeletonMethodBase& method) noexcept
+{
+    auto method_it = skeleton_base_.methods_.find(method_name);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(
+        method_it != skeleton_base_.methods_.cend(),
+        "SkeletonBaseView::UpdateMethod failed to update method because the requested method doesn't exist");
+
+    method_it->second = method;
+}
+
+const SkeletonBase::SkeletonEvents& SkeletonBaseView::GetEvents() const noexcept
+{
+    return skeleton_base_.events_;
+}
+
+const SkeletonBase::SkeletonFields& SkeletonBaseView::GetFields() const noexcept
+{
+    return skeleton_base_.fields_;
+}
+
+const SkeletonBase::SkeletonMethods& SkeletonBaseView::GetMethods() const noexcept
+{
+    return skeleton_base_.methods_;
+}
+
+bool SkeletonBaseView::AreBindingsValid() const
+{
+    return skeleton_base_.AreBindingsValid();
 }
 
 score::cpp::optional<InstanceIdentifier> GetInstanceIdentifier(const InstanceSpecifier& specifier)
--- a/score/mw/com/impl/skeleton_base.h
+++ b/score/mw/com/impl/skeleton_base.h
@@ -23,9 +23,6 @@
 
 #include "score/result/result.h"
 
-#include "score/mw/log/logging.h"
-
-#include <score/assert.hpp>
 #include <score/optional.hpp>
 #include <score/span.hpp>
 
@@ -87,7 +84,7 @@ class SkeletonBase
     }
 
   protected:
-    bool AreBindingsValid() const noexcept;
+    [[nodiscard]] bool AreBindingsValid() const noexcept;
 
     // Suppress "AUTOSAR C++14 A11-3-1", The rule states: "Friend declarations shall not be used".
     // Design decision. This class provides a view to the private members of this class.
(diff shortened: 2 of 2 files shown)


score/mw/com/impl/skeleton_base.cpp
```diff
@@ -86,46 +86,41 @@ SkeletonBinding::SkeletonFieldBindings GetSkeletonFieldBindingsMap(const Skeleto
...
 
+// ── Move constructor ──────────────────────────────────────────────────────────
+//
+// Step 1: Move every data member from `other` to `*this`.
+//         SkeletonServiceElements has a defaulted move constructor that moves the
+//         three maps; it does NOT touch back-references.
+// Step 2: Call UpdateSkeletonReferences so that every registered event, field, and
+//         method now points to the new SkeletonBase address (*this).
+//         This is the only reason this constructor cannot be defaulted.
+//
```
LittleHuba: Please change to a doxygen style.

score/mw/com/impl/skeleton_base.cpp
```diff
@@ -86,46 +86,41 @@ SkeletonBinding::SkeletonFieldBindings GetSkeletonFieldBindingsMap(const Skeleto
...
 SkeletonBase::SkeletonBase(SkeletonBase&& other) noexcept
     : binding_{std::move(other.binding_)},
-      events_{std::move(other.events_)},
-      fields_{std::move(other.fields_)},
-      methods_{std::move(other.methods_)},
+      service_elements_{std::move(other.service_elements_)},
       instance_id_{std::move(other.instance_id_)},
-      skeleton_mock_{std::move(other.skeleton_mock_)},
+      skeleton_mock_{other.skeleton_mock_},
```
LittleHuba: Why drop the move here?

score/mw/com/impl/skeleton_base.cpp
```diff
@@ -86,46 +86,41 @@ SkeletonBinding::SkeletonFieldBindings GetSkeletonFieldBindingsMap(const Skeleto
...
 
-// Suppress "AUTOSAR C++14 A6-2-1" rule violation. The rule states "Move and copy assignment operators shall either move
-// or respectively copy base classes and data members of a class, without any side effects." Due to architectural
-// decisions, SkeletonBase must perform a cleanup and update references to itself in its events and fields. Therefore,
-// side effects are required.
+// ── Move assignment operator ──────────────────────────────────────────────────
+//
+// Same two-step approach as the move constructor.  The AUTOSAR A6-2-1 suppression
+// is kept because we intentionally update back-references as a required side-effect.
+//
```
LittleHuba: Change to doxygen style.

score/mw/com/impl/skeleton_base.cpp
```diff
@@ -135,34 +130,42 @@ SkeletonBase& SkeletonBase::operator=(SkeletonBase&& other) noexcept
...
     {
         field.second.get().UpdateSkeletonReference(*this);
     }
-    for (auto& method : methods_)
+    for (auto& method : service_elements_.GetMethods())
     {
         method.second.get().UpdateSkeletonReference(*this);
     }
-    return *this;
 }
```
LittleHuba: This feels like the wrong place. You should be able to take a pointer to SkeletonBase as parameter in the function. This allows you to work with a predeclaration in SkeletonServiceElements and avoid the circular dependency.
Then do the implementation in cpp. You might have to use `impl_deps` in bazel to avoid the circular dependency there.

score/mw/com/impl/skeleton_service_elements.h
```diff
@@ -0,0 +1,108 @@
...
+    /// \brief Not copyable — mirrors SkeletonBase's copy-deletion requirement.
+    SkeletonServiceElements(const SkeletonServiceElements&) = delete;
+    SkeletonServiceElements& operator=(const SkeletonServiceElements&) = delete;
+
+    /// \brief Default move: moves the three maps; does NOT update skeleton back-references.
+    /// The owning SkeletonBase must call UpdateSkeletonReferences(*this) after the move.
+    SkeletonServiceElements(SkeletonServiceElements&&) noexcept = default;
+    SkeletonServiceElements& operator=(SkeletonServiceElements&&) noexcept = default;
+
+    // ── Registry operations ───────────────────────────────────────────────────
```
LittleHuba: Please avoid this comment style. Use doxygen if the comment adds value and structure with empty lines or composition if necessary.
Such section comments are very prone to get outdated.

LittleHuba: Please fix in other places as well.

score/mw/com/impl/skeleton_service_elements_test.cpp
LittleHuba, line 1: I would have expected a cleanup of the tests of SkeletonBase. Since you moved functionality, it makes little sense to test it in both places now.

score/mw/com/impl/skeleton_service_elements_test.cpp
```diff
@@ -0,0 +1,291 @@
...
+
+    SkeletonServiceElements dest;
+    dest = std::move(source);
+
+    EXPECT_TRUE(source.GetEvents().empty());
+    ASSERT_EQ(dest.GetEvents().size(), 1U);
+    EXPECT_EQ(&dest.GetEvents().at("ev").get(), &event);
+}
+
+// ── Multiple elements ─────────────────────────────────────────────────────────
```
LittleHuba: `using SkeletonServiceElementsWithMultipleElementsFixture = SkeletonServiceElementsFixture` is a better way to section tests. This makes also GTest show correct sections.

rudresh-systream:
Created a new PR , and fixingthe PR comments in the new created PR https://github.com/eclipse-score/communication/pull/342

rudresh-systream:
@LittleHuba rebased it with main branch and this PR was 100+ commits behind and other files are also changed according to the main branch in my pr, so i kindly request you to take the https://github.com/eclipse-score/communication/pull/368
as this is updated with the review comments u have given, i will try to reduce this delay from next PR

LittleHuba:
Closing as superseeded by #368.

As stated in our chat conversation, opening a new PR to answer review comments on an existing PR is not appreciated by us as maintainers.
This leads to overhead on our side, which reduces the overall PRs we can review in time.

Please push to the branch of the existing PR in the future to update a PR after it was reviewed.

LittleHuba:
As requested by author, reopening the PR.

score/mw/com/impl/skeleton_base.cpp
rudresh-systream: Fixed. Replaced the banner-style // ── Move constructor ── block with proper /// \brief / /// \details doxygen comments.

score/mw/com/impl/skeleton_base.cpp
rudresh-systream: Fixed. The non-move copy line skeleton_mock_{other.skeleton_mock_} was a leftover from an earlier draft. Removed — skeleton_mock_ is now correctly std::moved along with all other members.

score/mw/com/impl/skeleton_base.cpp
rudresh-systream: Fixed. Function header converted to /// \brief / /// \details doxygen style. The // Suppress line immediately before the coverity annotation is intentionally kept as a plain // comment — the Coverity tool requires it on the line directly preceding the annotation.

score/mw/com/impl/skeleton_base.cpp
rudresh-systream: Thank you for the suggestion. A pointer parameter + forward declaration does resolve the C++ header include cycle, but skeleton_service_elements.cpp would still need to #include "skeleton_base.h" for the full type definition, which creates a circular Bazel dependency (skeleton_base → skeleton_service_elements → skeleton_base). Resolving that requires impl_deps / implementation_deps, which is not used anywhere in this project's BUILD files today — introducing it would be out of scope for this PR. The current placement in skeleton_base.cpp keeps the function private to SkeletonBase, called only from its move constructor and move assignment operator, and the Bazel dependency direction remains correct. Happy to pursue the impl_deps approach as a follow-up if the team decides to enable it project-wide.

score/mw/com/impl/skeleton_service_elements_test.cpp
rudresh-systream: Fixed. Removed RegisteringServiceElementStoresReferenceInMap and MoveConstructingUpdatesReferencesToServiceElements from skeleton_base_test.cpp — both are now fully covered by skeleton_service_elements_test.cpp. Kept MoveAssigningUpdatesReferencesToServiceElements (has SCR-17432438 requirement traceability and tests the full SkeletonBase move-assign path) and MoveAssigningToItselfDoesNotDoAnything (tests the self-assignment guard in SkeletonBase, not covered in SSE tests). Both remaining test suites pass with --cache_test_results=no.

score/mw/com/impl/skeleton_service_elements_test.cpp
rudresh-systream: Fixed. Added using SkeletonServiceElementsWithMultipleElementsFixture = SkeletonServiceElementsFixture; and the multi-element tests now use that alias. GTest will display the correct fixture name as a distinct section in test output.

score/mw/com/impl/skeleton_service_elements.h
rudresh-systream: Fixed. All // ── section banner ── style comments replaced with per-method /// \brief doxygen comments throughout skeleton_service_elements.h. Applied consistently across all added methods and the class-level doc block.

rudresh-systream:
@LittleHuba  as stated, resolved the comments , pls check

score/mw/com/impl/skeleton_service_elements.h
```diff
@@ -0,0 +1,105 @@
...
+///          1. All map management (Register / Update / Get) is in one place.
+///          2. SkeletonBase's move constructor and move assignment operator each only need to
+///             call UpdateAllServiceElementReferences() after delegating the member-move to the
+///             compiler-generated default move of this class.
+///
+/// The move constructor and move assignment operator of this class are intentionally defaulted:
+/// they simply move the three maps.  The caller (SkeletonBase) is responsible for calling
+/// UpdateAllServiceElementReferences() after the move, because the correct new-owner address
+/// is only available at the SkeletonBase level.
+class SkeletonServiceElements
```
crimson11: After looking deeper into it, I feel, that introducing this wrapper class has no added value.
Within `SkeletonBase` you are still (in the update logic) accessing through into the members (`events_`, `fields_`, ...). So this wrapper/encapsulation imho brings nothing.

score/mw/com/impl/skeleton_base.cpp
```diff
@@ -330,4 +317,28 @@ score::cpp::optional<InstanceIdentifier> GetInstanceIdentifier(const InstanceSpe
...
 
+/// \brief Re-points every registered service element back-reference to this SkeletonBase.
+///
+/// \details This method is intentionally defined in skeleton_base.cpp rather than in
+///          skeleton_service_elements.cpp because it requires the complete definition of
+///          SkeletonBase.  Including skeleton_base.h in skeleton_service_elements.cpp would
+///          create a circular dependency, since skeleton_base already depends on
+///          skeleton_service_elements.  The pointer approach suggested in code review is
+///          equivalent in effect but does not resolve the Bazel dependency cycle.
+void SkeletonBase::UpdateAllServiceElementReferences() noexcept
```
crimson11, line 295: Introducing this method makes sense, to avoid code-duplication in move-ctor/move-assign-op.
But it doesn't need the `SkeletonServiceElements` wrapper. See my comment there.

score/mw/com/impl/skeleton_base.h
```diff
@@ -131,81 +149,47 @@ class SkeletonBaseView
...
     void UpdateEvent(const std::string_view event_name, SkeletonEventBase& event) noexcept
     {
-        // Suppress "AUTOSAR C++14 A15-4-2" rule finding. This rule states: "If a function is declared to be
-        // noexcept, noexcept(true) or noexcept(<true condition>), then it shall not exit with an exception"
-        // This function is indirectly called by the move constructor and as we register the event name in the
-        // events_ container during SkeletonEvent construction so we are sure the event name already exists, so no
-        // way for throwing std::out_of_range which leds to std::terminate().
-        // coverity[autosar_cpp14_a15_4_2_violation] : FALSE]
-        skeleton_base_.events_.at(event_name) = event;
+        skeleton_base_.service_elements_.UpdateEvent(event_name, event);
```
crimson11: If you follow my proposal to remove the `SkeletonServiceElements` wrapper again ... you will roughly recover this code ... then pls double-check, whether the asymmetry between `UpdateEvent` and `UpdateField`/`UpdateMethod` (still/really) makes sense ... i.e. the "old" comment, that in case of "event" we are sure, that event-name exists ... raising my attention. Out of my head I would say, that there is no difference in the case of field/method, thus the code should be symmetric! It currently already always leads to a `terminate()` ...
The extensive logging is also overdone. So, if we now refactor anyhow, I would prefer to have a simple

```
auto field_name_it = skeleton_base_.fields_.find(field_name);
SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(field_name_it != skeleton_base_.fields_.cend(), "SkeletonBaseView::UpdateField failed to update field because the requested field doesn't exist");
        if (field_name_it == skeleton_base_.fields_.cend())
```

score/mw/com/impl/skeleton_base_test.cpp
```diff
@@ -720,65 +720,6 @@ class SkeletonBaseServiceElementReferencesFixture : public ::testing::Test
     SkeletonMethodBase method_1_{skeleton_, method_name_1_, std::make_unique<mock_binding::SkeletonMethod>()};
 };
 
-TEST_F(SkeletonBaseServiceElementReferencesFixture, RegisteringServiceElementStoresReferenceInMap)
```
crimson11, line 726: Why was this massive refactoring of the tests needed? The API/signature of `SkeletonBase`/ `SkeletonBaseView` didn't change? So - were the tests wrong or missing?

crimson11 (changes requested):
The commit-history is massive and hard to understand. Please rebase, leave out any merge-commits.
For the extent of this change I would not expect more than 2/3 commits.

score/mw/com/impl/skeleton_service_elements.h
rudresh-systream: Agreed. I removed the SkeletonServiceElements wrapper.

After rechecking the design, the wrapper did not add useful encapsulation because SkeletonBase still needs to work directly with events_, fields_, and methods_ for service offering, stop-offering, tracing callback setup, binding validation, and post-move reference fix-up.

The updated solution keeps these maps directly in SkeletonBase and removes skeleton_service_elements.h/.cpp and the related test target.

score/mw/com/impl/skeleton_base.cpp
rudresh-systream: Done. I kept the useful part of the refactor: UpdateAllServiceElementReferences().

Both the move constructor and move assignment operator now call this helper after moving the service-element maps. This removes the duplicated UpdateSkeletonReference loops without introducing the SkeletonServiceElements wrapper.

score/mw/com/impl/skeleton_base.h
rudresh-systream: Done. I removed the wrapper usage and made UpdateEvent, UpdateField, and UpdateMethod symmetric.

All three now use find(), SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(), and the same defensive std::terminate() fallback when the requested service element does not exist. This also removes the earlier asymmetry where UpdateEvent used std::map::at() while field and method had different handling.

score/mw/com/impl/skeleton_base_test.cpp
rudresh-systream: Agreed. Since the public SkeletonBase / SkeletonBaseView API did not change, the large test refactor was unnecessary.

I removed the separate SkeletonServiceElements tests and kept the coverage at the SkeletonBase level, where the behavior is actually observable: registration, move construction, move assignment, and service-element back-reference fix-up.

rudresh-systream:
> The commit-history is massive and hard to understand. Please rebase, leave out any merge-commits. For the extent of this change I would not expect more than 2/3 commits.

Thank you @crimson11  for reviewing the PR and giving the valuable reviews, i resolved all your review comments and also squashed the commits

score/mw/com/impl/skeleton_base.h
```diff
@@ -151,30 +153,28 @@ class SkeletonBaseView
...
-        // This function is indirectly called by the move constructor and as we register the event name in the
-        // events_ container during SkeletonEvent construction so we are sure the event name already exists, so no
-        // way for throwing std::out_of_range which leds to std::terminate().
-        // coverity[autosar_cpp14_a15_4_2_violation] : FALSE]
-        skeleton_base_.events_.at(event_name) = event;
+        auto event_name_it = skeleton_base_.events_.find(event_name);
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(
+            event_name_it != skeleton_base_.events_.cend(),
+            "SkeletonBaseView::UpdateEvent failed to update event because the requested event doesn't exist");
+        if (event_name_it == skeleton_base_.events_.cend())
```
crimson11: This is redundant?
Line 159 already does the check/assert and will terminate in case assert fails.

score/mw/com/impl/skeleton_base.h
```diff
@@ -151,30 +153,28 @@ class SkeletonBaseView
...
+        event_name_it->second = event;
     }
 
     void UpdateField(const std::string_view field_name, SkeletonFieldBase& field) noexcept
     {
         auto field_name_it = skeleton_base_.fields_.find(field_name);
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE(
+            field_name_it != skeleton_base_.fields_.cend(),
+            "SkeletonBaseView::UpdateField failed to update field because the requested field doesn't exist");
         if (field_name_it == skeleton_base_.fields_.cend())
```
crimson11: see above - check/assert/terminate is already done by `SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE`

score/mw/com/impl/skeleton_base.h
```diff
@@ -151,30 +153,28 @@ class SkeletonBaseView
                                                 "Method cannot be registered as it already exists.");
     }
 
-    // Suppress "AUTOSAR C++14 A15-5-3" rule findings. This rule states: "The std::terminate() function shall not be
-    // called implicitly". std::terminate() is implicitly called from std::map::at in case the container doesn't
-    // have the mapped element. This function is called by the move constructor and as we register the event name in
-    // the events_ container during SkeletonEvent construction so we are sure that the event_name already exists, so
-    // no way for throwing std::out_of_range which leds to std::terminate().
-    // coverity[autosar_cpp14_a15_5_3_violation : FALSE]
     void UpdateEvent(const std::string_view event_name, SkeletonEventBase& event) noexcept
```
crimson11: Now just got aware: I think we can shift the implementations of the methods of `SkeletonBaseView` into the .cpp file. Can you clean this up?
None of these funcs/methods is used in our "hot paths". So there is no need for them being inlined as an optimization ...

crimson11 (changes requested):
I had some minor comments left ...
and btw - you have still 7 commits or so ;)
In the end I would like to see two commits in this PR: One for shifting SkeletonBaseView implementations to .cpp (see my separate comment) and the 2nd commit for the functional extensions you did.

score/mw/com/impl/skeleton_base.h
rudresh-systream: Removed the redundant defensive if (... == cend()) { std::terminate(); } checks after SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE in the SkeletonBaseView::Update* methods.

score/mw/com/impl/skeleton_base.h
rudresh-systream: Moved the SkeletonBaseView method implementations from skeleton_base.h into skeleton_base.cpp, since these methods are not hot-path relevant and do not need to stay inline.

score/mw/com/impl/skeleton_base.h
rudresh-systream: Kept the functional extraction through SkeletonBase::UpdateAllServiceElementReferences(), so the duplicated post-move service-element reference update logic is centralized

rudresh-systream:
@crimson11  Updated the PR based on the latest review comments.

Changes included:

Removed the redundant defensive checks after SCORE_LANGUAGE_FUTURECPP_ASSERT_MESSAGE.
Moved the SkeletonBaseView method implementations from skeleton_base.h to skeleton_base.cpp.
Kept the extracted UpdateAllServiceElementReferences() helper to centralize the post-move service-element back-reference update logic.
Squashed the branch history into a single clean commit.

Validation:

bazel build //score/mw/com/impl:skeleton_base passed.
bazel test //score/mw/com/impl:skeleton_base_test passed.
bazel test //score/mw/com/impl/methods:skeleton_method_test passed.
bazel test //score/mw/com/impl/methods:skeleton_method_base_test passed.
bazel test //score/mw/com/test/concurrent_skeleton_creation/integration_test:concurrent_skeleton_creation passed.

crimson11 (dismissed):

crimson11 (dismissed):

crimson11 (approved):

PR #262, opened 2026-03-31
Delete .github/ISSUE_TEMPLATE directory

anmittag:
to disable local issue types


castler (changes requested):
Sorry, we do not really see why our issue templates should be deleted. Especially as your commit message does not explain any alternatives?

For us it is important to get certain information (for example for bugs), which gets hinted towards the user with these issue templates.

Thus, we see no reason why to remove them.

PR #268, opened 2026-04-02
Brem move transaction log set

bemerybmw:


score/mw/com/impl/bindings/lola/transaction_log.h
crimson11, line 56: Why is this an optional?
`subscription_max_sample_count_` should always have a `value >= 1`, No?

score/mw/com/impl/bindings/lola/transaction_log_local_view_test.cpp
```diff
@@ -61,6 +61,36 @@ class TransactionLogFixture : public ::testing::Test
     StrictMock<MockFunction<void(TransactionLog::MaxSampleCountType)>> unsubscribe_callback_{};
 };
 
+TEST_F(TransactionLogFixture, SubscriptionTransactionUpdatesPointedToTransactionLog)
+{
+    // Given a TransactionLogLocalView pointing to a valid TransactionLog
+
+    // When calling SubscribeTransactionBegin on the TransactionLogLocalView
+    EXPECT_FALSE(transaction_log_.subscribe_transactions_.GetTransactionBegin());
```
crimson11: // expect, that it initially has no transaction-begin and transaction-end

score/mw/com/impl/bindings/lola/transaction_log_local_view_test.cpp
```diff
@@ -61,6 +61,36 @@ class TransactionLogFixture : public ::testing::Test
...
+}
+
+TEST_F(TransactionLogFixture, ReferenceTransactionUpdatesPointedToTransactionLog)
+{
+    // Given a TransactionLogLocalView pointing to a valid TransactionLog
+
+    // When calling ReferenceTransactionBegin on the first slot in the TransactionLogLocalView
+    constexpr std::size_t slot_index{0U};
+    auto& slot = transaction_log_.reference_count_slots_.at(slot_index);
+    EXPECT_FALSE(slot.GetTransactionBegin());
```
crimson11: document expectation here?

score/mw/com/impl/bindings/lola/transaction_log_local_view_test.cpp
```diff
@@ -61,6 +61,36 @@ class TransactionLogFixture : public ::testing::Test
...
+    EXPECT_FALSE(transaction_log_.subscribe_transactions_.GetTransactionEnd());
+}
+
+TEST_F(TransactionLogFixture, ReferenceTransactionUpdatesPointedToTransactionLog)
+{
+    // Given a TransactionLogLocalView pointing to a valid TransactionLog
+
+    // When calling ReferenceTransactionBegin on the first slot in the TransactionLogLocalView
+    constexpr std::size_t slot_index{0U};
+    auto& slot = transaction_log_.reference_count_slots_.at(slot_index);
```
crimson11: const?

score/mw/com/impl/bindings/lola/transaction_log_set.h
crimson11, line 280: After looking at the code for some time ... it feels "wrong" to have the `skeleton_tracing_transaction_log_` stored in here ... in the long-run I guess it makes more sense to only store the N proxy related logs within a `TransactionLogSet` and rename it to `ProxyTransactionLogSet` and keeping the skeleton transaction log separate... then we can also get rid of the sentinel-value for a skeleton transaction log and so on ... nothing to do for now ... but maybe a @ToDo for later?

score/mw/com/design/partial_restart/transaction_log_model.puml
```diff
@@ -21,8 +21,8 @@ class "lola::TransactionLog" {
 }
 
 class "lola::TransactionLogSet" {
-  +RollbackProxyTransactions(const TransactionLogId&, TransactionLog::DereferenceSlotCallback,TransactionLog::UnsubscribeCallback) : ResultBlank
-  +RollbackSkeletonTracingTransactions(TransactionLog::DereferenceSlotCallback) : ResultBlank
+  +RollbackProxyTransactions(const TransactionLogId&, TransactionLogLocalView::DereferenceSlotCallback,TransactionLogLocalView::DereferenceSlotCallback) : ResultBlank
+  +RollbackSkeletonTracingTransactions(TransactionLogLocalView::DereferenceSlotCallback) : ResultBlank
```
crimson11: But this central PUML class diagram is now vastly outdated? https://www.plantuml.com/plantuml/proxy?src=https://raw.githubusercontent.com/eclipse-score/communication/refs/heads/main/score/mw/com/design/events_fields/event_lola_model.puml

score/mw/com/impl/bindings/lola/skeleton_service_data_control_local_view.h
crimson11: In the context of this change/commit ... `SkeletonEventDataControlLocalView -> roviderEventControlLocalView` ... I ask myself, whether we shouldn't also rename `SkeletonServiceDataControlLocalView` -> `ProviderServiceDataControlLocalView` ? Why we didn't?

score/mw/com/impl/bindings/lola/transaction_log_registration_guard.cpp
```diff
@@ -68,33 +68,34 @@ TransactionLogRegistrationGuard::TransactionLogRegistrationGuard(
...
-        is_movable_or_copyable<ProxyEventDataControlLocalView<>>::value;
-    static_assert(!is_proxy_event_data_control_local_view_movable_or_copyable,
-                  "unregister_on_destruction_operation_ takes a reference to the ProxyEventDataControlLocalView, so we "
-                  "need to make sure that > is not movable or copyable to avoid dangling references.");
+    constexpr bool is_consumer_event_data_control_local_view_movable_or_copyable =
+        is_movable_or_copyable<ConsumerEventDataControlLocalView<>>::value;
+    static_assert(
+        !is_consumer_event_data_control_local_view_movable_or_copyable,
+        "unregister_on_destruction_operation_ takes a reference to the ConsumerEventDataControlLocalView, so we "
+        "need to make sure that > is not movable or copyable to avoid dangling references.");
```
crimson11: > that > is not movable
->
that **it** is not mov**e**able

score/mw/com/impl/bindings/lola/provider_event_data_control_local_view.cpp
```diff
@@ -46,6 +46,15 @@ template <template <class> class AtomicIndirectorType>
 auto ProviderEventDataControlLocalView<AtomicIndirectorType>::AllocateNextSlot() noexcept
     -> std::optional<SlotIndexType>
 {
+    auto log_performance_metrics = [](std::uint64_t& retry_counter) {
```
crimson11: since this lamda doesn't capture anything from the local scope, I would move it outside this function (hot path) instead of recreate it in every call?

score/mw/com/impl/bindings/lola/provider_event_data_control_local_view_test.cpp
```diff
@@ -177,6 +180,26 @@ TEST_F(ProviderEventDataControlLocalViewFixture, CanAllocateOneSlotWhenReference
     EXPECT_EQ(slot.value(), 1);
 }
 
+TEST_F(ProviderEventDataControlLocalViewFixture, CannotAllocateSlotIfExceedingMaximumRetriesDueToContention)
+{
+    GivenAProviderEventDataControlLocalViewUsingMockedAtomics(kMaxSlots);
+
+    // and given that load returns that the slot is in ready for writing
```
crimson11: this wording did confuse me a bit:
> the slot is in ready for writing
We have a defined state, which is `IN_WRITING` ... this you don't mean here.
You expect a state, which is:

- either INVALID (not used and also never initialized)
- or UNUSED (once initialized, but currently not used)

So you decided in your test to go the 2nd option -> UNUSED (timestamp 1 and 0 refcount)

So I would prefer:
 // and given that load returns that the slot is currently unused

score/mw/com/impl/bindings/lola/event_data_control_composite.cpp
```diff
@@ -71,69 +70,29 @@ auto EventDataControlComposite<AtomicIndirectorType>::GetNextFreeMultiSlot() con
         const EventSlotStatus slot_qm{
             asil_qm_control_local_.get().state_slots_[slot_index].load(std::memory_order_acquire)};
         const EventSlotStatus slot_b{asil_b_control_local_->state_slots_[slot_index].load(std::memory_order_acquire)};
-        if (slot_b.IsInvalid() || ((slot_qm.IsUsed() == false) && (slot_b.IsUsed() == false)))
+
+        // This situation normally could never happen because if we allocate a slot, we will _always_ record the
+        // allocation in the ASIL-B control section, marking the slot as _not_ invalid. However, if the QM consumer has
+        // misbehaved and modified the control memory region then the qm slot could be _not_ invalid. In this case, we
+        // exit early and will mark the qm consumer as misbehaving.
```
crimson11: > // exit early and will mark the qm consumer as misbehaving.

->
// exit early with an empty optional, thus the caller will mark the qm consumer as misbehaving.

score/mw/com/impl/bindings/lola/event_data_control_composite.h
```diff
@@ -128,9 +127,8 @@ class EventDataControlComposite
     bool ignore_qm_control_;
 
     // Algorithms that operate on multiple control blocks
-    std::optional<SlotWithTimeStamp> GetNextFreeMultiSlot() const noexcept;
+    std::optional<MultiSlotInfo> GetNextFreeMultiSlot() const noexcept;
```
crimson11: Where is the "added value" in this signature change? I mean `MultiSlotInfo` would always contain the same index for qm_slot_info/asil_b_slot_info? You gave-up the timestamp info ... if this was never needed -> fine ...

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -115,13 +112,13 @@ class SkeletonMemoryManager final
...
+    /// \brief Opens an EventControl for QM and optionally for ASIL-B (if the Skeleton is ASIL-B) for a specific
+    /// event that were created by a previous skeleton.
     ///
-    /// The EventDataControlComposite and TransactionLogSet are retrieved from the ServiceDataControl in the shared
-    /// memory region that was opened with OpenExistingSharedMemory.
-    auto OpenEventDataControlCompositeAndTransactionLogSetFromOpenedSharedMemory(const ElementFqId element_fq_id)
-        -> std::pair<EventDataControlComposite<>, std::reference_wrapper<EventControl>>;
+    /// The EventControls are retrieved from the ServiceDataControl in the shared memory regions that were opened with
+    /// OpenExistingSharedMemory.
+    auto OpenEventControlsFromOpenedSharedMemory(const ElementFqId element_fq_id)
```
crimson11: Nitpicky:
`OpenEventControlsFromOpenedSharedMemory` -> `RetrieveEventControlsFromOpenedSharedMemory`

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/transaction_log.h
bemerybmw: needed for unsubscribe in eventsubscriptioncontrol so we know how many slots are no longer used, so it needs to be set only when we record a subscription transaction.

score/mw/com/impl/bindings/lola/transaction_log_local_view_test.cpp
bemerybmw: IMO, this wouldn't improve the readability of the test. This is basically a sanity check that the initial state is as expected and is related to the main expectation of the test i.e. that the transaction begin is only true _after_ calling subscribe. expect is used for mock expectations.

score/mw/com/impl/bindings/lola/transaction_log_local_view_test.cpp
bemerybmw: i don't think so. the important part is that we're calling ReferenceTransactionBegin on the first slot in the TransactionLogLocalView. We also have some sanity checks but I don't think that should be documented. same reason as https://github.com/eclipse-score/communication/pull/268#discussion_r3131690074

score/mw/com/impl/bindings/lola/skeleton_service_data_control_local_view.h
bemerybmw: The reason was that these really were only on skeleton / proxy side, we didn't have the proxy one on skeleton side like the eventdatacontrollocalview. but anyway, it's now removed and the proxy one will be removed soon.

score/mw/com/impl/bindings/lola/transaction_log_registration_guard.cpp
bemerybmw: moveable and movable are both correct. movable is used in this file so i'll leave it

score/mw/com/design/partial_restart/transaction_log_model.puml
bemerybmw: True, I guess a lot of diagrams will need to be updated and should add some explanations about design decisions.

score/mw/com/impl/bindings/lola/transaction_log_local_view_test.cpp
bemerybmw: Will change to assert and will move to given part.

score/mw/com/impl/bindings/lola/transaction_log_set.h
bemerybmw: Manuel will create a ticket.

score/mw/com/impl/bindings/lola/event_data_control_composite.h
bemerybmw: GetNextFreeMultiSlot returns SlotInfo, GetNextFreeMultiSlot asserts that qm and asil b values are the same.

score/mw/com/impl/bindings/lola/transaction_log_set.h
crimson11: Ticket-258982

score/mw/com/design/partial_restart/transaction_log_model.puml
crimson11: Ticket-258986

score/mw/com/impl/bindings/lola/provider_event_data_control_local_view.h
```diff
@@ -87,47 +92,42 @@ class SkeletonEventDataControlLocalView final
...
 
   private:
     /// \brief Finds oldest unused slot within control slots, if there is any.
     /// \return if an unused slot is found, returns its index, otherwise, an empty optional is returned.
-    std::optional<SlotIndexType> FindOldestUnusedSlot() const noexcept;
+    std::optional<ProviderEventDataControlLocalView::SlotInfo> FindOldestUnusedSlot() const noexcept;
 
-    LocalEventControlSlots state_slots_;
+    /// \brief Logs performance metrics for slot allocation attempts.
+    void LogPerformanceMetrics(std::uint64_t retry_counter) noexcept
```
crimson11: Why did you put it in the header? It's only used in the .cpp file - so if the compiler wants to inline it ... it could do so?

crimson11 (changes requested):

crimson11 (approved):

PR #274, opened 2026-04-08
Miscellaneous cleanup

limdor:


LittleHuba (approved):
Discussed offline. Minor follow-up to add comments explaining the structuring of the MODULE file needed. Since the cleanup is ongoing, this will be done in a follow-up PR.

PR #276, opened 2026-04-08
Rust::com General Optimization and cleanup of RUST-API

bharatGoswami8:
- cleanup for internal type naming of rust apis
- Updated init async flag
- added NonNull type for internal types of consumer and producer

### DIFF
--- a/score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
@@ -335,7 +335,7 @@ impl<T: CommData + Debug> Subscriber<T, LolaRuntimeImpl> for SubscribableImpl<T>
             max_num_samples,
             instance_info,
             waker_storage: Arc::default(),
-            async_init_status: AtomicBool::new(false),
+            async_init_status: std::sync::Once::new(),
             _proxy: self.proxy_instance.clone(),
             _phantom: PhantomData,
         })
@@ -441,7 +441,7 @@ where
     max_num_samples: usize,
     instance_info: LolaConsumerInfo,
     waker_storage: Arc<AtomicWaker>,
-    async_init_status: AtomicBool,
+    async_init_status: std::sync::Once,
     _proxy: ProxyInstanceManager,
     _phantom: PhantomData<T>,
 }
@@ -455,9 +455,8 @@ impl<T: CommData + Debug> Drop for SubscriberImpl<T> {
         // and then unsubscribe from the event to clean up resources on the C++ side.
         let mut guard = self.event.get_proxy_event();
         unsafe {
-            if self
-                .async_init_status
-                .load(std::sync::atomic::Ordering::Relaxed)
+            if self.async_init_status.is_completed()
+            // Check if the async receive callback was initialized
             {
                 bridge_ffi_rs::clear_event_receive_handler(guard.deref_mut(), T::ID);
             }
@@ -564,18 +563,11 @@ where
             // on the same subscriber instance.
             let mut event_guard = self.event.get_proxy_event();
             // Initialize the async receive callback only once when the first receive call is made
-            if !self
-                .async_init_status
-                .load(std::sync::atomic::Ordering::Relaxed)
-            {
-                if let Err(_e) = self.init_async_receive(&mut event_guard) {
-                    return Err(Error::ReceiveError(
-                        ReceiveFailedReason::InitializationFailed,
-                    ));
-                }
-                self.async_init_status
-                    .store(true, std::sync::atomic::Ordering::Relaxed);
-            }
+            // We are using std::sync::Once to ensure that the callback is set only once.
+            self.async_init_status.call_once(|| {
+                self.init_async_receive(&mut event_guard)
+                    .expect("Failed to initialize async receive callback");
+            });
             ReceiveFuture {
                 event_guard: Some(event_guard),
                 waker_storage: Arc::clone(&self.waker_storage),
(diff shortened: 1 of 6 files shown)


rpreddyhv (approved):
looks okay to me.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -564,18 +557,11 @@ where
...
-                if let Err(_e) = self.init_async_receive(&mut event_guard) {
-                    return Err(Error::ReceiveError(
-                        ReceiveFailedReason::InitializationFailed,
-                    ));
-                }
-                self.async_init_status
-                    .store(true, std::sync::atomic::Ordering::Relaxed);
-            }
+            // We are using std::sync::Once to ensure that the callback is set only once.
+            self.async_init_status.call_once(|| {
```
arvindr19, line 561: `std::sync::Once::call_once` is a blocking, thread-parking primitive. Calling it inside an async block violates the fundamental async contract: futures must never block the executor thread.

suggestion: OnceLock (lock-free, non-blocking) or keep previous `AtomicBool`

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -259,23 +258,18 @@ impl NativeProxyEventBase {
     pub fn new(proxy: *mut ProxyBase, interface_id: &str, identifier: &str) -> Result<Self> {
```
arvindr19: NonNull instead of *mut ?

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
```diff
@@ -338,18 +339,16 @@ impl NativeSkeletonEventBase {
...
                 instance_info.interface_id,
                 identifier,
             )
         };
-        if skeleton_event_ptr.is_null() {
-            return Err(Error::ProducerError(
-                ProducerFailedReason::SkeletonCreationFailed,
-            ));
-        }
+        let skeleton_event_ptr = std::ptr::NonNull::new(raw_event_ptr).ok_or(
```
arvindr19, line 349: line 310 used ok_or_else, here ok_or, seems inconsistent. please check other places as well.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
```diff
@@ -338,18 +339,16 @@ impl NativeSkeletonEventBase {
...
+            Error::ProducerError(ProducerFailedReason::SkeletonCreationFailed),
```
arvindr19, line 350: SkeletonCreationFailed? is this correct?

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
```diff
@@ -287,7 +291,7 @@ impl std::fmt::Debug for SkeletonInstanceManager {
 /// And the lifetime is managed correctly
 /// As it has Send and Sync unsafe impls, it must not expose any mutable access to the skeleton handle
 pub struct NativeSkeletonHandle {
-    pub handle: *mut SkeletonBase,
+    pub handle: NonNull<SkeletonBase>,
```
arvindr19: pub? comment say `it must not expose any mutable access to the skeleton handle`

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Async receive enforces single-threaded access via ProxyEventManager's in_progress acquire/release lock, concurrent calls panic immediately. call_once executes only once on first receive() call, blocking for microseconds (FFI callback setup only) because init should be done before returning future.
And also `sync::once` give clear indication to developer this is one time call.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
bharatGoswami8: yes

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Yes why not, we can take as `&NonNull`

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
bharatGoswami8: Yes it was only in this line updated.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
bharatGoswami8: Thank you for pointing out this. Looks like in producer side few places still pub keyword was there for field.

arvindr19 (approved):
LTGM :)

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
```diff
@@ -370,10 +369,10 @@ impl std::fmt::Debug for NativeSkeletonEventBase {
 
 #[derive(Debug)]
 pub struct Publisher<T> {
-    pub identifier: String,
-    pub skeleton_event: NativeSkeletonEventBase,
-    pub _data: PhantomData<T>,
-    pub _skeleton_instance: SkeletonInstanceManager,
+    _identifier: String,
```
darkwisebear: This one doesn't seem to be used. Why is it there?

bharatGoswami8: Yes we are directly passing identifier to FFI call and after that we are not using, removed now.

darkwisebear (approved):

castler (approved):

PR #277, opened 2026-04-08
Changes related to service method integration tests

gdadunashvili:
These are changes that are related to integration tests for methods.
Either bugfixes that were discovered through methods tests, or infrastructure code that will be used by integration tests

### DIFF
--- a/score/mw/com/test/common_test_resources/command_line_parser.cpp
+++ b/score/mw/com/test/common_test_resources/command_line_parser.cpp
@@ -1,0 +1,62 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#include "score/mw/com/test/common_test_resources/command_line_parser.h"
+#include <iostream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+namespace score::mw::com::test
+{
+
+auto ParseCommandLineArguments(int argc,
+                               const char** argv,
+                               const std::vector<std::pair<std::string, std::string>>& param_names)
+    -> CommandLineArgsMapType
+{
+    namespace po = boost::program_options;
+    po::options_description options;
+    options.add_options()("help,h", "display the help message");
+    for (const auto& [param_name, param_doc] : param_names)
+    {
+        options.add_options()(
+            param_name.data(), po::value<std::string>(), ("specify " + std::string{param_doc}).data());
+    }
+
+    boost::program_options::variables_map args;
+    const auto parsed_args =
+        po::command_line_parser{argc, argv}
+            .options(options)
+            .allow_unregistered()
+            .style(po::command_line_style::unix_style | po::command_line_style::allow_long_disguise)
+            .run();
+    po::store(parsed_args, args);
+    po::notify(args);
+
+    CommandLineArgsMapType provided_args{};
+
+    for (const auto& [key, value] : args)
+    {
+        provided_args[key] = args[key].as<std::string>();
+    }
+
+    if (args.count("help") > 0U)
+    {
+        std::cerr << options << '\n';
+    }
+
+    return provided_args;
+}
+
+}  // namespace score::mw::com::test
--- a/score/mw/com/test/common_test_resources/command_line_parser.h
+++ b/score/mw/com/test/common_test_resources/command_line_parser.h
@@ -41,9 +41,9 @@ using CommandLineArgsMapType = std::unordered_map<std::string, std::string>;
 template <typename ReturnType>
 auto GetValueIfProvided(const CommandLineArgsMapType& args, const std::string& arg_string) -> Result<ReturnType>
 {
-    std::string error_msg = "could not find the requested parameter: " + arg_string;
     if (args.count(arg_string) == 0U)
     {
+        std::string error_msg = "could not find the requested parameter: " + arg_string;
         return score::MakeUnexpected<ReturnType>(
             MakeError(TestErrorCode::kParsingCommandLineArgumentFailed, error_msg));
     }
--- a/score/mw/com/test/common_test_resources/command_line_parser.h
+++ b/score/mw/com/test/common_test_resources/command_line_parser.h
@@ -1,0 +1,133 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#ifndef SCORE_MW_COM_TEST_COMMON_TEST_RESOURCES_COMMAND_LINE_PARSER_H
+#define SCORE_MW_COM_TEST_COMMON_TEST_RESOURCES_COMMAND_LINE_PARSER_H
+
+#include "score/mw/com/test/common_test_resources/test_error_domain.h"
+#include <boost/program_options.hpp>
+
+#include <charconv>
+#include <score/optional.hpp>
+
+#include "score/result/result.h"
+#include <string>
+#include <type_traits>
+#include <vector>
+
+namespace score::mw::com::test
+{
+
+using CommandLineArgsMapType = std::unordered_map<std::string, std::string>;
+
+/// \brief Get a value represented as a string, from an arguments map and parse it into the appropriate type
+///
+/// \tparam ReturnType Target type of the parsed argument. Must be std::string, integral or floating point type.
+/// \param args Map of provided command line arguments, where the key is the argument name and the value is the argument
+/// value as a string.
+/// \param arg_string The name of the argument to retrieve and parse from the map.
+/// \return The parsed argument value in case of success, or an error if the argument was not provided or could not be
+/// parsed.
+template <typename ReturnType>
+auto GetValueIfProvided(const CommandLineArgsMapType& args, const std::string& arg_string) -> Result<ReturnType>
+{
+    std::string error_msg = "could not find the requested parameter: " + arg_string;
+    if (args.count(arg_string) == 0U)
+    {
+        return score::MakeUnexpected<ReturnType>(
+            MakeError(TestErrorCode::kParsingCommandLineArgumentFailed, error_msg));
+    }
+
+    std::string value_str = args.at(arg_string);
+
+    if constexpr (std::is_same_v<ReturnType, std::string>)
+    {
+        return value_str;
+    }
+    else if constexpr (std::is_same_v<ReturnType, bool>)
+    {
+        if (value_str == "false" || value_str == "0")
+        {
+            return false;
+        }
+        if (value_str == "true" || value_str == "1")
+        {
+            return true;
+        }
+        return score::MakeUnexpected<ReturnType>(
+            MakeError(TestErrorCode::kParsingCommandLineArgumentFailed,
+                      "Failed during parsing of: " + arg_string + " . Provided value " + value_str +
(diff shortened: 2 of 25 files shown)


score/mw/com/types.h
```diff
@@ -89,6 +89,11 @@ using SamplePtr = impl::SamplePtr<SampleType>;
 template <typename SampleType>
 using SampleAllocateePtr = impl::SampleAllocateePtr<SampleType>;
 
+/// \api
+/// A pointer type which carries a pointer to the method return value in shared memory.
+template <typename ReturnType>
+using MethodReturnTypePtr = impl::MethodReturnTypePtr<ReturnType>;
```
bemerybmw, line 96: Why not also MethodInArgTypePtr?

score/mw/com/test/common_test_resources/command_line_parser.cpp
```diff
@@ -0,0 +1,55 @@
...
+#include <vector>
+
+namespace score::mw::com::test
+{
+auto ParseCommandLineArguments(int argc,
+                               const char** argv,
+                               const std::vector<std::pair<std::string, std::string>>& param_names)
+    -> CommandLineArgsMapType
+{
+    // uid is needed internally by sctftestrunner
```
bemerybmw: Remove?

score/mw/com/test/common_test_resources/command_line_parser.cpp
```diff
@@ -0,0 +1,55 @@
...
+    namespace po = boost::program_options;
+    po::options_description options;
+    options.add_options()("help,h", "display the help message");
+    for (const auto& [param_name, param_doc] : param_names)
+    {
+        options.add_options()(
+            param_name.data(), po::value<std::string>(), ("specify " + std::string{param_doc}).data());
+    }
+
+    po::variables_map args;
```
bemerybmw: This should use the CommandLineArgsMapType alias.

score/mw/com/test/common_test_resources/command_line_parser.h
```diff
@@ -0,0 +1,68 @@
...
+#include <type_traits>
+#include <vector>
+
+namespace score::mw::com::test
+{
+
+using CommandLineArgsMapType = boost::program_options::variables_map;
+
+template <typename ParsedType>
+auto GetValueIfProvided(const boost::program_options::variables_map& args, const std::string& arg_string)
```
bemerybmw: please add docstrings to these functions and mention when they should be used.

score/mw/com/test/common_test_resources/command_line_parser.cpp
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#include "score/mw/com/test/common_test_resources/command_line_parser.h"
+#include <vector>
+
+namespace score::mw::com::test
+{
+auto ParseCommandLineArguments(int argc,
```
bemerybmw, line 23: This code is extracted from SctfTestRunner? IMO, SctfTestRunner should be updated to use this, otherwise we have two ways of parsing command line arguments. And if we then want to for example remove the boost dependency in the future, we only have that dependency in this target and so only have to refactor this code and not the SctfTestRunner.

score/mw/com/test/common_test_resources/command_line_parser.h
```diff
@@ -0,0 +1,68 @@
...
+    -> Result<ParsedType>
+{
+    std::string error_msg = "could not find the requested parameter: " + arg_string;
+    if constexpr (std::is_same_v<ParsedType, std::string>)
+    {
+        return (args.count(arg_string) > 0U) ? args[arg_string].as<std::string>()
```
bemerybmw: This check is duplicated in each if. Can't we just check if args.count(arg_string) == 0U once at the start?

score/mw/com/test/common_test_resources/command_line_parser.h
```diff
@@ -0,0 +1,68 @@
...
+#include <string>
+#include <type_traits>
+#include <vector>
+
+namespace score::mw::com::test
+{
+
+using CommandLineArgsMapType = boost::program_options::variables_map;
+
+template <typename ParsedType>
```
bemerybmw: Instead of templating with ParsedType and then checking within the function that it's a string, int or float, why not template with:
enum class ParsedType
{
    INVALID,
    INT,
STRING,
FLOAT
}

score/mw/com/test/common_test_resources/fail_test.h
```diff
@@ -0,0 +1,60 @@
...
+    }
+    else
+    {
+        fail_test_(std::move(strstr));
+    }
+}
+}  // namespace detail
+
+template <typename... Args>
+void fail_test(Args&&... args)
```
bemerybmw: Please add function docs

score/mw/com/test/common_test_resources/fail_test.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include <sstream>
+#include <utility>
+
+namespace score::mw::com::test
+{
+
+namespace detail
+{
+void fail_test_(std::stringstream&& strstr);
```
bemerybmw: Why lowercase and trailing underscore? Why not FailTest or FailTestImpl or something like that?

score/mw/com/test/common_test_resources/fail_test.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+#include <cstdlib>
+#include <iostream>
+
+namespace score::mw::com::test::detail
+{
+
+void fail_test_(std::stringstream&& strstr)
+{
+    std::move(strstr) << "\033[0m  \033[1m\033[41mTEST FAILED\033[0m\n";
```
bemerybmw: Why are you moving here?

bemerybmw (changes requested):
The final merge commit can be removed?

score/mw/com/types.h
gdadunashvili: done

score/mw/com/test/common_test_resources/command_line_parser.h
gdadunashvili: Doing this is not possible since the return type can not be deduced anymore. Also since I changed the implementation to from_chars there are not internal branches anymore

score/mw/com/test/common_test_resources/fail_test.cpp
gdadunashvili: I was mistaken

score/mw/com/test/common_test_resources/fail_test.h
gdadunashvili: changed it

score/mw/com/test/common_test_resources/command_line_parser.cpp
gdadunashvili: I addressed this issue

score/mw/com/test/common_test_resources/command_line_parser.h
```diff
@@ -0,0 +1,139 @@
...
+{
+    INVALID,
+    STRING,
+    INTEGRAL,
+    FLOATINGPOINT
+};
+
+using CommandLineArgsMapType = std::unordered_map<std::string, std::string>;
+
+/// \brief Get a value reperesednted as a string, from an arguments map and parse it into the appropriate type
```
bemerybmw:
```suggestion
/// \brief Get a value represented as a string, from an arguments map and parse it into the appropriate type
```

score/mw/com/test/common_test_resources/command_line_parser.cpp
bemerybmw: Still not done?

score/mw/com/test/common_test_resources/command_line_parser.h
```diff
@@ -0,0 +1,139 @@
...
+///
+/// \tparam ParsedType Target type of the parsed argument. Must be std::string, integral or floating point type.
```
bemerybmw:
```suggestion
/// \tparam ReturnType Target type of the parsed argument. Must be std::string, integral or floating point type.
```

score/mw/com/test/common_test_resources/command_line_parser.cpp
gdadunashvili: Now CommandLineArgsMapType is a std::unordered map and boost command line options does not escape the body of this function

bemerybmw: Still it would make sense to use the alias here. Otherwise if you change the alias type, you would have to change it here.

gdadunashvili: ok this comment refers to the new `provided_args` variables defined later. I updated its type to the alias

score/mw/com/test/common_test_resources/command_line_parser.h
```diff
@@ -0,0 +1,132 @@
...
+/// \tparam ReturnType Target type of the parsed argument. Must be std::string, integral or floating point type.
+/// \param args Map of provided command line arguments, where the key is the argument name and the value is the argument
+/// value as a string.
+/// \param arg_string The name of the argument to retrieve and parse from the map.
+/// \return The parsed argument value in case of success, or an error if the argument was not provided or could not be
+/// parsed.
+template <typename ReturnType>
+auto GetValueIfProvided(const CommandLineArgsMapType& args, const std::string& arg_string) -> Result<ReturnType>
+{
+    std::string error_msg = "could not find the requested parameter: " + arg_string;
```
bemerybmw: Minor point: error_msg is created even if there's no error. Should be inside the if condition.

score/mw/com/test/common_test_resources/fail_test.h
```diff
@@ -0,0 +1,68 @@
...
+    }
+    else
+    {
+        FailTest(std::move(strstr));
+    }
+}
+}  // namespace detail
+
+/// \brief Fail a test by exiting the program with EXIT_FAILURE and printing a message to stderr.
+/// \tparam Args variadic number of argument types, each one must be streamable to a standard stringstream.
```
bemerybmw: Note for future: we don't need to add docs for every single thing, only where it adds value. So here for example, I would only add a docstring for Args or args.

score/mw/com/test/common_test_resources/sctf_test_runner.cpp
```diff
@@ -29,52 +35,66 @@ namespace
...
 
-std::string ParameterToString(const Parameters parameter)
+template <typename ParsedType, typename SavedType = ParsedType>
+auto ParseAndPackage(const std::unordered_map<std::string, std::string>& args, const std::string& name)
+    -> score::cpp::optional<SavedType>
+{
+    auto result = GetValueIfProvided<ParsedType>(args, name);
+    if (!result.has_value())
+    {
+        return score::cpp::optional<SavedType>{};
```
bemerybmw, line 45: Log error?

score/mw/com/test/common_test_resources/sctf_test_runner.cpp
```diff
@@ -29,52 +35,66 @@ namespace
 using Parameters = SctfTestRunner::RunParameters::Parameters;
 
 template <typename ParsedType, typename SavedType = ParsedType>
-score::cpp::optional<SavedType> GetValueIfProvided(const boost::program_options::variables_map& args,
-                                                   std::string arg_string)
+score::cpp::optional<SavedType> GetValueIfProvided_old(const boost::program_options::variables_map& args,
```
bemerybmw: Can be removed?

score/mw/com/test/common_test_resources/sctf_test_runner.cpp
```diff
@@ -234,84 +254,23 @@ SctfTestRunner::RunParameters SctfTestRunner::ParseCommandLineArguments(
     std::vector<Parameters> allowed_parameters_with_uid(allowed_parameters.begin(), allowed_parameters.end());
     allowed_parameters_with_uid.push_back(Parameters::UID);
 
-    namespace po = boost::program_options;
-
-    po::options_description options;
-    // clang-format off
-    options.add_options()
-        ("help,h", "Display the help message");
+    std::vector<std::pair<std::string, std::string>> prameter_descriptor_paris{{"help", "Display the help message"}};
```
bemerybmw:
```suggestion
    std::vector<std::pair<std::string, std::string>> prameter_descriptor_pairs{{"help", "Display the help message"}};
```

score/mw/com/test/common_test_resources/fail_test.h
gdadunashvili: I got carried away here 😆

bemerybmw (dismissed):

crimson11 (dismissed):

bemerybmw (approved):

PR #283, opened 2026-04-10
lola: Initialize method InArgs and Return values on creation

bemerybmw:
Previously, we were creating the type erased buffer for the InArgs and
Return values and then reinterpreting the buffer as InArgs / Return
values. This is UB and also means that non-trivially default
constructible types were not properly initialized. We now explicitly
call the constructors of each of these args.

### DIFF
--- a/score/mw/com/impl/methods/proxy_method_test.cpp
+++ b/score/mw/com/impl/methods/proxy_method_test.cpp
@@ -67,10 +67,10 @@ class ProxyMethodTestFixture : public ::testing::Test
     {
         ProxyMethodBindingFactory<MethodType>::InjectMockBinding(&proxy_method_binding_factory_mock_);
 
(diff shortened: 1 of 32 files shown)


score/mw/com/impl/methods/proxy_method_test.cpp
```diff
@@ -127,15 +141,18 @@ template <typename T>
 using ProxyMethodWithoutInArgsTestFixture = ProxyMethodTestFixture<T>;
 TYPED_TEST_SUITE(ProxyMethodWithoutInArgsTestFixture, WithoutInArgs, );
 
-template <typename T>
```
crimson11, line 130: Why has this been removed?

crimson11 (approved):

score/mw/com/impl/methods/proxy_method_test.cpp
bemerybmw: It's unused

PR #289, opened 2026-04-13
Provide simple integration demonstration

### LINKED ISSUE #287: download_utils marked as dev_dependency in v0.1.4 but used at module level — breaks transitive dependents
### Description

In score_communication@0.1.4, download_utils was changed to dev_dependency = True, but two use_repo_rule calls that reference @download_utils remain at module level (not guarded by dev_dependency). This causes bazel mod deps to fail for any module that depends on score_communication.

https://github.com/eclipse-score/logging/pull/86

### Error Occurrence  Rate

None

### How to reproduce

_No response_

### Supporting Information

_No response_

### Classification

major

### Affected Version

0.5

### Expected Closure Version

0.5

### Category

- [ ] Safety Related
- [ ] Security Related

castler:
Fixes https://github.com/eclipse-score/communication/issues/287


examples/README.md
```diff
@@ -1,137 +1,5 @@
...
-
-## Configuration
-
-The communication behavior is configured via `score/mw/com/example/ipc_bridge/etc/mw_com_config.json`:
-
-- Service type definitions and bindings
-- Event definitions with IDs
-- Instance-specific configuration (shared memory settings, subscriber limits)
-- ASIL level and process ID restrictions
+Please note these are the first steps. Are mor elaborated guide will be added in future.
```
LittleHuba:
```suggestion
Please note these are the first steps. A more elaborated guide will be added in future.
```

examples/.bazelrc
```diff
@@ -0,0 +1,6 @@
+common --registry=https://raw.githubusercontent.com/eclipse-score/bazel_registry/refs/heads/main/
+common --registry=https://bcr.bazel.build
+
+common --@score_baselibs//score/json:base_library=nlohmann
+
+common --@score_communication//score/mw/com/flags:tracing_library=@score_baselibs//score/analysis/tracing/generic_trace_library/stub_implementation
```
LittleHuba: Missing newline

.github/workflows/build_and_test_host.yml
```diff
@@ -105,3 +105,7 @@ jobs:
         - name: Bazel test communication targets
           run: |
             bazel test ${{ matrix.config }} //... --build_tests_only
+        - name: Bazel test integration as module
+          run: |
+            # for the beginning, we do this only with the default configuration
+            cd example && bazel build //... && bazel mod deps --lockfile_mode=update
```
LittleHuba: Missing newline

examples/.bazelrc
```diff
@@ -0,0 +1,6 @@
+common --registry=https://raw.githubusercontent.com/eclipse-score/bazel_registry/refs/heads/main/
+common --registry=https://bcr.bazel.build
+
+common --@score_baselibs//score/json:base_library=nlohmann
```
limdor, line 4: This will be tested with the C++ toolchain autogenerated from Bazel with whatever is detected in the system. Even though we still want to support the autodetected toolchains when possible, I would build the example in the CI with the same hermetic toolchain that we use one folder up.
To make sure we do not accidentally use it, you can add also here the changes from https://github.com/eclipse-score/communication/pull/291/changes/eae02a60d06e7fc7aa190df2afb4c608c11fd364

examples/README.md
```diff
@@ -1,137 +1,5 @@
 # Examples
 
-This directory contains aliases to example applications demonstrating the S-CORE Communication middleware API.
+This directory contains example applications demonstrating the integration of S-CORE Communication middleware API.
 
-## Available Examples
```
limdor, line 5: To not loose this content, it would make sense to create a README.md file in https://github.com/eclipse-score/communication/tree/e291a1ba502a6f4b4c17c492399a202829fbd348/score/mw/com/example with the content we had here.

.github/workflows/build_and_test_host.yml
```diff
@@ -105,3 +105,7 @@ jobs:
         - name: Bazel test communication targets
           run: |
             bazel test ${{ matrix.config }} //... --build_tests_only
+        - name: Bazel test integration as module
+          run: |
+            # for the beginning, we do this only with the default configuration
+            cd examples && bazel build //... && bazel mod deps --lockfile_mode=update
```
limdor, line 111: With this change, when building the upper level with `//...` it will still try to build the examples folder (even though it is a different module). Not so sure why this is happening, in the past AFAIK if a nested folder had another WORKSPACE file then the full folder was ignored.
Even worst, if you first built examples from inside, then when you are in the upper directory and try to build `//...` then it will try to build packages inside `//examples/bazel-*`.
The only way that I know to solve this problem is to add
```
common --deleted_packages=examples
```
This is something that is done also in rules_python
https://github.com/bazel-contrib/rules_python/blob/main/.bazelrc.deleted_packages

limdor (commented):
Just some minor things.
Because fixing MODULE.bazel and having a module integration test is the top priority and this PR achieves that, I would say that the comments can be addressed in a follow up PR.

limdor (approved):

.github/workflows/build_and_test_host.yml
limdor: Fixed with https://github.com/eclipse-score/communication/pull/293/changes

PR #295, opened 2026-04-13
enforcing lock file

balajiyalam2:


castler:
Converting to draft as this has conflicts.

In general - thank you for your contribution. And indeed this is a next step that we have to go and want to go.
But it would be great if you try to have more descriptive PR messages. Also it would be interesting to come up with an idea, how we ensure that in future we do not just forget in the CI to set this flag.

Thus, I am wondering if an action, to write a custom RC file with the lock-file mode set to error, would make more sense? Or do you have other ideas how todo avoid human mistake in future?

PR #296, opened 2026-04-14
Score: Add field method get/set Differentiation

KrishaDeshkool:


score/mw/com/impl/bindings/lola/methods/BUILD
```diff
@@ -119,6 +120,32 @@ cc_library(
...
 )
 
+cc_library(
+    name = "unique_method_identifier",
+    srcs = ["unique_method_identifier.cpp"],
+    hdrs = ["unique_method_identifier.h"],
+    features = COMPILER_WARNING_FEATURES,
+    tags = ["FFI"],
+    visibility = [
+        "//score/mw/com:__subpackages__",
```
KrishaDeshkool: single line

score/mw/com/impl/bindings/lola/methods/BUILD
```diff
@@ -119,6 +120,32 @@ cc_library(
...
+        "//score/mw/com/impl/configuration:lola_method_or_field_id",
+        "@score_logging//score/mw/log",
+    ],
+)
+
+cc_gtest_unit_test(
+    name = "unique_method_identifier_test",
+    srcs = ["unique_method_identifier_test.cpp"],
+    deps = [
+        ":unique_method_identifier",
```
KrishaDeshkool: single line

score/mw/com/impl/method_type.h
```diff
@@ -0,0 +1,53 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef PLATFORM_AAS_MW_COM_IMPL_METHOD_TYPE_H
```
KrishaDeshkool, line 13: This should be score not platform

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -617,11 +618,42 @@ void Proxy::UnregisterEventBinding(const std::string_view service_element_name)
...
+    // queue size > 1.
+    constexpr LolaMethodInstanceDeployment::QueueSize kFieldMethodQueueSize{1U};
+
+    const auto& lola_service_type_deployment = GetLoLaServiceTypeDeployment(handle_);
+    const auto& lola_service_instance_deployment = GetLoLaInstanceDeployment(handle_);
+    for (const auto& [field_name, field_instance_deployment] : lola_service_instance_deployment.fields_)
+    {
+        const auto field_id = GetServiceElementId<ServiceElementType::FIELD>(lola_service_type_deployment, field_name);
+
+        if (kUseGetIfAvailable && proxy_methods_.count({field_id, MethodType::kGet}) != 0U)
```
KrishaDeshkool, line 639: This is the binding layer map that would have registered itself when the proxy filed constructs the proxy methods inside it.

score/mw/com/impl/methods/proxy_method_with_in_args.h
```diff
@@ -125,7 +127,6 @@ template <typename... ArgTypes>
 ProxyMethod<void(ArgTypes...)>::ProxyMethod(ProxyMethod&& other) noexcept
     : ProxyMethodBase(std::move(other)), are_in_arg_ptrs_active_{std::move(other.are_in_arg_ptrs_active_)}
 {
-    // Since the address of this method has changed, we need update the address stored in the parent proxy.
```
KrishaDeshkool, line 128: removed by mistake add it back

score/mw/com/impl/method_identifier.h
```diff
@@ -0,0 +1,49 @@
...
+
+#include <string_view>
+#include <tuple>
+
+namespace score::mw::com::impl
+{
+
+/// \brief Lookup key for a method on a Proxy/Skeleton, formed from the method name and its MethodType.
+/// A field's Get, a field's Set and a regular method may share the same name the type field disambiguates them.
+struct UniqueMethodIdentifier
```
KrishaDeshkool, line 26: This should go away.
removed on later commits

score/mw/com/impl/method_type.h
```diff
@@ -0,0 +1,53 @@
...
+
+#include <cstdint>
+#include <string_view>
+
+namespace score::mw::com::impl
+{
+
+namespace detail
+{
+
```
crimson11, line 24: Maybe refactor into its own header: `field_tag_types.h` .
At least add a small comment, that these tag-types are used on ProxyField/SkeletonField level to accomplish overload-resolution for various signatures, which depend on the availability of Get/Set/Notifier.

score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
```diff
@@ -82,27 +85,32 @@ lola::TypeErasedCallQueue::TypeErasedElementInfo GetTypeErasedElementInfo(Handle
         in_arg_type_info = CreateDataTypeSizeInfoFromTypes<ArgTypes...>();
     }
 
-    auto queue_size = GetQueueSize(parent_handle, method_name_str);
+    // TODO : Field Get/Set methods use a fixed queue size of 1 for now, once it has been fully implemented, this should
```
crimson11: I would change this:
Imho we currently already have the general restriction, that we only support a queue size of 1. Because we do have synchronous/blocking method calls!
I looked at the configureation-schema. Interesstingly there we already support 1..255 for method-call queue-size. I did NOT check whether/where we WARN if the user configures a queue size > 1?

Bottom-line:
I would like you NOT to do the "special" handling here. Instead extend `GetQueueSize` with the additional `method_type` arg here and within `GetQueueSize` you do the short-cut/special handling to return 1 in case of Get/Set type ...

score/mw/com/impl/plumbing/proxy_field_binding_factory_impl.h
```diff
@@ -62,6 +70,24 @@ inline std::unique_ptr<ProxyEventBinding<SampleType>> ProxyFieldBindingFactoryIm
                                      ServiceElementType::FIELD>(parent, field_name);
 }
 
+template <typename SampleType>
+inline std::unique_ptr<ProxyMethodBinding> ProxyFieldBindingFactoryImpl<SampleType>::CreateGetMethodBinding(
+    ProxyBase& parent,
+    const std::string_view field_name) noexcept
+{
+    return ProxyMethodBindingFactory<SampleType()>::Create(
```
crimson11, line 85: Hm. Now I lost track. Why do we have this asymetric approach? `CreateEventBinding` some lines above dispatches to `CreateProxyServiceElement` ... but here we are dispatching to `ProxyMethodBindingFactory<SampleType()>::Create` ? So - to me it seems, that either `CreateProxyServiceElement` (and therefore entire `proxy_service_element_binding_factory_impl.h` should be removed OR we always dispatch to it ... also in the method case!?
A `proxy_service_element_binding_factory_impl.h`, which only handles events/fields but not methods makes no sense to me - it is rather confusing!
Can you discuss this with Brendan? I find our approach hard to understand right now ...?

score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
```diff
@@ -111,16 +119,26 @@ std::unique_ptr<ProxyMethodBinding> ProxyMethodBindingFactoryImpl<ReturnType(Arg
...
-        constexpr auto element_type{ServiceElementType::METHOD};
-
-        const auto lola_service_element_id = GetServiceElementId<element_type>(lola_type_deployment, method_name_str);
-        const lola::ElementFqId element_fq_id{
-            lola_type_deployment.service_id_, lola_service_element_id, lola_service_instance_id->GetId(), element_type};
+        LolaMethodOrFieldId lola_service_element_id{};
+        if (method_type == MethodType::kGet || method_type == MethodType::kSet)
+        {
+            lola_service_element_id =
+                GetServiceElementId<ServiceElementType::FIELD>(lola_type_deployment, method_name_str);
```
crimson11: How do you handle the differentiation between a Get and Set method here?
They will boil down exactly to the same `lola_service_element_id `. Is this intended?
I mean the enclosing function gets called WITH `method_type` ... so you could differ here ... but you do not?
So in case this Create-API gets called for a field-method (get/set), `method_name_str` contains the field name!
Maybe document it here - that is maybe noteworthy ...

... and then some lines down (line 137) you create a `ProxyMethodInstanceIdentifier` which additionally gets the `method_type` ... so the resulting `ProxyMethodInstanceIdentifier` does contain the get/set distinction?

So - maybe all is fine and is intended - then maybe add some small comments (see above), which make it a bit clearer.

score/mw/com/impl/traits.h
```diff
@@ -127,7 +127,9 @@ class ProxyTrait
     template <typename SampleType>
     using Event = ProxyEvent<SampleType>;
 
-    template <typename SampleType>
+    // Note : at the moment the get side implementation is not in the branch which means the skeleton and proxy side
```
crimson11: .. the SkeletonField::Get implementation is not ...

crimson11 (changes requested):

score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
KrishaDeshkool: i see it could be better documented,
we can already see now this `lola_service_element_id` is of type LolaMethodOrFieldId so it is intentional that it would have both method and field ids in this.
will add a clarifying comments.

score/mw/com/impl/plumbing/proxy_field_binding_factory_impl.h
KrishaDeshkool: ah! i see your point now. let me think about it

KrishaDeshkool: hmm! so i moved out only the part that is genuinely shared across all of them like validating the parent is a lola proxy, std visiting the deployment variant, and resolving the element id into an ElementFqId. removed this misleading proxy_service_element_binding_factory_impl header and then every factory events field notifier, and methods now starts with that one call and then does its own make_unique with whatever args its type needs.

score/mw/com/impl/method_type.h
KrishaDeshkool: i will just add a comment now once everything fits together (from other ppl PRs) then we can either move it out or keep it here since,

score/mw/com/impl/traits.h
KrishaDeshkool: done

score/mw/com/impl/configuration/lola_method_or_field_id.h
```diff
@@ -0,0 +1,25 @@
...
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_IMPL_CONFIGURATION_LOLA_METHOD_OR_FIELD_ID_H
+#define SCORE_MW_COM_IMPL_CONFIGURATION_LOLA_METHOD_OR_FIELD_ID_H
+
+#include <cstdint>
+
+namespace score::mw::com::impl
+{
+
+using LolaMethodOrFieldId = std::uint16_t;
```
crimson11: this needs to be either
`LolaEventOrFieldOrMethodId` ... or `LoLaServiceEelementId` ... which in case of LoLa are all uint16

score/mw/com/impl/plumbing/proxy_lola_element_lookup.h
```diff
@@ -0,0 +1,63 @@
...
+
+/// \brief Result of resolving a named service element against the lola deployment on the proxy side.
+/// \details Carries everything a binding constructor needs that comes from deployment/handle lookup.
+///          This is intentionally agnostic to what kind of binding is being built (event, field notifier,
+///          method); the construction step lives with the caller.
+struct LolaProxyElementLookup
+{
+    lola::Proxy& parent;
+    LolaServiceInstanceId::InstanceId instance_id;
+    LolaMethodOrFieldId element_id;
```
crimson11: see my comment in the type definition.
...but here you are redundant!?

`ElementFqId` already contains your newly introduced `LolaMethodOrFieldId`

What I propose:
- you keep your newly introduced `LolaMethodOrFieldId` but follow my comment in this type definition ->rename it.
- then in `ElementFqId` you fix `using ElementId = LolaEventId;` to `using ElementId = LoLaServiceElementId;`
- then you can remove `element_id` from here as `element_fq_id` contains everything.
Makes sense?

score/mw/com/impl/plumbing/proxy_lola_element_lookup.h
```diff
@@ -0,0 +1,63 @@
...
+#include <string_view>
+
+namespace score::mw::com::impl
+{
+
+/// \brief Result of resolving a named service element against the lola deployment on the proxy side.
+/// \details Carries everything a binding constructor needs that comes from deployment/handle lookup.
+///          This is intentionally agnostic to what kind of binding is being built (event, field notifier,
+///          method); the construction step lives with the caller.
+struct LolaProxyElementLookup
```
crimson11: Please rename to `LoLaProxyElementBuildingBlocks` or `LoLaProxyElementCreateParts` ... makes it more clear

score/mw/com/impl/plumbing/proxy_lola_element_lookup.h
```diff
@@ -0,0 +1,63 @@
...
+{
+    lola::Proxy& parent;
+    LolaServiceInstanceId::InstanceId instance_id;
+    LolaMethodOrFieldId element_id;
+    lola::ElementFqId element_fq_id;
+};
+
+/// \brief Validate the parent is a lola proxy and resolve the element name against the lola deployment.
```
crimson11: >against the lola deployment
-> against the lola deployment contained in the given handle.

crimson11 (changes requested):

score/mw/com/impl/configuration/lola_method_or_field_id.h
KrishaDeshkool: renamed to LoLaServiceEelementId

score/mw/com/impl/plumbing/proxy_lola_element_lookup.h
KrishaDeshkool: true, will update it

score/mw/com/impl/configuration/lola_method_or_field_id.h
KrishaDeshkool: rename and make this a base for all other aliases,

score/mw/com/impl/plumbing/proxy_lola_element_lookup.h
KrishaDeshkool: renamed to LoLaProxyElementBuildingBlocks

crimson11 (approved):

PR #305, opened 2026-04-15
GenericSkeleton: Fix writing to DynamicArray header instead it's data…

pawelrutkaq:
… region


crimson11:
Hi @pawelrutkaq  - I will answer in https://github.com/eclipse-score/communication/issues/261 ... these topics are connected. I think it makes no sense to merge your fix, because the `GenericSkeleton` is broken in several areas and needs some refactoring, which would render your change superflous ...

LittleHuba:
Closing as the fix being prepared in #261 is more complete. As @crimson11 stated, applying this patch without fixing the other problems in the GenericSkeleton would not improve the situation well enough.

PR #313, opened 2026-04-16
Suya add get and set and notification support to the fields in the configuration

soldier-sky:
Add useGetIfAvailable/useSetIfAvailable to LoLa field deployment

- Introduce dedicated LolaFieldInstanceDeployment class replacing the
  former LolaEventInstanceDeployment type alias, adding field-specific
  members use_get_if_available_ and use_set_if_available_.
- Fix mw_com_config_schema.json: move useGetIfAvailable and
  useSetIfAvailable into the field "properties" object so they are
  correctly validated when additionalProperties is false.
- Adapt existing test and add new tests for LolaFieldInstanceDeployment
- Update PUML diagram and readme document

Issue: SWP-250429

### DIFF
--- a/score/mw/com/design/configuration/structural_view.puml
+++ b/score/mw/com/design/configuration/structural_view.puml
@@ -68,13 +68,8 @@ package "configuration" {
     - number_of_sample_slots_ : std::optional<SampleSlotCountType>
     - number_of_tracing_slots_ : TracingSlotSizeType
     __
-    + LolaFieldInstanceDeployment(number_of_sample_slots : std::optional<SampleSlotCountType>, max_subscribers : std::optional<SubscriberCountType>, max_concurrent_allocations : std::optional<std::uint8_t>, enforce_max_samples : bool, number_of_tracing_slots : TracingSlotSizeType, use_get_if_available : bool, use_set_if_available : bool) noexcept
     {static} + CreateFromJson(json_object : const score::json::Object&) : LolaFieldInstanceDeployment
     + Serialize() const : score::json::Object
-    + SetNumberOfSampleSlots(number_of_sample_slots : SampleSlotCountType) noexcept : void
-    + GetNumberOfSampleSlots() const noexcept : std::optional<SampleSlotCountType>
-    + GetNumberOfSampleSlotsExcludingTracingSlot() const noexcept : std::optional<SampleSlotCountType>
-    + GetNumberOfTracingSlots() const noexcept : TracingSlotSizeType
     --
     <u>Type Aliases:</u>
     using SampleSlotCountType as std::uint16_t
--- a/score/mw/com/design/configuration/structural_view.puml
+++ b/score/mw/com/design/configuration/structural_view.puml
@@ -59,11 +59,27 @@ package "configuration" {
   }
 
   class "score::mw::com::impl::LolaFieldInstanceDeployment" {
-    + max_samples_: std::uint16_t
-    + max_subscribers_: std::uint8_t
-    + enforce_max_samples_ : score::cpp::optional<bool>
-    + is_set_configured: bool
-    + is_get_configured: bool
+    {static} + constexpr serializationVersion : std::uint32_t
+    + max_subscribers_ : std::optional<SubscriberCountType>
+    + max_concurrent_allocations_ : std::optional<std::uint8_t>
+    + enforce_max_samples_ : bool
+    + use_get_if_available_ : bool
+    + use_set_if_available_ : bool
+    - number_of_sample_slots_ : std::optional<SampleSlotCountType>
+    - number_of_tracing_slots_ : TracingSlotSizeType
+    __
+    + LolaFieldInstanceDeployment(number_of_sample_slots : std::optional<SampleSlotCountType>, max_subscribers : std::optional<SubscriberCountType>, max_concurrent_allocations : std::optional<std::uint8_t>, enforce_max_samples : bool, number_of_tracing_slots : TracingSlotSizeType, use_get_if_available : bool, use_set_if_available : bool) noexcept
+    {static} + CreateFromJson(json_object : const score::json::Object&) : LolaFieldInstanceDeployment
+    + Serialize() const : score::json::Object
+    + SetNumberOfSampleSlots(number_of_sample_slots : SampleSlotCountType) noexcept : void
+    + GetNumberOfSampleSlots() const noexcept : std::optional<SampleSlotCountType>
+    + GetNumberOfSampleSlotsExcludingTracingSlot() const noexcept : std::optional<SampleSlotCountType>
+    + GetNumberOfTracingSlots() const noexcept : TracingSlotSizeType
+    --
+    <u>Type Aliases:</u>
+    using SampleSlotCountType as std::uint16_t
+    using SubscriberCountType as std::uint8_t
+    using TracingSlotSizeType as std::uint8_t
   }
 
   class "score::mw::com::impl::LolaMethodInstanceDeployment" {
--- a/score/mw/com/impl/configuration/config_parser.cpp
+++ b/score/mw/com/impl/configuration/config_parser.cpp
@@ -79,6 +79,8 @@ constexpr auto kFieldEnforceMaxSamplesKey = "enforceMaxSamples"sv;
 constexpr auto kFieldMaxConcurrentAllocationsKey = "maxConcurrentAllocations"sv;
 constexpr auto kFieldUseGetIfAvailableKey = "useGetIfAvailable"sv;
 constexpr auto kFieldUseSetIfAvailableKey = "useSetIfAvailable"sv;
+constexpr auto kFieldUseGetIfAvailableDefault = false;
+constexpr auto kFieldUseSetIfAvailableDefault = false;
 constexpr auto kLolaShmSizeKey = "shm-size"sv;
 constexpr auto kLolaControlAsilBShmSizeKey = "control-asil-b-shm-size"sv;
 constexpr auto kLolaControlQmShmSizeKey = "control-qm-shm-size"sv;
@@ -488,10 +490,10 @@ auto ParseLolaFieldInstanceDeployment(const score::json::Object& json_map, LolaS
         const auto number_of_tracing_slots =
             deployment_parser.RetrieveJsonElement<NumberOfIpcTracingSlots_t>(kNumberOfIpcTracingSlotsKey)
                 .value_or(kNumberOfIpcTracingSlotsDefault);
-        const auto use_get_if_available =
-            deployment_parser.RetrieveJsonElement<bool>(kFieldUseGetIfAvailableKey).value_or(false);
-        const auto use_set_if_available =
-            deployment_parser.RetrieveJsonElement<bool>(kFieldUseSetIfAvailableKey).value_or(false);
+        const auto use_get_if_available = deployment_parser.RetrieveJsonElement<bool>(kFieldUseGetIfAvailableKey)
+                                              .value_or(kFieldUseGetIfAvailableDefault);
+        const auto use_set_if_available = deployment_parser.RetrieveJsonElement<bool>(kFieldUseSetIfAvailableKey)
+                                              .value_or(kFieldUseSetIfAvailableDefault);
 
         auto field_deployment = LolaFieldInstanceDeployment(number_of_sample_slots,
                                                             max_subscribers,
--- a/score/mw/com/impl/configuration/config_parser.cpp
+++ b/score/mw/com/impl/configuration/config_parser.cpp
@@ -77,6 +77,8 @@ constexpr auto kFieldNumberOfSampleSlotsKey = "numberOfSampleSlots"sv;
 constexpr auto kFieldMaxSubscribersKey = "maxSubscribers"sv;
 constexpr auto kFieldEnforceMaxSamplesKey = "enforceMaxSamples"sv;
 constexpr auto kFieldMaxConcurrentAllocationsKey = "maxConcurrentAllocations"sv;
+constexpr auto kFieldUseGetIfAvailableKey = "useGetIfAvailable"sv;
+constexpr auto kFieldUseSetIfAvailableKey = "useSetIfAvailable"sv;
 constexpr auto kLolaShmSizeKey = "shm-size"sv;
 constexpr auto kLolaControlAsilBShmSizeKey = "control-asil-b-shm-size"sv;
 constexpr auto kLolaControlQmShmSizeKey = "control-qm-shm-size"sv;
@@ -486,12 +488,18 @@ auto ParseLolaFieldInstanceDeployment(const score::json::Object& json_map, LolaS
         const auto number_of_tracing_slots =
             deployment_parser.RetrieveJsonElement<NumberOfIpcTracingSlots_t>(kNumberOfIpcTracingSlotsKey)
                 .value_or(kNumberOfIpcTracingSlotsDefault);
+        const auto use_get_if_available =
+            deployment_parser.RetrieveJsonElement<bool>(kFieldUseGetIfAvailableKey).value_or(false);
+        const auto use_set_if_available =
+            deployment_parser.RetrieveJsonElement<bool>(kFieldUseSetIfAvailableKey).value_or(false);
 
         auto field_deployment = LolaFieldInstanceDeployment(number_of_sample_slots,
                                                             max_subscribers,
                                                             kMaxConcurrentAllocationsDefault,
                                                             enforce_max_samples,
-                                                            number_of_tracing_slots);
+                                                            number_of_tracing_slots,
+                                                            use_get_if_available,
+                                                            use_set_if_available);
         const auto emplace_result = service.fields_.emplace(std::piecewise_construct,
                                                             std::forward_as_tuple(std::move(field_name_value)),
                                                             std::forward_as_tuple(field_deployment));
--- a/score/mw/com/impl/configuration/config_parser_test.cpp
+++ b/score/mw/com/impl/configuration/config_parser_test.cpp
@@ -2103,6 +2103,262 @@ TEST(ConfigParser, LolaFieldOptionalEnforceMaxSamples)
     EXPECT_EQ(deploymentInfo.fields_.at("CurrentTemperatureFrontLeft").enforce_max_samples_, false);
 }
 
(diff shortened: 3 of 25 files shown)


score/mw/com/impl/configuration/lola_field_instance_deployment.h
```diff
@@ -13,12 +13,79 @@
...
+
+#include <cstdint>
+#include <optional>
+#include <string>
 
 namespace score::mw::com::impl
 {
 
-using LolaFieldInstanceDeployment = LolaEventInstanceDeployment;
+class LolaFieldInstanceDeployment
```
bemerybmw, line 27: A field is a composition of an event and 0 to 2 methods. I would reflect this in the config object. i.e instead of dupcliating LolaEventInstanceDeployment here and adding field specific stuff, why not simply store a LolaEventInstanceDeployment as a member here. e.g.

```
LolaEventInstanceDeployment lola_event_instance_deployment_{};
bool use_get_if_available_{false};
// coverity[autosar_cpp14_m11_0_1_violation]
bool use_set_if_available_{false};
```

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -389,6 +377,18 @@
                                                 "default": 0,
                                                 "minimum": 0,
                                                 "maximum": 255
+                                            },
+                                            "useGetIfAvailable": {
+                                                "type": "boolean",
+                                                "title": "Use Field Getter If Available",
+                                                "description": "When true, the getter for this field will be used if the service type declares one.",
```
bemerybmw: "...if the service type declares one in the C++ service interface"

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -389,6 +377,18 @@
...
+                                                "default": false
+                                            },
+                                            "useSetIfAvailable": {
+                                                "type": "boolean",
+                                                "title": "Use Field Setter If Available",
+                                                "description": "When true, the setter for this field will be used if the service type declares one.",
```
bemerybmw: "...if the service type declares one in the C++ service interface"

score/mw/com/impl/configuration/lola_field_instance_deployment.h
```diff
@@ -13,12 +13,79 @@
...
+    // The struct is used as a config storage obtained by performing the parsing json object.
+    // Public access is more convenient to reach the following members of the struct.
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    std::optional<SubscriberCountType> max_subscribers_;
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    std::optional<std::uint8_t> max_concurrent_allocations_;
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    bool enforce_max_samples_;
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    bool use_get_if_available_{false};
```
bemerybmw: No need for this here. the value is initialized in the constructor.

score/mw/com/impl/configuration/lola_field_instance_deployment.h
```diff
@@ -13,12 +13,79 @@
...
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    bool use_set_if_available_{false};
```
bemerybmw: No need for this here. the value is initialized in the constructor.

score/mw/com/impl/configuration/config_parser.cpp
```diff
@@ -486,12 +488,18 @@ auto ParseLolaFieldInstanceDeployment(const score::json::Object& json_map, LolaS
         const auto number_of_tracing_slots =
             deployment_parser.RetrieveJsonElement<NumberOfIpcTracingSlots_t>(kNumberOfIpcTracingSlotsKey)
                 .value_or(kNumberOfIpcTracingSlotsDefault);
+        const auto use_get_if_available =
+            deployment_parser.RetrieveJsonElement<bool>(kFieldUseGetIfAvailableKey).value_or(false);
```
bemerybmw: the default value false should be a global constant (e.g. see kNumberOfIpcTracingSlotsDefault)

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
```diff
@@ -39,9 +39,15 @@ TEST_F(LolaFieldInstanceDeploymentFixture, CanCreateFromSerializedObjectWithoutO
     const std::optional<std::uint8_t> max_subscribers{13};
     const std::optional<std::uint8_t> max_concurrent_allocations{};
     const bool enforce_max_samples{true};
-
-    const LolaFieldInstanceDeployment unit{
-        MakeLolaFieldInstanceDeployment(max_samples, max_subscribers, max_concurrent_allocations, enforce_max_samples)};
+    const auto use_get_if_available{false};
```
bemerybmw: Make these true? More likely to catch if we're not passing the value to the data member

score/mw/com/impl/configuration/lola_service_instance_deployment_test.cpp
```diff
@@ -122,15 +122,15 @@ TEST(LolaServiceInstanceDeployment, ContainsEventReturnsFalseIfEventMissing)
 TEST(LolaServiceInstanceDeployment, ContainsFieldReturnsTrueIfEventPresent)
```
bemerybmw: ContainsFieldReturnsTrueIfFieldPresent

score/mw/com/impl/configuration/lola_service_instance_deployment_test.cpp
```diff
@@ -122,15 +122,15 @@ TEST(LolaServiceInstanceDeployment, ContainsEventReturnsFalseIfEventMissing)
...
 {
     LolaServiceInstanceDeployment unit{LolaServiceInstanceId{LolaServiceInstanceId{2U}}};
-    auto temp = MakeDefaultLolaEventInstanceDeployment();
+    auto temp = MakeLolaFieldInstanceDeployment();
     unit.fields_.emplace(std::make_pair("abc", temp));
     EXPECT_TRUE(unit.ContainsField("abc"));
 }
 
 TEST(LolaServiceInstanceDeployment, ContainsFieldReturnsFalseIfEventMissing)
```
bemerybmw: ContainsFieldReturnsFalseIfFieldMissing

score/mw/com/impl/tracing/configuration/tracing_filter_config.cpp
```diff
@@ -228,34 +228,41 @@ std::size_t FindNumberOfTracingSlots(
                 std::terminate();
             }
 
-            const auto& service_instance_map = [service_element_type, &lola_service_instance_deployment]() {
-                if (service_element_type == ServiceElementType::EVENT)
+            const auto service_element_name = service_element.service_element_name.data();
```
bemerybmw, line 231: Why did you change this? Isn't it adding duplicated code?

score/mw/com/impl/configuration/config_parser_test.cpp
```diff
@@ -2103,6 +2103,262 @@ TEST(ConfigParser, LolaFieldOptionalEnforceMaxSamples)
     EXPECT_EQ(deploymentInfo.fields_.at("CurrentTemperatureFrontLeft").enforce_max_samples_, false);
 }
```
bemerybmw, line 2107: Please also modify ParseExampleJson test

score/mw/com/impl/configuration/config_parser_test.cpp
```diff
@@ -2103,6 +2103,262 @@ TEST(ConfigParser, LolaFieldOptionalEnforceMaxSamples)
...
+                          "useGetIfAvailable": true
+                      }
+                  ]
+                }
+            ]
+        }
+    ]
+  }
+)"_json;
+    const auto config = score::mw::com::impl::configuration::Parse(std::move(j2));
```
bemerybmw, line 2164: PLease also add when, then sections based on our internal testing guideline

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
```diff
@@ -56,6 +63,106 @@ TEST_F(LolaFieldInstanceDeploymentFixture, CanCreateFromSerializedObjectWithoutO
...
+                                                                           kNumberOfTracingSlots,
+                                                                           /*use_get*/ true,
+                                                                           /*use_set*/ false)};
+
+    // When serialising and deserialising
+    const auto serialized_unit{unit.Serialize()};
+    const LolaFieldInstanceDeployment reconstructed_unit{serialized_unit};
+
+    // Then use_get_if_available is preserved and use_set_if_available is unaffected
+    EXPECT_TRUE(reconstructed_unit.use_get_if_available_);
```
bemerybmw: Better to extract values passed into unit constructor into variables and check the reconstructed value is the same. e.g. `EXPECT_EQ(reconstructed_unit.use_get_if_available_, is_get_enabled)`

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
```diff
@@ -56,6 +63,106 @@ TEST_F(LolaFieldInstanceDeploymentFixture, CanCreateFromSerializedObjectWithoutO
...
+    const LolaFieldInstanceDeployment reconstructed_unit{serialized_unit};
+
+    // Then both flags default to false
+    EXPECT_FALSE(reconstructed_unit.use_get_if_available_);
+    EXPECT_FALSE(reconstructed_unit.use_set_if_available_);
+}
+
+TEST_F(LolaFieldInstanceDeploymentFixture, UseGetAndUseSetFlagsAreIndependentOfEachOther)
+{
+    // Given four deployments covering all flag combinations
```
bemerybmw: this duplicates tests above? IMO, we just need these tests:
- UseGetIfAvailableIsTrueAfterRoundTripSerialisation
- UseSetIfAvailableIsTrueAfterRoundTripSerialisation

BothFlagsDefaultToFalseAfterRoundTripSerialisation is checked in CanCreateFromSerializedObjectWithoutOptionals

score/mw/com/design/configuration/structural_view.puml
```diff
@@ -59,11 +59,27 @@ package "configuration" {
   }
 
   class "score::mw::com::impl::LolaFieldInstanceDeployment" {
-    + max_samples_: std::uint16_t
-    + max_subscribers_: std::uint8_t
-    + enforce_max_samples_ : score::cpp::optional<bool>
-    + is_set_configured: bool
-    + is_get_configured: bool
+    {static} + constexpr serializationVersion : std::uint32_t
```
bemerybmw, line 62: The class diagrams are not meant to reflect the c++ code exactly. We only add as much detyail as is helpful. You can leave it as it was before IMO

bemerybmw (changes requested):

score/mw/com/design/configuration/structural_view.puml
soldier-sky: Yes LolaFieldInstanceDeployment appears to be too verbose. I've reduced boiler plate info and make it aligned to LolamethodInstanceDeployment.

score/mw/com/impl/configuration/config_parser.cpp
soldier-sky: corrected the default pattern

score/mw/com/impl/configuration/lola_field_instance_deployment.h
soldier-sky: I'd thought over this point and IMHO using composition approach will make coupling tight and LolaFieldInstanceDeployment will do more than 1 task (SRP breaks). Any changes in LolaEventInstanceDeployment e.g. member addition etc will appear in LolaFieldInstanceDeployment as well.  Also Serialization part cant be shared cleanly due to use of different JSON key (eventName vs fieldName), plus acess pattern will not be clean e.g. field.max_subscribers_  --->  field.event_.max_subscribers_ .

score/mw/com/impl/configuration/mw_com_config_schema.json
soldier-sky: Corrected the description

score/mw/com/impl/configuration/mw_com_config_schema.json
soldier-sky: Corrected the description

score/mw/com/impl/configuration/lola_field_instance_deployment.h
soldier-sky: Removed.

score/mw/com/impl/configuration/lola_service_instance_deployment_test.cpp
soldier-sky: Corrected the copy paste typo error.

score/mw/com/impl/tracing/configuration/tracing_filter_config.cpp
soldier-sky: Earlier LolaFieldInstanceDeployment was simply using alias of LolaEventInstanceDeployment and hence single lambda could handle the return type. With current state two distinct classes return different maps (events_ and fields_). Hence to for error handling duplicate logic. 

I'll rework on this part to optimize further.

score/mw/com/impl/configuration/config_parser_test.cpp
soldier-sky: Done. Added appropriate comments

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
soldier-sky: Updated the tests with use of local vars to validate against serialized <-> de-serialized values.

score/mw/com/impl/configuration/config_parser_test.cpp
soldier-sky: I've added assertions to ParseExampleJson for the defaults as example json omits the flags. Note:  dedicated tests for those flags (LolaFieldUseGetIfAvailableSetToTrue, LolaFieldUseSetIfAvailableSetToTrue, LolaFieldOmittingBothFlagsDefaultsBothToFalse) and hence only assertion test in ParseExampleJson with existing json file.

score/mw/com/impl/configuration/lola_field_instance_deployment.h
bemerybmw: These classes are tightly coupled either way. The question is just whether we duplicate all the code in LolaEventInstanceDeployment in LolaFieldInstanceDeployment. If we extend LolaEventInstanceDeployment then we want it to be also extended in LolaFieldInstanceDeployment because a field composes an event, and so this should be reflected in the configuration object IMO. 

>LolaFieldInstanceDeployment will do more than 1 task (SRP breaks)

What additional task is it doing when composing a LolaEventInstanceDeployment vs duplicating all the code within LolaFieldInstanceDeployment?

> Also Serialization part cant be shared cleanly due to use of different JSON key

IMO, this should not be the deciding factor about how we structure this code. Once we make a decision about the above point, we can discuss how the serialization code might look.

>plus acess pattern will not be clean e.g. field.max_subscribers_ ---> field.event_.max_subscribers_ .

IMO, that actually makes sense. I think it's exactly what we want to have something like:

```
field.event.max_subscribers_;
field.method.queue_size_;
...
```

@crimson11 what do you think about this?

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -389,6 +377,18 @@
                                                 "default": 0,
                                                 "minimum": 0,
                                                 "maximum": 255
+                                            },
+                                            "useGetIfAvailable": {
+                                                "type": "boolean",
+                                                "title": "Use Field Getter If Available",
+                                                "description": "When true, the getter for this field will be used if the service type declares one in the C++ service interface.",
```
crimson11: please annotate, that this property is only "relevant" for the proxy side/consumer side.
And should we re-think whether default is "false"? Maybe default should be "true" and you should explicitly disable the use?

And please also extend the `example/mw_com_confg.json` with the new properties.

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -389,6 +377,18 @@
...
+                                                "default": false
+                                            },
+                                            "useSetIfAvailable": {
```
crimson11, line 387: see above

score/mw/com/impl/configuration/lola_field_instance_deployment.h
crimson11: I'm also all in for "aggregation" ;)
Why: Because a field in mw::com IS already/explicitly an aggregate of a event plus optional methods. So why should this be "different" only on config level. ... and as you implied: It is the rule, that if we extend the event-config, we also need to extend the field-event config. I.e. updating the event-config and NOT updating the event-config part of a field would be an absolute exception to the rule.

As a side-note ... with the property naming `eventName`, `fieldName`, `methodName` ... we messed up imho :p - if it would be simpy `name` everywhere ... the serialization code would be generic ... but I guess it is too late already to change because our config structure is a "public interface", which is hard to change.

score/mw/com/impl/configuration/config_parser_test.cpp
crimson11: See my other comment ... I feel, that updating/extending the `example/mw_com_config.json` with some of the new properties makes sense. Why - because many people (including me) are sometimes too lazy to look up properties in the schema and check the `example/mw_com_config.json` instead ... it is more human-readable ;)

crimson11 (changes requested):

score/mw/com/impl/configuration/config_parser_test.cpp
soldier-sky: Updated the `example/mw_com_config.json` with setter/getter field values.

score/mw/com/impl/configuration/mw_com_config_schema.json
soldier-sky: Updated the description and changed default value to true i.e. now user need to explicitly flag getter/setter  false if given fields does not support them.

KrishaDeshkool:
This had been taken over and has been already merged,
just updating PR for closuer.

PR #318, opened 2026-04-17
Skeleton publishes method metainfo in shm

KrishaDeshkool:


KrishaDeshkool:
Putting this to draft since it was decided to park this ticket until we have a detailed design for this task

PR #322, opened 2026-04-20
Get enable methods list from config file

muhseth:


score/mw/com/impl/configuration/config_parser_methods_test.cpp
```diff
@@ -631,5 +631,202 @@ TEST_F(ConfigParserFixture, MethodQueueSizeCanBeSpecified)
     EXPECT_EQ(lola_deployment.methods_.at("SetPressure").queue_size_, 5);
 }
 
+TEST_F(ConfigParserFixture, MethodIsEnabledByDefaultWhenEnabledIsNotProvided)
```
bemerybmw: Can you test this change in `TEST_F(ConfigParserFixture, ParseExampleJson)`?

score/mw/com/impl/configuration/config_parser.cpp
```diff
@@ -523,7 +524,8 @@ auto ParseLolaMethodInstanceDeployment(const score::json::Object& json_map, Lola
         const auto& method_name = GetValueFromJson<std::string>(method_object, kMethodNameKey);
         const std::optional<LolaMethodInstanceDeployment::QueueSize> queue_size =
             GetOptionalValueFromJson<LolaMethodInstanceDeployment::QueueSize>(method_object, kMethodQueueSizeKey);
-        const LolaMethodInstanceDeployment method_deployment{queue_size};
+        const auto method_enabled = GetOptionalValueFromJson<bool>(method_object, kMethodEnabledKey);
+        const LolaMethodInstanceDeployment method_deployment{queue_size, method_enabled.value_or(true)};
```
bemerybmw: Default value should be a global constant.

score/mw/com/impl/configuration/lola_method_instance_deployment.h
```diff
@@ -39,8 +39,10 @@ class LolaMethodInstanceDeployment
      * @brief Construct LolaMethodInstanceDeployment with optional queue size, because LolaMethodInstanceDeployment for
      * a consumer will have a value while one for a provider will not.
      * @param queue_size The maximum number of pending method requests that can be queued.
+     * @param enabled Optional flag to disable/enable the method. Default value is true, which means the method is
+     * enabled.
      */
-    explicit LolaMethodInstanceDeployment(std::optional<QueueSize> queue_size);
+    explicit LolaMethodInstanceDeployment(std::optional<QueueSize> queue_size, bool enabled = true);
```
bemerybmw: The default value should be in the config parser only? Shouldn't this always be provided by the parser? Generally an optional in the config structs is for a param which is only used on skeleton side or proxy side but not both.

score/mw/com/impl/configuration/lola_method_instance_deployment.h
```diff
@@ -70,11 +72,12 @@ class LolaMethodInstanceDeployment
      * @brief The maximum number of method requests that can be queued on the server side.
      */
     std::optional<QueueSize> queue_size_{std::nullopt};
+    bool enabled_{true};
```
bemerybmw: This is initialized in the constructor. It shouldn't be default initialized here.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -731,29 +732,28 @@ memory::shared::SharedMemoryFactory::UserPermissions Proxy::GetSkeletonShmPermis
 }
 
 std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>
-Proxy::GetMethodIdAndQueueSizeFromNames(const std::vector<std::string_view>& enabled_method_names) const
+Proxy::GetMethodIdAndQueueSizeFromNames() const
```
bemerybmw: GetMethodIdAndQueueSizeForEnabledMethods

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -10,6 +10,7 @@
  *
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
+#include "gmock/gmock.h"
```
bemerybmw: Should be below with other system headers.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -154,11 +163,46 @@ class ProxyMethodHandlingFixture : public ProxyMockedMemoryFixture
 
     ProxyMethodHandlingFixture& GivenAProxy()
     {
-        InitialiseProxyWithConstructor(kConfigurationStore.GetInstanceIdentifier());
+        configuration_store_ = std::make_unique<ConfigurationStore>(CreateConfigurationStoreWithDisabledMethods());
```
bemerybmw: Why are methods disabled by default?

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -154,11 +163,46 @@ class ProxyMethodHandlingFixture : public ProxyMockedMemoryFixture
...
+        InitialiseProxyWithConstructor(configuration_store_->GetInstanceIdentifier());
         SCORE_LANGUAGE_FUTURECPP_ASSERT(proxy_ != nullptr);
         return *this;
     }
 
+    ProxyMethodHandlingFixture& WithEnabledMethodsInConfiguration(
```
bemerybmw: To disable all methods, we can simply pass an empty enabled_method_names? So we can remove CreateConfigurationStoreWithDisabledMethods completely? I would suggest:

have GivenAConfigurationWithEnabledMethods(const std::vector<std::string_view>& enabled_method_names). It creates the ConfigurationStore with methods enabled according to input vector. GivenAProxy() then creates the proxy using the ConfigurationStore provided in GivenAConfigurationWithEnabledMethods. 

GivenAProxy should then have an assertion that the ConfigurationStore was created by calling one of those methods.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -154,11 +163,46 @@ class ProxyMethodHandlingFixture : public ProxyMockedMemoryFixture
...
+            QualityType::kASIL_B,
+            kLolaServiceTypeDeploymentWithMethods,
+            LolaServiceInstanceDeployment{
+                kLolaInstanceId,
+                {},
+                {},
+                {{kDummyMethodName0,
+                  LolaMethodInstanceDeployment{
+                      kDummyQueueSize0,
+                      std::find(enabled_method_names.begin(), enabled_method_names.end(), kDummyMethodName0) !=
```
bemerybmw: extract this into a variable `is_dummy_method_0_enabled`. It's easy to miss this otherwise.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -154,11 +163,46 @@ class ProxyMethodHandlingFixture : public ProxyMockedMemoryFixture
...
+                      kDummyQueueSize1,
+                      std::find(enabled_method_names.begin(), enabled_method_names.end(), kDummyMethodName1) !=
+                          enabled_method_names.end()}},
+                 {kDummyMethodName2,
+                  LolaMethodInstanceDeployment{
+                      kDummyQueueSize2,
+                      std::find(enabled_method_names.begin(), enabled_method_names.end(), kDummyMethodName2) !=
+                          enabled_method_names.end()}}}}});
+
+        InitialiseProxyWithConstructor(configuration_store_->GetInstanceIdentifier());
```
bemerybmw: Should only be called in GivenAProxy()

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -872,22 +984,36 @@ TEST_P(ProxySetupMethodsShmSizeParamaterizedFixture, AllocatesEveryByteInSpecifi
...
     // When calling SetupMethods with a ProxyMethod name which does not correspond
     // to the registered ProxyMethod Then the program terminates
-    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(score::cpp::ignore = proxy_->SetupMethods({kDummyMethodName1}));
+    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(score::cpp::ignore = proxy_->SetupMethods());
 }
 
 TEST_F(ProxyMethodHandlingFixture, EnablingMethodsThatAreNotInTheConfigurationTerminates)
 {
-    GivenAProxy().GivenAMockedSharedMemoryResource();
+    const LolaServiceInstanceDeployment lola_service_instance_deployment_with_unknown_method{
```
bemerybmw, line 979: This is a completely different test. Please rename it. Should also have a test for the opposite situation i.e. its in LolaServiceInstanceDeployment but not in LolaServiceTypeDeployment?

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -872,22 +984,36 @@ TEST_P(ProxySetupMethodsShmSizeParamaterizedFixture, AllocatesEveryByteInSpecifi
 TEST_F(ProxyMethodHandlingFixture, EnablingMethodsThatWereNotRegisteredTerminates)
```
bemerybmw: This test no longer makes sense with this name. Please rename it.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
bemerybmw, line 1032: This configuration should enable the methods?

bemerybmw (changes requested):

score/mw/com/impl/configuration/lola_method_instance_deployment.h
muhseth: As I see it this object is created from parser + deserialization + tests/helpers.
Without a type default, each path has to duplicate fallback behavior.
Keeping it here preserves backward compatibility when enabled is missing

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
muhseth: remove this

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
muhseth: rename to GivenAConfigurationWithEnabledMethods

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
muhseth: creatr proxy with configuartionstore created in GivenAConfigurationWithEnabledMethods.

example test:
,
GivenAConfigurationWithEnabledMethods((method_name_0, method_name_1)).GivenAProxy()

muhseth: i have taken care of it as discussed.

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -413,6 +413,11 @@
                                                 "description": "Optional Binding specific consumer/proxy side setting for maximum number of pending method requests that can be queued on the server side before new requests are rejected.",
                                                 "minimum": 1,
                                                 "maximum": 255
+                                            },
+                                            "enabled": {
+                                                "type": "boolean",
+                                                "description": "Optional flag to disable/enable method. Default value is true, which means the method is enabled.",
```
bemerybmw: Please also specify that it's only relevant on proxy side. It will be ignored if specified in the configuration of a skeleton field.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -413,6 +413,7 @@ Proxy::Proxy(std::shared_ptr<memory::shared::ManagedMemoryResource> control,
              score::filesystem::Filesystem filesystem,
              ProxyInstanceIdentifier::ProxyInstanceCounter proxy_instance_counter) noexcept
     : ProxyBinding{},
+
```
bemerybmw: remove

score/mw/com/impl/configuration/lola_method_instance_deployment.h
```diff
@@ -39,8 +39,10 @@ class LolaMethodInstanceDeployment
      * @brief Construct LolaMethodInstanceDeployment with optional queue size, because LolaMethodInstanceDeployment for
      * a consumer will have a value while one for a provider will not.
      * @param queue_size The maximum number of pending method requests that can be queued.
+    * @param enabled Optional flag to disable/enable the method.
```
bemerybmw: Specify that it's always filled on proxy side and alwyas empty on skeleton side.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -861,33 +951,71 @@ TEST_P(ProxySetupMethodsShmSizeParamaterizedFixture, AllocatesEveryByteInSpecifi
...
 
     // When calling SetupMethods with a ProxyMethod name which does not correspond
     // to the registered ProxyMethod Then the program terminates
-    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(score::cpp::ignore = proxy_->SetupMethods({kDummyMethodName1}));
+    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(score::cpp::ignore = proxy_->SetupMethods());
+}
+
+TEST_F(ProxyMethodHandlingFixture, EnabledMethodPresentOnlyInInstanceDeploymentTerminates)
+{
+    const LolaServiceInstanceDeployment lola_service_instance_deployment_with_unknown_method{
```
bemerybmw, line 979: Please add a descriptive given comment.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -861,33 +951,71 @@ TEST_P(ProxySetupMethodsShmSizeParamaterizedFixture, AllocatesEveryByteInSpecifi
...
+    // When calling SetupMethods for a method that is enabled in the instance deployment but missing in the type
+    // deployment Then the program terminates
+    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(score::cpp::ignore = proxy_->SetupMethods());
 }
 
-TEST_F(ProxyMethodHandlingFixture, EnablingMethodsThatAreNotInTheConfigurationTerminates)
+TEST_F(ProxyMethodHandlingFixture, MethodPresentOnlyInTypeDeploymentDoesNotCreateSharedMemory)
 {
-    GivenAProxy().GivenAMockedSharedMemoryResource();
+    const LolaServiceInstanceDeployment lola_service_instance_deployment_without_method{kLolaInstanceId, {}, {}, {}};
```
bemerybmw, line 1002: Please fix given/when/then comments

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -744,29 +745,28 @@ memory::shared::SharedMemoryFactory::UserPermissions Proxy::GetSkeletonShmPermis
 }
 
 std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>
-Proxy::GetMethodIdAndQueueSizeFromNames(const std::vector<std::string_view>& enabled_method_names) const
+Proxy::GetMethodIdAndQueueSizeForEnabledMethods() const
 {
     std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>> method_data{};
```
bemerybmw: This should be renamed since we have a class called MethodData which is unrelated. I'd call it something like: enabled_method_ids_and_queue_sizes

bemerybmw (changes requested):

bemerybmw (approved):

PR #329, opened 2026-04-21
Rust::com fix for aarche64 gcc build for rust example file

bharatGoswami8:
* Added rustc flag for build
https://github.com/eclipse-score/communication/issues/328

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -35,6 +35,11 @@ common:linux_x86_64_score_gcc_12_2_0_posix --extra_toolchains=@score_gcc_x86_64_
 common:linux_aarch64_score_gcc_12_2_0_posix --platforms=@score_bazel_platforms//:aarch64-linux-gcc_12.2.0-posix
 common:linux_aarch64_score_gcc_12_2_0_posix --extra_toolchains=@score_gcc_aarch64_toolchain//:aarch64-linux
 common:linux_aarch64_score_gcc_12_2_0_posix --extra_toolchains=@score_toolchains_rust//toolchains/ferrocene:ferrocene_aarch64_unknown_linux_gnu
+# The score aarch64 GCC toolchain is built without --enable-default-pie, so static libs are non-PIC.
+# -no-pie disables PIE; -lstdc++ links the C++ stdlib; -lgcc provides aarch64 outline atomic helpers
+# (__aarch64_ldadd4_acq_rel etc.) from the toolchain's libgcc.a (sysroot's libgcc_s.so lacks them).
+# ticket - https://github.com/eclipse-score/toolchains_rust/issues/27
+common:linux_aarch64_score_gcc_12_2_0_posix --@rules_rust//:extra_rustc_flags=-Clink-arg=-no-pie,-Clink-arg=-lgcc
 
 # In order to build for QNX, you need:
 # * An account on qnx.com
--- a/examples/.bazelrc
+++ b/examples/.bazelrc
@@ -9,3 +9,5 @@ common --registry=https://bcr.bazel.build
 common --@score_baselibs//score/json:base_library=nlohmann
 
 common --@score_communication//score/mw/com/flags:tracing_library=@score_baselibs//score/analysis/tracing/generic_trace_library/stub_implementation
+
+common --@rules_rust//:extra_rustc_flags=-Clink-arg=-no-pie,-Clink-arg=-lgcc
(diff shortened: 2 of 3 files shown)


LittleHuba (commented):
This should also be added to the examples/.bazelrc

I would also ask that you open a bug ticket in the toolchain. Such workarounds should not be a longterm solution.

.bazelrc
```diff
@@ -35,6 +35,10 @@ common:linux_x86_64_score_gcc_12_2_0_posix --extra_toolchains=@score_gcc_x86_64_
 common:linux_aarch64_score_gcc_12_2_0_posix --platforms=@score_bazel_platforms//:aarch64-linux-gcc_12.2.0-posix
 common:linux_aarch64_score_gcc_12_2_0_posix --extra_toolchains=@score_gcc_aarch64_toolchain//:aarch64-linux
 common:linux_aarch64_score_gcc_12_2_0_posix --extra_toolchains=@score_toolchains_rust//toolchains/ferrocene:ferrocene_aarch64_unknown_linux_gnu
+# The score aarch64 GCC toolchain is built without --enable-default-pie, so static libs are non-PIC.
+# -no-pie disables PIE; -lstdc++ links the C++ stdlib; -lgcc provides aarch64 outline atomic helpers
+# (__aarch64_ldadd4_acq_rel etc.) from the toolchain's libgcc.a (sysroot's libgcc_s.so lacks them).
+common:linux_aarch64_score_gcc_12_2_0_posix --@rules_rust//:extra_rustc_flags=-Clink-arg=-no-pie,-Clink-arg=-lstdc++,-Clink-arg=-lgcc
```
NEOatNHNG:
```suggestion
common:linux_aarch64_score_gcc_12_2_0_posix --@rules_rust//:extra_rustc_flags=-Clink-arg=-no-pie,-Clink-arg=-lgcc
```
Linkage to stdc++ already solved in the updated rust toolchain.

NEOatNHNG (dismissed):
Fine as a workaround for now, but probably we should fix this is in the score rust toolchain, right?

.bazelrc
bharatGoswami8: yes, removed stdc++

bharatGoswami8:
> This should also be added to the examples/.bazelrc
> 
> I would also ask that you open a bug ticket in the toolchain. Such workarounds should not be a longterm solution.

@LittleHuba , Created ticket on score rust toolchain.

NEOatNHNG (approved):

examples/.bazelrc
```diff
@@ -9,3 +9,5 @@ common --registry=https://bcr.bazel.build
 common --@score_baselibs//score/json:base_library=nlohmann
 
 common --@score_communication//score/mw/com/flags:tracing_library=@score_baselibs//score/analysis/tracing/generic_trace_library/stub_implementation
+
+common --@rules_rust//:extra_rustc_flags=-Clink-arg=-no-pie,-Clink-arg=-lgcc
```
LittleHuba, line 13: I think this is going in the wrong direction. We should instead enable PIC compatible builds.
From what I see the issues about relocation come from ACL.
Can you create a bugticket in baselibs, that the ACL library must be provided with `-fPIC`?

LittleHuba, line 13: Regarding the linkage to gcc, we have to investigate further.
I'd first check whether bumping the toolchain version helps here. @nradakovic mentioned in a talk that they link libgcc. Maybe we are on a version that does not yet do that.
We need to look at the bazel subcommands to investigate.
Can you please do a bazel run with the flag `-s` and post the logs here?

bharatGoswami8: I think this is not just ACL library issue, it is a score aarch64 GCC toolchain tool-chain issuw which build without` --enable-default-pie`.
Because if resolve in ACL next static library may have same kind of issue.

bharatGoswami8: I have added logs in the mentioned ticket as a comment - https://github.com/eclipse-score/communication/issues/328#issuecomment-4325916988

bharatGoswami8: If keeping `no-pie` flag on tool-chain is not a ideal solution then we can move to user level - https://github.com/eclipse-score/communication/pull/361

bharatGoswami8: As discussed created ticket for ACL lib in baselibs - https://github.com/eclipse-score/baselibs/issues/172

bharatGoswami8:
fix should be done from toolchain side.

PR #332, opened 2026-04-21
mw/com: Add method configuration support to SOME/IP

sahithi-nukala:
- Extend the SOME/IP configuration to support method deployment alongside existing events and fields.
- This includes new SomeIpMethodInstanceDeployment class, SomeIpServiceTypeDeployment struct, new SOME/IP ID type aliases (event, field, method, service), updated BUILD targets, parser logic for method parsing, and documentation updates.
- All related tests have been updated to reflect the new functionality, including updating the stale TerminateOnParsingSomeIP test now that SOME/IP configuration parsing is fully implemented.

### DIFF
--- a/score/mw/com/design/configuration/README.md
+++ b/score/mw/com/design/configuration/README.md
@@ -129,7 +129,13 @@ An example of such a mapping is shown here:
                             "eventId": 633
                         }
                     ],
-                    "fields": []
+                    "fields": [],
+                    "methods": [
+                        {
+                            "methodName": "SetPressure",
+                            "methodId": 40
+                        }
+                    ]
                 },
                 {
                     "binding": "SHM",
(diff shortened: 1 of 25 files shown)


NEOatNHNG:
For SOME/IP we already have the config work started as part of the eclipse-score/inc_someip_gateway. Should we then sync this somehow? Use the same subschema, or something like that?

NEOatNHNG (changes requested):
At least the JSON schema should be also extended.

sahithi-nukala:
> For SOME/IP we already have the config work started as part of the eclipse-score/inc_someip_gateway. Should we then sync this somehow? Use the same subschema, or something like that?

yes, using the same subschema makes sense. After this PR gets merged, we can take the changes on top of it.

sahithi-nukala:
> At least the JSON schema should be also extended.

yes its done

PR #357, opened 2026-04-24
mw::com: Migrate from boolean enablers to struct Tags to enable Get/Set/Notify functionality in Fields.

KrishaDeshkool:
* This PR aims to migrate from boolean based enabelment of notifier/getter/setter on a field to a more user friendly and strictly typed Tag based mechanism.
* This change would let the users not to worry about the ordering of the boolean flags that would enable user required features on a field instead, they can just name what they want to enable with WithSetter/WithNotifer/WithGetter.
* if any of the above mentioned tags are not provided it is assumed that users wants those respective features to be disabled.

### DIFF
--- a/score/mw/com/impl/field_tags.h
+++ b/score/mw/com/impl/field_tags.h
@@ -1,0 +1,69 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_IMPL_FIELD_TAGS_H
+#define SCORE_MW_COM_IMPL_FIELD_TAGS_H
+
+#include <type_traits>
+
+namespace score::mw::com::impl
+{
+
+/// \brief Tag types used on ProxyField/SkeletonField level to accomplish SFINAE gating of various signatures,
+/// which depend on the availability of Get/Set/Notifier.
+struct WithGetter
+{
+};
+
+struct WithSetter
+{
+};
+
+struct WithNotifier
+{
+};
+
+template <typename TargetType, typename... Tags>
+struct contains_type : std::disjunction<std::is_same<TargetType, Tags>...>
+{
+};
+
+/// \brief Gates a member function on whether a capability tag was declared for this field.
+///
+/// Two things have to be true for the overload to work:
+///   - \c RequiredTag is somewhere(order does not matter) in \c EnabledTags.
+///   - \c FieldType == \c ClassFieldType (see below)
+///
+/// The second condition (\c FieldType == \c ClassFieldType) is a SFINAE workaround: \c FieldType
+/// must be a deduced local parameter of the member function template (defaulting to \c ClassFieldType)
+/// so the check stays dependent and the compiler does not harderr at class instantiation time.
+///
+/// In practice you use it like this inside \c Foo<FieldType, Tags...>:
+/// \code
+///   template <typename F = FieldType,
+///             std::enable_if_t<is_tag_enabled<F, FieldType, WithSetter, Tags...>::value, int> = 0>
+///   void Set(F value);
+/// \endcode
+///
+/// \tparam FieldType       Local deduced copy of the field type, must default to \c ClassFieldType.
+/// \tparam ClassFieldType  The actual field type the enclosing class was instantiated with.
+/// \tparam RequiredTag     The tag that must be present (e.g. \c WithSetter, \c WithGetter).
+/// \tparam EnabledTags     All tags the field was declared with.
+template <typename FieldType, typename ClassFieldType, typename RequiredTag, typename... EnabledTags>
+struct is_tag_enabled
+    : std::conjunction<contains_type<RequiredTag, EnabledTags...>, std::is_same<FieldType, ClassFieldType>>
+{
+};
+
+}  // namespace score::mw::com::impl
+
+#endif  // SCORE_MW_COM_IMPL_FIELD_TAGS_H
--- a/score/mw/com/impl/mocking/proxy_event_mock_test.cpp
+++ b/score/mw/com/impl/mocking/proxy_event_mock_test.cpp
@@ -66,7 +66,7 @@ struct ProxyEventStruct
 
 struct ProxyFieldStruct
 {
-    using ProxyEventField = ProxyField<TestSampleType>;
+    using ProxyEventField = ProxyField<TestSampleType, WithGetter, WithSetter, WithNotifier>;
     using ProxyEventFieldMock = ProxyEventMock<TestSampleType>;
 };
 
--- a/score/mw/com/impl/mocking/proxy_event_mock_test.cpp
+++ b/score/mw/com/impl/mocking/proxy_event_mock_test.cpp
@@ -54,8 +54,8 @@ class ProxyEventFieldMockFixture : public ::testing::Test
     ProxyEventFieldMock proxy_service_element_mock_{};
     ProxyBase proxy_base_{nullptr, MakeFakeHandle(1U)};
     ProxyEventField unit_{proxy_base_,
-                          std::unique_ptr<ProxyEventBinding<TestSampleType>>{nullptr},
-                          kDummyEventFieldName};
+                          kDummyEventFieldName,
+                          std::unique_ptr<ProxyEventBinding<TestSampleType>>{nullptr}};
 };
 
 struct ProxyEventStruct
--- a/score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
+++ b/score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
@@ -54,7 +54,8 @@ class MyInterface : public InterfaceTrait::Base
 
     typename InterfaceTrait::template Event<TestEventType> some_event{*this, kEventName};
     typename InterfaceTrait::template Event<TestEventType2> some_event_2{*this, kEventName2};
-    typename InterfaceTrait::template Field<TestFieldType> some_field{*this, kFieldName};
+    typename InterfaceTrait::template Field<TestFieldType, WithGetter, WithNotifier, WithSetter> some_field{*this,
+                                                                                                            kFieldName};
 };
 using MyProxy = AsProxy<MyInterface>;
 
@@ -244,7 +245,8 @@ class FieldOnlyInterface : public InterfaceTrait::Base
   public:
     using InterfaceTrait::Base::Base;
 
-    typename InterfaceTrait::template Field<TestFieldType> some_field{*this, kFieldName};
+    typename InterfaceTrait::template Field<TestFieldType, WithGetter, WithNotifier, WithSetter> some_field{*this,
+                                                                                                            kFieldName};
 };
 using FieldOnlyProxy = AsProxy<FieldOnlyInterface>;
 
--- a/score/mw/com/impl/mocking/skeleton_field_mock_test.cpp
+++ b/score/mw/com/impl/mocking/skeleton_field_mock_test.cpp
@@ -41,7 +41,7 @@ class SkeletonFieldMockFixture : public ::testing::Test
 
     SkeletonFieldMock<TestSampleType> skeleton_field_mock_{};
     SkeletonBase skeleton_base_{nullptr, MakeFakeInstanceIdentifier(1U)};
-    SkeletonField<TestSampleType> unit_{skeleton_base_, kDummyFieldName, nullptr};
+    SkeletonField<TestSampleType, WithGetter, WithNotifier, WithSetter> unit_{skeleton_base_, kDummyFieldName, nullptr};
 };
 
 TEST_F(SkeletonFieldMockFixture, AllocateDispatchesToMockAfterInjectingMock)
--- a/score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
+++ b/score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
@@ -52,14 +52,14 @@ class NamedSkeletonEventMock
     SkeletonEventMock<EventType> mock;
 };
 
-template <typename FieldType>
+template <typename FieldType, typename... Tags>
 class NamedSkeletonFieldMock
 {
   public:
     NamedSkeletonFieldMock(const std::string_view field_name_in) : field_name{field_name_in}, mock{} {}
 
     std::string_view field_name;
-    SkeletonFieldMock<FieldType> mock;
+    SkeletonFieldMock<FieldType, Tags...> mock;
 };
 
 template <typename SkeletonWrapperClass>
@@ -85,10 +85,10 @@ class SkeletonWrapperClassTestView
     /// @param event_mocks A tuple containing a NamedEventMock per event in the interface.
     /// @param field_mocks A tuple containing a NamedFieldMock per event in the interface.
     /// @return SkeletonWrapperClass containing mocked skeleton and mocked events / fields.
-    template <typename... EventTypes, typename... FieldTypes>
+    template <typename... EventTypes, typename... FieldTypes, typename... Tags>
     static SkeletonWrapperClass Create(SkeletonBaseMock& skeleton_mock,
                                        std::tuple<NamedSkeletonEventMock<EventTypes>...>& event_mocks,
-                                       std::tuple<NamedSkeletonFieldMock<FieldTypes>...>& field_mocks)
+                                       std::tuple<NamedSkeletonFieldMock<FieldTypes, Tags...>...>& field_mocks)
     {
         // Create service element binding factory guards which inject mocks into the binding factories. We rely on the
         // default behaviour that calls to Create on the factories will return nullptrs. This is required since the real
@@ -117,9 +117,9 @@ class SkeletonWrapperClassTestView
     }
 
     /// @brief Test-only Create call which can be used when an interface does not contain any events.
-    template <typename... FieldTypes>
+    template <typename... FieldTypes, typename... Tags>
     static SkeletonWrapperClass Create(SkeletonBaseMock& skeleton_mock,
-                                       std::tuple<NamedSkeletonFieldMock<FieldTypes>...>& field_mocks)
+                                       std::tuple<NamedSkeletonFieldMock<FieldTypes, Tags...>...>& field_mocks)
     {
         // Since empty_event_mocks is passed to Create as a non-const reference, when this function returns, the
         // reference to empty_event_mocks will be dangling. However, since its empty, the tuple and its contents are not
@@ -156,19 +156,20 @@ class SkeletonWrapperClassTestView
         typed_event->InjectMock(mock);
     }
 
-    template <typename SampleType>
-    static void InjectFieldMock(SkeletonFieldBase& field_base, SkeletonFieldMock<SampleType>& mock)
+    template <typename SampleType, typename... Tags>
+    static void InjectFieldMock(SkeletonFieldBase& field_base, SkeletonFieldMock<SampleType, Tags...>& mock)
     {
-        auto* typed_field = dynamic_cast<SkeletonField<SampleType>*>(&field_base);
+        auto* typed_field = dynamic_cast<SkeletonField<SampleType, Tags...>*>(&field_base);
         SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(typed_field != nullptr,
                                                           "field_base should always be a fully typed SkeletonField");
         typed_field->InjectMock(mock);
     }
 
-    template <typename... EventTypes, typename... FieldTypes>
+    template <typename... EventTypes, typename... FieldTypes, typename... Tags>
     static void InjectEventAndFieldMocks(SkeletonWrapperClass& skeleton,
                                          std::tuple<NamedSkeletonEventMock<EventTypes>...>& event_mocks,
-                                         std::tuple<NamedSkeletonFieldMock<FieldTypes>...>& field_mocks)
+                                         std::tuple<NamedSkeletonFieldMock<FieldTypes, Tags...>...>& field_mocks)
+
     {
         auto& events = SkeletonBaseView{skeleton}.GetEvents();
         auto& fields = SkeletonBaseView{skeleton}.GetFields();
--- a/score/mw/com/impl/proxy_field.h
+++ b/score/mw/com/impl/proxy_field.h
@@ -13,6 +13,7 @@
 #ifndef SCORE_MW_COM_IMPL_PROXY_FIELD_H
 #define SCORE_MW_COM_IMPL_PROXY_FIELD_H
 
+#include "score/mw/com/impl/field_tags.h"
 #include "score/mw/com/impl/methods/proxy_method_with_in_args_and_return.h"
 #include "score/mw/com/impl/methods/proxy_method_with_return_type.h"
 #include "score/mw/com/impl/plumbing/proxy_field_binding_factory.h"
@@ -38,140 +39,105 @@ namespace score::mw::com::impl
 // Suppress "AUTOSAR C++14 M3-2-3" rule finding. This rule states: "An identifier with external linkage shall have
 // exactly one definition". This a forward class declaration.
 // coverity[autosar_cpp14_m3_2_3_violation]
-template <typename FieldType>
+template <typename FieldType, typename... Tags>
 class ProxyFieldAttorney;
 
 /// \brief This is the user-visible class of a field that is part of a proxy. It delegates all functionality to
 /// ProxyEvent.
 ///
-/// \tparam SampleDataType Type of data that is transferred by the event.
-/// \tparam EnableGet Whether the get method shall be enabled for this field. If true, a Get() method will be available.
-/// \tparam EnableSet Whether the set method shall be enabled for this field. If true, a Set() method will be available.
-/// \tparam EnableNotifier Whether the notifier functionality shall be enabled for this field. Whether this has an
-/// effect, depends on the binding that is used. The LoLa/shm-binding ignores this template parameter.
-template <typename SampleDataType,
-          const bool EnableGet = false,
-          const bool EnableSet = false,
-          const bool EnableNotifier = false>
+/// \tparam SampleDataType Type of data carried by the field.
+/// \tparam Tags Pack of marker tags from field_tags.h. Any combination of WithGetter, WithSetter, WithNotifier is
+///         supported, subject to the static_assert below: at least one of WithGetter or WithNotifier must be
+///         present (otherwise the value is invisible to consumers). WithGetter enables Get(), WithSetter enables
+///         Set(), WithNotifier enables the notifier methods Subscribe(), Unsubscribe(), GetSubscriptionState(),
+///         GetFreeSampleCount(), GetNumNewSamplesAvailable(), SetReceiveHandler(), UnsetReceiveHandler(),
+///         GetNewSamples().
+template <typename SampleDataType, typename... Tags>
 class ProxyField final : public ProxyFieldBase
 {
+
+    // A ProxyField must enable WithGetter or WithNotifier. A field with only WithSetter (or no tags at all)
+    // gives the consumer no way to observe the value it can write, making the field useless.
+    static_assert(
+        contains_type<WithNotifier, Tags...>::value || contains_type<WithGetter, Tags...>::value,
+        "A field must either have WithGetter or WithNotifier enabled otherwise, there is no way for the consumer to "
+        "receive the field values.");
+
     // Suppress "AUTOSAR C++14 A11-3-1", The rule declares: "Friend declarations shall not be used".
     // Design decision: The "*Attorney" class is a helper, which sets the internal state of this class accessing
     // private members and used for testing purposes only.
     // coverity[autosar_cpp14_a11_3_1_violation]
-    friend class ProxyFieldAttorney<SampleDataType>;
+    friend class ProxyFieldAttorney<SampleDataType, Tags...>;
 
   public:
     using FieldType = SampleDataType;
 
-    /// Constructor that allows to set the binding directly (both EnableGet and EnableSet are true).
-    ///
-    /// This is used for testing only. Allows for directly setting the bindings, and usually the mock binding is used
-    /// here.
+    /// \brief Testing ctor that delegates to the production ctor that accepts mock bindings directly, this can only be
+    /// used in test code.
+    /// \details The template tags do not gate this overload; they continue to control which parts of the public
+    ///          API (Get, Set, Subscribe, ...) are available on the resulting object. The event binding parameter
+    ///          is required (non default) to disambiguate this overload from the production ctor.
+    /// \param proxy_base Parent proxy that owns this field's registration.
+    /// \param field_name Field's name as it appears in the deployment.
+    /// \param event_binding Mock event binding. Must be provided (use nullptr if no event binding is needed).
+    /// \param set_method_binding Optional mock Set-method binding. If nullptr, no ProxyMethod is built for Set
+    ///        (same MarkServiceElementBindingInvalid avoidance as get_method_binding).
+    /// \param get_method_binding Optional mock Get-method binding. If nullptr, no ProxyMethod is built for Get;
+    ///        this avoids the ProxyMethod ctor calling ProxyBaseView::MarkServiceElementBindingInvalid() on the
+    ///        parent, which would otherwise make AreBindingsValid() return false during proxy construction.
     ProxyField(ProxyBase& proxy_base,
                const std::string_view field_name,
                std::unique_ptr<ProxyEventBinding<FieldType>> event_binding,
-               std::unique_ptr<ProxyMethodBinding> get_method_binding = nullptr,
                std::unique_ptr<ProxyMethodBinding> set_method_binding = nullptr,
-               detail::EnableBothTag = {})
-        : ProxyField{
-              proxy_base,
-              field_name,
-              std::make_unique<ProxyEvent<FieldType>>(proxy_base, field_name, std::move(event_binding)),
-              // If a binding is not provided, then we don't create the method. This ensures that the ProxyMethod
-              // doesn't report that the binding is invalid (via proxy_base_view.MarkServiceElementBindingInvalid())
-              get_method_binding == nullptr ? nullptr
-                                            : std::make_unique<ProxyMethod<FieldType()>>(
-                                                  proxy_base,
-                                                  field_name,
-                                                  std::move(get_method_binding),
-                                                  typename ProxyMethod<FieldType()>::FieldOnlyConstructorEnabler{}),
-              set_method_binding == nullptr
-                  ? nullptr
-                  : std::make_unique<ProxyMethod<FieldType(FieldType)>>(
-                        proxy_base,
-                        field_name,
-                        std::move(set_method_binding),
-                        typename ProxyMethod<FieldType(FieldType)>::FieldOnlyConstructorEnabler{})}
+               std::unique_ptr<ProxyMethodBinding> get_method_binding = nullptr)
+        : ProxyField{proxy_base,
+                     field_name,
+                     std::make_unique<ProxyEvent<FieldType>>(proxy_base, field_name, std::move(event_binding)),
+                     set_method_binding == nullptr
+                         ? nullptr
+                         : std::make_unique<ProxyMethod<FieldType(FieldType)>>(
+                               proxy_base,
+                               field_name,
+                               std::move(set_method_binding),
+                               typename ProxyMethod<FieldType(FieldType)>::FieldOnlyConstructorEnabler{}),
+                     get_method_binding == nullptr
+                         ? nullptr
+                         : std::make_unique<ProxyMethod<FieldType()>>(
+                               proxy_base,
+                               field_name,
+                               std::move(get_method_binding),
+                               typename ProxyMethod<FieldType()>::FieldOnlyConstructorEnabler{})}
     {
     }
 
-    /// \brief Constructs a ProxyField (both EnableGet and EnableSet are true). Normal ctor that is used in production
-    /// code.
-    template <bool EG = EnableGet, bool ES = EnableSet, typename = std::enable_if_t<EG && ES>>
-    ProxyField(ProxyBase& proxy_base, const std::string_view field_name, detail::EnableBothTag = {})
+    /// \brief Production ctor used by the end users.
+    /// \details The Make*IfEnabled helpers consult the tag pack at compile time and return nullptr when their
+    ///          corresponding tag is absent, so the same body produces the correct dispatch for every valid
+    ///          tag combination. The class-level static_assert ensures that at least one of WithGetter or
+    ///          WithNotifier is present.
+    /// \param proxy_base Parent proxy that owns this field's registration.
+    /// \param field_name Field's name as it appears in the deployment.
+    ProxyField(ProxyBase& proxy_base, const std::string_view field_name)
         : ProxyField{proxy_base,
                      field_name,
-                     std::make_unique<ProxyEvent<FieldType>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateEventBinding(proxy_base, field_name),
-                         typename ProxyEvent<FieldType>::FieldOnlyConstructorEnabler{}),
-                     std::make_unique<ProxyMethod<FieldType()>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateGetMethodBinding(proxy_base, field_name),
-                         typename ProxyMethod<FieldType()>::FieldOnlyConstructorEnabler{}),
-                     std::make_unique<ProxyMethod<FieldType(FieldType)>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateSetMethodBinding(proxy_base, field_name),
-                         typename ProxyMethod<FieldType(FieldType)>::FieldOnlyConstructorEnabler{})}
+                     MakeEventDispatchIfEnabled(proxy_base, field_name),
+                     MakeSetMethodDispatchIfEnabled(proxy_base, field_name),
+                     MakeGetMethodDispatchIfEnabled(proxy_base, field_name)}
     {
-    }
-
-    /// \brief Constructs a ProxyField (only EnableGet is true). Normal ctor that is used in production code.
-    template <bool EG = EnableGet, bool ES = EnableSet, typename = std::enable_if_t<EG && !ES>>
-    ProxyField(ProxyBase& proxy_base, const std::string_view field_name, detail::EnableGetOnlyTag = {})
-        : ProxyField{proxy_base,
-                     field_name,
-                     std::make_unique<ProxyEvent<FieldType>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateEventBinding(proxy_base, field_name),
-                         typename ProxyEvent<FieldType>::FieldOnlyConstructorEnabler{}),
-                     std::make_unique<ProxyMethod<FieldType()>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateGetMethodBinding(proxy_base, field_name),
-                         typename ProxyMethod<FieldType()>::FieldOnlyConstructorEnabler{}),
-                     nullptr}
-    {
-    }
-
-    /// \brief Constructs a ProxyField (only EnableSet is true). Normal ctor that is used in production code.
-    template <bool EG = EnableGet, bool ES = EnableSet, typename = std::enable_if_t<!EG && ES>>
-    ProxyField(ProxyBase& proxy_base, const std::string_view field_name, detail::EnableSetOnlyTag = {})
-        : ProxyField{proxy_base,
-                     field_name,
-                     std::make_unique<ProxyEvent<FieldType>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateEventBinding(proxy_base, field_name),
-                         typename ProxyEvent<FieldType>::FieldOnlyConstructorEnabler{}),
-                     nullptr,
-                     std::make_unique<ProxyMethod<FieldType(FieldType)>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateSetMethodBinding(proxy_base, field_name),
-                         typename ProxyMethod<FieldType(FieldType)>::FieldOnlyConstructorEnabler{})}
-    {
-    }
-
-    /// \brief Constructs a ProxyField (both EnableGet and EnableSet are false). Normal ctor that is used in production
-    /// code.
-    template <bool EG = EnableGet, bool ES = EnableSet, typename = std::enable_if_t<!EG && !ES>>
-    ProxyField(ProxyBase& proxy_base, const std::string_view field_name, detail::EnableNeitherTag = {})
-        : ProxyField{proxy_base,
-                     field_name,
-                     std::make_unique<ProxyEvent<FieldType>>(
-                         proxy_base,
-                         field_name,
-                         ProxyFieldBindingFactory<FieldType>::CreateEventBinding(proxy_base, field_name),
-                         typename ProxyEvent<FieldType>::FieldOnlyConstructorEnabler{}),
-                     nullptr,
-                     nullptr}
-    {
+        // Each Make*IfEnabled must have produced a non-null dispatch when the
+        // corresponding tag is in the pack.
+        if constexpr (kHasNotifier)
+        {
+            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(proxy_event_dispatch_ != nullptr);
+        }
+        if constexpr (kHasSetter)
+        {
+            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(proxy_method_set_dispatch_ != nullptr);
+        }
+        if constexpr (kHasGetter)
+        {
+            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(proxy_method_get_dispatch_ != nullptr);
+        }
     }
 
     /// \brief A ProxyField shall not be copyable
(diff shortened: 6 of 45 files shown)


KrishaDeshkool:
Once the CI is green, i will make this PR open and can we reviewed.

score/mw/com/impl/skeleton_field.h
```diff
@@ -112,14 +114,12 @@ class SkeletonField : public SkeletonFieldBase
...
-    //
-    // \ brief Registers a handler that is invoked by the middleware whenever a proxy calls the field setter.
-    //
-    // \tparam CallableType Any callable (std::function, score::cpp::callback, lambda, ...) with the signature:
-    //         void(FieldType& new_value)
-    //   - new_value : the value requested by the proxy.
-    template <bool ES = EnableSet, typename std::enable_if<ES, int>::type = 0, typename CallableType>
+    /// Registers a handler invoked when a proxy issues the field setter.
+    /// The handler receives the proxy-requested value by reference and may modify it in-place;
+    /// the (possibly modified) value is then broadcast via Update().
```
crimson11: See teams-chat:
>the (possibly modified) value is then broadcast via Update().

-> the (possibly modified) value is then stored as the "latest value" via call to SkeletonField::Update().

score/mw/com/impl/skeleton_field.h
```diff
@@ -159,50 +159,45 @@ class SkeletonField : public SkeletonFieldBase
...
-    using SetHandlerStorageType =
-        std::conditional_t<EnableSet, score::cpp::callback<void(FieldType&)>, detail::EnableSetOnlyTag>;
-    SetHandlerStorageType set_handler_{};
-
-    // Tracks whether RegisterSetHandler() has been called. Zero-cost when EnableSet=false.
-    using IsSetHandlerRegisteredType = std::conditional_t<EnableSet, bool, detail::EnableSetOnlyTag>;
-    IsSetHandlerRegisteredType is_set_handler_registered_{};
-
-    // EnableSet=true: checks the flag; EnableSet=false: no setter, no handler required.
     bool IsSetHandlerRegistered() const noexcept override
```
crimson11: SFINAE out if no WithSetter tag exists. -> RegisterSetHandler()

crimson11 (changes requested):

score/mw/com/impl/proxy_field.h
```diff
@@ -67,11 +86,11 @@ class ProxyField final : public ProxyFieldBase
...
     using FieldType = SampleDataType;
 
-    /// Testing ctor: bindings are passed in directly (used with mock bindings). The event binding is required
-    /// (no default) to disambiguate this overload from the production ctors when called as `{proxy, field_name}`.
-    /// Method bindings default to nullptr; passing nullptr means the corresponding ProxyMethod is not built.
+    /// Testing ctor: bindings are passed in directly (used with mock bindings). Disambiguated from the production
+    /// ctors via the detail::WithTestTag marker, required as the 3rd argument. Bindings default to nullptr;
     ProxyField(ProxyBase& proxy_base,
                const std::string_view field_name,
+               detail::WithTestTag,
```
KrishaDeshkool: This is the topic i wanted to discuss,
if we do not have this Tag then if we make all other bindings as nullptr, in other words make these defaults,
then i have the following issue,
when all the bingngs are nullptr then the all of them are defaulted and depending on the tag combination , the real production ctr would clash with this test ctr and since the prod ctr is templated the compiler would pick this non templated ctr and hence we run into issues in production, this is why i have introduced this WithTestTag to clearly say this is test only ctr, i think it also reads much better than adding a comment on the ctr that this is a test only ctr.

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: done

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: added SFINAE cond on RegisterSetHandler

crimson11: Ah. ok. You need this `IsSetHandlerRegistered()` to always exist ... as in `PrepareOffer` impl. you don't want to distinguish (can't easily distinguish) between WithGetter and not WithGetter ....

score/mw/com/impl/traits_test.cpp
```diff
@@ -549,7 +549,7 @@ TEST_F(GeneratedSkeletonCreationInstanceSpecifierTestFixture,
     EXPECT_CALL(skeleton_field_binding_factory_mock_guard_.factory_mock_,
                 CreateEventBinding(identifier_with_valid_binding_, _, kFieldName))
         .WillOnce(Return(ByMove(std::move(skeleton_field_binding_mock_ptr))));
-    // Field get method binding (not yet fully implemented; set method only exists when EnableSet=true)
+    // Field set method binding (only created when the SkeletonField is tagged WithSetter)
     EXPECT_CALL(skeleton_method_binding_factory_mock_guard_.factory_mock_, Create(_, _, _, _))
```
crimson11: You should be more concise here? You are expecting a call to `Create` with `MethodType` `kSet`? ... and eventually also test/make clear, what "method-name" we are expecting here? Hint - in this case the method-name is the field-name afaik ...

I guess - currently gtest is smart enough to correctly assign the EXPECT in line 553 vs line 559 by the arg `kMethodName` ...

crimson11 (changes requested):

score/mw/com/impl/field_tags.h
```diff
@@ -0,0 +1,51 @@
...
+template <typename TargetType, typename... Tags>
+struct contains_type : std::disjunction<std::is_same<TargetType, Tags>...>
+{
+};
+
+// SFINAE only works when the condition depends on the function-template parameters, not the class-template
+// parameters. The std::is_same<FieldType, ClassFieldType> line ties substitution to a function-template
+// parameter so the check fires at overload resolution instead of class instantiation.
+template <typename FieldType, typename ClassFieldType, typename TargetTag, typename... Tags>
+struct is_tag_enabled : std::conjunction<contains_type<TargetTag, Tags...>, std::is_same<FieldType, ClassFieldType>>
```
bemerybmw: Since is_tag_enabled is a public function to be used by other targets, it shouldn't be in the detail namespace.

score/mw/com/impl/field_tags.h
```diff
@@ -0,0 +1,51 @@
...
+{
```
bemerybmw, line 64: We should have unit tests for any public function (so definitely is_tag_enabled and if contains_type is publicly exposed i.e. not in the detail namespace then also for it). You can look at the unit tests that I wrote for method_handler_checker_test as a reference if you want: https://github.com/eclipse-score/communication/pull/369/changes#diff-a5298ded87ed55bcf9aceb680edbe7e0c9b81700b7a5e0f24feb7b3600dac64d

bemerybmw: We can only support "positive" tests for the moment, but you should also write the negative tests e.g. https://github.com/eclipse-score/communication/blob/b336e39c038a3aadf42cbab3fa4c2741cc3d9673/score/mw/com/impl/methods/method_handler_checker_test.cpp#L41

score/mw/com/impl/proxy_field.h
```diff
@@ -45,14 +47,9 @@ class ProxyFieldAttorney;
...
-/// \tparam EnableGet Whether the get method shall be enabled for this field. If true, a Get() method will be available.
-/// \tparam EnableSet Whether the set method shall be enabled for this field. If true, a Set() method will be available.
-/// \tparam EnableNotifier Whether the notifier functionality shall be enabled for this field. Whether this has an
-/// effect, depends on the binding that is used. The LoLa/shm-binding ignores this template parameter.
-template <typename SampleDataType,
-          const bool EnableGet = false,
-          const bool EnableSet = false,
-          const bool EnableNotifier = false>
+/// \tparam Tags Optional tag pack; presence of WithGetter / WithSetter enables Get() / Set() respectively.
+///              The notifier surface (Subscribe, GetNewSamples, ...) is part of every field.
```
bemerybmw: What does "surface" mean here?

score/mw/com/impl/proxy_field.h
```diff
@@ -45,14 +47,9 @@ class ProxyFieldAttorney;
...
+template <typename SampleDataType, typename... Tags>
```
bemerybmw, line 55: Maybe be explicit in the documentation that the supported tags are any combination of the tags defined in field_tags.h (i.e. WithGetter, WithSetter, WithNotifier)

score/mw/com/impl/proxy_field.h
```diff
@@ -64,16 +61,13 @@ class ProxyField final : public ProxyFieldBase
   public:
     using FieldType = SampleDataType;
 
-    /// Constructor that allows to set the binding directly (both EnableGet and EnableSet are true).
-    ///
-    /// This is used for testing only. Allows for directly setting the bindings, and usually the mock binding is used
-    /// here.
+    /// Testing ctor: bindings are passed in directly (used with mock bindings).
```
bemerybmw: Please use the doxygen format.

```
/// \brief ...
///
/// More details here.
```

score/mw/com/impl/proxy_field.h
```diff
@@ -64,16 +61,13 @@ class ProxyField final : public ProxyFieldBase
...
+    /// Method bindings default to nullptr; passing nullptr means the corresponding ProxyMethod is not built.
```
bemerybmw: I would remove "Method bindings default to nullptr;". This is already in the code directly below, so there's no benefit to repeating it here.

bemerybmw: I would also mention that the template tags do not affect this test constructor. They will only change whether certain parts of the public API (i.e. Get, Set etc.) are available.

score/mw/com/impl/skeleton_field.h
```diff
@@ -35,37 +37,37 @@
...
-    /// \param field_name Field name of the field.
-    /// \param detail::EnableNeitherTag This parameter is only used for constructor overload disambiguation and
-    /// has no semantic meaning.
-    template <bool ES = EnableSet, typename = std::enable_if_t<!ES>>
-    SkeletonField(SkeletonBase& parent, const std::string_view field_name, detail::EnableNeitherTag = {});
-
-    /// Constructor that allows to set the binding directly.
-    ///
-    /// This is used only used for testing.
+    /// Normal ctor, setter enabled.
```
bemerybmw: Please use doxygen format. IMO, should keep "Constructs a SkeletonField with setter enabled. Normal ctor used in production code."

score/mw/com/impl/skeleton_field.h
```diff
@@ -35,37 +37,37 @@
...
+    template <typename U = SampleDataType,
+              typename = std::enable_if_t<detail::is_tag_enabled<U, SampleDataType, WithSetter, Tags...>::value>>
+    SkeletonField(SkeletonBase& parent, const std::string_view field_name, WithSetter = {})
+        : SkeletonField{parent, MakeSkeletonEvent(parent, field_name), field_name, WithSetter{}}
+    {
+    }
+
+    /// Normal ctor, setter not enabled.
```
bemerybmw: Same here

score/mw/com/impl/skeleton_field.h
```diff
@@ -35,37 +37,37 @@
...
+    template <typename U = SampleDataType,
+              typename = std::enable_if_t<!detail::is_tag_enabled<U, SampleDataType, WithSetter, Tags...>::value>>
+    SkeletonField(SkeletonBase& parent, const std::string_view field_name)
+        : SkeletonField{parent, MakeSkeletonEvent(parent, field_name), field_name}
+    {
+    }
+
+    /// Testing ctor: binding is passed in directly (used with mock bindings).
```
bemerybmw: Similar comments to the test constructor in proxy_field.h

score/mw/com/impl/skeleton_field.h
```diff
@@ -112,14 +114,12 @@ class SkeletonField : public SkeletonFieldBase
...
 
-    // SFINAE-gated: only exists when EnableSet == true
-    //
-    // \ brief Registers a handler that is invoked by the middleware whenever a proxy calls the field setter.
-    //
-    // \tparam CallableType Any callable (std::function, score::cpp::callback, lambda, ...) with the signature:
-    //         void(FieldType& new_value)
-    //   - new_value : the value requested by the proxy.
-    template <bool ES = EnableSet, typename std::enable_if<ES, int>::type = 0, typename CallableType>
+    /// Registers a handler invoked when a proxy issues the field setter.
```
bemerybmw: Please use doxygen formatting

score/mw/com/impl/skeleton_field.h
```diff
@@ -112,14 +114,12 @@ class SkeletonField : public SkeletonFieldBase
...
+    /// The handler receives the proxy-requested value by reference and may modify it in-place;
```
bemerybmw: Why did you remove the information about CallableType?

score/mw/com/impl/skeleton_field.h
```diff
@@ -159,50 +159,45 @@ class SkeletonField : public SkeletonFieldBase
...
+            typename SkeletonEvent<FieldType>::FieldOnlyConstructorEnabler{});
+    }
+
     /// \brief Private delegating constructor used by the setter-enabled public ctor.
-    template <bool ES = EnableSet, typename = std::enable_if_t<ES>>
     SkeletonField(SkeletonBase& parent,
                   std::unique_ptr<SkeletonEvent<FieldType>> skeleton_event_dispatch,
                   const std::string_view field_name,
-                  detail::EnableSetOnlyTag);
+                  WithSetter);
```
bemerybmw: We should have one private constructor like in proxyField which takes the skeleton_event_dispatch and also the methods (or a nullptr if they're not enabled)

score/mw/com/impl/traits.h
```diff
@@ -100,18 +101,16 @@ class ProxyWrapperClassTestView;
...
 ///
 ///     typename Trait::template Method<void(InArgType1)> struct_method_1_{*this, method_name_0};
 ///     typename Trait::template Method<ReturnType()> struct_method_2_{*this, method_name_1};
 ///
 /// };
-/// Notes regarding template args: A field has (besides its data type arg) three bool template args (enable_getter,
-/// enable_setter and enable_notifier). The enable_notifier template parameter is only relevant for certain bindings,
-/// e.g. the LoLa binding does not distinguish between true/false of this template parameter.
+/// Notes regarding template args: a field takes its data type plus an optional pack of tags. Presence of WithGetter
+/// or WithSetter enables Get() / Set() respectively on the field. The notifier surface (Subscribe/GetNewSamples on
```
bemerybmw: Again, what does surface mean here?

score/mw/com/impl/mocking/proxy_event_mock_test.cpp
```diff
@@ -64,7 +64,7 @@ struct ProxyEventStruct
 
 struct ProxyFieldStruct
 {
-    using ProxyEventField = ProxyField<TestSampleType>;
+    using ProxyEventField = ProxyField<TestSampleType, WithGetter, WithNotifier>;
```
bemerybmw: Why getter / notifier but no setter?

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
```diff
@@ -54,7 +54,7 @@ class MyInterface : public InterfaceTrait::Base
 
     typename InterfaceTrait::template Event<TestEventType> some_event{*this, kEventName};
     typename InterfaceTrait::template Event<TestEventType2> some_event_2{*this, kEventName2};
-    typename InterfaceTrait::template Field<TestFieldType> some_field{*this, kFieldName};
+    typename InterfaceTrait::template Field<TestFieldType, WithGetter, WithNotifier> some_field{*this, kFieldName};
```
bemerybmw: Why getter / notifier but no setter? Same question in all other places in this commit.

score/mw/com/impl/mocking/skeleton_field_mock_test.cpp
```diff
@@ -41,7 +41,7 @@ class SkeletonFieldMockFixture : public ::testing::Test
 
     SkeletonFieldMock<TestSampleType> skeleton_field_mock_{};
     SkeletonBase skeleton_base_{nullptr, MakeFakeInstanceIdentifier(1U)};
-    SkeletonField<TestSampleType> unit_{skeleton_base_, kDummyFieldName, nullptr};
+    SkeletonField<TestSampleType, WithGetter, WithNotifier> unit_{skeleton_base_, kDummyFieldName, nullptr};
```
bemerybmw: Why getter / notifier but no setter? This is calling the test constructor and not actually creating the event / method bindings, right?

score/mw/com/impl/field_tags.h
```diff
@@ -19,7 +19,8 @@ namespace score::mw::com::impl
 {
 
 /// \brief Tag types used on ProxyField/SkeletonField level to accomplish overload-resolution for various signatures,
-/// which depend on the availability of Get/Set/Notifier.
+/// which depend on the availability of Get/Set/Notifier. A field must enable WithGetter or WithNotifier
```
bemerybmw: This comment "A field must enable WithGetter or WithNotifier (a write-only field would be invisible to consumers)." isn't really relevant here IMO. this should rather be in the documentation of the field itself. The tags themselves don't care about that invariant. The invairnat only matters in the place where the tags are used i.e. the actual field.

score/mw/com/impl/proxy_field.h
```diff
@@ -52,11 +52,17 @@ class ProxyFieldAttorney;
 template <typename SampleDataType, typename... Tags>
 class ProxyField final : public ProxyFieldBase
 {
+
+    static_assert(
```
bemerybmw, line 61: I guess we should have the same assertion in skeleton field?

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -63,12 +67,20 @@ TEST(ProxyFieldTest, ClassTypeDependsOnFieldDataType)
...
-    using SecondProxyFieldType = ProxyField<std::uint16_t>;
+    using FirstProxyFieldType = ProxyField<bool, WithGetter, WithNotifier>;
+    using SecondProxyFieldType = ProxyField<std::uint16_t, WithGetter, WithNotifier>;
     static_assert(!std::is_same_v<FirstProxyFieldType, SecondProxyFieldType>,
                   "Class type does not depend on field data type");
 }
 
+TEST(ProxyFieldTest, ClassTypeDependsOnFieldTagType)
+{
+    using FirstProxyFieldType = ProxyField<bool, WithGetter, WithNotifier>;
```
bemerybmw: The test above is needed to test a specific requirement we have. I don't think this test adds any value.

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -94,12 +98,20 @@ TEST(SkeletonFieldTest, ClassTypeDependsOnFieldDataType)
...
-    using SecondSkeletonFieldType = SkeletonField<std::uint16_t>;
+    using FirstSkeletonFieldType = SkeletonField<bool, WithGetter, WithNotifier>;
+    using SecondSkeletonFieldType = SkeletonField<std::uint16_t, WithGetter, WithNotifier>;
     static_assert(!std::is_same_v<FirstSkeletonFieldType, SecondSkeletonFieldType>,
                   "Class type does not depend on field data type");
 }
 
+TEST(SkeletonFieldTest, ClassTypeDependsOnFieldTagType)
+{
+    using FirstSkeletonFieldType = SkeletonField<bool, WithGetter, WithNotifier>;
```
bemerybmw: The test above is needed to test a specific requirement we have. I don't think this test adds any value.

score/mw/com/impl/traits_test.cpp
```diff
@@ -67,7 +67,7 @@ class MyInterface : public InterfaceTrait::Base
     using InterfaceTrait::Base::Base;
 
     typename InterfaceTrait::template Event<TestSampleType> some_event{*this, kEventName};
-    typename InterfaceTrait::template Field<TestSampleType> some_field{*this, kFieldName};
+    typename InterfaceTrait::template Field<TestSampleType, WithNotifier> some_field{*this, kFieldName};
```
bemerybmw: Why just WithNotifier?

score/mw/com/types.h
```diff
@@ -105,6 +105,12 @@ using MethodInArgTypePtr = impl::MethodInArgPtr<ReturnType>;
 /// \requirement SWS_CM_00309
 using EventReceiveHandler = impl::EventReceiveHandler;
 
+/// \api
+/// \brief Field tag types used in service-interface definitions to enable Get/Set/Notifier on a field.
+using WithGetter = impl::WithGetter;
```
bemerybmw, line 111: Need include?

score/mw/com/impl/proxy_field.h
```diff
@@ -47,8 +47,8 @@ class ProxyFieldAttorney;
 /// ProxyEvent.
 ///
 /// \tparam SampleDataType Type of data that is transferred by the event.
-/// \tparam Tags Optional tag pack; presence of WithGetter / WithSetter enables Get() / Set() respectively.
-///              The notifier surface (Subscribe, GetNewSamples, ...) is part of every field.
+/// \tparam Tags Optional tag pack; presence of WithGetter / WithSetter / WithNotifier enables Get() / Set() /
```
bemerybmw: Comment does not make sense currently. Can you please use full sentences. It needs to be clear that when WithNotifier is disabled, Subscribe() etc are not available.

score/mw/com/impl/proxy_field.h
```diff
@@ -203,12 +203,63 @@ class ProxyField final : public ProxyFieldBase
      * \param max_num_samples Maximum number of samples to return via the given callable.
      * \return Number of samples that were handed over to the callable or an error.
      */
-    template <typename F>
+    template <typename F,
```
bemerybmw: Can you rename F to ReceiverType?

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -96,5 +201,35 @@ TEST(ProxyFieldTest, ProxyFieldContainsPublicFieldType)
         "Incorrect FieldType.");
 }
 
+TEST(ProxyFieldNotifierGatingTest, NotifierSurfaceExistsWhenWithNotifierTagIsPresent)
+{
+
+    using NotifierField = ProxyField<TestSampleType, WithNotifier>;
+    static_assert(has_subscribe<NotifierField>::value);
```
bemerybmw, line 244: Can we use std::is_member_function_pointer_v instead of these custom traits? E.g. how we do it in `SamplePtrGenericTypeTest, InterfaceMatchesRequirements`

score/mw/com/impl/proxy_field.h
```diff
@@ -290,7 +292,10 @@ class ProxyField final : public ProxyFieldBase
           proxy_method_get_dispatch_{std::move(proxy_method_get_dispatch)},
           proxy_method_set_dispatch_{std::move(proxy_method_set_dispatch)}
     {
-        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(proxy_event_dispatch_ != nullptr);
+        if constexpr (kHasNotifier)
+        {
+            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(proxy_event_dispatch_ != nullptr);
```
bemerybmw: We should have similar checks for method dispatch objects?

score/mw/com/impl/proxy_field.h
```diff
@@ -67,96 +67,119 @@ class ProxyField final : public ProxyFieldBase
   public:
     using FieldType = SampleDataType;
 
-    /// Testing ctor: bindings are passed in directly (used with mock bindings).
+    /// Testing ctor: bindings are passed in directly (used with mock bindings). The event binding is required
+    /// (no default) to disambiguate this overload from the production ctors when called as `{proxy, field_name}`.
```
bemerybmw: But there is a default agrument for event_binding?

score/mw/com/impl/proxy_field.h
```diff
@@ -67,96 +67,119 @@ class ProxyField final : public ProxyFieldBase
...
     ProxyField(ProxyBase& proxy_base,
                const std::string_view field_name,
                std::unique_ptr<ProxyEventBinding<FieldType>> event_binding = nullptr,
                std::unique_ptr<ProxyMethodBinding> get_method_binding = nullptr,
                std::unique_ptr<ProxyMethodBinding> set_method_binding = nullptr)
-        : ProxyField{
-              proxy_base,
-              field_name,
-              std::make_unique<ProxyEvent<FieldType>>(proxy_base, field_name, std::move(event_binding)),
-              // If a binding is not provided, then we don't create the method. This ensures that the ProxyMethod
```
bemerybmw: We need to document the seond sentence "This ensures that the ProxyMethod doesn't report that the binding is invalid (via proxy_base_view.MarkServiceElementBindingInvalid())" somewhere. Either here or in the docstring for the function.

score/mw/com/impl/proxy_field.h
```diff
@@ -282,6 +308,45 @@ class ProxyField final : public ProxyFieldBase
         }
     }
 
+    /// Builds the Get-method dispatch via the binding factory when WithGetter is enabled. Returns nullptr
+    /// without invoking the factory when WithGetter is disabled.
+    static std::unique_ptr<ProxyMethod<FieldType()>> MakeGetMethodDispatchFromFactory(ProxyBase& proxy_base,
+                                                                                      const std::string_view field_name)
+    {
+        if constexpr (kHasGetter)
```
bemerybmw: I think it would be clearer to only call this function when a getter is enabled. I.e. the construtor either calls MakeGetMethodDispatchFromFactory or explicitly passes a nullptr. That way it's clear from the constructor whether a method dispatch is created or not.

On second thought, what could make more sense is to have one production constructor and the dispatch objects are created according to the tags in these "Make" functions. Because currently all those constructors do the same thing, i.e. simply dispatch to these functions. In this case, maybe you can rename them to `MakeGetMethodDispatchIfUsed` or `MakeGetMethodDispatchIfEnabled` or something.

score/mw/com/impl/mocking/proxy_event_mock_test.cpp
```diff
@@ -45,15 +45,27 @@ class ProxyEventFieldMockFixture : public ::testing::Test
     using ProxyEventFieldMock = typename T::ProxyEventFieldMock;
 
     ProxyEventFieldMockFixture()
+        : unit_{[this] {
```
bemerybmw: I don't think this commit makes sense. I don't think we should add a tag for a test only "helper" constructor. I think we just live with the limitation that we have to always provide the event binding (either with a real binding or a nullptr).

score/mw/com/impl/traits_test.cpp
```diff
@@ -549,13 +549,14 @@ TEST_F(GeneratedSkeletonCreationInstanceSpecifierTestFixture,
     EXPECT_CALL(skeleton_field_binding_factory_mock_guard_.factory_mock_,
                 CreateEventBinding(identifier_with_valid_binding_, _, kFieldName))
         .WillOnce(Return(ByMove(std::move(skeleton_field_binding_mock_ptr))));
-    // Field set method binding (only created when the SkeletonField is tagged WithSetter)
-    EXPECT_CALL(skeleton_method_binding_factory_mock_guard_.factory_mock_, Create(_, _, _, _))
+    // Field get method binding always created for every SkeletonField regardless of tags, because get_method_ is
```
bemerybmw: This test is anyway fixed by https://github.com/eclipse-score/communication/pull/373/changes#diff-412e28d5d756f0c6692791281ab7995b8b0ec4935a6d836fd68ee6cdb71a2b7e. So I think you can remove this commit.

bemerybmw (changes requested):

score/mw/com/impl/proxy_field.h
bemerybmw: See comment here: https://github.com/eclipse-score/communication/pull/357#discussion_r3201178767

KrishaDeshkool:
putting back to draft until i fix the review comments

score/mw/com/impl/proxy_field.h
KrishaDeshkool: okay, i wanted to have uniformity where all can be set to nullptrs, (this comes with a cost of new TestOnlyTag),
fine for me to remove it.

score/mw/com/impl/field_tags.h
KrishaDeshkool: really? this is just a internal helper to findout if some tag is enabled or not, why should this be a public API?
details here means it is used by all our(internal) targets, public API to me is what we expose to users.
we dont want to make this trait available to users.
(in tunneling we have seen many instances of users depending on such helpers and once we refactor these helpers user code would break)

score/mw/com/impl/field_tags.h
KrishaDeshkool: i'm still not convinced that these are public methods,
but i have added tests.

score/mw/com/impl/proxy_field.h
KrishaDeshkool: API surface, i meant all the methods that are part of public API,
reworded it to be more explicit.

score/mw/com/impl/proxy_field.h
KrishaDeshkool: done

score/mw/com/impl/proxy_field.h
KrishaDeshkool: reworded the whole comment block with doxygen style

KrishaDeshkool: reworded the whole comment block with doxygen style

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: ah! that was a mistake, adding it back

score/mw/com/impl/traits.h
KrishaDeshkool: API surface,

score/mw/com/impl/field_tags.h
KrishaDeshkool: true, this is not relevant here,
removed.

score/mw/com/impl/proxy_field.h
KrishaDeshkool: ah! this i did not do because SkeletonWrapperClassTestView's InjectFieldMock() was doing dynamic_cast<SkeletonField<SampleType>*>(&field_base) 
Adding the assert broke that cast at compile time since SkeletonField<SampleType> with an empty pack no longer instantiates.
to fix this we have to trickledown the tags to APIs like InjectFieldMock and NamedSkeletonFieldMock.
what my assumption is,
since the only way the user would get skeleton and proxy is with AsProxy and AsSkeleton on same interface (which would mean both proxy and skeleton have same tags) when proxy side is created with some tags that enable for a user to observe something the same tags are auto applied on skeleton side so there is no way skeleton is created with different tags?)

score/mw/com/impl/mocking/proxy_event_mock_test.cpp
KrishaDeshkool: Removed TestOnlyTag as discussed in some comment earlier.

score/mw/com/impl/proxy_field.h
KrishaDeshkool: good idea, i will adapt the ctrs

score/mw/com/impl/proxy_field.h
KrishaDeshkool: okay, this now feels like a footgun,
we rely on the expectation that the proxy is also created along with skeleton for a given interface,
This might not always be true, (even though this is not the case today).
i will add those in.

score/mw/com/impl/field_tags.h
```diff
@@ -0,0 +1,56 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef PLATFORM_AAS_MW_COM_IMPL_FIELD_TAGS_H
```
KrishaDeshkool: SCORE

score/mw/com/impl/mocking/skeleton_field_mock_test.cpp
KrishaDeshkool: you are right.

score/mw/com/impl/field_tags.h
KrishaDeshkool: removed

score/mw/com/impl/proxy_field.h
KrishaDeshkool: done

score/mw/com/impl/proxy_field.h
KrishaDeshkool: done

score/mw/com/impl/mocking/proxy_event_mock_test.cpp
KrishaDeshkool: added.

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
KrishaDeshkool: added in all the places.

score/mw/com/impl/proxy_field_test.cpp
KrishaDeshkool: okay, removed

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
KrishaDeshkool: missing

score/mw/com/impl/proxy_field_test.cpp
KrishaDeshkool: when SFINAE disables some method then if we check if that method is present or not we get a hard compiler error we cannot test in tests,
this is a SFINAE friendly pattern i followed (we have used such approach in tunneling before).
i gave this a try but it did not work as expected. https://godbolt.org/z/Y897bxGa8

score/mw/com/types.h
KrishaDeshkool: added

score/mw/com/impl/field_tags.h
bemerybmw: I'm here referring to the public api of the target, not of mw/com. The impl namespace is what we use to denote internals (detail namespace is something which is only used within the header of a given file and should not be used by anyone else. We have an aou that says that users cannot use anything in the impl namespace. If a user ignores that, then there's no reason why they wouldn't also ignore the additional detail namespace.

score/mw/com/impl/field_tags.h
bemerybmw: See my comment above. If you still disagree, let's have a quick call.

score/mw/com/impl/field_tags.h
```diff
@@ -0,0 +1,53 @@
...
+namespace detail
+{
+
+template <typename TargetType, typename... Tags>
+struct contains_type : std::disjunction<std::is_same<TargetType, Tags>...>
+{
+};
+
+// SFINAE helper to check if a specific tag is contained in the provided tags.
+template <typename FieldType, typename ClassFieldType, typename TargetTag, typename... Tags>
```
bemerybmw: Can you please write a bit more documentation? Particularly why we have FieldType and ClassFieldType.

score/mw/com/impl/proxy_field.h
```diff
@@ -39,337 +39,404 @@ namespace score::mw::com::impl
...
 
   public:
     using FieldType = SampleDataType;
 
-    /// Constructor that allows to set the binding directly (both EnableGet and EnableSet are true).
-    ///
-    /// This is used for testing only. Allows for directly setting the bindings, and usually the mock binding is used
-    /// here.
-    template <bool EG = EnableGet, bool ES = EnableSet, typename = std::enable_if_t<EG && ES>>
+    /// \brief Testing ctor that accepts mock bindings directly.
```
bemerybmw: This isn't a testing ctor. It's also delegated to in production code. But it can only be called directly in test code.

score/mw/com/impl/proxy_field.h
```diff
@@ -39,337 +39,404 @@ namespace score::mw::com::impl
...
-    ///
-    /// This is used for testing only. Allows for directly setting the bindings, and usually the mock binding is used
-    /// here.
-    template <bool EG = EnableGet, bool ES = EnableSet, typename = std::enable_if_t<EG && !ES>>
-    ProxyField(ProxyBase& proxy_base,
-               std::unique_ptr<ProxyEventBinding<FieldType>> event_binding,
-               std::unique_ptr<ProxyMethodBinding> get_method_binding,
-               const std::string_view field_name,
-               detail::EnableGetOnlyTag = {})
+    /// \brief Production ctor used by generated code and end users.
```
bemerybmw: We don't have any code generators?

score/mw/com/impl/skeleton_field.h
```diff
@@ -35,37 +36,58 @@
...
-    /// \param field_name Field name of the field.
-    /// \param detail::EnableNeitherTag This parameter is only used for constructor overload disambiguation and
-    /// has no semantic meaning.
-    template <bool ES = EnableSet, typename = std::enable_if_t<!ES>>
-    SkeletonField(SkeletonBase& parent, const std::string_view field_name, detail::EnableNeitherTag = {});
-
-    /// Constructor that allows to set the binding directly.
-    ///
-    /// This is used only used for testing.
+    /// \brief Testing ctor: wraps the supplied mock event binding in a SkeletonEvent with nullptr method
```
bemerybmw: same point that this isn't test only but is delegated to

score/mw/com/impl/traits.h
bemerybmw: I've never heard this. Is it a commonly used term? Any reason why we can't just use public API since we use that in all other places?

score/mw/com/impl/field_tags.h
KrishaDeshkool: ah okay, i did not know about such AoUs.
okay, i saw this in proxy methods header that defines some methods in details and used across the other headers,
but, the difference there is that, all headers that share those methods are in same bazel target unlike mine.
okay, then i dont have any other points here.

KrishaDeshkool: pushed outside details.

score/mw/com/impl/proxy_field_base.h
```diff
@@ -51,124 +59,79 @@ class ProxyFieldBase
...
-     * \api
-     * \brief Removes any ReceiveHandler registered via SetReceiveHandler.
-     */
     Result<void> UnsetReceiveHandler() noexcept
     {
         return proxy_event_base_dispatch_->UnsetReceiveHandler();
     }
 
-  protected:
+    void Unsubscribe() noexcept
```
bemerybmw: Can you leave unsubscribe next to subscribe?

score/mw/com/impl/proxy_field_base.h
```diff
@@ -51,124 +59,79 @@ class ProxyFieldBase
...
     }
 
-    /**
-     * \api
-     * \brief Subscribe to the field.
-     * \param max_sample_count Specify the maximum number of concurrent samples that this event shall
-     *                          be able to offer to the using application.
-     * \return On failure, returns an error code.
-     */
+  protected:
```
bemerybmw: Can you please document why these are protected and why we need the view class below?

score/mw/com/impl/proxy_field_base.h
```diff
@@ -51,124 +59,79 @@ class ProxyFieldBase
...
+            proxy_event_base_dispatch_->Unsubscribe();
+        }
+    }
+
     std::reference_wrapper<ProxyBase> proxy_base_;
     ProxyEventBase* proxy_event_base_dispatch_;
     std::string_view field_name_;
 };
 
+/// \brief View pattern to access ProxyFieldBase's private members and functions from ProxyBase without making them
```
bemerybmw: This explains _what_ but it's important to explain _why_. i.e. we need this because we want to be able to gate the access to these methods to the user based on the field tags but still need to expose these to ProxyBase since it call Unsubscribe unconditionally (since it doesn't know the field tags) in deinit.

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -79,9 +207,80 @@ TEST(ProxyFieldTest, ProxyFieldContainsPublicFieldType)
...
+    // Given a ProxyField with the WithNotifier tag
+    // When checking the notifier API
+    // Then all notifier-related members should be present
+    using NotifierField = ProxyField<TestSampleType, WithNotifier>;
+    static_assert(has_subscribe<NotifierField>::value);
+    static_assert(has_unsubscribe<NotifierField>::value);
+    static_assert(has_get_subscription_state<NotifierField>::value);
+    static_assert(has_get_free_sample_count<NotifierField>::value);
+    static_assert(has_get_num_new_samples_available<NotifierField>::value);
+    static_assert(has_set_receive_handler<NotifierField>::value);
```
bemerybmw, line 251: missing state change handler checks?

score/mw/com/impl/skeleton_field.h
```diff
@@ -193,13 +210,74 @@ class SkeletonField : public SkeletonFieldBase
...
+        if constexpr (!kHasSetter)
         {
             return false;
         }
         return !is_set_handler_registered_;
     }
 
+    static constexpr bool kHasGetter = contains_type<WithGetter, Tags...>::value;
+    static constexpr bool kHasSetter = contains_type<WithSetter, Tags...>::value;
+    // No kHasNotifier: the SkeletonEvent is always built because it provides Update/Allocate.
```
bemerybmw: This should probably go inside the implementation of MakeSkeletonEvent

score/mw/com/impl/traits.h
```diff
@@ -100,19 +101,20 @@ class ProxyWrapperClassTestView;
...
 ///     typename Trait::template Method<ReturnType()> struct_method_2_{*this, method_name_1};
 ///
 /// };
-/// Notes regarding template args: A field has (besides its data type arg) three bool template args (enable_getter,
-/// enable_setter and enable_notifier). The enable_notifier template parameter is only relevant for certain bindings,
-/// e.g. the LoLa binding does not distinguish between true/false of this template parameter.
-/// A method has a template arg describing the method signature in the form ReturnType(InArgType1, InArgType2, ...).
+/// Notes regarding template args: a field takes its data type plus a pack of tags from {WithGetter, WithSetter,
+/// WithNotifier}. WithGetter / WithSetter enable Get() / Set() on the proxy. WithNotifier enables the proxy notifier
+/// API surface (Subscribe, Unsubscribe, GetSubscriptionState, GetFreeSampleCount, GetNumNewSamplesAvailable,
```
bemerybmw: same question here about surface

score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
```diff
@@ -79,22 +81,23 @@ class SkeletonWrapperClassTestView
...
     /// SkeletonWrapperClass
     /// @param event_mocks A tuple containing a NamedEventMock per event in the interface.
     /// @param field_mocks A tuple containing a NamedFieldMock per event in the interface.
     /// @return SkeletonWrapperClass containing mocked skeleton and mocked events / fields.
-    template <typename... EventTypes, typename... FieldTypes>
+    template <typename... EventTypes, typename... FieldMocks>
     static SkeletonWrapperClass Create(SkeletonBaseMock& skeleton_mock,
                                        std::tuple<NamedSkeletonEventMock<EventTypes>...>& event_mocks,
-                                       std::tuple<NamedSkeletonFieldMock<FieldTypes>...>& field_mocks)
+                                       std::tuple<FieldMocks...>& field_mocks)
```
bemerybmw: Why did you remove  NamedSkeletonFieldMock?

bemerybmw (changes requested):

score/mw/com/impl/proxy_field.h
KrishaDeshkool: this might sound like Jinja generators

score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
```diff
@@ -79,22 +81,22 @@ class SkeletonWrapperClassTestView
     /// fields_tuple);
     ///
     /// @tparam EventTypes Variadic template pack containing the types of ALL events specified in the interface
-    /// @tparam FieldTypes Variadic template pack containing the types of ALL fields specified in the interface
+    /// @tparam FieldMocks Variadic pack of NamedSkeletonFieldMock instantiations, one per field in the interface.
```
bemerybmw: Revert

score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
```diff
@@ -52,14 +52,16 @@ class NamedSkeletonEventMock
...
 };
 
-template <typename FieldType>
+template <typename FieldType, typename... Tags>
 class NamedSkeletonFieldMock
 {
   public:
     NamedSkeletonFieldMock(const std::string_view field_name_in) : field_name{field_name_in}, mock{} {}
 
+    using SampleType = FieldType;
```
bemerybmw: Is this still needed?

score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
```diff
@@ -79,22 +81,22 @@ class SkeletonWrapperClassTestView
...
     /// @tparam EventTypes Variadic template pack containing the types of ALL events specified in the interface
-    /// @tparam FieldTypes Variadic template pack containing the types of ALL fields specified in the interface
+    /// @tparam FieldMocks Variadic pack of NamedSkeletonFieldMock instantiations, one per field in the interface.
     /// @param skeleton_mock SkeletonBaseMock which will be injected into the SkeletonBase of the constructed
     /// SkeletonWrapperClass
     /// @param event_mocks A tuple containing a NamedEventMock per event in the interface.
     /// @param field_mocks A tuple containing a NamedFieldMock per event in the interface.
     /// @return SkeletonWrapperClass containing mocked skeleton and mocked events / fields.
-    template <typename... EventTypes, typename... FieldTypes>
+    template <typename... EventTypes, typename... FieldType, typename... Tags>
```
bemerybmw:
```suggestion
    template <typename... EventTypes, typename... FieldTypes, typename... Tags>
```

score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
```diff
@@ -156,25 +158,26 @@ class SkeletonWrapperClassTestView
...
-                                         std::tuple<NamedSkeletonFieldMock<FieldTypes>...>& field_mocks)
+                                         std::tuple<NamedSkeletonFieldMock<FieldTypes, Tags...>...>& field_mocks)
+
     {
         auto& events = SkeletonBaseView{skeleton}.GetEvents();
         auto& fields = SkeletonBaseView{skeleton}.GetFields();
 
-        // Note. this assert only checks if there were additional EventTypes / FieldTypes provided (or if an there are
-        // multiple EventTypes / FieldTypes of the same type, but at least one event / field of that type is not
+        // Note. this assert only checks if there were additional EventTypes / FieldMocks provided (or if an there are
```
bemerybmw: Revert

score/mw/com/impl/proxy_field_base.h
```diff
@@ -51,124 +59,91 @@ class ProxyFieldBase
...
 
-    /**
-     * \api
-     * \brief Removes any ReceiveHandler registered via SetReceiveHandler.
-     */
     Result<void> UnsetReceiveHandler() noexcept
     {
         return proxy_event_base_dispatch_->UnsetReceiveHandler();
     }
+    /// \}
```
bemerybmw, line 126: remove

score/mw/com/impl/proxy_field_base.h
```diff
@@ -51,124 +59,91 @@ class ProxyFieldBase
...
-     * \api
-     * \brief End subscription to a field and release needed resources.
-     * \details It is illegal to call this method while data is still held by the application in the form of SamplePtr.
-     *          Doing so will result in undefined behavior. After a call to this method, the field behaves as if it had
-     *          just been constructed.
-     */
     void Unsubscribe() noexcept
     {
-        proxy_event_base_dispatch_->Unsubscribe();
+        // if the WithNotifier tag is not set, proxy_event_base_dispatch_ will be nullptr.
```
bemerybmw: It seems a bit weird that we have this check here but all the other functions don't check proxy_event_base_dispatch_. Should we add a precondition check to the other functions?

score/mw/com/impl/proxy_field_base.h
```diff
@@ -51,124 +59,91 @@ class ProxyFieldBase
...
     std::reference_wrapper<ProxyBase> proxy_base_;
     ProxyEventBase* proxy_event_base_dispatch_;
     std::string_view field_name_;
 };
 
+/// \brief Controlled access to ProxyFieldBase internals for use within the impl namespace.
+///
+/// ProxyBase needs to call certain ProxyFieldBase methods (e.g. \c Unsubscribe) during its own
+/// lifecycle, but those methods must not be reachable by end users of the public API.  Making
+/// them \c public would expose them too broadly, making ProxyBase a \c friend directly would
```
bemerybmw: Should mention why we don't want to publicly expose them. i.e. because the ProxyField which inherits from is class decides which functions are exposed to the user based on its template params.

score/mw/com/impl/skeleton_field.h
```diff
@@ -193,13 +210,74 @@ class SkeletonField : public SkeletonFieldBase
...
         return !is_set_handler_registered_;
     }
 
+    static constexpr bool kHasGetter = contains_type<WithGetter, Tags...>::value;
+    static constexpr bool kHasSetter = contains_type<WithSetter, Tags...>::value;
+
+    static std::unique_ptr<SkeletonEvent<FieldType>> MakeSkeletonEvent(SkeletonBase& parent,
+                                                                       const std::string_view field_name)
+    {
+        // No kHasNotifier: the SkeletonEvent is always built because it provides Update/Allocate.
```
bemerybmw, line 226:
```suggestion
        // No kHasNotifier: the SkeletonEvent is always built because it provides actually set the value of the field via Update/Allocate. This is either needed to be used directly by the user via the event api in case WithNotifier is provided or via the set handler if WithSetter is enabled.
```

score/mw/com/impl/skeleton_field.h
```diff
@@ -35,37 +36,58 @@
 namespace score::mw::com::impl
 {
 
-template <typename SampleDataType, const bool EnableSet = false, const bool EnableNotifier = false>
+template <typename SampleDataType, typename... Tags>
 class SkeletonField : public SkeletonFieldBase
 {
+    // SkeletonField must enable WithGetter or WithNotifier. Without one the consumer has no way to
+    // observe the value, so the field is useless. WithSetter alone does not count.
+    static_assert(
```
bemerybmw, line 44: Do we also need a check that WithNotifier or WithSetter is enabled? Because otherwise the user has no way of setting the value of the field.

score/mw/com/impl/proxy_field_base.h
```diff
@@ -38,10 +38,10 @@ class ProxyFieldBase
     friend class ProxyFieldBaseView;
 
   public:
+    /// \param proxy_event_base_dispatch May be nullptr when the field's tag pack does not include WithNotifier.
```
bemerybmw, line 41: In MakeSkeletonEvent, you always create the SkeletonEvent, regardless of WithNotifier. This seems to contradict this change ("mw/com: allow null dispatch in ProxyFieldBase when WithNotifier is absent")

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: now that i rendered, i was using wrong type of comment tag, it should be /name not /note,
with that this is how it would look like
image

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: wait, maybe i was confused before,
we do not gate the Update() method based on the Notifier tag presence on skeleton side, 
so there is no restriction on the skeleton side to set/update a value,

originally we only had the static assert on the consumer side.
This whole assertion is only relavent on the consumer side, we would need atleast Notifer methods(subscribe and so on...) or getter(Get method) for the consumer to get any value,

then we mirrored to have the same static assert on the skeleton side because (eventhough we use same interface AsProxy and AsSkeleton so if we set some tags on proxy we would have same on skeleton side)
we added these asserts on skeleton side for symmetry and avoid situations when proxy and skelton are somehow instantiated with different tags.

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: `In MakeSkeletonEvent, you always create the SkeletonEvent, `
Because, we do not gate the Update() on the skeleton side so there is always a way for the skeleton side to set a value.
but on the proxy side.
The MakeProxyEvent is not always creating ProxyEvent because if user is not interested in notifier methods like (Subscribe, ....) then he can get to know the Field value with Get methods (that is why we have an assert on the proxy side that either notifier or getter should be enabled).

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: i think it is quite obvious if a derived class is delegating calls to the base class protected method it means it is trying to restrict direct calls to base class methods based on some condition(something being the tag gating in this case)
but, i added your reasoning  instead of saying exposing too broadly.

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: this is due to our current architecture, we have to call this method from Deinit() from proxyBase and since ProxyBase does not know about which tags are enabled we have this check here so the Unsub() call is idempotent from PorxyBase POV, 
what we can do is we can push this check to the View class below because, there is no way a call from ProxyField can happen when Notifer Tag is disabled. (so it would never be called with null proxy_event_dispatch_)

score/mw/com/impl/mocking/skeleton_wrapper_class_test_view.h
KrishaDeshkool: nope

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: i dont get this point,
`This is either needed to be used directly by the user via the event api in case WithNotifier`
This is not what we agreed on before, we said that we do not have any gating for Update/Allocate methods.

if we gate the Update/Allocate with a notifer tag then if a skeleon was created with (Setter and Getter tags),
then is there a way to set a initial value ?
i think with SetHandler it is not possible to set the initial value, no?

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: agreed to update the documentation to make it explicit about the behaviour of these tags, added a section in readme

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: agreed to update the documentation to make it explicit about the behaviour of these tags, added a section in readme

score/mw/com/impl/skeleton_field.h
KrishaDeshkool: agreed to update the documentation to make it explicit about the behaviour of these tags, added a section in readme

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: pushed the checks to the View class since that is only place where denint call comes to,
the real Unsub does not  need these extra check since it is impossible to get called from field method public API with a null dispatch (when notifier is disabled there wont be any call to unsub and when there is a notifier there would always be a event_dispath (we have an assertion otherwise)).

score/mw/com/impl/proxy_field_base.h
```diff
@@ -52,123 +60,93 @@ class ProxyFieldBase
...
-     * \return On failure, returns an error code.
+     * \name Tag-gated interface
+     *
+     * These methods are protected rather than public because the derived \c ProxyField class
+     * template decides which of them are exposed to the end user, based on the capability tags
+     * (\c WithGetter, \c WithSetter, \c WithNotifier) the field was declared with.  Keeping them
+     * protected here prevents direct access while still letting the derived class forward them
+     * selectively — only when the matching tag is present.
      */
+  protected:
```
bemerybmw: protected: is in the middle of the comment and doxygen scope?

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: both would produce a same render on doxygen, but i will move it out of protected for clarity.

score/mw/com/impl/proxy_field_base.h
KrishaDeshkool: it would look like this

bemerybmw (approved):

PR #358, opened 2026-04-24
Send notification to Skeleton on Proxy destruction to "Unsubscribe" method.

Tejveerpratap2803:
**Notify Skeleton on Proxy destruction to unsubscribe method resources**

When a Proxy is destroyed, TeardownMethods() now sends an UnsubscribeServiceMethod IPC message to the Skeleton, which removes the proxy's shared memory region and unregisters all method call handlers for that proxy.

- Added UnsubscribeServiceMethod IPC path (local + remote) and RegisterOnServiceMethodUnsubscribedHandler to IMessagePassingService
- Skeleton registers the unsubscription handler in PrepareOffer; on notification, it calls MethodResourceMap::Remove and unregisters all associated method call handlers
- Added MethodUnsubscriptionRegistrationGuard RAII guard to manage handler lifetime; destroyed in PrepareStopOffer
- Added SkeletonMethod::OnProxyMethodUnsubscribeFinished to complete the unsubscription flow
- Refactored registration_guards_ from an ApplicationId-keyed map (with inner vector of guards) to a flat unordered_map<ProxyMethodInstanceIdentifier, pair<pid_t, MethodCallRegistrationGuard>> — one entry per proxy method instance; unsubscribe now erases by the full ProxyMethodInstanceIdentifier, preventing accidental removal of entries from other proxies sharing the same ApplicationId, while crash-detection cleanup still iterates by application_id extracted from the key to correctly remove all stale entries from a crashed process
- Added unit tests covering all new IPC paths, the unsubscription flow, and the two-proxies-from-the-same-application scenario

### DIFF
--- a/score/mw/com/impl/bindings/lola/messaging/i_message_passing_service.h
+++ b/score/mw/com/impl/bindings/lola/messaging/i_message_passing_service.h
@@ -16,6 +16,7 @@
 #include "score/mw/com/impl/bindings/lola/element_fq_id.h"
 #include "score/mw/com/impl/bindings/lola/messaging/method_call_registration_guard.h"
 #include "score/mw/com/impl/bindings/lola/messaging/method_subscription_registration_guard.h"
+#include "score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard.h"
 #include "score/mw/com/impl/bindings/lola/methods/proxy_method_instance_identifier.h"
 #include "score/mw/com/impl/bindings/lola/proxy_instance_identifier.h"
 #include "score/mw/com/impl/bindings/lola/skeleton_instance_identifier.h"
@@ -47,6 +48,7 @@ class IMessagePassingService
 {
     friend class MethodSubscriptionRegistrationGuardFactory;
     friend class MethodCallRegistrationGuardFactory;
+    friend class MethodUnsubscriptionRegistrationGuardFactory;
 
   public:
     using HandlerRegistrationNoType = std::uint32_t;
@@ -85,6 +87,19 @@ class IMessagePassingService
     ///        restarted.
     using ServiceMethodSubscribedHandler = safecpp::CopyableScopedFunction<
         score::Result<void>(ProxyInstanceIdentifier proxy_instance_identifier, uid_t proxy_uid, pid_t proxy_pid)>;
+
+    /// \brief Handler which will be called when the proxy process sends a message that it has unsubscribed from a
+    /// service method.
+    ///
+    /// When a Proxy is destroyed, it will call UnsubscribeServiceMethod which will send a message to the Skeleton
+    /// process. The Skeleton process will then call the ServiceMethodUnsubscribedHandler registered in
+    /// RegisterOnServiceMethodUnsubscribedHandler. The handler should close the proxy's shared memory region and
+    /// unregister the method call handler corresponding to the proxy's ProxyMethodInstanceIdentifier from each
+    /// SkeletonMethod.
+    ///
+    /// \param proxy_instance_identifier unique identifier of the Proxy instance which unsubscribed from the method.
+    using ServiceMethodUnsubscribedHandler =
+        safecpp::CopyableScopedFunction<score::ResultBlank(ProxyInstanceIdentifier proxy_instance_identifier)>;
 
     /// \brief Handler which will be called when the proxy process sends a message that it has called a method.
     ///
@@ -197,6 +212,24 @@ class IMessagePassingService
         ServiceMethodSubscribedHandler subscribed_callback,
         AllowedConsumerUids allowed_proxy_uids) = 0;
 
+    /// \brief Register a handler on Skeleton side which will be called when UnsubscribeServiceMethod is called by a
+    /// Proxy.
+    ///
+    /// When a Proxy is destroyed, it will call UnsubscribeServiceMethod. When this message is received in the
+    /// Skeleton process, the handler registered in this function will be called.
+    ///
+    /// The handler is scoped to on_service_method_subscribed_handler_scope_ on the Skeleton side, so that it is
+    /// not called after the Skeleton has started its StopOffer sequence. This is our synchronisation point.
+    ///
+    /// \param asil_level ASIL level of method.
+    /// \param skeleton_instance_identifier to identify which ServiceMethodUnsubscribedHandler to call when
+    ///        UnsubscribeServiceMethod is called on the Proxy side
+    /// \param unsubscribed_callback callback that will be called when UnsubscribeServiceMethod is called
+    virtual Result<MethodUnsubscriptionRegistrationGuard> RegisterOnServiceMethodUnsubscribedHandler(
+        const QualityType asil_level,
+        const SkeletonInstanceIdentifier skeleton_instance_identifier,
+        ServiceMethodUnsubscribedHandler unsubscribed_callback) = 0;
+
     /// \brief Register a handler on Skeleton side which will be called when CallMethod is called by a ProxyMethod.
     ///
     /// When a user calls a method on a ProxyMethod, it will put the InArgs in shared memory (if there are any) and then
@@ -283,6 +316,21 @@ class IMessagePassingService
                                                 const ProxyInstanceIdentifier& proxy_instance_identifier,
                                                 const pid_t target_node_id) = 0;
 
+    /// \brief Best-effort call made by a Proxy on destruction to notify the Skeleton that the proxy is being
+    /// destroyed so that the Skeleton can clean up the proxy's shared memory region.
+    ///
+    /// This mirrors SubscribeServiceMethod. If this call fails, some resources will be leaked on the Skeleton side
+    /// but will be cleaned up when the Proxy process restarts and the Skeleton detects the PID change.
+    ///
+    /// \param asil_level ASIL level of method.
+    /// \param skeleton_instance_identifier identification of the Skeleton corresponding to the Proxy.
+    /// \param proxy_instance_identifier identification of the Proxy which is being destroyed.
+    /// \param target_node_id PID of the Skeleton process which the unsubscribe call is sent to.
+    virtual ResultBlank UnsubscribeServiceMethod(const QualityType asil_level,
+                                                 const SkeletonInstanceIdentifier& skeleton_instance_identifier,
+                                                 const ProxyInstanceIdentifier& proxy_instance_identifier,
+                                                 const pid_t target_node_id) = 0;
+
     /// \brief Blocking call which is called on Proxy side to trigger the Skeleton to process a method call. The
     /// callback registered with RegisterOnServiceMethodSubscribed will be called on the Skeleton side and a response
     /// will be returned
@@ -330,6 +378,19 @@ class IMessagePassingService
     /// \param proxy_method_instance_identifier to identify which registered MethodCallHandler to unregister
     virtual void UnregisterMethodCallHandler(const QualityType asil_level,
                                              ProxyMethodInstanceIdentifier proxy_method_instance_identifier) = 0;
+
+    /// \brief Unregister handler that was registered with RegisterOnServiceMethodUnsubscribedHandler
+    ///
+    /// This function is private and will only be called by MethodUnsubscriptionRegistrationGuardFactory on
+    /// destruction.
+    ///
+    /// \pre Shall only be called after RegisterOnServiceMethodUnsubscribedHandler was successfully called.
+    ///
+    /// \param asil_level ASIL level of method.
+    /// \param skeleton_instance_identifier to identify which registered ServiceMethodUnsubscribedHandler to unregister
+    virtual void UnregisterOnServiceMethodUnsubscribedHandler(
+        const QualityType asil_level,
+        SkeletonInstanceIdentifier skeleton_instance_identifier) = 0;
 };
 
 }  // namespace score::mw::com::impl::lola
--- a/score/mw/com/impl/bindings/lola/messaging/i_message_passing_service.h
+++ b/score/mw/com/impl/bindings/lola/messaging/i_message_passing_service.h
@@ -92,12 +92,10 @@ class IMessagePassingService
     /// service method.
     ///
     /// When a Proxy is destroyed, it will call UnsubscribeServiceMethod which will send a message to the Skeleton
-    /// process. The Skeleton process will then call the ServiceMethodUnsubscribedHandler registered in
-    /// RegisterOnServiceMethodUnsubscribedHandler. The handler should close the proxy's shared memory region and
-    /// unregister the method call handler corresponding to the proxy's ProxyMethodInstanceIdentifier from each
-    /// SkeletonMethod.
+    /// process. The Skeleton process will then call the ServiceMethodUnsubscribedHandler to close the proxy's shared
+    /// memory region and unregister the method call handler corresponding to the proxy's ProxyMethodInstanceIdentifier
+    /// from each SkeletonMethod.
     ///
-    /// \param proxy_instance_identifier unique identifier of the Proxy instance which unsubscribed from the method.
     using ServiceMethodUnsubscribedHandler =
         safecpp::CopyableScopedFunction<score::ResultBlank(ProxyInstanceIdentifier proxy_instance_identifier)>;
 
--- a/score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp
+++ b/score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp
@@ -66,6 +66,12 @@ struct SubscribeServiceMethodUnserializedPayload
     ProxyInstanceIdentifier proxy_instance_identifier;
 };
 
+struct UnsubscribeServiceMethodUnserializedPayload
+{
+    SkeletonInstanceIdentifier skeleton_instance_identifier;
+    ProxyInstanceIdentifier proxy_instance_identifier;
+};
+
 struct MethodCallUnserializedPayload
 {
     ProxyMethodInstanceIdentifier proxy_method_instance_identifier;
@@ -411,6 +417,10 @@ score::Result<void> MessagePassingServiceInstance::MessageCallbackWithReply(
         {
             return HandleSubscribeServiceMethodMsg(payload, sender_uid, sender_pid);
         }
+        case score::cpp::to_underlying(MessageWithReplyType::kUnsubscribeServiceMethod):
+        {
+            return HandleUnsubscribeServiceMethodMsg(payload, sender_pid);
+        }
         case score::cpp::to_underlying(MessageWithReplyType::kCallMethod):
         {
             return HandleCallMethodMsg(payload, sender_uid);
@@ -605,6 +615,20 @@ score::Result<void> MessagePassingServiceInstance::HandleSubscribeServiceMethodM
                                              sender_node_id);
 }
 
+score::ResultBlank MessagePassingServiceInstance::HandleUnsubscribeServiceMethodMsg(
+    const score::cpp::span<const std::uint8_t> payload,
+    const pid_t /*sender_node_id*/)
+{
+    UnsubscribeServiceMethodUnserializedPayload unserialized_payload{};
+    if (!DeserializeFromPayload(payload, unserialized_payload))
+    {
+        return MakeUnexpected(MethodErrc::kUnexpectedMessageSize);
+    }
+
+    return CallUnsubscribeServiceMethodLocally(unserialized_payload.skeleton_instance_identifier,
+                                               unserialized_payload.proxy_instance_identifier);
+}
+
 score::Result<void> MessagePassingServiceInstance::HandleCallMethodMsg(
     const score::cpp::span<const std::uint8_t> payload,
     const uid_t sender_uid)
@@ -657,6 +681,36 @@ score::Result<void> MessagePassingServiceInstance::CallSubscribeServiceMethodLoc
             << "Invocation of subscribe service method handler failed as scope has been destroyed: "
                "SkeletonMethod has already been destroyed.";
         return MakeUnexpected(MethodErrc::kSkeletonAlreadyDestroyed);
+    }
+    return invocation_result.value();
+}
+
+score::ResultBlank MessagePassingServiceInstance::CallUnsubscribeServiceMethodLocally(
+    const SkeletonInstanceIdentifier& skeleton_instance_identifier,
+    const ProxyInstanceIdentifier& proxy_instance_identifier)
+{
+    // A copy of the handler is made under lock and called outside the lock. See CallSubscribeServiceMethodLocally for
+    // a detailed explanation of the pattern.
+    std::shared_lock<std::shared_mutex> read_lock{unsubscribe_service_method_handlers_mutex_};
+    auto handler_it = unsubscribe_service_method_handlers_.find(skeleton_instance_identifier);
+    if (handler_it == unsubscribe_service_method_handlers_.cend())
+    {
+        // The skeleton may have already called StopOffer, which expired the scope used for the handler, or may not
+        // have registered a handler at all. This is a best-effort call so we treat this as success.
+        mw::log::LogInfo("lola") << "Unsubscribe method handler has not been registered for this SkeletonMethod. "
+                                    "Skeleton may have already stopped offering.";
+        return {};
+    }
+
+    auto handler_copy = handler_it->second;
+    read_lock.unlock();
+
+    const auto invocation_result = std::invoke(handler_copy, proxy_instance_identifier);
+    if (!(invocation_result.has_value()))
+    {
+        mw::log::LogInfo("lola") << "Invocation of unsubscribe service method handler was a no-op: scope has been "
+                                    "destroyed (skeleton already stopped offering).";
+        return {};
     }
     return invocation_result.value();
 }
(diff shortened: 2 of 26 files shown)


gdadunashvili (commented):
For the future reference please split your PRs in thematically meaningful commits to make the review easier.
In the description you outline 6 major bullet points that have been addressed in this PR. Some of this points are sufficiently different that they could have gotten their own commits.

score/mw/com/impl/bindings/lola/messaging/i_message_passing_service.h
```diff
@@ -86,6 +88,19 @@ class IMessagePassingService
...
         score::Result<void>(ProxyInstanceIdentifier proxy_instance_identifier, uid_t proxy_uid, pid_t proxy_pid)>;
 
+    /// \brief Handler which will be called when the proxy process sends a message that it has unsubscribed from a
+    /// service method.
+    ///
+    /// When a Proxy is destroyed, it will call UnsubscribeServiceMethod which will send a message to the Skeleton
+    /// process. The Skeleton process will then call the ServiceMethodUnsubscribedHandler registered in
+    /// RegisterOnServiceMethodUnsubscribedHandler. The handler should close the proxy's shared memory region and
+    /// unregister the method call handler corresponding to the proxy's ProxyMethodInstanceIdentifier from each
+    /// SkeletonMethod.
```
gdadunashvili:
```suggestion
    /// When a Proxy is destroyed, it will call UnsubscribeServiceMethod which will send a message to the Skeleton
    /// process. The Skeleton process will then call the ServiceMethodUnsubscribedHandler to close the proxy's shared memory region and
    /// unregister the method call handler corresponding to the proxy's ProxyMethodInstanceIdentifier from each
    /// SkeletonMethod.
```

gdadunashvili: I think putting too much detail in code documentation can be detrimental. I would only keep things that are necessary to understand the purpose of the described code

score/mw/com/impl/bindings/lola/messaging/i_message_passing_service.h
```diff
@@ -86,6 +88,19 @@ class IMessagePassingService
...
+    ///
+    /// \param proxy_instance_identifier unique identifier of the Proxy instance which unsubscribed from the method.
```
gdadunashvili:
```suggestion
```

gdadunashvili: We should not add excessive documentation that does not add information. 
The parameter name proxy_instance_identifier already contains the information that is added as a comment

score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    // destruction of the guard and that the scope of the guard is correctly handled. The more complex tests about
+    // testing whether the handler is called when move constructing / move assigning the guard is handled in the tests
+    // for ScopeExit.
+    static_assert(std::is_same_v<MethodUnsubscriptionRegistrationGuard,
+                                 utils::ScopeExit<safecpp::MoveOnlyScopedFunction<void()>>>);
+}
+
+TEST_F(MethodUnsubscriptionRegistrationGuardFixture, CreatingGuardDoesNotCallUnregister)
+{
+    // When creating a MethodUnsubscriptionRegistrationGuard
```
gdadunashvili:
```suggestion
```

gdadunashvili: Please do not mix up the given/when/then clauses.
I know this is just a semantic difference but keeping it consistent makes it easier to read and review tests.

gdadunashvili: If a fixture function has a given in a name and has a descriptive name then we treat it as a given clause and usually omit the comment

score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    GivenAMethodUnsubscriptionRegistrationGuard();
+
+    // Expecting that UnregisterOnServiceMethodUnsubscribedHandler is never called
```
gdadunashvili:
```suggestion
    // When UnregisterOnServiceMethodUnsubscribedHandler is never called
```

gdadunashvili: Here it could have been either. Sometimes we have Expect clauses instead of when clauses. In this particular case I think when fits better, but this is not a strict rule

score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard_test.cpp
```diff
@@ -0,0 +1,116 @@
...
+    // Then UnregisterOnServiceMethodUnsubscribedHandler is not called
+    EXPECT_FALSE(unregister_on_service_method_unsubscribed_handler_called_);
+}
+
+TEST_F(MethodUnsubscriptionRegistrationGuardFixture, DestroyingGuardCallsUnregister)
+{
+    GivenAMethodUnsubscriptionRegistrationGuard();
+
+    // Expecting that UnregisterOnServiceMethodUnsubscribedHandler is called with the asil_level and
+    // SkeletonInstanceIdentifier used to create the guard
```
gdadunashvili:
```suggestion
    // Expecting that UnregisterOnServiceMethodUnsubscribedHandler is called
```

gdadunashvili: It is not necessary to describe the signature of the call in the expect clause, unless the passed parameters directly relate to the test

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp
```diff
@@ -1210,8 +1326,8 @@ void MessagePassingServiceInstance::RegisterEventNotificationRemote(const Elemen
         {
             score::mw::log::LogError("lola")
                 << "MessagePassingService: RegisterEventNotificationRemote called for event" << event_id.ToString()
-                << "and node_id" << target_node_id << "although event is "
-                << " currently located at node" << registration_count_inserted.first->second.node_id;
+                << "and node_id" << target_node_id << "although event is " << " currently located at node"
```
gdadunashvili:
```suggestion
                << "and node_id" << target_node_id << "although event is currently located at node"
```

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.h
```diff
@@ -136,7 +147,8 @@ class MessagePassingServiceInstance : public IMessagePassingServiceInstance
     enum class MessageWithReplyType : std::uint8_t
     {
         kSubscribeServiceMethod = 1U,
-        kCallMethod,
+        kCallMethod = 2U,
+        kUnsubscribeServiceMethod = 3U,
```
gdadunashvili:
```suggestion
        kCallMethod,
        kUnsubscribeServiceMethod,
```

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance_methods_test.cpp
```diff
@@ -1330,5 +1359,272 @@ TEST_F(MessagePassingServiceInstanceUnregisterSubscribeMethodHandlerTest,
         unit_->UnregisterOnServiceMethodSubscribedHandler(kSkeletonInstanceIdentifier));
 }
 
+using MessagePassingServiceInstanceRegisterUnsubscribeHandlerTest = MessagePassingServiceInstanceMethodsFixture;
+TEST_F(MessagePassingServiceInstanceRegisterUnsubscribeHandlerTest, ReregisteringHandlerReturnsError)
```
gdadunashvili:
```suggestion
TEST_F(MessagePassingServiceInstanceRegisterUnsubscribeHandlerTest, ReregisteringHandlerWhenOneIsAlreadyRegisteredReturnsError)
```

score/mw/com/impl/bindings/lola/methods/method_resource_map.h
```diff
@@ -79,6 +79,13 @@ class MethodResourceMap
     auto CleanUpOldRegions(const ProxyInstanceIdentifier proxy_instance_identifier, const pid_t proxy_pid)
         -> CleanUpResult;
 
+    /// \brief Removes the ISharedMemoryResource corresponding to the provided ProxyInstanceIdentifier.
+    ///
+    /// This is called on Skeleton side when the Proxy notifies that it is being destroyed, so that the Skeleton can
+    /// close the proxy's shared memory region. If the provided ProxyInstanceIdentifier does not exist in the map, this
+    /// is a no-op (it may have already been cleaned up e.g. by CleanUpOldRegions).
```
gdadunashvili:
```suggestion
    /// is a no-op (it may have already been cleaned up).
```

gdadunashvili: I am not sure if `CleanUpOldRegions` would ever interfere with `Remove`. Remove will be called at the end of the life-time of a proxy that is shutting down gracefully. CleanUpOldRegions will be called after a restart

score/mw/com/impl/bindings/lola/methods/method_resource_map_test.cpp
```diff
@@ -246,6 +246,71 @@ TEST_F(MethodResourceMapContainsFixture, ContainsReturnsTrueWhenElementMatchingK
...
+                                                kDummyPid1));
+}
+
+TEST_F(MethodResourceMapRemoveFixture, RemoveForNonExistentEntryIsANoOp)
+{
+    GivenAMethodResourceMap().WithAnInsertedRegion(ProxyInstanceIdentifier{kProcessIdentifier1, kProxyInstanceCounter1},
+                                                   kDummyPid1);
+
+    // When removing an element that was never inserted (different process identifier)
+    // Then the program should not terminate and should remain a no-op
```
gdadunashvili:
```suggestion
```

score/mw/com/impl/bindings/lola/methods/method_resource_map_test.cpp
```diff
@@ -246,6 +246,71 @@ TEST_F(MethodResourceMapContainsFixture, ContainsReturnsTrueWhenElementMatchingK
...
+    method_resource_map_->Remove(ProxyInstanceIdentifier{kProcessIdentifier2, kProxyInstanceCounter1});
+
+    // and the originally inserted element should still be contained in the map
```
gdadunashvili:
```suggestion
    // Then the program should not terminate and the originally inserted element should still be contained in the map
```

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -772,6 +739,44 @@ score::Result<void> Proxy::SetupMethods()
...
+    }
+
+    // Subscription state is tracked directly in Proxy (not via ProxyMethod references) because ProxyMethod
+    // objects may have already been destroyed by the time TeardownMethods() is called from ~Proxy().
+    const bool is_subscribed = are_proxy_methods_subscribed_.load();
+    if (is_subscribed)
+    {
+        // Best-effort: notify the Skeleton to clean up its registration guards for this Proxy's methods.
+        // On failure the guards remain until the Skeleton stops offering or restarts, at which point
+        // they will be cleaned up.
```
gdadunashvili:
```suggestion
```

gdadunashvili: I would not add a comment here on how the Skeleton menages it's resources internally, since that can change and no one will remember to update this comment

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -347,11 +367,32 @@ auto Skeleton::PrepareOffer(SkeletonEventBindings& events,
...
+            QualityType::kASIL_B,
+            skeleton_instance_identifier,
+            IMessagePassingService::ServiceMethodUnsubscribedHandler{
+                on_service_method_subscribed_handler_scope_,
+                [this](const ProxyInstanceIdentifier proxy_instance_identifier) -> ResultBlank {
+                    return OnServiceMethodsUnsubscribed(proxy_instance_identifier);
+                }});
+        if (!(asil_b_unsubscription_result.has_value()))
+        {
+            method_subscription_registration_guard_qm_.reset();
```
gdadunashvili: why are the qm subscription and unsubscription guards reset here as well?

score/mw/com/impl/bindings/lola/skeleton_method.cpp
```diff
@@ -151,17 +151,23 @@ void SkeletonMethod::Call(const std::optional<score::cpp::span<std::byte>> in_ar
...
-    auto found_handler_cleanup_package_it = registration_guards_.find(application_id);
-    if (found_handler_cleanup_package_it == registration_guards_.end())
-    {
-        return;
-    };
-
-    if (found_handler_cleanup_package_it->second.proxy_pid != proxy_pid)
+    // Linear scan to erase all stale guards from a crashed application (same application_id, different pid).
+    // If benchmarking identifies this as a bottleneck, the data structure can be revisited.
+    for (auto it = registration_guards_.begin(); it != registration_guards_.end();)
```
gdadunashvili: This whole algorithm needs to change. We should not use raw iterator loops and we definitely should not modify the length of an array while iterating on it!

gdadunashvili (changes requested):

score/mw/com/impl/bindings/lola/skeleton_method.cpp
gdadunashvili: I would suggest using the erase-remove idiom https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

That way you avoid modifying the `lenght` while iterating,
and also avoid pointer arithmetic.

score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/messaging/method_unsubscription_registration_guard_test.cpp
Tejveerpratap2803: DOne

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.h
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance_methods_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/methods/method_resource_map_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/methods/method_resource_map_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/proxy.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/methods/method_resource_map.h
Tejveerpratap2803: Make sense :)

score/mw/com/impl/bindings/lola/skeleton.cpp
Tejveerpratap2803: For this case, If ASIL-B unsubscription registration fails, we're returning an error from PrepareOffer() meaning the whole offer failed. But by that point, the QM handlers were already successfully registered. If we don't reset those guards, they stay alive and keep firing even though the skeleton never fully offered. So we clean them up here to undo everything and leave no partial state behind.

score/mw/com/impl/bindings/lola/skeleton_method.cpp
Tejveerpratap2803: Done, Updated the CleanUpOldHandlers() logic.

gdadunashvili (approved):

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -772,6 +739,42 @@ score::Result<void> Proxy::SetupMethods()
...
+                << __func__ << " " << __LINE__
+                << " TeardownMethods: UnsubscribeServiceMethod failed with error: " << result.error();
+        }
+    }
+
+    // Unlink the method SHM name from the filesystem, then release our mapping.
+    const auto method_shm_path_name = GetMethodChannelShmName();
+    memory::shared::SharedMemoryFactory::Remove(method_shm_path_name);
+    method_shm_resource_.reset();
+    memory::shared::SharedMemoryFactory::RemoveStaleArtefacts(method_shm_path_name);
```
bemerybmw: I don't think this should be called here. It's basically doing the same thing as Remove above except it's intended to be called when the memory region was not created by the process calling it (which is not the case here).

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -772,6 +739,42 @@ score::Result<void> Proxy::SetupMethods()
...
+{
+    // Skip teardown if method SHM was never created (method_shm_resource_ is null, nothing to clean up).
+    if (!are_proxy_methods_setup_.load())
+    {
+        return;
+    }
+
+    // Subscription state is tracked directly in Proxy (not via ProxyMethod references) because ProxyMethod
+    // objects may have already been destroyed by the time TeardownMethods() is called from ~Proxy().
+    const bool is_subscribed = are_proxy_methods_subscribed_.load();
```
bemerybmw, line 748: We currently have a race condition between are_proxy_methods_subscribed_ being set and checking it here. It's possible that after a skeleton re-offers, we will call SubscribeServiceMethod and before we set are_proxy_methods_subscribed_ to true, the destructor is called and we check the value here which would still be false.

I think we want to make sure that this function is never called concurrently with the ServiceAvailabilityChangeHandler to avoid any race conditions. So before TeardownMethods is called, we should call find_service_guard_.reset() or better yet, create a guard object which calls the methods cleanup on destruction and ensure that it is destroyed after the find_service_guard_.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -333,6 +329,26 @@ auto Skeleton::PrepareOffer(SkeletonEventBindings& events,
     }
     method_subscription_registration_guard_qm_.emplace(std::move(qm_registration_result).value());
 
+    // Register an unsubscription handler for QM proxies. The handler uses the same scope as the subscribe handler
```
bemerybmw: The PrepareOffer function is getting extremely long. Can you please create a new function called RegisterMethodHandlers or something similar? I think it should take a quality type and return a pair of subscribe / unsubscribe handler registration guards. You can then call it for qm and optionally for asil b and store the results in member variables if there are no errors in either (that way you don't have to manually reset the qm guards in case the asil b registration fails).

Tejveerpratap2803: Done.

score/mw/com/impl/bindings/lola/proxy.cpp
Tejveerpratap2803: Done.

score/mw/com/impl/bindings/lola/proxy.cpp
Tejveerpratap2803: Done.

score/mw/com/impl/bindings/lola/proxy.h
```diff
@@ -259,10 +260,38 @@ class Proxy : public ProxyBinding
...
+
     score::filesystem::Filesystem filesystem_;
 
-    // We make find_service_guard_ the last member variable since it registers a handler which accesses member variables
-    // of this class, so they should be initialised first.
+    // RAII guard that invokes TeardownMethods() on destruction. It must be declared before find_service_guard_ so
+    // that, in reverse-destruction order, find_service_guard_ is torn down first (stopping
+    // ServiceAvailabilityChangeHandler callbacks) before TeardownMethods() runs, eliminating the race on
+    // are_proxy_methods_subscribed_.
+    class MethodsCleanupGuard
```
bemerybmw: Please use scope_exit instead of creating the guard yourself ( https://github.com/eclipse-score/baselibs/blob/main/score/scope_exit/scope_exit.h). You can create an alias to a scope exit and create a factory like we do in https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/messaging/method_subscription_registration_guard.h. You can put this in the anonymous namespace of this file next to FindServiceGuard.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -310,54 +310,32 @@ auto Skeleton::PrepareOffer(SkeletonEventBindings& events,
...
+        RegisterMethodHandlers(QualityType::kASIL_QM, skeleton_instance_identifier, lola_message_passing);
+    if (!qm_handlers_result.has_value())
     {
-        score::mw::log::LogError("lola") << "Could not register QM service method handler. Returning error.";
-        return MakeUnexpected<void>(qm_registration_result.error());
+        return MakeUnexpected<void>(qm_handlers_result.error());
     }
-    method_subscription_registration_guard_qm_.emplace(std::move(qm_registration_result).value());
+
+    auto qm_handlers = std::move(qm_handlers_result).value();
```
bemerybmw: Put this below right before it's used.

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/skeleton.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/proxy.h
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -307,6 +307,24 @@ class FindServiceGuard final
...
 };
 
+namespace method_cleanup_guard
+{
+
+using ProxyMethodsCleanupGuard = utils::ScopeExit<score::cpp::callback<void()>>;
+
+}  // namespace method_cleanup_guard
+
+class ProxyMethodsCleanupGuardFactory
```
bemerybmw: Just noticed this but probably the names of "TeardownMethods" and this factory should be aligned. E.g. Change TeardownMethods to CleanupMethods.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -772,6 +795,42 @@ score::Result<void> Proxy::SetupMethods()
     return subscription_result;
 }
 
+void Proxy::TeardownMethods() noexcept
```
bemerybmw: As discussed in the call, you should do the same thing that I mentioned in point 2 of https://github.com/eclipse-score/communication/pull/385/changes/ed71f9b487ae18305253dfb8367c0fe4590b3b9b#r3257711431

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/proxy.cpp
Tejveerpratap2803: Done, Change the name of method to CleanupMethods:)

score/mw/com/impl/bindings/lola/proxy.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -929,4 +927,58 @@ void Proxy::RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod&
     SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(was_inserted, "Method IDs must be unique!");
 }
 
+void Proxy::PrepareDeinitialize()
+{
+    StopProxyAutoReconnect();
```
bemerybmw, line 939: Please add a comment that explains why StopProxyAutoReconnect must be called before CleanupMethods.

Tejveerpratap2803: Done

bemerybmw (approved):

bemerybmw (changes requested):
Changes look good. Please clean up commits before merging

bemerybmw (approved):

PR #361, opened 2026-04-27
Rust::com fix for aarche64 gcc build for rust binary at user level

bharatGoswami8:
* Added rustc flag


bharatGoswami8:
fix should be done from toolchain side.

PR #367, opened 2026-04-28
fix(com/impl): fix clang-tidy warnings for fields

sahithi-nukala:
- Replace std::move with std::forward in skeleton_field.h RegisterSetHandler to correctly handle forwarding references and fix **bugprone-move-forwarding-reference warning**

### DIFF
--- a/score/mw/com/impl/skeleton_field.h
+++ b/score/mw/com/impl/skeleton_field.h
@@ -125,7 +125,7 @@ class SkeletonField : public SkeletonFieldBase
         static_assert(std::is_invocable_v<CallableType, FieldType&>,
                       "RegisterSetHandler: handler must be callable as void(FieldType& value). "
                       "The argument initially holds the proxy-requested value and may be modified in-place.");
-        set_handler_ = std::move(handler);
+        set_handler_ = std::forward<CallableType>(handler);
 
         auto wrapped_callback = [this](FieldType& new_value) -> FieldType {
             // Allow user to validate/modify the value in-place


score/mw/com/impl/configuration/someip_field_instance_deployment.h
```diff
@@ -24,7 +24,7 @@ class SomeIpFieldInstanceDeployment
     SomeIpFieldInstanceDeployment() noexcept = default;
     explicit SomeIpFieldInstanceDeployment(const score::json::Object& /* json_object */) noexcept {}
 
-    json::Object Serialize() const noexcept
```
bemerybmw, line 27: Why is this made static?

score/mw/com/impl/plumbing/i_proxy_field_binding_factory.h
```diff
@@ -35,31 +34,17 @@ class IProxyFieldBindingFactory
     virtual ~IProxyFieldBindingFactory() noexcept = default;
 
     IProxyFieldBindingFactory(IProxyFieldBindingFactory&&) = delete;
```
bemerybmw, line 37: We should discuss whether this clang tidy warning makes sense. I don't think it does in this case.

score/mw/com/impl/plumbing/i_proxy_field_binding_factory.h
```diff
@@ -35,31 +34,17 @@ class IProxyFieldBindingFactory
...
+    auto operator=(IProxyFieldBindingFactory&&) -> IProxyFieldBindingFactory& = delete;
     IProxyFieldBindingFactory(const IProxyFieldBindingFactory&) = delete;
-    IProxyFieldBindingFactory& operator=(const IProxyFieldBindingFactory&) = delete;
+    auto operator=(const IProxyFieldBindingFactory&) -> IProxyFieldBindingFactory& = delete;
 
     /// Creates instances of the event binding of a proxy field with a particular data type.
     /// \tparam SampleType Type of the data that is exchanges
     /// \param handle The handle containing the binding information.
     /// \param field_name The binding unspecific name of the event inside the proxy denoted by handle.
     /// \return An instance of ProxyEventBinding or nullptr in case of an error.
```
bemerybmw, line 46: Why did you remove const?

bemerybmw (changes requested):

.clang-tidy
```diff
@@ -86,8 +86,6 @@ Checks: >
   clang-analyzer-cplusplus.NewDelete,
   # --- qnx-checks ---
   bugprone-reserved-identifier,
-  # --- extra ---
```
bemerybmw, line 89: Why was this removed?

limdor: I would make sense to instead of removing it from here, to directly remove the clang tidy extra check file
https://github.com/eclipse-score/communication/blob/main/score/mw/com/.clang-tidy-extra and leave it here and apply it everywhere. I do not think we want to have in score/communication different clang tidy checks depending on the folder.

sahithi-nukala: moved the   **modernize-concat-nested-namespaces,** to .clang-tidy & remove the .clang-tidy-extra

limdor: Lets move please the removal of clang-tidy-extra in a different PR to keep topics separate.
Apart from the file we also need to remove the target
https://github.com/eclipse-score/communication/blob/main/score/mw/com/BUILD#L175

bemerybmw (commented):
The change in skeleton_field.h is fine for me. @limdor, you know the clang-tidy stuff better than me so can you please take over the review and approve once you're happy with the changes in .clang-tidy?

bemerybmw (approved):

PR #368, opened 2026-04-28
Feature/extraction skeleton service elements clean

rudresh-systream:


LittleHuba:
Clsoing as author requested to continue work in context of #259

PR #369, opened 2026-04-28
mw/com: Fix method signature to avoid copies of return type

bemerybmw:


score/mw/com/impl/methods/method_handler_checker.h
```diff
@@ -0,0 +1,572 @@
...
+    // method.func([](int& ret, const int& value, const double& value2){});
+
+    // using TupleType = get_method_in_arg_type_tuple<std::function<void(int&, float)>>::type;
+    // int j{};
+    // TupleType val{j, 2.0};
+
+    // std::tuple<int> v{1};
+
+    SkeletonMethod<int, int, double> method{};
+    method.func([](int&, const int&, const double&){});
```
bemerybmw: remove

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -189,20 +182,35 @@ Result<void> SkeletonMethod<ReturnType(ArgTypes...)>::RegisterHandler(Callable&&
             constexpr bool is_return_type_not_void = !std::is_same_v<ReturnType, void>;
             if constexpr (is_return_type_not_void)
             {
+                const auto typed_return_ptr_tuple = Deserialize<ReturnType>(type_erased_return.value());
```
crimson11: you could use `DeserializeArg<ReturnType>()` directly ... much simpler?

score/mw/com/impl/methods/skeleton_method_test.cpp
```diff
@@ -119,113 +125,92 @@ struct MyDataStruct
...
+struct MethodTypeAndHandlerType
+{
+    using MethodType = MethodTypeIn;
+    using HandlerType = HandlerTypeIn;
+};
+using RegisteredFunctionTypes =
+    ::testing::Types<MethodTypeAndHandlerType<TestMethodType, TestMethodHandlerType>,
+                     MethodTypeAndHandlerType<void(int), void(const int&)>,
+                     MethodTypeAndHandlerType<void(int, int), void(const int&, const int&)>,
+                     MethodTypeAndHandlerType<void(const double, int), void(const double&, const int&)>,
```
crimson11: Interessting. So you have a `const` in front of an in-arg .... in the user provided method-signature ... "technically" this is uneeded as with treat in-args as const anyhow! But here you are intentionally checking, that we also can cope with such a signature?

score/mw/com/impl/methods/skeleton_method_test.cpp
```diff
@@ -99,17 +83,39 @@ TEST(SkeletonMethodTest, ClassTypeDependsOnMethodType)
                   "Class type does not depend on event data type");
```
crimson11: > event data type

method signature?

score/mw/com/impl/methods/skeleton_method_test.cpp
```diff
@@ -99,17 +83,39 @@ TEST(SkeletonMethodTest, ClassTypeDependsOnMethodType)
...
+    }
+
+    SkeletonMethodTypedTest& WithAMockedMethodBindingfactory()
+    {
+        skeleton_method_binding_factory_mock_guard_ = std::make_unique<SkeletonMethodBindingFactoryMockGuard>();
+        return *this;
+    }
+
+    EmptySkeleton empty_skeleton_{std::make_unique<mock_binding::Skeleton>(), kInstanceIdWithLolaBinding};
+    std::string method_name_{"dummy_method"};
```
crimson11: const

score/mw/com/impl/proxy_field.h
```diff
@@ -259,7 +259,7 @@ class ProxyField final : public ProxyFieldBase
     }
 
     template <typename T = SampleDataType, typename = std::enable_if_t<EnableSet && std::is_same_v<T, SampleDataType>>>
-    score::Result<MethodReturnTypePtr<T>> Set(SampleDataType& new_field_value) noexcept
+    score::Result<MethodReturnTypePtr<T>> Set(const SampleDataType& new_field_value) noexcept
```
crimson11, line 323: Hm. Now I got aware, that we don't expose the zero-copy/allocate variant of method calls to the user ... I guess this would be just a small addition ... but let's wait til someone really asks for it :)

score/mw/com/test/methods/methods_test_resources/method_provider.h
```diff
@@ -109,10 +109,11 @@ bool MethodProvider<Skeleton>::RegisterMethodHandlerWithInArgsAndReturn()
 {
     SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(skeleton_ != nullptr);
 
-    auto handler_with_in_args_and_return = [](std::int32_t a, std::int32_t b) -> std::int32_t {
-        std::cout << "Provider: with_in_args_and_return called with " << a << " + " << b << std::endl;
-        return a + b;
-    };
+    auto handler_with_in_args_and_return =
+        [](std::int32_t& return_value, const std::int32_t& a, const std::int32_t& b) {
```
crimson11: maybe add a `-> void` ... makes it "clearer"?

crimson11 (changes requested):

score/mw/com/impl/methods/skeleton_method_test.cpp
bemerybmw: Yep, exactly.

score/mw/com/impl/methods/skeleton_method_test.cpp
```diff
@@ -237,34 +222,38 @@ TEST_F(SkeletonMethodTestFixture, ACallbackWithAPointerAsStateCanBeRegistered)
...
     {
         constexpr std::size_t in_type_1_size = sizeof(InType1);
+        constexpr std::size_t in_type_2_size = sizeof(InType2);
 
         std::byte* write_head = in_args_buffer_.begin();
         new (write_head) InType1(in_arg_1);
         write_head += in_type_1_size;
         new (write_head) InType2(in_arg_2);
+        write_head += in_type_2_size;
+        new (write_head) InType2(in_arg_3);
```
KrishaDeshkool, line 352: the in_args_buffer_ is still sized with 2 elements here.
`static constexpr std::size_t in_args_buffer_size = sizeof(InType1) + sizeof(InType2);`
shouldn't this be
`static constexpr std::size_t in_args_buffer_size = sizeof(InType1) + sizeof(InType2) + sizeof(InType2);` ?

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -192,17 +185,31 @@ Result<void> SkeletonMethod<ReturnType(ArgTypes...)>::RegisterHandler(Callable&&
                 SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
                     type_erased_return.has_value(),
                     "ReturnType is non void. Thus, type_erased_result needs to have a value!");
-                ReturnType res = std::apply(callable_invoker, std::forward<InArgPtrTuple>(typed_in_arg_ptrs));
-                SerializeArgs<ReturnType>(type_erased_return.value(), res);
+                auto* const typed_return_ptr = DeserializeArg<ReturnType>(type_erased_return.value());
```
KrishaDeshkool: ah! we populate inplace, cool,
but, does it not weaken the "safe for trivially copyable" check?
in SerializeArgs route before we had this explicit check static_assert(std::is_trivially_copyable_v<T> before we populate type_erased_return.
i think this is fine today though.

score/mw/com/impl/util/type_erased_storage.h
```diff
@@ -239,12 +239,26 @@ void SerializeArgs(score::cpp::span<std::byte> target_buffer, T& arg, Args&... a
...
+/// \brief Overload of DeserializeArgs which takes a single argument type.
+///
+/// This overload is added for convenience since it returns Arg* instead of a tuple with one element. See docstring of
+/// DeserializeArgs for more details.
+template <typename Arg>
+auto DeserializeArg(score::cpp::span<std::byte> src_buffer) -> typename std::add_pointer<Arg>::type
+{
+    detail::MemoryBufferAccessor memory_buffer_accessor{src_buffer};
+    auto tuple_of_args = detail::Deserialize<Arg>(memory_buffer_accessor);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(std::tuple_size_v<decltype(tuple_of_args)> == 1,
```
KrishaDeshkool: this can be a static assert?

score/mw/com/impl/methods/skeleton_method_test.cpp
```diff
@@ -119,113 +125,92 @@ struct MyDataStruct
...
+{
+    using MethodType = MethodTypeIn;
+    using HandlerType = HandlerTypeIn;
+};
+using RegisteredFunctionTypes =
+    ::testing::Types<MethodTypeAndHandlerType<TestMethodType, TestMethodHandlerType>,
+                     MethodTypeAndHandlerType<void(int), void(const int&)>,
+                     MethodTypeAndHandlerType<void(int, int), void(const int&, const int&)>,
+                     MethodTypeAndHandlerType<void(const double, int), void(const double&, const int&)>,
+                     MethodTypeAndHandlerType<void(int, int), void(const int&, const int&)>,
```
KrishaDeshkool: This looks like a dup of https://github.com/eclipse-score/communication/pull/369/changes#diff-40f76737b28b11683a2795d7bc4658afc427cd2d105ba193b4f35133dc9fadbaR138
is this intended?

score/mw/com/impl/skeleton_field.h
```diff
@@ -123,24 +124,27 @@ class SkeletonField : public SkeletonFieldBase
     template <bool ES = EnableSet, std::enable_if_t<ES, int> = 0, typename CallableType>
     Result<void> RegisterSetHandler(CallableType&& set_handler)
     {
-        static_assert(std::is_invocable_v<CallableType, FieldType&>,
+        static_assert(std::is_invocable_r_v<void, CallableType, FieldType&>,
```
KrishaDeshkool: does this really do something?
somehow, my static asserts [here](https://godbolt.org/z/7oj46cWTv) always work even when the ret type is non void.

score/mw/com/impl/methods/skeleton_method_test.cpp
```diff
@@ -119,113 +125,92 @@ struct MyDataStruct
...
 
     // Expecting that the register call is dispatched to the binding without errors
-    EXPECT_CALL(mock_method_binding, RegisterHandler(_));
+    EXPECT_CALL(this->mock_method_binding_, RegisterHandler(_));
 
-    // When a Register call is issued at the binding independent level
-    score::cpp::callback<FixtureMethodType> test_callback{};
-
-    std::ignore = method.RegisterHandler(std::move(test_callback));
+    // // When a Register call is issued at the binding independent level
```
KrishaDeshkool:
```suggestion
    // When a Register call is issued at the binding independent level
```

score/mw/com/impl/methods/skeleton_method.h
bemerybmw: Hmmm, I actually don't think we have that requirement for deserializing. Deserializing is basically just casting the type, there's no copying done. In practice, we always serialize with the function in this file which already confirms that the type is trivially copyable but in theory, you could create a type erased buffer containing a non trivially copyable type (but the caller has to take  that it's correctly created in the buffer) and then deserialize it without making a copy which should not be a problem. If the user then wants to copy it out of the buffer, it's now strongly typed and so the copy constructor can be called.

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -946,282 +972,162 @@ TEST_F(SkeletonFieldSetHandlerTest, PrepareOfferSucceedsWithoutHandlerWhenEnable
...
-                    })
-                    .has_value());
+    // When calling the set handler that was captured by the method binding
+    auto [in_span, out_span] = CreateFieldSetterInArgAndReturnSpans(kDummySetValue, TestSampleType{});
+    captured_set_handler_.value()(in_span, out_span);
+}
 
-    ASSERT_TRUE(unit.my_setter_field_.Update(TestSampleType{1U}).has_value());
-    ASSERT_TRUE(unit.my_setter_field_.PrepareOffer().has_value());
+TEST_F(SkeletonFieldSetHandlerTest, CallingPrepareOfferWithoutRegisteringSetHandlerReturnsError)
```
KrishaDeshkool: isn't this the same test as here https://github.com/eclipse-score/communication/blob/c9c64382866d8262fcbd3ddd0ca4e14856595f6a/score/mw/com/impl/skeleton_field_test.cpp#L925

score/mw/com/impl/skeleton_field.h
```diff
@@ -121,30 +122,50 @@ class SkeletonField : public SkeletonFieldBase
...
-
-            // Return the accepted value to the proxy
-            return new_value;
         };
 
         is_set_handler_registered_ = true;
-        return set_method_->RegisterHandler(std::move(wrapped_callback));
+
+        auto state = std::make_unique<std::pair<SkeletonField*, CallableType>>(
+            this, std::forward<CallableType>(user_set_handler));
```
KrishaDeshkool: wont this cause the same "use after free" issue?
since we store "this" during the registration into the stored set_handler and i dont think we update this when we move the SkeletonField then won't this still be pointing to the moved from object?
or am i missing something?

score/mw/com/impl/skeleton_field.h
```diff
@@ -121,30 +122,50 @@ class SkeletonField : public SkeletonFieldBase
...
+        static_assert(std::is_same_v<ExpectedCallableReturnType, ActualCallableReturnType>,
+                      "Registered method callable must have void return type!");
+
+        using ActualCallableInArgType = typename get_callable_args_types<CallableType>::type;
+        using ExpectedCallableInArgTypes = std::tuple<FieldType&>;
+        static_assert(std::is_same_v<ExpectedCallableInArgTypes, ActualCallableInArgType>,
+                      "Registered method callable must have a single non-const reference argument of type FieldType&!");
+
+        // Since set_handler can be an lvalue reference or an rvalue reference, we ideally would store it as a
+        // universal reference in the type_erased_handler. However, in C++17, this is not supported. Instead, we create
```
KrishaDeshkool: to me this sounds like this is C++17 only issue (storing a universal ref), 
It is only used by compiler to figure out what is the type of the args, once it is figures that out it would be either an L or an R value, so can we just drop that confusing line ?
https://stackoverflow.com/questions/14757474/how-to-store-universal-references

score/mw/com/impl/skeleton_field.h
```diff
@@ -121,30 +122,50 @@ class SkeletonField : public SkeletonFieldBase
...
+        // an callable here which will be called below by another lambda which explicitly stores the callback as either
+        // an lvalue reference or an rvalue reference depending on how it was passed in.
```
KrishaDeshkool: does this also have lvalue vs rvalue branch like above? https://github.com/eclipse-score/communication/pull/369/changes#diff-5067b694039ba429e1f705db5d97521e3f0b3b0b07232117860ecd84ad9ebc80R212

score/mw/com/impl/methods/test/callable_traits_resources.h
```diff
@@ -0,0 +1,214 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_IMPL_METHODS_TEST_CALLABLE_TRAITS_RESOURCES_H
+#define SCORE_MW_COM_IMPL_METHODS_TEST_CALLABLE_TRAITS_RESOURCES_H
+
+namespace score::mw::com::impl
+{
+
+/// \brief Factory that constructs a default-initialised CallableWrapper<HandlerSignature> via a static Create().
```
crimson11: OK. Now I am NITPICKY. ;)
>Factory that constructs a default-initialised CallableWrapper<HandlerSignature> 
better and shorter
>Factory that default-constructs a CallableWrapper<HandlerSignature> 

The factory is calling the default-ctor here. Default-initializing is a broader process, where depending on the type, which is being default-initialized eventually also the default-ctor gets called

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -0,0 +1,206 @@
...
+/// - If `ReturnType` is not `void` and there are in arguments:
+///   `void(ReturnType&, const ArgTypes&...)`
+/// - If `ReturnType` is not `void` and there are no in arguments:
+///   `void(ReturnType&)`
+/// - If `ReturnType` is `void` and there are in arguments:
+///   `void(const ArgTypes&...)`
+/// - If `ReturnType` is `void` and there are no in arguments:
+///   `void()`
+template <FailureMode FailureMode, typename Callable, typename ReturnType, typename... ArgTypes>
+constexpr void AssertCallableMatchesMethodSignature()
```
crimson11: Thankfully you had the extensive doc on top ;)
Because the name could be a bit misleading ... 
>AssertCallableMatchesMethodSignature
...
>AssertMethodHandlerSupportsMethodSignature

... because matches (no native speaker ;) ) sounds like "equals" to me ... and Callable is so generic ... we exactly know, which callable type we talk about.

crimson11 (dismissed):

score/mw/com/impl/skeleton_field.h
bemerybmw: Yep, you're right. I wasn't trying to avoid the use after free issue here, it was rather about fitting all the state in the callable. I missed this point that we're on the binding independent layer in which the field is moveable. It's also a problem that we don't have a test that detects this...

score/mw/com/impl/skeleton_field.h
bemerybmw: This was a copy paste mistake. I was originally planning on using the same approach (with the if constexpr) but since I anyway need to put the state into a unique_ptr to fit it into the method handler (which is a score::cpp::callback of fixed size), we don't have the same problem (because the unique_ptr will either store an lvalue or rvalue reference depending on the template type of the CallableType).

score/mw/com/impl/skeleton_field.h
bemerybmw: This is no longer relevant here because I've removed the comment. But maybe it still applies to the comment in skeleton_method. But I don't fully understand your point, the issue with c++17 is not being able to "store" a universal reference in a lambda because we don't have templated lambdas. The link you sent is about "storing" a universal ref in a class. Probably easier if we just have a quick call and I can explain the problem / solution.

castler (dismissed):

castler (dismissed):

score/mw/com/impl/skeleton_field.h
bemerybmw: Replaced this with some more explicit checks using some helpers that I added.

score/mw/com/impl/methods/skeleton_method_test.cpp
bemerybmw: Completely reworked the tests.

score/mw/com/impl/skeleton_field.h
```diff
@@ -145,30 +148,48 @@ class SkeletonField : public SkeletonFieldBase
...
+                      "Registered method callable must have void return type!");
+
+        using ActualCallableInArgType = typename get_callable_args_types<CallableType>::type;
+        using ExpectedCallableInArgTypes = std::tuple<FieldType&>;
+        static_assert(std::is_same_v<ExpectedCallableInArgTypes, ActualCallableInArgType>,
+                      "Registered method callable must have a single non-const reference argument of type FieldType&!");
+
+        // Since the pointer to SkeletonField and the user callable don't fit in the state of the type-erased handler,
+        // we have to allocate them on the heap and store a pointer to them.
+        auto state = std::make_unique<
```
KrishaDeshkool, line 165: I think we already had this talk before,
since we store the ref to the user callback, is it already documented to user somewhere that he should keep his callback alive ? (even though it is a trivial expecation).

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -87,9 +82,13 @@ class SkeletonMethod<ReturnType(ArgTypes...)> final : public SkeletonMethodBase
                                                    method_type),
               method_type)
     {
+        AssertMethodSignatureDoesNotContainPointersOrReferences<FailureMode::COMPILE_TIME, ReturnType, ArgTypes...>();
```
KrishaDeshkool, line 85: maybe i'm missing something here,
now you removed the in class checks ` static constexpr bool no_in_args_are_pointers = (... && !std::is_pointer_v<ArgTypes>);`
and we only check this condition here in the constructor that is called only by Field methods,
then this check is not done for other normal users of methods?
if method user would do somwthing like this,
`SkeletonMethod<int*<int*>>` that would work today?

score/mw/com/impl/skeleton_field.h
bemerybmw: We have an AoU about this on proxy side. We should create a similar one on skeleton side: SWP-269665

score/mw/com/impl/methods/skeleton_method.h
bemerybmw: This was just a mistake by me. I added it to the delegated constructor so it should be hit in all construction paths.

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -0,0 +1,206 @@
...
+
+// TODO: We currently use FailureMode as a workaround to allow testing AssertMethodHandlerSupportsMethodSignature at
+// runtime since we don't currently have infrastructure for compile time testing. When compile time testing is enabled
+// in SWP-46885, we can remove FailureMode and the runtime tests and replace them with compile time tests.
+template <bool condition, FailureMode failure_mode>
+void CompileOrRuntimeAssert(const char* message)
+{
+    if constexpr (failure_mode == FailureMode::COMPILE_TIME)
+    {
+        static_assert(condition);
```
KrishaDeshkool, line 78: `static_assert(condition, ,message);`

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -118,57 +117,86 @@ SkeletonMethod<ReturnType(ArgTypes...)>::SkeletonMethod(SkeletonBase& skeleton_b
...
-    static_assert(std::is_rvalue_reference_v<decltype(callback)>,
-                  "Callbeck provided to register has to be an rvalue reference");
-
-    // A move would only be problematic if the forwarding refference Callback&& evealuates to an LValue and and is moved
-    // from (which the caller of the function might not expect). We static assert that the callback is an RValue
-    // refference and can always be movef from.
-    // NOLINTNEXTLINE(bugprone-move-forwarding-reference) The callback is asserted to be an RValue reference
-    auto callable_invoker = [callable = std::move(callback)](auto&&... ptrs) -> decltype(auto) {
-        return std::invoke(callable, (*ptrs)...);
+    AssertMethodCallableIsNotStdBind<FailureMode::COMPILE_TIME, Callable>();
```
KrishaDeshkool, line 128: in AssertMethodCallableIsNotStdBind when the  Callable collapses to a reference type Callable& the [std::bind check would not catch the ref types,](https://godbolt.org/z/KjMTThMrc).

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -961,279 +987,180 @@ TEST_F(SkeletonFieldSetHandlerTest, PrepareOfferSucceedsWithoutHandlerWhenWithSe
...
-                   "Calling RegisterSetHandler() a second time shall replace the previously stored "
-                   "user callback. When the wrapped handler is subsequently invoked, only the second "
-                   "callback shall be called.");
-    RecordProperty("TestType", "Requirements-based test");
-    RecordProperty("Priority", "1");
-    RecordProperty("DerivationTechnique", "Analysis of requirements");
+    testing::MockFunction<void(TestSampleType&)> user_callback_mock{};
+
+    // Given that the user registered callback will be called which modifies the value in-place
+    TestSampleType expected_set_value{kDummySetValue};
```
KrishaDeshkool, line 1112: we don't have a end to end test that proves that a register and then move and then invoke the previously set handler so we can prove that move worked as expected, no?
we can test that in a unit test level ?

score/mw/com/impl/methods/callable_traits.cpp
```diff
@@ -0,0 +1,13 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/impl/methods/method_handler_traits.h"
```
KrishaDeshkool:
```suggestion
#include "score/mw/com/impl/methods/callable_traits.h"
```

score/mw/com/impl/methods/test/BUILD
```diff
@@ -0,0 +1,24 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_cc//cc:defs.bzl", "cc_library")
+load("//score/mw:common_features.bzl", "COMPILER_WARNING_FEATURES")
+
+cc_library(
+    name = "callable_traits_resources",
+    hdrs = ["callable_traits_resources.h"],
```
KrishaDeshkool, line 20: missing srcs ?

score/mw/com/impl/methods/method_traits_checker.h
bemerybmw: We can only provide a string literal in c++20. I know this is not ideal, but this is a workaround until we have compile time testing.

score/mw/com/impl/methods/skeleton_method.h
bemerybmw: Added std::remove_reference and added a test.

score/mw/com/impl/skeleton_field_test.cpp
bemerybmw: Tejveer is working on an integration test for this. For the moment, we only have unit tests. Added a unit test for this.

KrishaDeshkool:
LGTM, i see you have added a note that the new commit would be squashed,
so approving from content POV.

KrishaDeshkool (approved):

LittleHuba (changes requested):
Blocked since the PR is too big and would break other ongoing workstreams.
Will unblock when critical stage is over.

anmittag:
@bemerybmw is your PR still valid?

bemerybmw:
> @bemerybmw is your PR still valid?

@anmittag Yes, just waiting for the green light from @LittleHuba to merge.

LittleHuba (approved):
Block removed. Formal approval based on review by @KrishaDeshkool

PR #373, opened 2026-04-29
mw/com: Various field fixes

bemerybmw:


score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -12,6 +12,7 @@
  ********************************************************************************/
 #include "score/mw/com/impl/skeleton_field.h"
 
+#include "method_type.h"
```
bemerybmw: Fix include path

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -969,9 +842,8 @@ class MySetterSkeleton : public SkeletonBase
     SkeletonField<TestSampleType, /*EnableSet=*/true> my_setter_field_{*this, kFieldName};
 };
 
-// Static type-trait tests for RegisterSetHandler availability
-
-TEST(SkeletonFieldSetHandlerTest, RegisterSetHandlerOnlyExistsWhenEnableSetIsTrue)
+/// gtodo: What is this test testing??
```
bemerybmw: Address this

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -1161,45 +932,22 @@ TEST(SkeletonFieldSetHandlerTest, PrepareOfferSucceedsAfterRegisterSetHandler)
...
     const auto result = unit.my_dummy_field_.PrepareOffer();
+
+    // Then it succeeds
     EXPECT_TRUE(result.has_value());
 }
 
-// RegisterSetHandler accepts any callable with the expected signature
-
-TEST(SkeletonFieldSetHandlerTest, RegisterSetHandlerAcceptsAnyCallable)
+// gtodo: What is this testing???
```
bemerybmw: and this

score/mw/com/impl/traits_test.cpp
```diff
@@ -67,7 +68,7 @@ class MyInterface : public InterfaceTrait::Base
     using InterfaceTrait::Base::Base;
 
     typename InterfaceTrait::template Event<TestSampleType> some_event{*this, kEventName};
-    typename InterfaceTrait::template Field<TestSampleType> some_field{*this, kFieldName};
+    typename InterfaceTrait::template Field<TestSampleType, true> some_field{*this, kFieldName};
```
bemerybmw, line 71: Will also need tests somewhere for different combinations of setter / getter / notifier?

score/mw/com/impl/skeleton_field.h
```diff
@@ -229,8 +224,19 @@ SkeletonField<SampleDataType, EnableSet, EnableNotifier>::SkeletonField(Skeleton
                                                          parent,
                                                          field_name),
                                                      typename SkeletonEvent<FieldType>::FieldOnlyConstructorEnabler{}),
-          field_name,
-          detail::EnableSetOnlyTag{}}
+          std::make_unique<SkeletonMethod<SetMethodSignature>>(
```
crimson11, line 228: I'm maybe missing some context here ... and it is not an issue with your PR ... But does this ctor call of `SkeletonMethod` work? I just looked at the code in `SkeletonMethod` and it looks like it would register now this getter/setter under wrong name/type in the `SkeletonBase`?

crimson11 (dismissed):

crimson11 (approved):

score/mw/com/impl/skeleton_field.h
bemerybmw: As discussed, they won't be registered as methods but rather the field registers itself with the parent and it takes care of the event/method dispatch objects.

Issue #311, opened 2026-04-16
Bug: GenericSkeleton interaction with typed Proxies not working


crimson11:
### Description

Presumably the interaction between the (newly introduced) `GenericSkeleton` and an existing "normal" proxy (not a `GenericProxy`) seems to be broken.
The root-cause is, that both sides (the `GenericSkeleton` and the "normal" proxy) do interprete the `EventDataStorage` within the shared-memory differently/incompatible.

Details:
In case of non-generic/typed skeletons the `EventDataStorage` gets created within method-template

```
template <typename SampleType>
SkeletonMemoryManager::CreateEventDataFromOpenedSharedMemory(const ElementFqId element_fq_id,
                                                                  const SkeletonEventProperties& element_properties)
```
while in case of `GenericSkeleton`/`GenericSkeletonEvent` tzhe `EventDataStorage` gets created within non-template method

```
CreateEventDataFromOpenedSharedMemory(
        const ElementFqId element_fq_id,
        const SkeletonEventProperties& element_properties,
        size_t sample_size,
        size_t sample_alignment)
```

In the non-generic/typed approach, the `EventDataStorage` gets created like this:

```
auto* typed_event_data_storage_ptr = storage_resource_->construct<EventDataStorage<SampleType>>(
        element_properties.number_of_slots,
        memory::shared::PolymorphicOffsetPtrAllocator<SampleType>(*storage_resource_));
```
This essentially creates a `DynamicArray` of `element_properties.number_of_slots` of type `SampleType`.

In the generic approach the `EventDataStorage` gets created like this:
```
auto* data_storage = storage_resource_->construct<EventDataStorage<std::max_align_t>>(
        num_max_align_elements, memory::shared::PolymorphicOffsetPtrAllocator<std::max_align_t>(*storage_resource_));
```

These two representations are different/incompatible. The underlying raw-byte-array being the storage for the `DynamicArray` has the same size.
Nowever the type T is differtent, which is reflected in the size/number_of_elements member variable in the `DynamicArray`! I.e. in the case of the generic skeleton/skeleton event, the number of elements gets calculated on the basis of `std::max_align_t` type - which typically is much smaller than the real type used in case of typed skeletons (and proxies), thus leading to a much larger calculated number of elements (`num_max_align_elements`).

[...]

PR #379, opened 2026-05-04
Generic skeleton integration test

ShoroukRamzy:
**Overview**

This PR significantly expands the integration testing suite by introducing tests for both Generic-Typed (type-erased Skeleton Provider with Strongly-Typed Proxy Consumer) and Generic-Generic (type-erased Skeleton Provider with type-erased Proxy Consumer) interactions.

**This addresses this issue https://github.com/eclipse-score/communication/issues/261 and  https://github.com/eclipse-score/communication/issues/311 by rigorously validating the underlying shared memory (SHM) communication mechanisms with various payload sizes (64, 32, 16, and 8 bytes) and configurations.** 

The tests strictly evaluate type-erased memory striding, boundary enforcements, and data integrity by spinning up a provider that sends 30 consecutive samples into a heavily constrained 5-slot ring buffer, forcing multiple buffer wrap-arounds.

**Test Scope & Root Causes of Current Failures** 

Currently, all introduced test variants intentionally fail, successfully exposing critical bugs within the Generic skeleton implementation related to shared memory allocation and addressing. The failure modes depend on the payload size relative to std::max_align_t (which is assumed to be 32 bytes on the current architecture).

The identified root causes for these failures are:

**1. Data Corruption (Expected: X, got: 0 or Y)**

**Affected Tests**: All Generic-Generic interaction tests (64, 32, 8-byte payloads) and Generic-Typed tests with payloads > std::max_align_t (64-byte, 32-byte payloads).
**Root Cause**: The fundamental issue was an incorrect base pointer being returned for event data. The SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory and Skeleton::RegisterGeneric functions were erroneously returning a pointer to the EventDataStorage object itself, instead of the actual data buffer managed within that object (EventDataStorage::data()). This led both providers and consumers to miscalculate offsets, resulting in writes to unintended (often zero-initialized) memory locations and reads from those incorrect, zero-filled locations.

**2. Boundary Check Crashes (Exit Code 134 - SIGABRT)**

**Affected Tests**: Generic-Typed interaction tests with payloads < std::max_align_t (16-byte, 8-byte payloads).
**Root Cause:** This failure was caused by an incorrect capacity calculation for the shared memory array. The EventDataStorage's internal DynamicArray was being initialized with a capacity based on num_max_align_elements. For smaller payloads, this calculated capacity was often less than the numberOfSampleSlots specified in the configuration. When  the consumer attempted to retrieve one beyond this physically undersized capacity, it resulted in out-of-bounds memory access, triggering assertions and process crashes.

### DIFF
--- a/score/mw/com/test/generic_skeleton/generic_generic_interaction_app.cpp
+++ b/score/mw/com/test/generic_skeleton/generic_generic_interaction_app.cpp
@@ -1,0 +1,204 @@
+#include "score/mw/com/impl/generic_proxy.h"
+#include "score/mw/com/impl/generic_skeleton.h"
+#include "score/mw/com/impl/instance_specifier.h"
+#include "score/mw/com/runtime.h"
+#include "score/mw/com/runtime_configuration.h"
+#include "score/mw/log/logging.h"
+
+#include <chrono>
+#include <cstdlib>
+#include <iostream>
+#include <string>
+#include <string_view>
+#include <thread>
+#include <vector>
+
+// Default to 64-byte if not specified by the build system
+#ifndef PAYLOAD_SIZE
+#define PAYLOAD_SIZE 64
+#endif
+
+namespace
+{
+
+struct MyEventData
+{
+    uint64_t counter;
+#if PAYLOAD_SIZE > 8
+    char padding[PAYLOAD_SIZE - 8];
+#endif
+};
+
+constexpr std::string_view kInstanceSpecifier = "/test/generic/generic/interaction";
+constexpr int kSamplesToProcess = 30;
+constexpr int kSamplesToSubscribe = 5;
+
+#if PAYLOAD_SIZE == 64
+constexpr std::string_view kEventName = "Event64Byte";
+#elif PAYLOAD_SIZE == 32
+constexpr std::string_view kEventName = "Event32Byte";
+#elif PAYLOAD_SIZE == 16
+constexpr std::string_view kEventName = "Event16Byte";
+#elif PAYLOAD_SIZE == 8
+constexpr std::string_view kEventName = "Event8Byte";
+#else
+#error "Unsupported payload size configured."
+#endif
+
+int run_provider()
+{
+    const auto instance_specifier = score::mw::com::impl::InstanceSpecifier::Create(kInstanceSpecifier).value();
+    std::cout << "[PROVIDER] Instance specifier created." << std::endl;
+    const score::mw::com::impl::DataTypeMetaInfo meta{sizeof(MyEventData), alignof(MyEventData)};
+    std::cout << "[PROVIDER] DataTypeMetaInfo created (size=" << sizeof(MyEventData)
+              << ", align=" << alignof(MyEventData) << ")." << std::endl;
+    const std::vector<score::mw::com::impl::EventInfo> events = {{kEventName, meta}};
+    std::cout << "[PROVIDER] EventInfo vector created for event: " << kEventName << std::endl;
+
+    score::mw::com::impl::GenericSkeletonServiceElementInfo create_params;
+    create_params.events = events;
+    std::cout << "[PROVIDER] GenericSkeletonServiceElementInfo prepared." << std::endl;
+
+    std::cout << "[PROVIDER] Calling GenericSkeleton::Create..." << std::endl;
+    auto skeleton_res = score::mw::com::impl::GenericSkeleton::Create(instance_specifier, create_params);
+    if (!skeleton_res.has_value())
+    {
+        std::cerr << "[PROVIDER] GenericSkeleton::Create FAILED." << std::endl;
+        return 1;
+    }
+    auto& skeleton = skeleton_res.value();
+    std::cout << "[PROVIDER] GenericSkeleton created." << std::endl;
+
+    std::cout << "[PROVIDER] Calling skeleton.OfferService()..." << std::endl;
+    if (!skeleton.OfferService().has_value())
+    {
+        std::cerr << "[PROVIDER] OfferService FAILED." << std::endl;
+        return 1;
+    }
+    std::cout << "[PROVIDER] OfferService SUCCEEDED." << std::endl;
+
+    std::cout << "[PROVIDER] Getting event reference for " << kEventName << "..." << std::endl;
+    auto it = skeleton.GetEvents().find(kEventName);
+    if (it == skeleton.GetEvents().cend())
+    {
+        std::cerr << "[PROVIDER] Could not find event: " << kEventName << std::endl;
+        return 1;
+    }
+    std::cout << "[PROVIDER] Event reference obtained." << std::endl;
+
+    // Get reference to the GenericSkeletonEvent
+    auto& generic_event = const_cast<score::mw::com::impl::GenericSkeletonEvent&>(it->second);
+
+    std::cout << "[PROVIDER] Generic-Generic " << PAYLOAD_SIZE << "-byte - Waiting 5s for consumer to subscribe..."
+              << std::endl;
+    std::this_thread::sleep_for(std::chrono::seconds(5));
+    std::cout << "[PROVIDER] Finished initial 5s sleep." << std::endl;
+
+    for (int i = 0; i < kSamplesToProcess; ++i)
+    {
+        auto sample_res = generic_event.Allocate();
+        if (!sample_res.has_value())
+        {
+            std::cerr << "[PROVIDER] Allocation failed for sample: " << i << std::endl;
+            return 1;
+        }
+        std::cout << "[PROVIDER] Sample " << i << " allocated." << std::endl;
+
+        auto* typed_sample = static_cast<MyEventData*>(sample_res.value().Get());
+        typed_sample->counter = i;
+
+        std::cout << "[PROVIDER] Sending sample: " << i << std::endl;
+        generic_event.Send(std::move(sample_res.value()));
+        std::cout << "[PROVIDER] " << PAYLOAD_SIZE << "-byte Event Sent sample: " << i << std::endl;
+        std::this_thread::sleep_for(std::chrono::milliseconds(10));
+    }
+
+    std::cout << "[PROVIDER] All samples sent." << std::endl;
+    std::this_thread::sleep_for(std::chrono::seconds(15));
+    std::cout << "[PROVIDER] Finished post-send 15s sleep. Calling StopOfferService()..." << std::endl;
+    skeleton.StopOfferService();
+    std::cout << "[PROVIDER] StopOfferService() completed." << std::endl;
+    return 0;
+}
+
+int run_consumer()
+{
+    const auto instance_specifier = score::mw::com::impl::InstanceSpecifier::Create(kInstanceSpecifier).value();
+
+    score::Result<score::mw::com::ServiceHandleContainer<score::mw::com::impl::GenericProxy::HandleType>> handles_res;
+    int retries = 0;
+    while (retries < 50)
+    {
+        handles_res = score::mw::com::impl::GenericProxy::FindService(instance_specifier);
+        if (handles_res.has_value() && !handles_res.value().empty())
+            break;
+        std::this_thread::sleep_for(std::chrono::milliseconds(100));
+        retries++;
+    }
+    if (!handles_res.has_value() || handles_res.value().empty())
+        return 1;
+
+    auto proxy_res = score::mw::com::impl::GenericProxy::Create(handles_res.value()[0]);
+    if (!proxy_res.has_value())
+        return 1;
+    auto& proxy = proxy_res.value();
+
+    auto event_it = proxy.GetEvents().find(kEventName);
+    if (event_it == proxy.GetEvents().cend())
+        return 1;
+
+    // Get reference to the GenericProxyEvent
+    auto& generic_event = event_it->second;
+    generic_event.Subscribe(kSamplesToSubscribe);
+
+    uint64_t expected = 0;
+    uint64_t received = 0;
+    int data_mismatches = 0;
+
+    while (received < kSamplesToProcess)
+    {
+        // std::cout << "[CONSUMER] " << PAYLOAD_SIZE << "-byte Waking up, calling GetNewSamples..." << std::endl;
+
+        // The receiver callback operates on type-erased memory (SamplePtr<const void>)
+        generic_event.GetNewSamples(
+            [&](auto sample) {
+                auto* typed_sample = static_cast<const MyEventData*>(sample.get());
+                if (typed_sample->counter != expected)
+                {
+                    std::cerr << "[CONSUMER] " << PAYLOAD_SIZE << "-byte Data mismatch! Expected: " << expected
+                              << ", got: " << typed_sample->counter << std::endl;
+                    data_mismatches++;
+                }
+                else
+                {
+                    std::cout << "[CONSUMER] " << PAYLOAD_SIZE
+                              << "-byte Event Received sample: " << typed_sample->counter << std::endl;
+                }
+                expected++;
+                received++;
+            },
+            kSamplesToSubscribe);
+        std::this_thread::sleep_for(std::chrono::milliseconds(5));
+    }
+    if (data_mismatches > 0)
+    {
+        std::cerr << "[CONSUMER] Test failed with " << data_mismatches << " mismatches." << std::endl;
+        return 1;
+    }
+    return 0;
+}
+}  // namespace
+
+int main(int argc, const char* argv[])
+{
+    std::string mode;
+    for (int i = 1; i < argc; ++i)
+        if (std::string(argv[i]) == "--mode" && i + 1 < argc)
+            mode = argv[++i];
+    score::mw::com::runtime::InitializeRuntime(score::mw::com::runtime::RuntimeConfiguration(argc, argv));
+    if (mode == "provider")
+        return run_provider();
+    if (mode == "consumer")
+        return run_consumer();
+    return 1;
+}
--- a/score/mw/com/test/generic_skeleton/generic_generic_interaction_app.cpp
+++ b/score/mw/com/test/generic_skeleton/generic_generic_interaction_app.cpp
@@ -157,8 +157,6 @@ int run_consumer()
 
     while (received < kSamplesToProcess)
     {
-        // std::cout << "[CONSUMER] " << PAYLOAD_SIZE << "-byte Waking up, calling GetNewSamples..." << std::endl;
-
         // The receiver callback operates on type-erased memory (SamplePtr<const void>)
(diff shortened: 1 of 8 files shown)


ShoroukRamzy:
Hi @crimson11,
I will confirm the main reasons for the above two issues soon.

ShoroukRamzy:
> Hi @crimson11, I will confirm the main reasons for the above two issues soon.

@crimson11, Done

crimson11:
I will have a look at it tomorrow!

crimson11:
@ShoroukRamzy 
I had now a 1st look. I do have the following question:
To me it is unexpected, that the Generic-Generic interaction fails?
Because what you have written above:
>The fundamental issue was an incorrect base pointer being returned for event data. The SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory and Skeleton::RegisterGeneric functions were erroneously returning a pointer to the EventDataStorage object itself, ...

should have no effect in this case. Because a typed proxy/proxy event uses the `DataStoragePointer` here: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/generic_proxy_event.cpp#L156 , which the `GenericSkeleton` has placed into the `events_metainfo_` here:
https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp#L228
... and this is the "correct" raw-data-storage pointer (without the enclosing `DynamicArray` control structure).
So imho the Generic-Generic interaction should work, because:
- the GenericSkeleton/provider is storing the raw-pointer (not the pointer to a `DynamicArray`) into the "meta-info-map"
- the GenericProxyEvent/consumer is getting this raw-.pointer from the "meta-info-map"

Am I missing something here? I did not yet re-run/use your tests as picking it from the fork is "tedious" ;)

ShoroukRamzy:
> @ShoroukRamzy I had now a 1st look. I do have the following question: To me it is unexpected, that the Generic-Generic interaction fails? Because what you have written above:
> 
> > The fundamental issue was an incorrect base pointer being returned for event data. The SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory and Skeleton::RegisterGeneric functions were erroneously returning a pointer to the EventDataStorage object itself, ...
> 
> should have no effect in this case. Because a typed proxy/proxy event uses the `DataStoragePointer` here: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/generic_proxy_event.cpp#L156 , which the `GenericSkeleton` has placed into the `events_metainfo_` here: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp#L228 ... and this is the "correct" raw-data-storage pointer (without the enclosing `DynamicArray` control structure). So imho the Generic-Generic interaction should work, because:
> 
> * the GenericSkeleton/provider is storing the raw-pointer (not the pointer to a `DynamicArray`) into the "meta-info-map"
> * the GenericProxyEvent/consumer is getting this raw-.pointer from the "meta-info-map"
> 
> Am I missing something here? I did not yet re-run/use your tests as picking it from the fork is "tedious" ;)

@crimson11, You are absolutely correct, the consumer is looking at the correct memory location (data_storage->data()). However, the failure in the Generic-Generic interaction is caused by the provider side (GenericSkeletonEvent).
If we trace what gets returned to the provider during initialization:

- In SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory, it returns data_storage (which is a pointer to the EventDataStorage object itself, i.e., the DynamicArray control block), instead of data_storage->data().
https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp#L235

- This object pointer is propagated back through Skeleton::RegisterGeneric to GenericSkeletonEvent::PrepareOffer, where it is saved as event_data_storage_.

https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/generic_skeleton_event.cpp#L38

- Then, during GenericSkeletonEvent::Allocate(), the provider computes the memory address to write to by adding the offset directly to event_data_storage_ (the object pointer)

https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/generic_skeleton_event.cpp#L69

Because of this, the provider and consumer are using two different base addresses:

- Consumer: Reads from data_storage->data() (the actual allocated data buffer).

- Provider: Writes to data_storage + offset (overwriting the DynamicArray control block and corrupting shared memory).

crimson11:
@ShoroukRamzy
Thank you for the clarification! So right now we have these two issues:

1. `GenericSkeleton` / `GenericSkeletonEvent` is "completely broken" independent, whether the event-data gets consumed by a `GenericProxy` or a "typed" `Proxy`:
  - `GenericSkeletonEvent::Allocate()` allocates memory starting from a "base-pointer" pointing at the `EventDataStorage` type aka `DynamicArray<T>`, but thinks/expects, that its "base pointer" points to the underlying raw-array stored within `DynamicArray<T>`. Thus, it potentially overwrites the control-data part of the `DynamicArray<T>` and generally writes data to the wrong locations, since it uses the wrong base-pointer.
2. even if this general problem under (1) would NOT exist, a typed `Proxy` / `ProxyEvent` consuming the Event provided by the `GenericSkeleton` would currently access the event-slots by interpreting the `EventDataStorage` / `DynamicArray<T>`, but since the `GenericSkeletonEvent` and the typed `ProxyEvent` have a different interpretation of T and therefore the number of slots, the access of the typed `ProxyEvent` to the underlying data is "off".

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -731,4 +731,4 @@ EventMetaInfo& SkeletonMemoryManager::EmplaceEventMetaInfo(const ElementFqId ele
     return inserted_meta_info.first->second;
 }
 
-}  // namespace score::mw::com::impl::lola
+}  // namespace score::mw::com::impl::lola
```
crimson11: Remove this file from the commit ... you only provide tests in this PR.

score/mw/com/test/generic_skeleton/mw_com_config_generic_generic.json
```diff
@@ -0,0 +1,74 @@
+{
```
crimson11: Why did you duplicate the config file? Why not use the same config for all combinations of interactions?

ShoroukRamzy:
> @ShoroukRamzy Thank you for the clarification! So right now we have these two issues:
> 
> 1. `GenericSkeleton` / `GenericSkeletonEvent` is "completely broken" independent, whether the event-data gets consumed by a `GenericProxy` or a "typed" `Proxy`:
> 
> * `GenericSkeletonEvent::Allocate()` allocates memory starting from a "base-pointer" pointing at the `EventDataStorage` type aka `DynamicArray<T>`, but thinks/expects, that its "base pointer" points to the underlying raw-array stored within `DynamicArray<T>`. Thus, it potentially overwrites the control-data part of the `DynamicArray<T>` and generally writes data to the wrong locations, since it uses the wrong base-pointer.
> 
> 2. even if this general problem under (1) would NOT exist, a typed `Proxy` / `ProxyEvent` consuming the Event provided by the `GenericSkeleton` would currently access the event-slots by interpreting the `EventDataStorage` / `DynamicArray<T>`, but since the `GenericSkeletonEvent` and the typed `ProxyEvent` have a different interpretation of T and therefore the number of slots, the access of the typed `ProxyEvent` to the underlying data is "off".

yes exactly @crimson11

score/mw/com/test/generic_skeleton/generic_typed_interaction_32_byte_app.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+    if (!handles_res.has_value() || handles_res.value().empty())
+        return 1;
+
+    auto proxy_res = MyTestServiceProxy::Create(handles_res.value()[0]);
+    if (!proxy_res.has_value())
+        return 1;
+    auto& proxy = proxy_res.value();
+
+    uint64_t received = 30;
```
crimson11: why does `received` start with 30? You are using it as the loop-variable below? It should start with 0 (or 1)?
Otherwise it is confused with the "counter-value" sent by the provider ... which for some reason also starts with 30? What is the magic with 30 everywhere? :P

score/mw/com/test/generic_skeleton/generic_typed_interaction_app.cpp
```diff
@@ -0,0 +1,251 @@
+/********************************************************************************
```
crimson11, line 1: You need only this test application for generic/typed interaction! You can remove all the other variants for generic/typed interaction.
See further comment below.

score/mw/com/test/generic_skeleton/generic_typed_interaction_app.cpp
```diff
@@ -0,0 +1,251 @@
...
+    score::mw::log::LogInfo("GenericSkeletonProvider") << "Finished sending. Waiting for consumer to process...";
+    std::this_thread::sleep_for(std::chrono::seconds(2));
+
+    skeleton.StopOfferService();
+    score::mw::log::LogInfo("GenericSkeletonProvider") << "Shutting down.";
+    return 0;
+}
+
+template <typename Trait>
+class MyTestService : public Trait::Base
```
crimson11, line 119: See my initial comment! If you feel the need to show correct interaction with different event-data-sizes, then you just need this test and you add to your service-interface `MyTestService` additional event types (16 and 32 byte)

crimson11 (changes requested):

score/mw/com/test/generic_skeleton/generic_typed_interaction_app.cpp
ShoroukRamzy: @crimson11 , This was my first version of the test, then I separated them in different processes to be able to see the full behavior in terms of data integrity and number of slots check. Separation is needed as crashes happen due to the capacity (no. of slots issue) and prevent other events from being processed as well. we want to test on different event sizes and see the full behavior for each. Sorry I had to remove this file from the commit.

score/mw/com/test/generic_skeleton/generic_typed_interaction_app.cpp
ShoroukRamzy: @crimson11, see my comment below

score/mw/com/test/generic_skeleton/generic_typed_interaction_32_byte_app.cpp
ShoroukRamzy: yes, I did this for further investigation on the data integrity check. this follows the same test scenario as we send 30 values starting from 30 not 0. the idea I was trying to prove that if  we started to send data by any other value rather than 0, the offset ptr inside EventDataStorage header will be fully corrupted by the provider and proxy will read a garbage in the first read. Of course we can keep it just like other tests but this makes the issue more clear.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
ShoroukRamzy: Done

score/mw/com/test/generic_skeleton/mw_com_config_generic_generic.json
ShoroukRamzy: Done

score/mw/com/test/generic_skeleton/generic_typed_interaction_32_byte_app.cpp
ShoroukRamzy: I adjusted it to start from 0

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
crimson11: Hm - I guess we were cross-talking? You now removed `skeleton_memory_manager.cpp` from the repo in your commit? I meant: This file shouldn't show up as changed in your commit ;) - no need to touch/change this file

score/mw/com/test/generic_skeleton/generic_typed_interaction_app.cpp
crimson11: OK - I trust you, that you need different/separate apps ;)
But you have immense code duplication here! All the files `generic_typed_interaction_XX_byte_app.cpp`  are almost IDENTICAL! Just do a diff ...
And you already have a mechanism to bring in the PAYLOAD_SIZE via a define!

So:
- just provide ONE `generic_typed_interaction_app.cpp` - it just need eventually some additional `#ifdef PAYLOAD_SIZE`
- The variation you want/need you then get with the bazel instantiations of the app.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
ShoroukRamzy: Sorry, my mistake. Done!

score/mw/com/test/generic_skeleton/generic_typed_interaction_app.cpp
ShoroukRamzy: Yes I agree with you. I changed it to the macro version just like the generic-generic case.

ShoroukRamzy:
@crimson11, I marked the tests as xfail so the CI checks pass and we can merge this. I will remove the markers in a follow-up PR once the fix PR is merged

score/mw/com/test/generic_skeleton/generic_generic_interaction_app.cpp
```diff
@@ -0,0 +1,204 @@
...
+    auto& generic_event = event_it->second;
+    generic_event.Subscribe(kSamplesToSubscribe);
+
+    uint64_t expected = 0;
+    uint64_t received = 0;
+    int data_mismatches = 0;
+
+    while (received < kSamplesToProcess)
+    {
+        // std::cout << "[CONSUMER] " << PAYLOAD_SIZE << "-byte Waking up, calling GetNewSamples..." << std::endl;
```
crimson11: why commented out? Either remove or comment-in.

crimson11 (changes requested):
@ShoroukRamzy 
I'm basically happy, but there are the following issues left:

1. The general design of your test expects a certain "synchronization" between provider and consumer! I.e. afais - you expect, that the provider runs and provides N samples within a certain cycle and the consumer also cyclically polls and tries to receive the same amount N.
This is VERY brittle! In the CI/github, where these test will be run you can't control timing/scheduling! I.e. it could easily happen, that your consumer app gets NOT scheduled in time, will miss a sample and therefore will NEVER receive N samples ... leading to a test-failure!
How you should "re-design" your test-sequence:
   - provider/skeleton side shall run/produce samples forever. (have a look at bigdata-test).
   - in this case the provider has never to wait/sleep for its consumer! It is just "producing" and at the end, the test-infrastructure will just abort it.
    - for this you just have to "extend" your provider-code to react on abortion/termination. Have a look at `big_datatype_application.cpp` ... it basically relies on `SctfTestRunner::SetupSigTermHandler()` for this functionality - i.e. you can copy/paste this mechanism.
    - the consumer then does a configured amount of sample-receptions (like it is doing now). But since the producer is producing forever, there is no risk anymore, when the consumer misses samples!. If the consumer is done (N samples received), it exists successfully.
    - the enclosing python testrunner - see `wait_on_exit` in the python code - basically waits just for the exit of the consumer ... and then kills the provider.

If you follow this approach, the setup is much more stable and not that errorprone against scheduling glitches/load on the github servers.

2. Please stash your commits! In this case, it makes sense to only have two commits! The 1st one with all your changes and the 2nd one with the marking of the integration test as currently "fail".

Since the tests are basically fine (and others might want to use/see them in action), it is ok for me:
- if you now just tackle the commit-stashing. Then we should merge the current state.
- if you do the test-flowredesign, which I explained you above in an additional/later PR
Just tell me, what you prefer :P

score/mw/com/test/generic_skeleton/generic_generic_interaction_app.cpp
ShoroukRamzy: Done

ShoroukRamzy:
> @ShoroukRamzy I'm basically happy, but there are the following issues left:
> 
> 1. The general design of your test expects a certain "synchronization" between provider and consumer! I.e. afais - you expect, that the provider runs and provides N samples within a certain cycle and the consumer also cyclically polls and tries to receive the same amount N.
>    This is VERY brittle! In the CI/github, where these test will be run you can't control timing/scheduling! I.e. it could easily happen, that your consumer app gets NOT scheduled in time, will miss a sample and therefore will NEVER receive N samples ... leading to a test-failure!
>    How you should "re-design" your test-sequence:
>    
>    * provider/skeleton side shall run/produce samples forever. (have a look at bigdata-test).
>    * in this case the provider has never to wait/sleep for its consumer! It is just "producing" and at the end, the test-infrastructure will just abort it.
>    * for this you just have to "extend" your provider-code to react on abortion/termination. Have a look at `big_datatype_application.cpp` ... it basically relies on `SctfTestRunner::SetupSigTermHandler()` for this functionality - i.e. you can copy/paste this mechanism.
>    * the consumer then does a configured amount of sample-receptions (like it is doing now). But since the producer is producing forever, there is no risk anymore, when the consumer misses samples!. If the consumer is done (N samples received), it exists successfully.
>    * the enclosing python testrunner - see `wait_on_exit` in the python code - basically waits just for the exit of the consumer ... and then kills the provider.
> 
> If you follow this approach, the setup is much more stable and not that errorprone against scheduling glitches/load on the github servers.
> 
> 2. Please stash your commits! In this case, it makes sense to only have two commits! The 1st one with all your changes and the 2nd one with the marking of the integration test as currently "fail".
> 
> Since the tests are basically fine (and others might want to use/see them in action), it is ok for me:
> 
> * if you now just tackle the commit-stashing. Then we should merge the current state.
> * if you do the test-flowredesign, which I explained you above in an additional/later PR
>   Just tell me, what you prefer :P

Hi @crimson11, thank you for your feedback. I think it would be much better to go for the first option to merge the current state with only two commits then open a new PR for the improvement. In this case I squashed the commits to be two commits as you suggested.

crimson11:
OK. As discussed today: We will merge the test as is for now and I will create a new issue to address the test design/make it more stable in a separate issue, which I will assign to @ShoroukRamzy

crimson11 (approved):

crimson11:
@ShoroukRamzy - I created this issue: https://github.com/eclipse-score/communication/issues/512
Can you assign it to you?

ShoroukRamzy:
> @ShoroukRamzy - I created this issue: #512 Can you assign it to you?

Yes please @crimson11

PR #385, opened 2026-05-05
Fix flaky test

KrishaDeshkool:
Issue: https://github.com/eclipse-score/communication/issues/386

### DIFF
--- a/score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
+++ b/score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
@@ -39,6 +39,16 @@ class LolaGenericProxyEventFixture : public LolaProxyEventResources
         return *this;
     }
 
+    // ProxyEventCommon no longer Unsubscribes on destruction; release state-machine state explicitly.
+    void TearDown() override
+    {
+        if (generic_proxy_event_ != nullptr)
+        {
+            generic_proxy_event_->Unsubscribe();
+        }
+        ProxyMockedMemoryFixture::TearDown();
+    }
+
     std::unique_ptr<GenericProxyEvent> generic_proxy_event_{nullptr};
 };
 TEST_F(LolaGenericProxyEventFixture, CanConstructAGenericProxyEvent)
--- a/score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
+++ b/score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
@@ -39,6 +39,16 @@ class LolaGenericProxyEventFixture : public LolaProxyEventResources
         return *this;
     }
 
+    // ProxyEventCommon no longer Unsubscribes on destruction, so do it explicitly.
+    void TearDown() override
+    {
+        if (generic_proxy_event_ != nullptr)
+        {
+            generic_proxy_event_->Unsubscribe();
+        }
+        ProxyMockedMemoryFixture::TearDown();
+    }
+
     std::unique_ptr<GenericProxyEvent> generic_proxy_event_{nullptr};
 };
 TEST_F(LolaGenericProxyEventFixture, CanConstructAGenericProxyEvent)
--- a/score/mw/com/impl/bindings/lola/proxy.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy.cpp
@@ -263,50 +263,6 @@ ServiceDataStorage& GetServiceDataStorage(const memory::shared::ManagedMemoryRes
 
 }  // namespace detail_proxy
 
-class FindServiceGuard final
-{
-  public:
-    // Suppress "AUTOSAR C++14 A15-5-3" rule findings. This rule states: "The std::terminate() function shall not be
-    // called implicitly". std::terminate() is implicitly called from 'find_service_handle_result.value()' in case
-    // find_service_handle_result doesn't have a value but as we check before with 'has_value()' so no way for throwing
-    // std::bad_optional_access which leds to std::terminate().
-    // coverity[autosar_cpp14_a15_5_3_violation : FALSE]
-    FindServiceGuard(FindServiceHandler<HandleType> find_service_handler,
-                     EnrichedInstanceIdentifier enriched_instance_identifier)
-        : service_availability_change_handle_{nullptr}
-    {
-        auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
-        const auto find_service_handle_result = service_discovery.StartFindService(
-            std::move(find_service_handler), std::move(enriched_instance_identifier));
-        if (!find_service_handle_result.has_value())
-        {
-            score::mw::log::LogFatal("lola")
-                << "StartFindService failed with error" << find_service_handle_result.error() << ". Terminating.";
-            std::terminate();
-        }
-        service_availability_change_handle_ = std::make_unique<FindServiceHandle>(find_service_handle_result.value());
-    }
-
-    ~FindServiceGuard() noexcept
-    {
-        auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
-        const auto stop_find_service_result = service_discovery.StopFindService(*service_availability_change_handle_);
-        if (!stop_find_service_result.has_value())
-        {
-            score::mw::log::LogError("lola")
-                << "StopFindService failed with error" << stop_find_service_result.error() << ". Ignoring error.";
-        }
-    }
-
-    FindServiceGuard(const FindServiceGuard&) = delete;
-    FindServiceGuard& operator=(const FindServiceGuard&) = delete;
-    FindServiceGuard(FindServiceGuard&& other) = delete;
-    FindServiceGuard& operator=(FindServiceGuard&& other) = delete;
-
-  private:
-    std::unique_ptr<FindServiceHandle> service_availability_change_handle_;
-};
-
 std::atomic<ProxyInstanceIdentifier::ProxyInstanceCounter> Proxy::current_proxy_instance_counter_{0U};
 
 ElementFqId Proxy::EventNameToElementFqIdConverter::Convert(const std::string_view event_name) const noexcept
@@ -430,20 +386,22 @@ Proxy::Proxy(std::shared_ptr<memory::shared::ManagedMemoryResource> control,
       offered_state_machine_{},
       are_proxy_methods_setup_{false},
       filesystem_{filesystem},
-      find_service_guard_{std::make_unique<FindServiceGuard>(
-          [this](ServiceHandleContainer<HandleType> service_handle_container, FindServiceHandle) {
-              // Suppress Autosar C++14 A8-5-3 states that auto variables shall not be initialized using braced
-              // initialization. This is a false positive, we don't use auto here
-              // coverity[autosar_cpp14_a8_5_3_violation : FALSE]
-              std::lock_guard lock{proxy_event_registration_mutex_};
-              is_service_instance_available_ = !service_handle_container.empty();
-              ServiceAvailabilityChangeHandler(is_service_instance_available_);
-          },
-          EnrichedInstanceIdentifier{handle_})}
+      find_service_handle_{},
+      prepare_deinitialize_called_{false},
+      finalize_deinitialize_called_{false}
 {
+    StartProxyAutoReconnect();
 }
 
-Proxy::~Proxy() = default;
+Proxy::~Proxy()
+{
+    if (!prepare_deinitialize_called_ || !finalize_deinitialize_called_)
+    {
+        score::mw::log::LogFatal("lola")
+            << "Proxy destroyed without prior PrepareDeinitialize + FinalizeDeinitialize. Terminating.";
+        std::terminate();
+    }
+}
 
 void Proxy::ServiceAvailabilityChangeHandler(const bool is_service_available)
 {
@@ -927,6 +885,59 @@ void Proxy::RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod& 
     const auto [ignorable, was_inserted] = proxy_methods_.insert({method_id, proxy_method});
     score::cpp::ignore = ignorable;
     SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(was_inserted, "Method IDs must be unique!");
+}
+
+void Proxy::PrepareDeinitialize()
+{
+    StopProxyAutoReconnect();
+    prepare_deinitialize_called_ = true;
+}
+
+void Proxy::FinalizeDeinitialize()
+{
+    {
+        std::lock_guard lock{proxy_event_registration_mutex_};
+        event_bindings_.clear();
+        is_service_instance_available_ = false;
+    }
+    {
+        std::lock_guard lock{proxy_method_registration_mutex_};
+        proxy_methods_.clear();
+    }
+    finalize_deinitialize_called_ = true;
+}
+
+void Proxy::StartProxyAutoReconnect()
+{
+    auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
+    const auto find_service_handle_result = service_discovery.StartFindService(
+        [this](ServiceHandleContainer<HandleType> service_handle_container, FindServiceHandle) {
+            std::lock_guard lock{proxy_event_registration_mutex_};
+            is_service_instance_available_ = !service_handle_container.empty();
+            ServiceAvailabilityChangeHandler(is_service_instance_available_);
+        },
+        EnrichedInstanceIdentifier{handle_});
+    if (!find_service_handle_result.has_value())
+    {
+        score::mw::log::LogFatal("lola") << "StartFindService failed with error" << find_service_handle_result.error()
+                                         << ". Terminating.";
+        std::terminate();
+    }
+    find_service_handle_ = find_service_handle_result.value();
+}
+
+void Proxy::StopProxyAutoReconnect()
+{
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(find_service_handle_.has_value(),
+                                                      "StopProxyAutoReconnect: find_service_handle_ must be set");
+    auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
+    const auto stop_find_service_result = service_discovery.StopFindService(*find_service_handle_);
+    if (!stop_find_service_result.has_value())
+    {
+        score::mw::log::LogError("lola") << "StopFindService failed with error" << stop_find_service_result.error()
+                                         << ". Ignoring error.";
+    }
+    find_service_handle_.reset();
 }
 
 }  // namespace score::mw::com::impl::lola
--- a/score/mw/com/impl/bindings/lola/proxy.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy.cpp
@@ -263,50 +263,6 @@ ServiceDataStorage& GetServiceDataStorage(const memory::shared::ManagedMemoryRes
 
 }  // namespace detail_proxy
 
-class FindServiceGuard final
-{
-  public:
-    // Suppress "AUTOSAR C++14 A15-5-3" rule findings. This rule states: "The std::terminate() function shall not be
-    // called implicitly". std::terminate() is implicitly called from 'find_service_handle_result.value()' in case
-    // find_service_handle_result doesn't have a value but as we check before with 'has_value()' so no way for throwing
-    // std::bad_optional_access which leds to std::terminate().
-    // coverity[autosar_cpp14_a15_5_3_violation : FALSE]
-    FindServiceGuard(FindServiceHandler<HandleType> find_service_handler,
-                     EnrichedInstanceIdentifier enriched_instance_identifier)
-        : service_availability_change_handle_{nullptr}
-    {
-        auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
-        const auto find_service_handle_result = service_discovery.StartFindService(
-            std::move(find_service_handler), std::move(enriched_instance_identifier));
-        if (!find_service_handle_result.has_value())
-        {
-            score::mw::log::LogFatal("lola")
-                << "StartFindService failed with error" << find_service_handle_result.error() << ". Terminating.";
-            std::terminate();
-        }
-        service_availability_change_handle_ = std::make_unique<FindServiceHandle>(find_service_handle_result.value());
-    }
-
-    ~FindServiceGuard() noexcept
-    {
-        auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
-        const auto stop_find_service_result = service_discovery.StopFindService(*service_availability_change_handle_);
-        if (!stop_find_service_result.has_value())
-        {
-            score::mw::log::LogError("lola")
-                << "StopFindService failed with error" << stop_find_service_result.error() << ". Ignoring error.";
-        }
-    }
-
-    FindServiceGuard(const FindServiceGuard&) = delete;
-    FindServiceGuard& operator=(const FindServiceGuard&) = delete;
-    FindServiceGuard(FindServiceGuard&& other) = delete;
-    FindServiceGuard& operator=(FindServiceGuard&& other) = delete;
-
-  private:
-    std::unique_ptr<FindServiceHandle> service_availability_change_handle_;
-};
-
 std::atomic<ProxyInstanceIdentifier::ProxyInstanceCounter> Proxy::current_proxy_instance_counter_{0U};
 
 ElementFqId Proxy::EventNameToElementFqIdConverter::Convert(const std::string_view event_name) const noexcept
@@ -430,20 +386,22 @@ Proxy::Proxy(std::shared_ptr<memory::shared::ManagedMemoryResource> control,
       offered_state_machine_{},
       are_proxy_methods_setup_{false},
       filesystem_{filesystem},
-      find_service_guard_{std::make_unique<FindServiceGuard>(
-          [this](ServiceHandleContainer<HandleType> service_handle_container, FindServiceHandle) {
-              // Suppress Autosar C++14 A8-5-3 states that auto variables shall not be initialized using braced
-              // initialization. This is a false positive, we don't use auto here
-              // coverity[autosar_cpp14_a8_5_3_violation : FALSE]
-              std::lock_guard lock{proxy_event_registration_mutex_};
-              is_service_instance_available_ = !service_handle_container.empty();
-              ServiceAvailabilityChangeHandler(is_service_instance_available_);
-          },
-          EnrichedInstanceIdentifier{handle_})}
+      find_service_handle_{},
+      prepare_deinitialize_called_{false},
+      finalize_deinitialize_called_{false}
 {
+    StartProxyAutoReconnect();
 }
 
-Proxy::~Proxy() = default;
+Proxy::~Proxy()
+{
+    if (!prepare_deinitialize_called_ || !finalize_deinitialize_called_)
+    {
+        score::mw::log::LogFatal("lola")
+            << "Proxy destroyed without prior PrepareDeinitialize + FinalizeDeinitialize. Terminating.";
+        std::terminate();
+    }
+}
 
 void Proxy::ServiceAvailabilityChangeHandler(const bool is_service_available)
 {
@@ -927,6 +885,72 @@ void Proxy::RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod& 
     const auto [ignorable, was_inserted] = proxy_methods_.insert({method_id, proxy_method});
     score::cpp::ignore = ignorable;
     SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(was_inserted, "Method IDs must be unique!");
+}
+
+void Proxy::PrepareDeinitialize()
+{
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(!prepare_deinitialize_called_,
+                                                "PrepareDeinitialize must only be called once.");
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
+        !finalize_deinitialize_called_,
+        "Defensive programming: FinalizeDeinitialize must not have been called before PrepareDeinitialize (This should "
+        "never fail since we check this in FinalizeDeinitialize).");
+
+    StopProxyAutoReconnect();
+    prepare_deinitialize_called_ = true;
+}
+
+void Proxy::FinalizeDeinitialize()
+{
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(!finalize_deinitialize_called_,
+                                                "FinalizeDeinitialize must only be called once.");
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
+        prepare_deinitialize_called_, "PrepareDeinitialize must have been called before FinalizeDeinitialize.");
+
+    {
+        std::lock_guard lock{proxy_event_registration_mutex_};
+        event_bindings_.clear();
+        is_service_instance_available_ = false;
+    }
+    {
+        std::lock_guard lock{proxy_method_registration_mutex_};
+        proxy_methods_.clear();
+    }
+    finalize_deinitialize_called_ = true;
+}
+
+void Proxy::StartProxyAutoReconnect()
+{
+    auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
+    const auto find_service_handle_result = service_discovery.StartFindService(
+        [this](ServiceHandleContainer<HandleType> service_handle_container, FindServiceHandle) {
+            std::lock_guard lock{proxy_event_registration_mutex_};
+            is_service_instance_available_ = !service_handle_container.empty();
+            ServiceAvailabilityChangeHandler(is_service_instance_available_);
+        },
+        EnrichedInstanceIdentifier{handle_});
+    if (!find_service_handle_result.has_value())
+    {
+        score::mw::log::LogFatal("lola") << "StartFindService failed with error" << find_service_handle_result.error()
+                                         << ". Terminating.";
+        std::terminate();
+    }
+    find_service_handle_ = find_service_handle_result.value();
+}
+
+void Proxy::StopProxyAutoReconnect()
+{
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(
+        find_service_handle_.has_value(),
+        "StopProxyAutoReconnect: find_service_handle_ must be set in StartProxyAutoReconnect");
+    auto& service_discovery = impl::Runtime::getInstance().GetServiceDiscovery();
+    const auto stop_find_service_result = service_discovery.StopFindService(*find_service_handle_);
+    if (!stop_find_service_result.has_value())
+    {
+        score::mw::log::LogError("lola") << "StopFindService failed with error" << stop_find_service_result.error()
+                                         << ". Ignoring error.";
+    }
+    find_service_handle_.reset();
 }
 
 }  // namespace score::mw::com::impl::lola
--- a/score/mw/com/impl/bindings/lola/proxy.h
+++ b/score/mw/com/impl/bindings/lola/proxy.h
@@ -27,6 +27,7 @@
 #include "score/mw/com/impl/configuration/lola_service_instance_id.h"
 #include "score/mw/com/impl/configuration/lola_service_type_deployment.h"
 #include "score/mw/com/impl/configuration/quality_type.h"
+#include "score/mw/com/impl/find_service_handle.h"
 #include "score/mw/com/impl/handle_type.h"
 #include "score/mw/com/impl/proxy_binding.h"
 #include "score/mw/com/impl/proxy_event_binding_base.h"
@@ -62,7 +63,6 @@ namespace score::mw::com::impl::lola
 {
 
 class IShmPathBuilder;
-class FindServiceGuard;
 
 namespace detail_proxy
 {
@@ -200,8 +200,14 @@ class Proxy : public ProxyBinding
 
     void RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod& proxy_method) noexcept;
 
+    void PrepareDeinitialize() override;
+    void FinalizeDeinitialize() override;
+
   private:
     static std::atomic<ProxyInstanceIdentifier::ProxyInstanceCounter> current_proxy_instance_counter_;
+
+    void StartProxyAutoReconnect();
+    void StopProxyAutoReconnect();
 
     void ServiceAvailabilityChangeHandler(const bool is_service_available);
     void InitializeSharedMemoryForMethods(
@@ -261,9 +267,13 @@ class Proxy : public ProxyBinding
 
     score::filesystem::Filesystem filesystem_;
 
-    // We make find_service_guard_ the last member variable since it registers a handler which accesses member variables
-    // of this class, so they should be initialised first.
-    std::unique_ptr<FindServiceGuard> find_service_guard_;
+    /// Handle returned by ServiceDiscovery::StartFindService which is called for proxy auto reconnect, Held so we can
+    /// call StopFindService later from StopProxyAutoReconnect().
+    std::optional<FindServiceHandle> find_service_handle_;
+
+    /// Set by PrepareDeinitialize / FinalizeDeinitialize respectively. ~Proxy terminates unless both were called.
+    bool prepare_deinitialize_called_;
+    bool finalize_deinitialize_called_;
 };
 
 template <typename EventSampleType>
--- a/score/mw/com/impl/bindings/lola/proxy.h
+++ b/score/mw/com/impl/bindings/lola/proxy.h
@@ -27,6 +27,7 @@
 #include "score/mw/com/impl/configuration/lola_service_instance_id.h"
 #include "score/mw/com/impl/configuration/lola_service_type_deployment.h"
 #include "score/mw/com/impl/configuration/quality_type.h"
+#include "score/mw/com/impl/find_service_handle.h"
 #include "score/mw/com/impl/handle_type.h"
 #include "score/mw/com/impl/proxy_binding.h"
 #include "score/mw/com/impl/proxy_event_binding_base.h"
@@ -62,7 +63,6 @@ namespace score::mw::com::impl::lola
 {
 
 class IShmPathBuilder;
-class FindServiceGuard;
 
 namespace detail_proxy
 {
@@ -200,8 +200,19 @@ class Proxy : public ProxyBinding
 
     void RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod& proxy_method) noexcept;
 
+    /// \brief Stops auto-reconnect for this proxy and marks it ready for destruction.
+    /// \pre Must be called exactly once on a given Proxy. Calling it a second time will terminate.
+    void PrepareDeinitialize() override;
+
+    /// \brief Clears event and method registration state and marks the proxy ready for destruction.
+    /// \note Idempotent: calling it more than once is safe.
+    void FinalizeDeinitialize() override;
+
   private:
     static std::atomic<ProxyInstanceIdentifier::ProxyInstanceCounter> current_proxy_instance_counter_;
+
+    void StartProxyAutoReconnect();
+    void StopProxyAutoReconnect();
 
     void ServiceAvailabilityChangeHandler(const bool is_service_available);
     void InitializeSharedMemoryForMethods(
@@ -261,9 +272,13 @@ class Proxy : public ProxyBinding
 
     score::filesystem::Filesystem filesystem_;
 
-    // We make find_service_guard_ the last member variable since it registers a handler which accesses member variables
-    // of this class, so they should be initialised first.
-    std::unique_ptr<FindServiceGuard> find_service_guard_;
+    /// Handle returned by ServiceDiscovery::StartFindService which is called for proxy auto reconnect, Held so we can
+    /// call StopFindService later from StopProxyAutoReconnect().
+    std::optional<FindServiceHandle> find_service_handle_;
+
+    /// Set by PrepareDeinitialize / FinalizeDeinitialize respectively. ~Proxy terminates unless both were called.
+    bool prepare_deinitialize_called_;
+    bool finalize_deinitialize_called_;
 };
 
 template <typename EventSampleType>
--- a/score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
@@ -100,6 +100,16 @@ class LolaProxyEventCommonFixture : public ProxyMockedMemoryFixture
                     UnregisterEventNotification(QualityType::kASIL_QM, kElementFqId, my_handler_no, kDummyPid));
     }
 
+    // ProxyEventCommon no longer Unsubscribes on destruction; release state-machine state explicitly.
+    void TearDown() override
+    {
+        if (proxy_event_ != nullptr)
+        {
+            proxy_event_->Unsubscribe();
+        }
+        ProxyMockedMemoryFixture::TearDown();
+    }
+
   protected:
     ::testing::MockFunction<void()> event_handler_{};
     std::unique_ptr<T> proxy_event_{nullptr};
@@ -235,53 +245,6 @@ TYPED_TEST(LolaProxyEventCommonFixture, UnsubscribingWillUnregisterEventHandler)
 
     // and when the ProxyEvent unsubscribes
     this->proxy_event_->Unsubscribe();
-
-    // Then the receive handler should be unregistered
-    EXPECT_TRUE(handler_unregistered);
-}
-
-TYPED_TEST(LolaProxyEventCommonFixture, DestroyingTheProxyEventWillUnregisterEventHandler)
-{
-    this->RecordProperty("Verifies", "SCR-20236391, SCR-20237033");
-    this->RecordProperty(
-        "Description",
-        "Checks that a registered event handler will be unregistered when the (generic) proxy event is destroyed.");
-    this->RecordProperty("TestType", "Requirements-based test");
-    this->RecordProperty("Priority", "1");
-    this->RecordProperty("DerivationTechnique", "Analysis of requirements");
-
-    const std::size_t max_sample_count{1U};
-    constexpr const IMessagePassingService::HandlerRegistrationNoType my_handler_no = 37U;
-    bool handler_unregistered{false};
-
-    // Expecting that a receive handler will be registered
-    EXPECT_CALL(*this->mock_service_,
-                RegisterEventNotification(QualityType::kASIL_QM, kElementFqId, _, this->kDummyPid))
-        .WillOnce(Return(my_handler_no));
-
-    // and the same receive handler will be unregistered
-    EXPECT_CALL(*this->mock_service_,
-                UnregisterEventNotification(QualityType::kASIL_QM, kElementFqId, my_handler_no, this->kDummyPid))
-        .WillOnce(InvokeWithoutArgs([&handler_unregistered]() {
-            handler_unregistered = true;
-        }));
-
-    // Given a proxy that unsubscribes while waiting for being subscribed correctly
-    this->InitialiseProxyAndEvent();
-
-    // When we subscribe (sending a subscribe message to the producer)
-    std::ignore = this->proxy_event_->Subscribe(max_sample_count);
-
-    // and Register a receive handler
-    safecpp::Scope<> event_receive_handler_scope{};
-    std::ignore =
-        this->proxy_event_->SetReceiveHandler(FromMockFunction(event_receive_handler_scope, this->event_handler_));
-
-    // Then the receive handler should not be unregistered
-    EXPECT_FALSE(handler_unregistered);
-
-    // and when the ProxyEvent is destroyed
-    this->proxy_event_.reset();
 
     // Then the receive handler should be unregistered
     EXPECT_TRUE(handler_unregistered);
--- a/score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
@@ -100,6 +100,16 @@ class LolaProxyEventCommonFixture : public ProxyMockedMemoryFixture
                     UnregisterEventNotification(QualityType::kASIL_QM, kElementFqId, my_handler_no, kDummyPid));
     }
 
+    // ProxyEventCommon no longer Unsubscribes on destruction, so do it explicitly.
+    void TearDown() override
+    {
+        if (proxy_event_ != nullptr)
+        {
+            proxy_event_->Unsubscribe();
+        }
+        ProxyMockedMemoryFixture::TearDown();
+    }
+
   protected:
     ::testing::MockFunction<void()> event_handler_{};
     std::unique_ptr<T> proxy_event_{nullptr};
@@ -194,7 +204,10 @@ TYPED_TEST(LolaProxyEventCommonFixture, UnsubscribeImmediatelyAfterSubscribing)
 
 TYPED_TEST(LolaProxyEventCommonFixture, UnsubscribingWillUnregisterEventHandler)
 {
-    this->RecordProperty("Verifies", "SCR-21293524");
+    // SCR-20236391 and SCR-20237033 are split with CallsUnsubscribeOnDestruction in traits_test.cpp.
+    // this test covers Unsubscribe triggering UnregisterEventNotification, the other covers proxy
+    // destruction triggering Unsubscribe on the events and fields.
+    this->RecordProperty("Verifies", "SCR-21293524, SCR-20236391, SCR-20237033");
     this->RecordProperty(
         "Description",
         "Checks that calling Unsubscribe while currently subscribed will unregister a registered event "
@@ -235,53 +248,6 @@ TYPED_TEST(LolaProxyEventCommonFixture, UnsubscribingWillUnregisterEventHandler)
 
     // and when the ProxyEvent unsubscribes
     this->proxy_event_->Unsubscribe();
-
-    // Then the receive handler should be unregistered
-    EXPECT_TRUE(handler_unregistered);
-}
-
-TYPED_TEST(LolaProxyEventCommonFixture, DestroyingTheProxyEventWillUnregisterEventHandler)
-{
-    this->RecordProperty("Verifies", "SCR-20236391, SCR-20237033");
-    this->RecordProperty(
-        "Description",
-        "Checks that a registered event handler will be unregistered when the (generic) proxy event is destroyed.");
-    this->RecordProperty("TestType", "Requirements-based test");
-    this->RecordProperty("Priority", "1");
-    this->RecordProperty("DerivationTechnique", "Analysis of requirements");
-
-    const std::size_t max_sample_count{1U};
-    constexpr const IMessagePassingService::HandlerRegistrationNoType my_handler_no = 37U;
-    bool handler_unregistered{false};
-
-    // Expecting that a receive handler will be registered
-    EXPECT_CALL(*this->mock_service_,
-                RegisterEventNotification(QualityType::kASIL_QM, kElementFqId, _, this->kDummyPid))
-        .WillOnce(Return(my_handler_no));
-
-    // and the same receive handler will be unregistered
-    EXPECT_CALL(*this->mock_service_,
-                UnregisterEventNotification(QualityType::kASIL_QM, kElementFqId, my_handler_no, this->kDummyPid))
-        .WillOnce(InvokeWithoutArgs([&handler_unregistered]() {
-            handler_unregistered = true;
-        }));
-
-    // Given a proxy that unsubscribes while waiting for being subscribed correctly
-    this->InitialiseProxyAndEvent();
-
-    // When we subscribe (sending a subscribe message to the producer)
-    std::ignore = this->proxy_event_->Subscribe(max_sample_count);
-
-    // and Register a receive handler
-    safecpp::Scope<> event_receive_handler_scope{};
-    std::ignore =
-        this->proxy_event_->SetReceiveHandler(FromMockFunction(event_receive_handler_scope, this->event_handler_));
-
-    // Then the receive handler should not be unregistered
-    EXPECT_FALSE(handler_unregistered);
-
-    // and when the ProxyEvent is destroyed
-    this->proxy_event_.reset();
 
     // Then the receive handler should be unregistered
     EXPECT_TRUE(handler_unregistered);
--- a/score/mw/com/impl/bindings/lola/proxy_test.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy_test.cpp
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
     EXPECT_DEATH(start_find_service_fails(), ".*");
 }
 
+TEST_F(ProxyAutoReconnectDeathTest, StopProxyAutoReconnectTerminatesWhenCalledAfterHandleReset)
+{
+    // Given a proxy whose ctor populated find_service_handle_
+    InitialiseProxyWithConstructor(identifier_);
+
+    // When PrepareDeinitialize is called twice
+    // Then StopProxyAutoReconnect's precondition fires on the second call
+    EXPECT_DEATH(
+        {
+            proxy_->PrepareDeinitialize();  // resets find_service_handle_
+            proxy_->PrepareDeinitialize();  // assertion fires here
+        },
+        ".*");
+}
+
+TEST_F(ProxyAutoReconnectDeathTest, ProxyDestructionTerminatesWhenPrepareDeinitializeNotCalled)
+{
+    // When a proxy is constructed and destroyed without PrepareDeinitialize
+    // Then ~Proxy terminates because prepare_deinitialize_called_ is still false
+    EXPECT_DEATH(
+        {
+            InitialiseProxyWithConstructor(identifier_);
+            proxy_.reset();
+        },
+        ".*");
+}
+
+TEST_F(ProxyAutoReconnectDeathTest, ProxyDestructionTerminatesWhenFinalizeDeinitializeNotCalled)
+{
+    // Given a proxy on which PrepareDeinitialize has been called
+    // When it is destroyed without FinalizeDeinitialize
+    // Then ~Proxy terminates because finalize_deinitialize_called_ is still false
+    EXPECT_DEATH(
+        {
+            InitialiseProxyWithConstructor(identifier_);
+            proxy_->PrepareDeinitialize();
+            proxy_.reset();
+        },
+        ".*");
+}
+
 TEST_F(ProxyEventBindingFixture, RegisteringEventBindingWillCallNotifyServiceInstanceChangedAvailabilityOnBinding)
 {
     mock_binding::ProxyEventBase mock_proxy_event_base_binding{};
--- a/score/mw/com/impl/bindings/lola/proxy_test.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy_test.cpp
@@ -26,6 +26,8 @@
 
 #include "score/result/result.h"
 
+#include <score/assert_support.hpp>
+
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
 #include <chrono>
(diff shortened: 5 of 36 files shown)


score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -658,6 +658,12 @@ void Proxy::UnregisterEventBinding(const std::string_view service_element_name)
     // This is a false positive, we don't use auto here
     // coverity[autosar_cpp14_a8_5_3_violation : FALSE]
     std::lock_guard lock{proxy_event_registration_mutex_};
+    if (event_bindings_.size() == 0U)
```
bemerybmw: We already do this check below with a warning log. Why do we do it again here?

score/mw/com/impl/traits.h
```diff
@@ -358,6 +358,38 @@ class ProxyWrapperClass : public Interface<Trait>
         return proxy_wrapper;
     }
 
+    ~ProxyWrapperClass()
```
bemerybmw, line 361: This could actually be a scope exit so that ProxyWrapperClass can have defaulted destructor / move ops. We use ScopeExit in a few places, you can check some existing uses.

bemerybmw: Could you do the same thing for SkeletonWrapperClass?

score/mw/com/impl/proxy_event_base.cpp
```diff
@@ -153,6 +153,11 @@ void ProxyEventBase::Unsubscribe() noexcept
         return;
     }
 
+    if (binding_base_ == nullptr)
```
bemerybmw: This is required in case the binding was not created so the ProxyWrapperClass returns an error? Could you add a comment?

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,12 @@ class ProxyBinding
     virtual void UnregisterEventBinding(const std::string_view service_element_name) noexcept = 0;
 
     virtual Result<void> SetupMethods() = 0;
+
+    /// Stops any asynchronous activity in the binding
+    virtual void PrepareUnsubscribe() noexcept = 0;
```
bemerybmw: New functions should not be noexcept.

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,12 @@ class ProxyBinding
     virtual void UnregisterEventBinding(const std::string_view service_element_name) noexcept = 0;
 
     virtual Result<void> SetupMethods() = 0;
+
+    /// Stops any asynchronous activity in the binding
```
bemerybmw: These are functions implemented by every binding so we should be clear about what should be done in one function vs the other. I think that they should be PrepareUnsubscribe and FinalizeUnsubscribe. PrepareUnsubscribe will always be called before destroying the service elements and FinalizeUnsubscribe will be called after destroying the service elements. We previously had something like that on skeleton side but we removed it. Maybe we should reintroduce it for symmetry? What do you think @crimson11?

bemerybmw: They need to be properly documented.

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,12 @@ class ProxyBinding
...
+    virtual void PrepareUnsubscribe() noexcept = 0;
+
+    /// Clears binding-internal registration state
```
bemerybmw: 1. Add comment about why we need these functions i.e. because of the weird inheritance hierarchy. 
2. Remove FindServiceGuard. Should have private functions which are called explicitly e.g. StartProxyAutoReconnect / StopProxyAutoReconnect.
3. Have flags which are set in PrepareUnsubscribe, ShutdownBinding and checked in ~Proxy (terminate if not called). This is defensive programming since they should always be called before destroying a Proxy.

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
```diff
@@ -280,7 +289,8 @@ TYPED_TEST(LolaProxyEventCommonFixture, DestroyingTheProxyEventWillUnregisterEve
     // Then the receive handler should not be unregistered
     EXPECT_FALSE(handler_unregistered);
 
-    // and when the ProxyEvent is destroyed
+    // and when the ProxyEvent is unsubscribed and destroyed
```
bemerybmw: Is this really done when destroying the event? Or just when unsubscribing?

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +329,117 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+
+    MyProxy CreateProxy()
+    {
+        auto proxy_result = MyProxy::Create(handle_);
+        EXPECT_TRUE(proxy_result.has_value());
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD(proxy_result.has_value());
+        return std::move(proxy_result).value();
+    }
+
+    GeneratedProxyUnsubscribeRaiiFixture& GivenAProxy()
```
bemerybmw: Why do we need CreateProxy in addition to this function?

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +329,117 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+
+    std::optional<MyProxy> proxy_{};
+    bool proxy_prepare_unsubscribe_called_{false};
+    bool proxy_event_unsubscribe_called_{false};
+    bool proxy_field_unsubscribe_called_{false};
+    bool proxy_shutdown_binding_called_{false};
+};
+
+using GeneratedProxyDestructionFixture = GeneratedProxyUnsubscribeRaiiFixture;
+TEST_F(GeneratedProxyDestructionFixture, CallsUnsubscribeOnDestructionOfOwnedProxy)
```
bemerybmw: Why OwnedProxy?

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +329,117 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+    EXPECT_CALL(proxy_event_binding_mock_, Unsubscribe()).Times(1);
+    EXPECT_CALL(proxy_field_binding_mock_, Unsubscribe()).Times(1);
+    EXPECT_CALL(proxy_binding_mock_, ShutdownBinding()).Times(1);
+
+    // When destroying the proxy
+    proxy_.reset();
+}
+
+using GeneratedProxyMoveConstructionFixture = GeneratedProxyUnsubscribeRaiiFixture;
+TEST_F(GeneratedProxyMoveConstructionFixture, MoveConstructingDoesNotCallUnsubscribe)
```
bemerybmw, line 416: When you change ProxyWrapperClass to use ScopeExit then you can also remove the move tests (assuming that you add in a test like this one: https://github.com/KrishaDeshkool/communication/blob/ed71f9b487ae18305253dfb8367c0fe4590b3b9b/score/mw/com/impl/bindings/lola/messaging/method_subscription_registration_guard_test.cpp#L60)

score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
bemerybmw, line 1: This commit should be squashed with the previous one since tests should be merged alongside the changes.

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
```diff
@@ -175,7 +175,8 @@ TEST_F(ProxyWrapperTestClassCreateFixture, CallingFunctionsOnMockProxyDispatches
     // Expecting that OfferService will be called on the Proxy mock and Unsubscribe on the event and field mocks
     EXPECT_CALL(proxy_event_mock, Unsubscribe());
     EXPECT_CALL(proxy_event_mock_2, Unsubscribe());
-    EXPECT_CALL(proxy_field_mock, Unsubscribe());
+    // Times(2): once from the explicit call below, once from the RAII destructor in ~ProxyWrapperClass
+    EXPECT_CALL(proxy_field_mock, Unsubscribe()).Times(2);
```
bemerybmw: We currently have nothing stopping Unsubscribe from being called multiple times or even from being called before subscribe: https://github.com/KrishaDeshkool/communication/blob/ed71f9b487ae18305253dfb8367c0fe4590b3b9b/score/mw/com/impl/proxy_event_base.cpp#L148. We have a requirement that says that in the latter case, nothing should happen (https://github.com/eclipse-score/communication/blob/56c36d4059d276e804c143d14012576ddf1b9e25/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc#L2107). We now will also call Unsubscribe in the destructor of a ProxyWrapperClass which would recall unsubscribe on an event if already called. I think that we should have a flag in ProxyEventBase that is set when subscribed and unset when unsubscribed like we do on skeleton side. (https://github.com/KrishaDeshkool/communication/blob/ed71f9b487ae18305253dfb8367c0fe4590b3b9b/score/mw/com/impl/proxy_event_base.cpp#L148)

bemerybmw: I guess that we should do similar checks on skeleton side / in all events / fields as described in point 3 of https://github.com/eclipse-score/communication/pull/385#discussion_r3257711431

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: ah! this would be needed if we still had the EventBindingRegistrationGuard around,
after ShutdownBinding() clears event_bindings_, each EventBindingRegistrationGuard dtor still fires and calls UnregisterEventBinding. without this early-return, the LogWarn below would spam once per event/field on every clean shutdown.

now that EventBindingRegistrationGuard would be removed, i can remove this as well.

score/mw/com/impl/traits.h
KrishaDeshkool: ok

KrishaDeshkool: hmm.. okay i think we cannot use ScopeExit here,
why?
if we want to call the Unsubscribe inside the callback we pass in to scope exit,
something like ` ScopeExit guard_{[this]() { this->Unsubscribe(); }};`
Then we have to capture "this",
then when we move the wrapper `this` stored inside the lambda would be invalidated and it crashes.
i think this approach worked in other places where we do not capture to this but we capture something that class in which this scopeexit object lives are captured.

score/mw/com/impl/proxy_binding.h
KrishaDeshkool: oh boy, a lot of tests needed adaptation, done

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
KrishaDeshkool: hmm, this test name does not fit well now,
now destroying the proxy_event_ does not implicitly call Unsub() anymore,
we already have UnsubscribingWillUnregisterEventHandler test, shall i just remove this?

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: leftover from an earlier change where i had one test that compared owned vs moved-from proxies in the same body i split it into separate tests (DestroyingMovedToProxyCallsUnsubscribe, DestroyingMovedFromProxyDoesNotCallUnsubscribe), the "owned"  here is redundant. will rename to CallsUnsubscribeOnDestruction.

score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
KrishaDeshkool: both commits would be merged at once anyway since this is one PR,
but if you are fine with all the things in one commit then i will do it like this,
usually ppl ask for separate commit for tests in the past. since you are reviewing it i will squash as you asked.

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: leaving it as is since scope exit does not work here (atleast not without lot of rework)

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: ah right! this is no longer required since i do not have the move tests which had createProxy to create a second proxy, will remove now.

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
KrishaDeshkool: added

score/mw/com/impl/proxy_event_base.cpp
```diff
@@ -142,14 +147,30 @@ Result<void> ProxyEventBase::Subscribe(const std::size_t max_sample_count) noexc
...
+    // Unsubscribe before a successful Subscribe is a silent no-op.
+    if (!is_subscribed_flag_.IsSet())
+    {
+        return;
+    }
+
     if (proxy_event_base_mock_ != nullptr)
     {
         proxy_event_base_mock_->Unsubscribe();
+        is_subscribed_flag_.Clear();
```
bemerybmw: Since we have multiple return points in this function, we should create a small raii guard object which takes a reference to a bool and clears it on destruction. Probably makes sense to have something like: `ScopeExit is_subscribed_flag_guard{[&is_subscribed_flag_]{is_subscribed_flag_ = false; };`

score/mw/com/impl/proxy_event_base.cpp
```diff
@@ -142,14 +147,30 @@ Result<void> ProxyEventBase::Subscribe(const std::size_t max_sample_count) noexc
...
+        return;
+    }
+
+    // binding_base_ is null when binding creation failed during ProxyWrapperClass::Create(). so we return early in this
```
bemerybmw: Actually, this should be an assertion, right? There's no way for is_subscribed_flag_ to be set if the binding was not created successfully?

score/mw/com/impl/proxy_event_base.h
```diff
@@ -215,6 +199,8 @@ class ProxyEventBase
 
     IProxyEventBase* proxy_event_base_mock_;
 
+    FlagOwner is_subscribed_flag_{};
```
bemerybmw: Should be private? Also, we usually either initialize all members inline with the declarations or all in the constructor. (Makes it easier to confirm that all members are initialized and in the same order). And should be initialized with false.

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,14 @@ class ProxyBinding
     virtual void UnregisterEventBinding(const std::string_view service_element_name) noexcept = 0;
 
     virtual Result<void> SetupMethods() = 0;
+
+    /// Called by ProxyBase::Unsubscribe before unsubscribing the service elements. Stops any asynchronous activity
```
bemerybmw: The second sentence is only relevant for the lola binding so should go in lola/proxy.h

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,14 @@ class ProxyBinding
...
+    /// in the binding so that no callback can race with the element teardown that follows.
+    virtual void PrepareUnsubscribe() = 0;
+
+    /// Called by ProxyBase::Unsubscribe after the service elements have been unsubscribed. Clears any remaining
```
bemerybmw: The second sentence is only relevant for the lola binding so should go in lola/proxy.h (although it's currently quite vague so should either be a bit more descriptive or removed).

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,14 @@ class ProxyBinding
     virtual void UnregisterEventBinding(const std::string_view service_element_name) noexcept = 0;
 
     virtual Result<void> SetupMethods() = 0;
+
+    /// Called by ProxyBase::Unsubscribe before unsubscribing the service elements. Stops any asynchronous activity
+    /// in the binding so that no callback can race with the element teardown that follows.
```
bemerybmw: Please add a comment describing in more detail why we need this because of our architecture. If you're unsure or need help with the wording, let me know.

score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
bemerybmw: Commits should be split up into small but complete parts. E.g. in this PR, you probably should have had one commit which adds PrepareUnsubscribe / FinalizeUnsubscribe and a second commit which adds the checks to ~Proxy() which modifies a lot of tests. (and then another commit to add checks to ~Skeleton etc).

Splitting the pr up serves a few purposes: 1) It makes it easier to review and also if we see that the pr is too big, makes it easier to split it up. 2) In the future, it makes it easier to understand a full change by checking the commit without having to try to find the pr that merged the change (this is also why these commits like "Address review comments" are not good). 3) If we ever want to revert a change, then you can just revert a single commit. If the change and tests are in multiple commits, you would have to revert all of them (and if you don't realise there are multiple commits then you might end up manually reverting all the tests). E.G. It's possible in the future that we prefer not to have the check in ~Proxy(). With all the changes together as they are now, we can't revert the change at all because we still need the unsubscribe change.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -929,4 +885,61 @@ void Proxy::RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod&
...
+        score::mw::log::LogFatal("lola")
+            << "StartFindService failed with error" << find_service_handle_result.error() << ". Terminating.";
+        std::terminate();
+    }
+    find_service_handle_ = find_service_handle_result.value();
+}
+
+void Proxy::StopProxyAutoReconnect()
+{
+    if (!find_service_handle_.has_value())
```
bemerybmw: Should be an assertion? If the guard fails to be created then we terminate in StartProxyAutoReconnect.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -929,4 +885,61 @@ void Proxy::RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod&
...
+    {
+        return;
+    }
+    // StopFindService waits for any in-flight handler to finish; must run before we take
```
bemerybmw: I don't really understand the point of the second part of this comment. The handler is run in a separate thread so I wouldn't expect a deadlock, but anyway, we have no reason to lock proxy_event_registration_mutex_ here?

score/mw/com/impl/bindings/lola/proxy.h
```diff
@@ -261,9 +267,13 @@ class Proxy : public ProxyBinding
 
     score::filesystem::Filesystem filesystem_;
 
-    // We make find_service_guard_ the last member variable since it registers a handler which accesses member variables
-    // of this class, so they should be initialised first.
-    std::unique_ptr<FindServiceGuard> find_service_guard_;
+    /// Handle returned by ServiceDiscovery::StartFindService. Held so we can call StopFindService later from
```
bemerybmw:
```suggestion
    /// Handle returned by ServiceDiscovery::StartFindService which is called for proxy auto reconnect. Held so we can call StopFindService later from
```

score/mw/com/impl/bindings/lola/proxy.h
```diff
@@ -261,9 +267,13 @@ class Proxy : public ProxyBinding
...
+    /// StopProxyAutoReconnect().
+    std::optional<FindServiceHandle> find_service_handle_;
+
+    /// Set by PrepareUnsubscribe / FinalizeUnsubscribe respectively. ~Proxy terminates unless both were called.
+    bool prepare_unsubscribe_called_{false};
```
bemerybmw: Initialise them in the constructor

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
bemerybmw: Yep, it should be removed since it's testing the raii semantics which no longer exist.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -443,6 +446,16 @@ auto Skeleton::PrepareStopOffer(std::optional<UnregisterShmObjectTraceCallback>
     memory_manager_.Reset();
 }
 
+Skeleton::~Skeleton() noexcept
+{
+    if (prepare_offer_called_ && !prepare_stop_offer_called_)
```
bemerybmw, line 453: Hmmmm, this starts to become a bit problematic IMO because we start adding logic here which is reliant on the logic of the layer above (i.e. SkeletonBase only calls PrepareStopOffer if the service was offered). But then on proxy side, unsubscribe is always called. We should discuss this.

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
```diff
@@ -279,6 +279,8 @@ TEST_F(SkeletonPrepareOfferFixture, PrepareOfferWillRegisterServiceMethodSubscri
 
     // Then a valid result is returned
     EXPECT_TRUE(result.has_value());
+
+    skeleton_->PrepareStopOffer({});
```
bemerybmw: I would have some comments about this, but I suspect that this might change anyway.

score/mw/com/impl/mocking/proxy_event_mock_test.cpp
```diff
@@ -104,7 +104,9 @@ TYPED_TEST(ProxyEventFieldMockFixture, SubscribeReturnsErrorWhenMockReturnsError
 
 TYPED_TEST(ProxyEventFieldMockFixture, UnsubscribeDispatchesToMockAfterInjectingMock)
 {
-    // Given a ProxyEvent constructed with an empty binding and an injected mock
+    // Given a subscribed ProxyEvent with an injected mock
```
bemerybmw:
```suggestion
    // Given a ProxyEvent constructed with an empty binding and an injected mock which is currently subscribed
```

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
```diff
@@ -217,10 +226,14 @@ TEST_F(ProxyWrapperTestClassEventsOnlyCreateFixture, CallingFunctionsOnMockProxy
     auto& proxy_event_mock = std::get<0>(events_tuple).mock;
     proxy.some_event.InjectMock(proxy_event_mock);
 
+    // Subscribe first so the subscribed flag is set; otherwise Unsubscribe is a silent no-op.
```
bemerybmw: Expecting that...

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -751,6 +751,9 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture, UnsubscribeCallsAreTracedWhenEna
     // When a Proxy containing a ProxyEvent is created based on a lola deployment
     this->CreateProxy();
 
+    // and the event is in the subscribed state
+    ProxyEventBaseAttorney{this->proxy_->my_service_element_}.MarkAsSubscribed();
```
bemerybmw: Why not just subscribe?

score/mw/com/impl/proxy_event_base_test.cpp
```diff
@@ -257,8 +257,9 @@ TYPED_TEST(ProxyEventBaseUnsubscribeFixture, CallingUnsubscribeWhileSubscribedCa
     this->RecordProperty("Priority", "1");
     this->RecordProperty("DerivationTechnique", "Analysis of requirements");
 
-    // Given a Service Element, that is connected to a mock binding
+    // Given a subscribed Service Element, that is connected to a mock binding
     this->CreateServiceElement();
+    ProxyEventBaseAttorney{*this->service_element_}.MarkAsSubscribed();
```
bemerybmw: I'm very hesitant to introduce an attorney function like this unless it's absolutely necessary because it's putting the Proxy into an invalid state which we would never see in production.

score/mw/com/impl/traits.h
bemerybmw: Ah yep you're right. That's also why I didn't add it on skeleton side, forgot about that.

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +319,142 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+        return *this;
+    }
+
+    std::optional<MyProxy> proxy_{};
+};
+
+using GeneratedProxyDestructionFixture = GeneratedProxyUnsubscribeRaiiFixture;
+TEST_F(GeneratedProxyDestructionFixture, CallsUnsubscribeOnDestruction)
+{
+    RecordProperty("Description", "Checks that destroying a proxy unsubscribes events/fields and tears down binding");
```
bemerybmw: We only add these RecordProperty tags when linking a requirement. The issue with them in general is that it's very easy for them to get out of sync with the test. The test should be clear from the name and the test body itself.

bemerybmw: So in this case, the name does not perfectly describe the test since it doesn't mentoin that it calls unsubscribe on the events / fields.

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +319,142 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+    RecordProperty("TestType", "Requirements-based test");
+    RecordProperty("Priority", "1");
+    RecordProperty("DerivationTechnique", "Analysis of requirements");
+
+    // Given a subscribed proxy (binding mocks transition state from kNotSubscribed to kSubscribed on Subscribe)
+    GivenAProxy();
+    EXPECT_CALL(proxy_event_binding_mock_, GetSubscriptionState())
```
bemerybmw: IMO, this is not relevant to the test (it's rather relevant to the implementation of Subscribe) so should be removed.

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +319,142 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
+        .WillRepeatedly(Return(SubscriptionState::kSubscribed));
+    EXPECT_CALL(proxy_field_binding_mock_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
+        .WillRepeatedly(Return(SubscriptionState::kSubscribed));
+    EXPECT_CALL(proxy_event_binding_mock_, Subscribe(_)).WillOnce(Return(score::Result<void>{}));
```
bemerybmw: This also is not what what we're testing in this test: we're testing what happens when we call unsubscribe, not what happens when we call subscribe. So these should be ON_CALLS in the fixture.

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +319,142 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+    EXPECT_CALL(proxy_field_binding_mock_, Subscribe(_)).WillOnce(Return(score::Result<void>{}));
+    std::ignore = proxy_->some_event.Subscribe(1);
+    std::ignore = proxy_->some_field.Subscribe(1);
+
+    // Expecting that the binding-independent ProxyBase::Unsubscribe dispatches in this exact order: stop async source,
```
bemerybmw: stop async source -> PrepareUnsubscribe
clear binding-internal state -> FinalizeUnsubscribe

score/mw/com/impl/traits_test.cpp
bemerybmw: Also remove RecordProperty tags in other tests.

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +319,142 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+    EXPECT_CALL(proxy_binding_mock_, PrepareUnsubscribe()).Times(1);
+    EXPECT_CALL(proxy_binding_mock_, FinalizeUnsubscribe()).Times(1);
+    EXPECT_CALL(proxy_event_binding_mock_, Unsubscribe()).Times(0);
+    EXPECT_CALL(proxy_field_binding_mock_, Unsubscribe()).Times(0);
+
+    // When destroying the moved-from proxy
+    proxy_.reset();
+}
+
+// TODO: add MoveAssignmentFixture tests once the use-after-free during proxy wrapper move-assignment is resolved.
```
bemerybmw, line 464: Please create and link a ticket so that we don't forget

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +319,142 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+    RecordProperty("TestType", "Requirements-based test");
+    RecordProperty("Priority", "1");
+    RecordProperty("DerivationTechnique", "Analysis of requirements");
+
+    GivenAProxy();
+
+    // Across the whole test, expect exactly one PrepareUnsubscribe/FinalizeUnsubscribe — the moved-to proxy's
+    // destructor at end of scope. If move construction wrongly called them, the count would be 2 and the test fails.
+    // Per-event/-field Unsubscribe is gated by is_subscribed_flag_ which is not set here.
+    EXPECT_CALL(proxy_binding_mock_, PrepareUnsubscribe()).Times(1);
```
bemerybmw: You can check GeneratedSkeletonMoveConstructionFixture to see how to do this in a better way (that allows differentiating between these functions being called on destruction vs when moving.

bemerybmw (changes requested):

score/mw/com/impl/proxy_event_base.cpp
KrishaDeshkool: ah! right this is perfect place for the scope exit. will do

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: i see, this "we" is misleading,
This seems a old relic from when these were in the same method  there, the order within the function mattered-> calling StopFindService while holding proxy_event_registration_mutex_ would have deadlocked against the worker thread (which holds worker_mutex_ during handler dispatch, and whose handler wants proxy_event_registration_mutex_), this does not make sense anymore, removing

score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
KrishaDeshkool: i fully agree, Scope shifted a few times mid-review (started as a single race fix, then grew into the two-phase teardown and the defensive flags), which is how the commits ended up bundled. ;)
happy to split them up. :D

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: yes, added assertion and also tests to cover the same

score/mw/com/impl/bindings/lola/proxy.h
KrishaDeshkool: done

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
KrishaDeshkool: removed

score/mw/com/impl/bindings/lola/skeleton.cpp
KrishaDeshkool: let me cover the open points then we can talk about this tommorrow at work

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
KrishaDeshkool: Expecting that Subscribe is called

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
KrishaDeshkool: since all the tests i added was focusing on unscubscribe behaviour. i did not want to bloat that with subscription expectations(also i just wanted to test unsub unit behavior) but this opens a back door also,
i'm 50/50 on this one, i'm okay to remove this and add back subscription expectations like other tests.

score/mw/com/impl/proxy_event_base_test.cpp
KrishaDeshkool: same reason as easiler this just to make sure we only focus on the unsub behaviour, how we flip the subscription part is not interesting here, so i added this backdoor,
from your comment i feel like this is not really needed there are other ways to go about it, i removed it

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: right, i can make them on call (This is one of the reason why i added that attorny to remove these long setups for setting up subscriptions) now we have to do this to wiggle through the conditions we have.

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: you mean to remove the assertion macro or remove this subscription dance?
we do need them now, since we have checks if we are subscribed before we call unsubscribe,

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: also this test doesn't have to fail if there is some problem with subscription part, there are other tests to test it directly.

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: This is something that is about to merged or already merged no?
you were fixing this and it was approved already? i can create another issue on score but since it is already fixed i did not. let me know the status of that ticket if it takes longer then i will create a ticket

score/mw/com/impl/proxy_binding.h
KrishaDeshkool: added class hirarchy and destruction order,
if you need more info let me know i will add that.

score/mw/com/impl/proxy_binding.h
KrishaDeshkool: removed

score/mw/com/impl/bindings/lola/skeleton.cpp
KrishaDeshkool: but that is asymmetry we designed right?
for a proxy as soon as it is created it starts to look for services so there is no created but not doing anything state,
but for the skeleton there is a state where it does not offer anything yet until someone calls offer service on it,
so this is something we cannot change in current architecture,
meaning,
proxy needs something to be cleaned up even if it did not subscribe to aynthing (we need to stop that handle looking for services)
but on the skeleton if no one offered there is no shm artifacts to clear no?

lets discuss this.

score/mw/com/impl/plumbing/proxy_event_field_binding_factory_test.cpp
```diff
@@ -140,6 +140,20 @@ class ProxyServiceElementBindingFactoryParamaterisedFixture : public lola::Proxy
     ServiceElementTypes service_element_type_{GetParam()};
     std::unique_ptr<ProxyBase> proxy_base_{nullptr};
     DummyInstanceIdentifierBuilder dummy_instance_identifier_builder_{};
+
+    void TearDown() override
```
bemerybmw, line 144: Should we use a guard here for calling shutdown functions or use the teardown function?

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
```diff
@@ -393,24 +399,32 @@ TEST_F(SkeletonPrepareOfferFixture, PrepareOfferWillNotRegisterServiceMethodSubs
...
 }
 
 TEST_F(SkeletonPrepareOfferFixture, PrepareOfferWillNotCallUnregisterSubscribedMethodHandler)
```
KrishaDeshkool, line 419: This test does not make sense now-

KrishaDeshkool: if we call PrepareStopOffer then we call the unregister but this is not because we called PrepareOffer,(test is about PrepareOfferWillNotCallUnregisterSubscribedMethodHandler)
so what i think we can do is,
keep the expecations like do prepare offer and check no unregister is called and clear the expecations and then call the PrepareStopOffer then our expectations are cleared so our test pass,
feels like a hack, thats why opening a dicussion.

score/mw/com/impl/test/proxy_resources.h
```diff
@@ -70,6 +70,11 @@ class ProxyEventBaseAttorney
         return proxy_event_base_.tracing_data_;
     }
 
+    void MarkAsSubscribed() noexcept
```
KrishaDeshkool: this is no longer requried

score/mw/com/impl/generic_proxy.h
```diff
@@ -82,6 +91,10 @@ class GenericProxy : public ProxyBase
     void FillEventMap(const std::vector<std::string_view>& event_names) noexcept;
 
     EventMap events_;
+
+    /// Flag which is checked before calling Unsubscribe in the destructor.
+    /// Cleared on move so the moved-from instance does not call Unsubscribe.
+    FlagOwner is_proxy_owner_{true};
```
KrishaDeshkool: class init

score/mw/com/impl/plumbing/proxy_event_field_binding_factory_test.cpp
KrishaDeshkool: after the discussion it was decided to leaving it as is for now.

score/mw/com/impl/traits_test.cpp
KrishaDeshkool: EXPECT_CALL(proxy_field_binding_mock_, GetSubscriptionState())
This has to be an Expect call so we can have different behaviour when called two times,
ON call only allows us to set one behaviour, no?

KrishaDeshkool:
test failed due to know issue : ERROR ../score_itf+ - OSError: [Errno 28] No space left on device: '/home/run...

KrishaDeshkool:
recheck

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
```diff
@@ -240,53 +250,6 @@ TYPED_TEST(LolaProxyEventCommonFixture, UnsubscribingWillUnregisterEventHandler)
     EXPECT_TRUE(handler_unregistered);
 }
 
-TYPED_TEST(LolaProxyEventCommonFixture, DestroyingTheProxyEventWillUnregisterEventHandler)
-{
-    this->RecordProperty("Verifies", "SCR-20236391, SCR-20237033");
```
bemerybmw, line 245: This test was previously linked to requirements. As far as I can see, these are no longer linked to another test. Although, I don't know if we still use the same codes in score. We should:
- check that these requirement codes are still valid.
- Link existing tests (I guess that we should link a test that shows that the handler is unregistered on unsubscribe i.e. UnsubscribingWillUnregisterEventHandler and another test (I guess in traits_test or proxy_base_test) which shows that unsubscribe is called on destruction.

score/mw/com/impl/proxy_event_base.cpp
```diff
@@ -142,17 +149,31 @@ Result<void> ProxyEventBase::Subscribe(const std::size_t max_sample_count) noexc
...
 
 void ProxyEventBase::Unsubscribe() noexcept
 {
+    // Unsubscribe before a successful Subscribe is a silent no-op.
+    if (!is_subscribed_flag_.IsSet())
+    {
+        return;
+    }
+
+    utils::ScopeExit clear_subscribed_flag_on_exit{[&is_subscribed_flag_ = is_subscribed_flag_] {
```
bemerybmw, line 164: I've never seen this way of capturing a member variable. Is there a good reason for doing this (i.e. a reason to do this instead of capturing this) or you just came up with it on a whim?

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,25 @@ class ProxyBinding
     virtual void UnregisterEventBinding(const std::string_view service_element_name) noexcept = 0;
 
     virtual Result<void> SetupMethods() = 0;
+
+    /// Called by ProxyBase::Deinitialize before unsubscribing the service elements.
```
bemerybmw:
```suggestion
    /// \brief Called by ProxyBase::Deinitialize before unsubscribing the service elements.
```

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -929,4 +887,57 @@ void Proxy::RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod&
...
+                                         << ". Terminating.";
+        std::terminate();
+    }
+    find_service_handle_ = find_service_handle_result.value();
+}
+
+void Proxy::StopProxyAutoReconnect()
+{
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(find_service_handle_.has_value(),
+                                                      "StopProxyAutoReconnect: find_service_handle_ must be set");
```
bemerybmw:
```suggestion
                                                      "StopProxyAutoReconnect: find_service_handle_ must be set in StartProxyAutoReconnect");
```

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
     EXPECT_DEATH(start_find_service_fails(), ".*");
 }
 
+TEST_F(ProxyAutoReconnectDeathTest, StopProxyAutoReconnectTerminatesWhenCalledAfterHandleReset)
```
bemerybmw: Here you're testing implementation details. Unit tests should generally be black box tests, i.e. they shouldn't know the implementation but rather should be checking behaviour of the class's public api. So in this case, I would rather call this: StopProxyAutoReconnectTerminatesWhenCallingPrepareDeinitializeTwice. The other test names are good.

bemerybmw: This should also be added as a precondition in the documentation of PrepareDeinitialize that it can only be called once.

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
...
+{
+    // Given a proxy whose ctor populated find_service_handle_
```
bemerybmw: Don't mention the impl detail of find_service_handle_

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
...
+    InitialiseProxyWithConstructor(identifier_);
+
+    // When PrepareDeinitialize is called twice
+    // Then StopProxyAutoReconnect's precondition fires on the second call
+    EXPECT_DEATH(
```
bemerybmw: This should be a precondition violation test instead of a death test.

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
...
+        {
+            proxy_->PrepareDeinitialize();  // resets find_service_handle_
```
bemerybmw: If you call both here, then this test will pass even if we crash in the first call to PrepareDeinitialize. I would structure the test like this:
```
// Given a proxy on which PrepareDeinitialize has already been called
 
// When calling PrepareDeinitialize a second time
// Then the program terminates
SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(proxy_->PrepareDeinitialize());
```

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
...
+        ".*");
+}
+
+TEST_F(ProxyAutoReconnectDeathTest, ProxyDestructionTerminatesWhenPrepareDeinitializeNotCalled)
+{
+    // When a proxy is constructed and destroyed without PrepareDeinitialize
+    // Then ~Proxy terminates because prepare_deinitialize_called_ is still false
+    EXPECT_DEATH(
+        {
+            InitialiseProxyWithConstructor(identifier_);
```
bemerybmw: athis should be part of the given section?

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +552,47 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
...
+}
+
+TEST_F(ProxyAutoReconnectDeathTest, ProxyDestructionTerminatesWhenFinalizeDeinitializeNotCalled)
+{
+    // Given a proxy on which PrepareDeinitialize has been called
+    // When it is destroyed without FinalizeDeinitialize
+    // Then ~Proxy terminates because finalize_deinitialize_called_ is still false
+    EXPECT_DEATH(
+        {
+            InitialiseProxyWithConstructor(identifier_);
```
bemerybmw: Same points as above

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
bemerybmw: This should be part of the teardown of the fixture

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
```diff
@@ -172,12 +172,21 @@ TEST_F(ProxyWrapperTestClassCreateFixture, CallingFunctionsOnMockProxyDispatches
     proxy.some_event_2.InjectMock(proxy_event_mock_2);
     proxy.some_field.InjectMock(proxy_field_mock);
 
-    // Expecting that OfferService will be called on the Proxy mock and Unsubscribe on the event and field mocks
+    // Subscribe first so the per-event subscribed flag is set; otherwise Unsubscribe is a silent no-op.
```
bemerybmw: Expecting that...

score/mw/com/impl/traits_test.cpp
```diff
@@ -328,6 +329,130 @@ TEST_F(GeneratedProxyCreationTestFixture, CallingSubscribeOnServiceElementsDispa
...
+        return *this;
+    }
+
+    GeneratedProxyUnsubscribeRaiiFixture& WhichHasSubscribedEventsAndFields()
+    {
+        std::ignore = proxy_->some_event.Subscribe(1);
+        std::ignore = proxy_->some_field.Subscribe(1);
+        // After Subscribe completes, transition the GetSubscriptionState defaults from kNotSubscribed to
+        // kSubscribed so subsequent Unsubscribe calls take the actually-unsubscribe branch.
+        ON_CALL(proxy_event_binding_mock_, GetSubscriptionState())
```
bemerybmw: Interleaving expectations with calls is UB in gtest. So I guess we should do what we discussed in the call i.e. put this in the given part of the tests EXPECT_CALL(*this->mock_service_element_binding_, GetSubscriptionState())
        .WillOnce(Return(SubscriptionState::kNotSubscribed))
        .WillOnce(Return(SubscriptionState::kSubscribed));

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
bemerybmw: Hmmm, I think that VerifyAndClearExpectations is intended to be used when you're not sure that a mock object will be destroyed (https://google.github.io/googletest/gmock_cook_book.html#forcing-a-verification), although I guess it should also work in this case. Not sure what happens with default behaviour set via ON_CALL after this is called though (i.e. default behaviour that we need for PrepareStopOffer). 

Usually what I would do is to invoke a lambda when UnregisterOnServiceMethodSubscribedHandler is called which sets a flag and then you can check that it's not called in between PrepareOffer and PrepareStopOffer. @limdor what do you think?

bemerybmw (changes requested):

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -786,6 +797,29 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture,
...
 
+    // Expect that GetSubscriptionState returns kNotSubscribed then kSubscribed
+    EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
+        .WillOnce(Return(SubscriptionState::kSubscribed));
+
+    // Expect that a trace call relating to Subscribe is called
+    tracing::ITracingRuntime::TracePointType subscribe_trace_point_type{
+        ProxyEventTracingFixture<TypeParam>::TracePointType::SUBSCRIBE};
+    EXPECT_CALL(tracing_runtime_mock,
```
bemerybmw: This shouldn't be needed for thit test?

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -786,6 +797,29 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture,
...
+                Trace(BindingType::kLoLa,
+                      expected_service_element_instance_identifier_view,
+                      subscribe_trace_point_type,
+                      std::optional<tracing::ITracingRuntime::TracePointDataId>{},
+                      _,
+                      _))
+        .WillOnce(Return(Result<void>{}));
+
+    // and that GetBindingType is called twice on the proxy event binding
+    EXPECT_CALL(*this->mock_proxy_event_binding_, GetBindingType()).Times(2).WillRepeatedly(Return(BindingType::kLoLa));
```
bemerybmw: This should be default behaviour

score/mw/com/impl/proxy_event_base.cpp
KrishaDeshkool: i did not want to give this lambda access to "this", we are only interested in this one flag so only capturing that by ref,
if it was more than one member of some other methods then capturing this would make sense,
i did not see a reason to capture "this" here.(just keeping the exposure small)

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
KrishaDeshkool: hmm, true but this is the same pattern in other tests in this file, not something we introduced here. i'll update them all now. these things keep blowing this PR up though :D

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
KrishaDeshkool: so, [here](https://github.com/eclipse-score/communication/pull/385/changes/56f9cc04d5fed2e5a9e47fc19b3ee4f7e9ff3dd9#diff-43a44d576c9ba7d7ef616ef82be57860c130cc421dc70c3cefe5a37756e92bc9L771) we say enable subscribe = true,
then when we call the Subscribe() it internally calls the 
`tracing::TraceSubscribe(tracing_data_, *binding_base_, max_sample_count);`
that eventually calls Trace that's why we have this,

you mean this shouldn't be an Expect call ? that i agree this can be ON CALL. but i disagree that this is not needed.

KrishaDeshkool: again, we have all these dance steps because we called Subscribe() unlike on main where we just called Unsubscribe() directly.

score/mw/com/impl/bindings/lola/proxy_test.cpp
KrishaDeshkool: nice, thanks will update it.

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
KrishaDeshkool: ah! this seems bad!
i could not just switch to ON Call since there are other Expect calls on this mock, 
tried making this a nice mock that also failed. 
do you have a better idea?

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
KrishaDeshkool: hmm, can we done but with a catch,
i dont think this skeleton_->PrepareStopOffer({}); is idempotent?
there are tests that call this mid test and also at the beginning then we need a way to keep track of that before we call this in teardown, that is another layer of indirection,
so i left it as is.
if you think that would be better let me know, we can talk about it.

KrishaDeshkool: Then on each test which calls PrepareStopOffer() we also have to set this boolean to keep tract of or create some wrapper that knows it was explicitly called so it can skip calling PrepareStopOffer again in teardown.
but again, this is more complexity in the long run.

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
bemerybmw: Ok, I didn't realise that mocks have that limitation.  I would then mark enable_subscribe as false at the start of the function so that we never get the trace call.

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
bemerybmw: In general, you don't have to fix other tests if it's not related to your pr. But we should make sure that any new tests that we add are the highest quality that they can be (unless writing them properly would be a huge time investment). So in this case, it also would have been fine to add the ON_CALL to the fixture and just remove these in this test.

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
KrishaDeshkool: okay, we have other set_receive state so we can verify that this was not changes,
2 should be enough for this test POV, 
will remove the subscription related boolean

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
KrishaDeshkool: yes but, mostly these type of followups are usually forgotten even if we create tickets we don't have enough time to come back to it,
i will do this fix in this PR, if there are more findings like these i will create tickets and link them as TODOs in code.

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
limdor: >Usually what I would do is to invoke a lambda when UnregisterOnServiceMethodSubscribedHandler is called which sets a flag and then you can check that it's not called in between PrepareOffer and PrepareStopOffer. @limdor what do you think?

That is also what I would have done. I do not know if there is a better way with gtest.

score/mw/com/impl/proxy_event_base.cpp
bemerybmw: Nice, I haven't seen it before but it makes a lot of sense.

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,29 @@ class ProxyBinding
...
+    ///   2. ~MyInterface body, then its members (some_event/some_field/some_method) destroyed
+    ///   3. ~ProxyBase body, then its members (events_ map of references, proxy_binding_) destroyed
+    ///
+    /// By step 3 the events and fields are gone, so any binding callback that still touches them would be
+    /// a use-after-free. PrepareDeinitialize has to run at step 1 while they are still around.
+    virtual void PrepareDeinitialize() = 0;
+
+    /// \brief Clears binding-internal registration state.
+    ///        Called by ProxyBase::Deinitialize after the service elements have been unsubscribed.
+    /// \pre Must be called after PrepareDeinitialize.
```
bemerybmw, line 86: If you're going to add this precondition, then we should check it in the code and have a test for it.

score/mw/com/impl/bindings/lola/proxy_test.cpp
```diff
@@ -552,6 +554,38 @@ TEST_F(ProxyAutoReconnectDeathTest, ProxyCreateWillTerminateIfStartFindServiceRe
     EXPECT_DEATH(start_find_service_fails(), ".*");
 }
 
+TEST_F(ProxyAutoReconnectDeathTest, StopProxyAutoReconnectTerminatesWhenCallingPrepareDeinitializeTwice)
```
bemerybmw:
```suggestion
TEST_F(ProxyAutoReconnectDeathTest, PrepareDeinitializeTerminatesWhenCalledTwice)
```

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
```diff
@@ -258,6 +266,9 @@ class SkeletonMethodHandlingFixture : public SkeletonMockedMemoryFixture
     std::optional<IMessagePassingService::ServiceMethodSubscribedHandler> captured_method_subscribed_handler_b_{};
 
     safecpp::Scope<> method_call_registration_guard_scope_{};
+
+  protected:
```
bemerybmw: - If you make it prepare_stop_offer_called_, then what happens if we have a test which doesn't call offer or stop offer? Then you'd have to set this to true artificially in order to not crash?
- We're not taking advantage of the fixture: we call WhichIsOffered() which would be the perfect place to mark prepare_offer_called_ so that it doesn't have to be called in each test. You would then just have to manually set that value in the tests in which you offer the service without using the fixture (when the offer service doesn't fail).

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
KrishaDeshkool: Does these Expect calls here make sense for this test?
https://github.com/eclipse-score/communication/pull/385/changes#diff-a71c8a31e7dcd475fad7f42d52ed966a2fdcd6a6b45650c76f5f63f228bea035R414

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
KrishaDeshkool: linked to two tests as discussed.

score/mw/com/impl/mocking/proxy_wrapper_class_test_view_test.cpp
```diff
@@ -172,12 +172,22 @@ TEST_F(ProxyWrapperTestClassCreateFixture, CallingFunctionsOnMockProxyDispatches
     proxy.some_event_2.InjectMock(proxy_event_mock_2);
     proxy.some_field.InjectMock(proxy_field_mock);
 
-    // Expecting that OfferService will be called on the Proxy mock and Unsubscribe on the event and field mocks
+    // and given that Subscribe on each event and field mock returns success (so the per-event subscribed
+    // flag is set; otherwise the later Unsubscribe is a silent no-op).
+    ON_CALL(proxy_event_mock, Subscribe(_)).WillByDefault(Return(score::Result<void>{}));
```
bemerybmw: This default behaviour should go in the fixture.same for other places in this commit

score/mw/com/impl/proxy_base.h
```diff
@@ -131,6 +131,11 @@ class ProxyBase
     ProxyBase(ProxyBase&& other) noexcept;
     ProxyBase& operator=(ProxyBase&& other) noexcept;
 
+    /// \brief Deinitialize all events and fields and tears down binding-level state.
+    /// \pre Must only be invoked from the wrapper class destructor (ProxyWrapperClass / GenericProxy).
```
bemerybmw: I don't think this is a precondition. It's about how it should be called.

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,29 @@ class ProxyBinding
     virtual void UnregisterEventBinding(const std::string_view service_element_name) noexcept = 0;
 
     virtual Result<void> SetupMethods() = 0;
+
+    /// \brief Stops async sources (e.g. FindService callbacks) so they cannot fire during teardown.
```
bemerybmw: This line is lola specific. This function can contain anything that needs to be called before cleaning up the events.

score/mw/com/impl/proxy_binding.h
```diff
@@ -62,6 +62,29 @@ class ProxyBinding
...
+    /// C++ destruction order:
+    ///   1. ~ProxyWrapperClass body
+    ///   2. ~MyInterface body, then its members (some_event/some_field/some_method) destroyed
+    ///   3. ~ProxyBase body, then its members (events_ map of references, proxy_binding_) destroyed
+    ///
+    /// By step 3 the events and fields are gone, so any binding callback that still touches them would be
+    /// a use-after-free. PrepareDeinitialize has to run at step 1 while they are still around.
+    virtual void PrepareDeinitialize() = 0;
+
+    /// \brief Clears binding-internal registration state.
```
bemerybmw: I don't really know what binding-internal registration state. Again, I think this should just say that it's any cleanup code that must be called after events are cleaned up.

score/mw/com/impl/bindings/lola/proxy.h
```diff
@@ -200,9 +200,20 @@ class Proxy : public ProxyBinding
 
     void RegisterMethod(const UniqueMethodIdentifier method_id, ProxyMethod& proxy_method) noexcept;
 
+    /// \brief Stops auto-reconnect for this proxy and marks it ready for destruction.
+    /// \pre Must be called exactly once on a given Proxy. Calling it a second time will terminate.
+    void PrepareDeinitialize() override;
+
+    /// \brief Clears event and method registration state and marks the proxy ready for destruction.
+    /// \note Idempotent: calling it more than once is safe.
```
bemerybmw, line 208: I don't think it makes sense that PrepareDeinitialize _must_ only be called once and this can be called multiple times. Either this should crash if called a second time (I think this makes sense so we can catch programming bugs) or PrepareDeinitialize should return quietly without doing anything if called a second time.

score/mw/com/impl/mocking/proxy_event_mock_test.cpp
```diff
@@ -104,7 +104,9 @@ TYPED_TEST(ProxyEventFieldMockFixture, SubscribeReturnsErrorWhenMockReturnsError
 
 TYPED_TEST(ProxyEventFieldMockFixture, UnsubscribeDispatchesToMockAfterInjectingMock)
 {
-    // Given a ProxyEvent constructed with an empty binding and an injected mock
+    // Given a ProxyEvent constructed with an empty binding and an injected mock which is currently subscribed
+    ON_CALL(this->proxy_service_element_mock_, Subscribe(kDummyMaxSampleCount)).WillByDefault(Return(Result<void>{}));
```
bemerybmw: fixture

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -726,6 +720,14 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture, UnsubscribeCallsAreTracedWhenEna
         typename ProxyEventTracingFixture<TypeParam>::TracePointType>(
         expected_enabled_trace_points, kServiceIdentifierStringView, kServiceElementName, kInstanceSpecifierStringView);
 
+    // Expect that GetSubscriptionState returns kNotSubscribed then kSubscribed
+    EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
+        .WillOnce(Return(SubscriptionState::kSubscribed));
+
+    // and that Subscribe will be called on the binding
+    EXPECT_CALL(*this->mock_proxy_event_binding_, Subscribe(max_sample_count));
```
bemerybmw: not relevant for test

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -786,6 +786,26 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture,
         typename ProxyEventTracingFixture<TypeParam>::TracePointType>(
         expected_enabled_trace_points, kServiceIdentifierStringView, kServiceElementName, kInstanceSpecifierStringView);
 
+    // Expect that GetSubscriptionState returns kNotSubscribed then kSubscribed
+    EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
+        .WillOnce(Return(SubscriptionState::kSubscribed));
+
+    // Expect that a trace call relating to Subscribe is called
+    tracing::ITracingRuntime::TracePointType subscribe_trace_point_type{
```
bemerybmw: disable enable_subscribe?

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -786,6 +786,26 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture,
...
+                Trace(BindingType::kLoLa,
+                      expected_service_element_instance_identifier_view,
+                      subscribe_trace_point_type,
+                      std::optional<tracing::ITracingRuntime::TracePointDataId>{},
+                      _,
+                      _))
+        .WillOnce(Return(Result<void>{}));
+
+    // and that Subscribe will be called on the binding
+    EXPECT_CALL(*this->mock_proxy_event_binding_, Subscribe(max_sample_count));
```
bemerybmw: not relevant to test

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -855,6 +873,26 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture,
...
         expected_enabled_trace_points, kServiceIdentifierStringView, kServiceElementName, kInstanceSpecifierStringView);
 
+    // Expect that GetSubscriptionState returns kNotSubscribed then kSubscribed
+    EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
+        .WillOnce(Return(SubscriptionState::kSubscribed));
+
+    // Expect that a trace call relating to Subscribe is called
+    tracing::ITracingRuntime::TracePointType subscribe_trace_point_type{
+        ProxyEventTracingFixture<TypeParam>::TracePointType::SUBSCRIBE};
```
bemerybmw: disable subscribe trace point

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -855,6 +873,26 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture,
...
+                Trace(BindingType::kLoLa,
+                      expected_service_element_instance_identifier_view,
+                      subscribe_trace_point_type,
+                      std::optional<tracing::ITracingRuntime::TracePointDataId>{},
+                      _,
+                      _))
+        .WillOnce(Return(Result<void>{}));
+
+    // and that Subscribe will be called on the binding
+    EXPECT_CALL(*this->mock_proxy_event_binding_, Subscribe(max_sample_count));
```
bemerybmw: not relevant

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -919,6 +952,17 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture, UnsubscribeCallsAreNotTracedWhen
...
 
+    const std::size_t max_sample_count{1U};
+    InSequence sequence{};
+
+    // Expect that GetSubscriptionState is called once and indicates that the Service Element is currently not
+    // subscribed
+    EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed));
+
+    // and that Subscribe will be called on the binding
```
bemerybmw: not relevant

score/mw/com/impl/plumbing/proxy_binding_factory_test.cpp
```diff
@@ -76,7 +76,21 @@ class SkeletonBindingGuard
     std::unique_ptr<SkeletonBinding> skeleton_binding_;
 };
 
-using ProxyBindingFactoryCreateFixture = lola::ProxyMockedMemoryFixture;
+class ProxyBindingFactoryCreateFixture : public lola::ProxyMockedMemoryFixture
```
bemerybmw, line 79: lola::ProxyMockedMemoryFixture already contains a proxy_ which is deinitialized in the teardown. Why not use that?

bemerybmw: We should check all the fixtures which have the teardown added / a proxy or event binding to see if its already supported by the existing fixture.

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -720,13 +722,13 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture, UnsubscribeCallsAreTracedWhenEna
...
-    // Expect that GetSubscriptionState returns kNotSubscribed then kSubscribed
+    // and given that GetSubscriptionState returns kNotSubscribed then kSubscribed
     EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
         .WillOnce(Return(SubscriptionState::kNotSubscribed))
         .WillOnce(Return(SubscriptionState::kSubscribed));
 
-    // and that Subscribe will be called on the binding
-    EXPECT_CALL(*this->mock_proxy_event_binding_, Subscribe(max_sample_count));
+    // and given that Subscribe on the binding returns success
+    ON_CALL(*this->mock_proxy_event_binding_, Subscribe(max_sample_count)).WillByDefault(Return(score::Result<void>{}));
```
bemerybmw: remove

score/mw/com/impl/tracing/test/proxy_event_tracing_test.cpp
```diff
@@ -953,21 +905,16 @@ TYPED_TEST(ProxyEventTracingUnsubscribeFixture, UnsubscribeCallsAreNotTracedWhen
...
-
-    // and that GetSubscriptionState is called once and indicates that the Service Element is currently subscribed
+    // and given that GetSubscriptionState returns kNotSubscribed then kSubscribed
     EXPECT_CALL(*this->mock_proxy_event_binding_, GetSubscriptionState())
+        .WillOnce(Return(SubscriptionState::kNotSubscribed))
         .WillOnce(Return(SubscriptionState::kSubscribed));
 
-    // and that Unsubscribe will be called on the binding
+    // and given that Subscribe on the binding returns success
+    ON_CALL(*this->mock_proxy_event_binding_, Subscribe(max_sample_count)).WillByDefault(Return(score::Result<void>{}));
```
bemerybmw: remove

score/mw/com/impl/proxy_event_test.cpp
```diff
@@ -283,11 +286,12 @@ TYPED_TEST(ProxyEventFixture, UnsubscribeWhileNotHoldingSamplePtrs)
     using Base = ProxyEventFixture<TypeParam>;
 
     // Given a subscribed proxy event that received a sample but no longer holds it
+    ON_CALL(Base::mock_proxy_event_, Subscribe(1U)).WillByDefault(Return(score::Result<void>{}));
```
bemerybmw: fixture

score/mw/com/impl/proxy_event_test.cpp
```diff
@@ -249,29 +249,32 @@ TYPED_TEST(ProxyEventDeathTest, DieOnUnsubscribingWhileHoldingSamplePtrs)
...
+    Base::RecordProperty("Description",
+                         "Checks whether if a user unsubscribes an event, which still holds sample ptr, that we die.");
     Base::RecordProperty("TestType", "Requirements-based test");
     Base::RecordProperty("Priority", "1");
     Base::RecordProperty("DerivationTechnique", "Analysis of requirements");
 
+    const std::size_t max_num_samples{1};
+
     // Given a subscribed proxy event that has delivered a sample which is still held
+    ON_CALL(Base::mock_proxy_event_, Subscribe(max_num_samples)).WillByDefault(Return(score::Result<void>{}));
```
bemerybmw: fixture

bemerybmw (changes requested):

score/mw/com/impl/proxy_base.h
KrishaDeshkool: done

score/mw/com/impl/proxy_binding.h
KrishaDeshkool: reworded

score/mw/com/impl/proxy_binding.h
KrishaDeshkool: after the discussion it was noted to not enforce this here but we would add a integration test that proves that moved from object is still valid and can call all public methods.

KrishaDeshkool: Create a ticket

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
KrishaDeshkool: okay, moved it to WhichIsOffered() and adapted the test.

score/mw/com/impl/proxy_binding.h
KrishaDeshkool: Created https://github.com/eclipse-score/communication/issues/483

bemerybmw (dismissed):

KrishaDeshkool:
fixed merge conflicts

bemerybmw (approved):

PR #388, opened 2026-05-06
Rust::com FFI mock implementation for unit test

bharatGoswami8:
- Added a native FFI bridge and a mock FFI bridge for deterministic local testing.
- Runtime, consumer, and producer APIs are now pluggable via a bridge abstraction to support interchangeable backends.
- Added an explicit error for requests that exceed the maximum allowed sample count.
- Added mock-backed tests

### DIFF
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
@@ -55,3 +55,14 @@ rust_library(
         ":bridge_ffi_rs",
     ],
 )
+
+rust_library(
+    name = "bridge_ffi_lola",
+    srcs = ["bridge_ffi_lola.rs"],
+    visibility = [
+        "//score/mw/com:__subpackages__",
+    ],
+    deps = [
+        ":bridge_ffi_rs",
+    ],
+)
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
@@ -35,6 +35,7 @@ cc_library(
 rust_library(
     name = "bridge_ffi_rs",
     srcs = ["bridge_ffi.rs"],
+    edition = "2024",
     visibility = [
         "//score/mw/com:__subpackages__",
     ],
@@ -48,6 +49,7 @@ rust_library(
 rust_library(
     name = "bridge_ffi_mock",
     srcs = ["bridge_ffi_mock.rs"],
+    edition = "2024",
     visibility = [
         "//score/mw/com:__subpackages__",
     ],
@@ -60,6 +62,7 @@ rust_library(
 rust_library(
     name = "bridge_ffi_lola",
     srcs = ["bridge_ffi_lola.rs"],
+    edition = "2024",
     visibility = [
         "//score/mw/com:__subpackages__",
     ],
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
@@ -53,6 +53,7 @@ rust_library(
     ],
     deps = [
         ":bridge_ffi_rs",
+        "//score/mw/com/impl/plumbing/rust:sample_allocatee_ptr_rs",
     ],
 )
 
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
@@ -44,3 +44,14 @@ rust_library(
         "//score/mw/com/impl/rust:mw_com",
     ],
 )
+
+rust_library(
+    name = "bridge_ffi_mock",
+    srcs = ["bridge_ffi_mock.rs"],
+    visibility = [
+        "//score/mw/com:__subpackages__",
+    ],
+    deps = [
+        ":bridge_ffi_rs",
+    ],
+)
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
@@ -56,6 +56,7 @@ rust_library(
     deps = [
         ":bridge_ffi_rs",
         "//score/mw/com/impl/plumbing/rust:sample_allocatee_ptr_rs",
+        "@score_communication_crate_index//:mockall",
     ],
 )
 
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
@@ -67,6 +67,7 @@
 // - Enables efficient zero-copy string passing for interface_id, event_id, and type_name parameters
 
 use core::fmt::{Debug, Formatter};
+use core::marker::Unpin;
 use std::ffi::c_char;
 
 /// Opaque C++ void* pointer wrapper
@@ -82,7 +83,10 @@ pub use mw_com::proxy::ProxyEventBase;
 pub use mw_com::proxy::ProxyWrapperClass;
 pub use mw_com::InstanceSpecifier;
 
-pub trait FFIBridge {
+/// FFIBridge trait defines the interface for FFI interactions between Rust COM-API and C++ Lola runtime.
+/// This trait abstracts the FFI calls and allows for different implementations
+/// e.g., Lola bridge, mock bridge for testing.
+pub trait FFIBridge: Send + Sync + Clone + Debug + 'static + Unpin {
     // Define any methods that the bridge should have here
     unsafe fn get_allocatee_ptr(
         event_ptr: *mut SkeletonEventBase,
@@ -568,6 +572,7 @@ extern "C" {
     fn mw_com_proxy_event_unsubscribe(event_ptr: *mut ProxyEventBase);
 }
 
+#[derive(Debug, Clone)]
 pub struct LolaFFIBridge;
 
 impl FFIBridge for LolaFFIBridge {
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
@@ -82,6 +82,94 @@ pub use mw_com::proxy::ProxyEventBase;
 pub use mw_com::proxy::ProxyWrapperClass;
 pub use mw_com::InstanceSpecifier;
 
+pub trait FFIBridge {
+    // Define any methods that the bridge should have here
+    unsafe fn get_allocatee_ptr(
+        event_ptr: *mut SkeletonEventBase,
+        allocatee_ptr: *mut std::ffi::c_void,
+        event_type: &str,
+    ) -> bool;
+
+    unsafe fn delete_allocatee_ptr(allocatee_ptr: *mut std::ffi::c_void, type_name: &str);
+
+    unsafe fn get_allocatee_data_ptr(
+        allocatee_ptr: *const std::ffi::c_void,
+        type_name: &str,
+    ) -> *mut std::ffi::c_void;
+
+    unsafe fn skeleton_event_send_sample_allocatee(
+        event_ptr: *mut SkeletonEventBase,
+        event_type: &str,
+        allocatee_ptr: *const std::ffi::c_void,
+    ) -> bool;
+
+    unsafe fn sample_ptr_get(
+        sample_ptr: *const std::ffi::c_void,
+        type_name: &str,
+    ) -> *const std::ffi::c_void;
+
+    unsafe fn sample_ptr_delete(sample_ptr: *mut std::ffi::c_void, type_name: &str);
+
+    unsafe fn skeleton_offer_service(skeleton_ptr: *mut SkeletonBase) -> bool;
+
+    unsafe fn skeleton_stop_offer_service(skeleton_ptr: *mut SkeletonBase);
+
+    unsafe fn create_proxy(interface_id: &str, handle_ptr: &HandleType) -> *mut ProxyBase;
+
+    unsafe fn create_skeleton(
+        interface_id: &str,
+        instance_spec: *const NativeInstanceSpecifier,
+    ) -> *mut SkeletonBase;
+
+    unsafe fn destroy_proxy(proxy_ptr: *mut ProxyBase);
+
+    unsafe fn destroy_skeleton(skeleton_ptr: *mut SkeletonBase);
+
+    unsafe fn get_event_from_proxy(
+        proxy_ptr: *mut ProxyBase,
+        interface_id: &str,
+        event_id: &str,
+    ) -> *mut ProxyEventBase;
+
+    unsafe fn get_event_from_skeleton(
+        skeleton_ptr: *mut SkeletonBase,
+        interface_id: &str,
+        event_id: &str,
+    ) -> *mut SkeletonEventBase;
+
+    unsafe fn get_samples_from_event(
+        event_ptr: *mut ProxyEventBase,
+        event_type: &str,
+        callback: &FatPtr,
+        max_samples: u32,
+    ) -> u32;
+
+    unsafe fn skeleton_send_event(
+        event_ptr: *mut SkeletonEventBase,
+        event_type: &str,
+        data_ptr: *const std::ffi::c_void,
+    ) -> bool;
+
+    unsafe fn subscribe_to_event(event_ptr: *mut ProxyEventBase, max_sample_count: u32) -> bool;
+
+    unsafe fn unsubscribe_to_event(event_ptr: *mut ProxyEventBase);
+
+    unsafe fn set_event_receive_handler(
+        proxy_event_ptr: *mut ProxyEventBase,
+        handler: &FatPtr,
+        event_type: &str,
+    ) -> bool;
+
+    unsafe fn clear_event_receive_handler(proxy_event_ptr: *mut ProxyEventBase, event_type: &str);
+
+    unsafe fn start_find_service(
+        callback: &FatPtr,
+        instance_spec: InstanceSpecifier,
+    ) -> *mut FindServiceHandle;
+
+    unsafe fn stop_find_service(handle: *mut FindServiceHandle);
+}
+
 /// Opaque proxy base struct
 #[repr(C)]
 pub struct ProxyBase {
@@ -174,7 +262,7 @@ impl From<&'_ str> for StringView {
 /// - `ptr` must point to a valid FatPtr
 /// - `sample_ptr` must point to valid placement-new storage containing SamplePtr<T>
 #[no_mangle]
-pub unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_sample(
+unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_sample(
     ptr: *const FatPtr,
     sample_ptr: *mut std::ffi::c_void,
 ) {
@@ -205,7 +293,7 @@ pub unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_sample(
 /// - `find_service_handle` must point to valid FindServiceHandle data that can be safely wrapped
 ///   in a NativeFindServiceHandle.
 #[no_mangle]
-pub unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_find_service(
+unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_find_service(
     ptr: *const FatPtr,
     service_handles: *mut NativeHandleContainer,
     find_service_handle: *mut FindServiceHandle,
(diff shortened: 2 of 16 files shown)


tejaswinijoshi89 (approved):

tejaswinijoshi89 (approved):

rpreddyhv (approved):
Looks good to me.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
```diff
@@ -82,6 +83,184 @@ pub use mw_com::proxy::ProxyEventBase;
...
+    /// # Safety
+    /// `proxy_event_ptr` must be a valid pointer to a `ProxyEventBase` obtained from
+    /// `get_event_from_proxy`. `event_type` must match the type used when the handler was set.
+    unsafe fn clear_event_receive_handler(proxy_event_ptr: *mut ProxyEventBase, event_type: &str);
+
+    /// # Safety
+    /// `callback` must be a valid `FatPtr` referencing a callable compatible with the
+    /// find-service callback signature. `instance_spec` must be a valid `InstanceSpecifier`.
+    /// The returned handle must eventually be passed to `stop_find_service`.
+    unsafe fn start_find_service(
```
darkwisebear, line 258: I think this unsafety is misplaced. Afaiu, the main point is that `FatPtr` is essentially a type-erased Rust fat ptr. However, we could turn this method into something safe if we wrap the `FatPtr` into a `FindServiceCallable` and embed the `FatPtr` inside this type. The invariant of this type would then be the guarantee that it points to the appropriate callback. If the Rust compiler cannot verify that, then the unsafety is pushed to the point where this new type is instantiated. And this is imo the better place, since it is directly at the point where the original type gets "encoded into the Rust type system".

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
```diff
@@ -82,6 +83,184 @@ pub use mw_com::proxy::ProxyEventBase;
...
+    /// # Safety
+    /// `handle` must be a valid pointer returned from `start_find_service` that has not
+    /// been stopped yet. The handle must not be used after this call.
+    unsafe fn stop_find_service(handle: *mut FindServiceHandle);
+
+    /// # Safety
+    /// `container` must be a valid reference to a `HandleContainer` that was produced by the
+    /// bridge (e.g. from `start_find_service` callback). Calling this on a sentinel mock pointer
+    /// is only safe when the mock implementation ignores the inner pointer.
+    unsafe fn handle_container_size(container: &HandleContainer) -> usize;
```
darkwisebear: Same. If it is possible to safely construct a HandleContainer, the method's safety should not rely on the content being correct. If there is a precondition for the HandleContainer to be correct, this shall be pushed to construction time.

darkwisebear: I just asked myself why these methods are there in the first place... Why not let `HandleContainer` do that job?

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
```diff
@@ -82,6 +83,184 @@ pub use mw_com::proxy::ProxyEventBase;
...
+
+    /// # Safety
+    /// Same as `handle_container_size`. `index` need not be in-bounds; the method must return
+    /// `None` when `index >= handle_container_size(container)`.
+    unsafe fn handle_container_get_at(
```
darkwisebear: `unsafe` also unclear.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
darkwisebear, line 1: Please enable Rust 2024 edition for all targets.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+// This file provides a mock implementation of the FFIBridge trait for testing purposes.
+// It is kind of stub code which we will change according to the needs of our tests.
+// As of now, it just provide basic mock implementation and returning values based on
+// input parameters. In future, we will enhance this mock implementation to verify
+// all the test cases and scenarios.
+#![doc(hidden)]
```
darkwisebear, line 15: Why hide the mock backend?

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,334 @@
...
+#![allow(clippy::missing_safety_doc)]
```
darkwisebear: Yes, it's test code, but I don't see how this justifies skipping the safety comments. I'd rather say we can afford things that allocate more and are less performant in exchange to using less `unsafe` code instead.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,334 @@
...
+
+use bridge_ffi_rs::{
+    FatPtr, FindServiceHandle, HandleContainer, HandleType, InstanceSpecifier,
+    NativeHandleContainer, NativeInstanceSpecifier, ProxyBase, ProxyEventBase, SkeletonBase,
+    SkeletonEventBase,
+};
+use std::cell::RefCell;
+
+// Type alias for the heap-pointer + drop-glue pair stored in backing thread-locals.
+type BackingEntry = (*mut std::ffi::c_void, fn(*mut std::ffi::c_void));
```
darkwisebear: I prefer a dedicated struct here. This way, you can name the parts and it's clear what part does what.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,334 @@
...
+
+// The thread-local storage is used to hold the backing data for the mock FFI bridge.
+thread_local! {
```
darkwisebear: Why `thread_local!`? Because each test is run in a separate thread..? I think it would make more sense to generalize all the instance pointers in the trait and use them to store mock data. Or you just add `self` parameters and use the usual mocking methods instead of relying on static data. Imo that's almost always a smell, also in test code.

score/mw/com/test/basic_rust_api/consumer_async_apis/consumer_app.rs
```diff
@@ -55,7 +55,7 @@ async fn async_main() {
     );
 
     // Initialise the Lola runtime.
-    let mut runtime_builder = LolaRuntimeBuilderImpl::new();
+    let mut runtime_builder: LolaRuntimeBuilderImpl = LolaRuntimeBuilderImpl::new();
```
darkwisebear, line 58: Isn't that redundant? Why is this necessary?

score/mw/com/impl/rust/com-api/com-api-runtime-lola/runtime.rs
```diff
@@ -23,15 +23,20 @@ use com_api_concept::{
...
 
-impl Runtime for LolaRuntimeImpl {
-    type ServiceDiscovery<I: Interface + Send> = LolaConsumerDiscovery<I>;
-    type Subscriber<T: CommData + Debug> = SubscribableImpl<T>;
-    type ProducerBuilder<I: Interface> = LolaProducerBuilder<I>;
-    type Publisher<T: CommData + Debug> = Publisher<T>;
-    type ProviderInfo = LolaProviderInfo;
-    type ConsumerInfo = LolaConsumerInfo;
+pub struct LolaRuntimeImpl<B: FFIBridge = LolaFFIBridge> {
+    _marker: PhantomData<B>,
```
darkwisebear: Why not just `B`?

darkwisebear (changes requested):
I think the idea is fine. However, the approach with the thread locals bothers me: Why not just instantiate the bridge type and hold test data inside the bridge type if necessary? The extra `self` parameter needed for the methods are ok imo.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/BUILD
bharatGoswami8: Added ffi `BUILD` as well as another COM-API lib related target also.

score/mw/com/test/basic_rust_api/consumer_async_apis/consumer_app.rs
bharatGoswami8: Compiler is not able to resolve the FFIBridge even though there is default and complaining about it 
```
note: cannot satisfy `_: bridge_ffi_rs::FFIBridge`
help: the trait `bridge_ffi_rs::FFIBridge` is implemented for `bridge_ffi_lola::LolaFFIBridge`
```
explicit type annotation gives the compiler enough information to resolve the generic parameter.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/runtime.rs
bharatGoswami8: we do not need a field with type with `B` , if we keep `B` then we need to construct instance and we don't need `FFIBridge` instance in `LolaRuntimeImpl` or runtime trait impl.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: changed to struct type

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
bharatGoswami8: Removed as not required.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
bharatGoswami8: `HandleContainer` is still part of the proxy_bridge(old API implementation) and we don't have generic or mock support, anyways we are going to clean-up the old API implementation and required part will move in bridge folder.
(https://github.com/eclipse-score/communication/issues/432)
For bridge as we have enabled the mock support so rather than calling direct from proxy_bridge, we are using from this file.

`Unsafe` removed from method signature.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
bharatGoswami8: I have created an issue/ task for this, will investigate and optimize the unsafe call in rust FFI.
https://github.com/eclipse-score/communication/issues/431

score/mw/com/impl/rust/com-api/com-api-runtime-lola/runtime.rs
darkwisebear: But only because you use static data to keep the mock data. If you had an instance here, you might keep the data there. Much cleaner imo.

bharatGoswami8: understood, optimized the `MockFFI` and Runtime implementation to use the instance of bridge and `&self` in method signature

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: yes it was because of test running in multithread environment but as you suggested i have optimized the `MockFFI` and Runtime implementation to use the instance of bridge and `&self` in ffi method signature.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Removed the clippy lint and added safety comment on required places.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: It was included because, as it is only needed for internal unit testing, so I didn’t see the need to expose it in the crate-level documentation.

bharatGoswami8:
> I think the idea is fine. However, the approach with the thread locals bothers me: Why not just instantiate the bridge type and hold test data inside the bridge type if necessary? The extra `self` parameter needed for the methods are ok imo.

@darkwisebear, I have updated the mock backend, Instead of using `thread_local!`, I have adopted the approach you suggested - instantiating the bridge type and storing the test data within it.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -53,31 +53,36 @@ use bridge_ffi_rs::*;
...
 
 #[derive(Clone, Debug)]
-pub struct LolaConsumerInfo {
+pub struct LolaConsumerInfo<B: FFIBridge> {
     handle_container: Arc<mw_com::proxy::HandleContainer>,
     handle_index: usize,
     interface_id: &'static str,
+    // Bridge is Arc-wrapped at creation for optimized cloning throughout the consumer lifecycle.
+    // For LolaFFIBridge (ZST), Arc overhead is minimal rather than cloning directly.
+    bridge: Arc<B>,
```
darkwisebear: The production case is a ZST, so why optimize for the test case by adding an `Arc` here? If cloning the type is an issue, the type itself should introduce this optimization, or the lib provides an `impl FFIBridge for Arc<B> where B: FFIBridge` so that the user may choose to use an `Arc` instead of hardwiring it.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,389 @@
...
+    ptr: *mut std::ffi::c_void,
+    // Drop function that knows how to properly free the data
+    drop_fn: fn(*mut std::ffi::c_void),
+}
+
+// SAFETY: BackingEntry is safe to Send and Sync because,
+// each test will typically use its own instance of MockFFIBridge and thus its own BackingEntry,
+// so there is no shared mutable state across threads.
+unsafe impl Send for BackingEntry {}
+unsafe impl Sync for BackingEntry {}
```
darkwisebear: This reasoning is flawed. You need to argue why marking the type as `Send` and / or `Sync` is safe although the compiler can't verify that. Arguing that "typically" the user will prevent UB because they happen to use the type in a safe way is just wrong and is prone to failure. Please reword if you believe that `Send` and `Sync` are safe under _any_ circumstances!

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,389 @@
...
+    //the mock to zero-fill the caller's slot correctly.
+    alloc_size: RefCell<usize>,
+    //SAMPLE_BACKING holds a pointer to the heap-allocated sample data and a drop function to free it.
+    sample_backing: RefCell<Option<BackingEntry>>,
+}
+
+// SAFETY: MockFFIBridge is safe to Send and Sync because each test will typically use its own instance of MockFFIBridge,
+// so there is no shared mutable state across threads.
+unsafe impl Send for MockFFIBridge {}
+unsafe impl Sync for MockFFIBridge {}
```
darkwisebear: Nope. Saying "hey, we can do this because tests never use threads and so declaring MockFFIBridge `Sync` knowing that `RefCell` is explicitly `!Sync` is fiiiiiiiiine" is just incorrect and will cause UB. Your safety reasoning already says that:
> tests will _typically_ use its own instance...
_typically_ means tests that aren't typical will do otherwise, and they will then expose UB.

So while `Send` _may_ be ok (provided the type inside the cells are `Send`), `Sync` is certainly wrong; please replace by safe alternatives, like an `Arc<Mutex<MockBridgeState>>` or the like.

darkwisebear (changes requested):

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
darkwisebear: Ok, this is reasonable. Maybe you want to clarify the comment above the attribute then?

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: We have `samplePtr` and `AllocateePtr`, which operate in a critical section. Cloning them during receive and allocation is not the most optimal approach. However, I got the point, since the production code uses ZSTs, there’s no clone overhead in practice. I’ll add a note/comment that if any fields are introduced in the `LoLaBridgeFFI` type in the future, this part should be revisited and optimized.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Yes, I initially started with `Arc<Mutex<MockBridgeState>>`, but later switched to `RefCell`, assuming it would be used in individual tests and that wrapping it in a Mutex inside an Arc might introduce unnecessary overhead. However, I agree that since this is a mock meant for testing, it’s better to prioritize safety over performance. I’ll revert back to using the Arc-based approach.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: I will update the safety comment for this.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Update the comment for attribute.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Updated the code changes.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Removed Sync as not needed because of Mock type now using thread safe field type.
Updated safety comment for Send.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Updated the code changes.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+#[derive(Debug, Clone)]
+pub struct MockFFIBridge {
+    //DATA_BACKING holds a pointer to the heap-allocated backing data and a drop function to free it.
+    //We are using this when user calls get_allocatee_data_ptr to return a pointer to this backing data.
+    data_backing: Arc<Mutex<Option<BackingEntry>>>,
+    //ALLOC_SIZE holds the size of the allocatee type for the next get_allocatee_ptr call, allowing
+    //the mock to zero-fill the caller's slot correctly.
+    alloc_size: Arc<Mutex<usize>>,
+    //SAMPLE_BACKING holds a pointer to the heap-allocated sample data and a drop function to free it.
+    sample_backing: Arc<Mutex<Option<BackingEntry>>>,
```
darkwisebear: Please merge these three entries by creating a struct that contains the three members and put it behind an `Arc<Mutex<MockFFIBridgeState>>`.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
```diff
@@ -82,6 +83,204 @@ pub use mw_com::proxy::ProxyEventBase;
...
+/// an because of that this bridge also need to be bound by these traits to be used in runtime implementation.
+pub trait FFIBridge: Send + Sync + Clone + Debug + 'static + Unpin + Default {
+    /// # Safety
+    /// `event_ptr` must be a valid, non-null pointer to a `SkeletonEventBase` obtained from
+    /// `get_event_from_skeleton`. `allocatee_ptr` must point to valid memory large enough to
+    /// hold the allocatee instance for `event_type`.
+    unsafe fn get_allocatee_ptr(
+        &self,
+        event_ptr: *mut SkeletonEventBase,
+        allocatee_ptr: *mut std::ffi::c_void,
```
darkwisebear, line 100: Sorry to ask this so late, but isn't the type of this pointer something known? Why does this have to be `c_void`?

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        _event_type: &str,
+    ) -> bool {
+        if event_ptr.is_null() || allocatee_ptr.is_null() {
+            return false;
+        }
+        let size = *self.alloc_size.lock().expect("Failed to lock alloc_size");
+        if size == 0 {
+            return false;
+        }
+        unsafe { std::ptr::write_bytes(allocatee_ptr as *mut u8, 0, size) };
```
darkwisebear: Why does this line exist?

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        &self,
+        event_ptr: *mut SkeletonEventBase,
+        allocatee_ptr: *mut std::ffi::c_void,
+        _event_type: &str,
+    ) -> bool {
+        if event_ptr.is_null() || allocatee_ptr.is_null() {
+            return false;
+        }
+        let size = *self.alloc_size.lock().expect("Failed to lock alloc_size");
+        if size == 0 {
```
darkwisebear: If 0 has a special meaning, you should use an `Option<NonZeroUsize>` instead of using a magic number.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        // Tests must not call any real C++ handle-container operations on the returned Arc.
+        let ptr = std::ptr::NonNull::<NativeHandleContainer>::dangling().as_ptr();
+        let hc = std::sync::Arc::new(HandleContainer::new(ptr));
+        // Prevent HandleContainer::drop (which would call real C++ delete) by pinning the refcount.
+        std::mem::forget(std::sync::Arc::clone(&hc));
+        hc
+    }
+
+    // `ProxyEventBase` is a ZST opaque type, a dangling non-null pointer is the
+    // canonical representation for "valid but empty" in the mock.
```
darkwisebear: In which universe is a dangling pointer a "valid but empty representation"..? I really think it's a wording issue but it's just so disturbing. Please consider the mock approach suggested above.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        *self.alloc_size.lock().expect("Failed to lock alloc_size") = 0;
+    }
+
+    // Returns a pointer to the heap-allocated backing data for the current allocatee, if any.
+    unsafe fn get_allocatee_data_ptr(
+        &self,
+        _allocatee_ptr: *const std::ffi::c_void,
+        _type_name: &str,
+    ) -> *mut std::ffi::c_void {
+        self.data_backing
```
darkwisebear: That's brittle. If someone calls `get_allocatee_data_ptr` and retains the pointer, then a call to `set_alloc_backing` will invalidate the pointer and we see a use-after-free.

tejaswinijoshi89 (approved):

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        _event_id: &str,
+    ) -> *mut ProxyEventBase {
+        if proxy_ptr.is_null() {
+            return std::ptr::null_mut();
+        }
+        // ProxyEventBase is a ZST opaque type; a non-null sentinel is sufficient.
+        // LIMITATION: The returned pointer is intentionally dangling and must NOT be
+        // dereferenced. It is only valid as a non-null sentinel for null-checks in the
+        // mock. Any test that dereferences this pointer will trigger undefined behavior.
+        std::ptr::NonNull::<ProxyEventBase>::dangling().as_ptr()
```
darkwisebear: That makes me feel uneasy... while I do get the intent, this will lead to something that _knowingly_ returns something bad. I anyway wonder why we cannot use the classic mock approach but we provide a type that intentionally does bad things in a global way. If we had a real mock type (as e.g. made by the `mockall` crate), we could do such things in a controlled way locally in the test that needs it.

I'd ask you to consider using a `mockall` mock or give reason why this isn't appropriate.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        _container: &'a HandleContainer,
+        _index: usize,
+    ) -> Option<&'a HandleType> {
+        None
+    }
+}
+
+impl MockFFIBridge {
+    // The handle container is backed by a heap-allocated zeroed stub so the pointer
+    // is stable and non-dangling.  An extra `Arc` clone is intentionally forgotten
```
darkwisebear: Ok, but why? isn't this something the test code may just as well do if it thinks that this is valid? (and under which circumstances is that necessary?). This strikes me again as unclean test code that makes many assumptions on how it's going to be used during testing. While this might work now, this is brittle when refactorings are being made.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,409 @@
...
+        // Replace any previous backing - Drop trait automatically cleans up the old entry
+        *self
+            .data_backing
+            .lock()
+            .expect("Failed to lock data_backing") = Some(BackingEntry {
+            ptr: data_ptr,
+            drop_fn,
+        });
+        *self.alloc_size.lock().expect("Failed to lock alloc_size") =
+            std::mem::size_of::<SampleAllocateePtr<T>>();
```
darkwisebear: I wonder why `alloc_size` is sized with the SampleAllocateePtr's size, whereas the actual data is of size `T`. Isn't that a mismatch..?

darkwisebear (commented):
We're not there yet. Please consider using a real mock for the bridge that implements all the trait methods and allows for setting call reactions on a per-test basis with defined values that are different from dangling pointers and other stuff that scares me.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: yes, updated with size field in `BackingEntry`

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Merged all the mock related types with an `MockFFIBridgeState` and using with one Arc instance under `MockFFIBridge`

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
bharatGoswami8: Type `T` known at rust side compile time but we were not using generic in FFI functions/methods in old static based function call, but with new trait and FFI instance-based approach, i will investigate with https://github.com/eclipse-score/communication/issues/431 with this ticket.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: I have updated allocation with setter function like `mockall` expect function.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Updated with `Option<NonZeroUsize>`

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Added unit type for `Proxy/Skeleton/Event` and constructing using setter methods, so each test can individually construct related types and use it as per test cases.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Added HashMap for all the types of Mock so it can be stored for different interface id and event id.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: I have added unit structs and setter functions for creating test-specific pointers. With this approach, we achieve similar per-test configurability as `mockall` but implemented manually, which allows us to extend it later as per requirements for coverage and unit testing scenarios. 

I have tried implementing with `mockall` as well, but it does not work properly for our use case because we have a pointer-heavy FFI interface. When setting expectations, mockall still requires creating raw pointers, and these don't create proper type-safe instances needed by the runtime. During test execution, we faced segmentation faults due to mockall's inability to properly manage FFI pointer lifetimes and ownership semantics.
Additionally, mockall cannot handle the generic lifetime parameter in `handle_container_get_at<'a>() `method, resulting in compilation errors which can be overcomed as of because we have only one function but in future we get more function with generic and lifetime then this will be a blocker for us,  so Mockall is not appropriate for FFI traits with generic lifetimes, complex bounds, and pointer-heavy unsafe APIs.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: Updated to the setter-based approach, which is providing necessary abstractions over FFI resource management. With this approach each test can run indivisibly and track resources. 

The mock encapsulates resource lifecycle management (allocation + Drop-based cleanup) and provides type safety through generic methods like `set_sample_backing<T>().` This ensures data type match FFI type at compile time. 

When the FFI layer changes, we update the mock implementation once rather than modifying every test. This centralized approach is more maintainable and refactoring-safe than having scattered pointer manipulation across test files.

bharatGoswami8:
> We're not there yet. Please consider using a real mock for the bridge that implements all the trait methods and allows for setting call reactions on a per-test basis with defined values that are different from dangling pointers and other stuff that scares me.

I have added unit structs and setter functions for creating test-specific pointers. With this approach, we achieve similar per-test configurability as mockall but implemented manually, which allows us to extend it later as per requirements for coverage and unit testing scenarios.
And each test can configure the value as per test scenario which can extend as per need.
No dangling pointer now in mock ffi implementation.
With generic methods like `set_sample_backing<T>()` we can construct the samplePtr and allocate the as per data types at compile time rather than run time in mockall.

bharatGoswami8:
> > We're not there yet. Please consider using a real mock for the bridge that implements all the trait methods and allows for setting call reactions on a per-test basis with defined values that are different from dangling pointers and other stuff that scares me.
> 
> I have added unit structs and setter functions for creating test-specific pointers. With this approach, we achieve similar per-test configurability as mockall but implemented manually, which allows us to extend it later as per requirements for coverage and unit testing scenarios. And each test can configure the value as per test scenario which can extend as per need. No dangling pointer now in mock ffi implementation. With generic methods like `set_sample_backing<T>()` we can construct the samplePtr and allocate the as per data types at compile time rather than run time in mockall.

We are able to resolve the issue with mockall, the major issue was for `clone`, so we have created another newtype which wrap mockffi type in `Arc<Mutex>` with this `clone` related all the issue resolved in runtime.
So now Mock FFI module is using Mockall crate with `mock!` macro.
Updated consumer and producer test cases also according to that.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,479 @@
...
+        &self,
+        event_ptr: *mut SkeletonEventBase,
+        allocatee_ptr: *mut std::ffi::c_void,
+        event_type: &str,
+    ) -> bool {
+        //Safety: This is just forwarding the call to the inner mock, which is expected to be configured correctly in tests using mockall's expectations.
+        unsafe {
+            self.0
+                .lock()
+                .expect("Failed to acquire lock")
```
darkwisebear: This could be moved to a private helper:

```rust
    fn locked(&self) -> std::sync::MutexGuard<MockFFIBridge> {
        self.0.lock().expect("Failed to acquire lock")
    }
```

So that you then write for each method:

```rust
    fn trait_method(&self, arg1: X, arg2: Y) -> Res {
        self.locked().trait_method(arg1, arg2)
    }
```

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -1142,120 +1212,167 @@ pub fn create_sample_callback<'a, T: CommData + Debug>(
...
-        let instance_specifier = com_api_concept::InstanceSpecifier::new("/test/vehicle")
-            .expect("Failed to create instance specifier");
+        let proxy = proxy_vec.clone();
+        mock.expect_destroy_proxy().returning(move |ptr| {
+            let mut proxy = proxy.lock().expect("not able to acquire lock on proxy_vec");
+            let pos = proxy
+                .iter_mut()
+                .position(|b| b.as_mut() as *mut ProxyBase == ptr)
+                .expect("destroy_proxy called with unknown pointer");
+            proxy.remove(pos);
```
darkwisebear:
```suggestion
            let extracted_element = proxy.extract_if(..,|b| b.as_mut() as *mut ProxyBase == ptr)).next();
            assert!(extracted_element.is_some());
```
is clearer, imo.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -1142,120 +1212,167 @@ pub fn create_sample_callback<'a, T: CommData + Debug>(
...
+        // we need Arc here because expect closures are take the ownership of the variables,
+        // and we need to share the state between multiple expectations.
+        // vec is required to store multiple proxies/events and simulate multiple calls to create_proxy/subscribe_to_event/get_event_from_proxy/unsubscribe_to_event/destroy_proxy.
+        // also cleanup the created proxies/events by removing them from the vec when destroy/unsubscribe is called, and assert the pointer passed in is valid by checking if it exists in the vec.
+        let proxy_vec = Arc::new(Mutex::new(Vec::new()));
+        let event_vec = Arc::new(Mutex::new(Vec::new()));
+
+        // Set up all expectations - all clones of SharedMockBridge will share these
+        let proxy = proxy_vec.clone();
+        mock.expect_create_proxy().returning(move |_, _| {
```
darkwisebear: You should probably also create a sequence for these calls, as it doesn't make sense to subscribe before creating the proxy or something like that,

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
```diff
@@ -462,22 +499,136 @@ impl<I: Interface> Builder<I::Producer<LolaRuntimeImpl>> for LolaProducerBuilder
...
+        }
+    }
+
     #[test]
-    #[ignore] // Test will be update with Ticket-242140
-    fn send_stuff() {
-        let _test_data = TestData(42);
+    fn test_provider_info_offer_service() {
+        let mut mock = MockFFIBridge::new();
+        let skeleton_vec = Arc::new(Mutex::new(Vec::new()));
```
darkwisebear: It might make sense to create a helper struct that provides an `allocate` and a `free` method and does the management of the boxes and the pointers internally instead of exposing it. Saying this because I see that pattern now over and over. Putting this into a common utility struct may help here.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: I think we can use `mockall::Sequence` for this.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8:
```
-> /work/.cache/ferrocene-src-ubuntu20-prof/library/alloc/src/vec/mod.rs:4064:11
help: provide the argument
    |
472 |         allocs.extract_if(|b: &mut Box<T>| b.as_mut() as *mut T == ptr, /* filter */).
```
Here : extract_if(range, filter) required two parameters, we have centralized all the allocation part, but we are using same approach , find the position and remove the ptr.

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
bharatGoswami8: yes agreed, updated same.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: updated expect call with `.in_sequence(&mut seq)` using `mockall::Sequence`.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/producer.rs
bharatGoswami8: Agreed, created `MockPointerAllocator` type which has `allocate` and `free` support method .

score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_mock.rs
```diff
@@ -0,0 +1,513 @@
...
+    /// - `true` if the pointer was found and freed
+    /// - `false` if the pointer was not found in tracked allocations
+    pub fn free(&self, ptr: *mut T) -> bool {
+        let mut allocs = self.locked();
+        if let Some(pos) = allocs.iter_mut().position(|b| b.as_mut() as *mut T == ptr) {
+            allocs.remove(pos);
+            true
+        } else {
+            false
+        }
```
darkwisebear:
```suggestion
        allocs.extract_if(.., |b| b.as_mut() as *mut T == ptr)
            .next()
            .is_some()
```

bharatGoswami8: Updated.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
darkwisebear: Seems to be a copy/paste error, the `..` are apparently missing.

darkwisebear (approved):

bemerybmw (approved):

PR #389, opened 2026-05-06
Add S-CORE quality maintenance proposal

Komal362000:


S-CORE_Quality_Maintenance_Proposal.md
```diff
@@ -0,0 +1,268 @@
...
+I suggest a **hybrid quality plan** for S-CORE.
+
+This means:
+- Run **fast checks on every PR** so developers get quick feedback.
+- Run **full quality checks at night** so we still catch deep issues.
+- Show results in a **shared quality dashboard** so everyone can track quality.
+
+**Proposed Model:**
+- **PR checks (fast):** Build, unit tests, formatting, and basic lint.
+- **Nightly checks (full):** CodeQL, full clang-tidy, sanitizers, coverage, and static analysis.
```
ahmed0mousa: In case we got some findings, When do you think we should fix it? Do we fix it immediately or When?

S-CORE_Quality_Maintenance_Proposal.md
```diff
@@ -0,0 +1,268 @@
...
+
+I suggest a **hybrid quality plan** for S-CORE.
+
+This means:
+- Run **fast checks on every PR** so developers get quick feedback.
+- Run **full quality checks at night** so we still catch deep issues.
+- Show results in a **shared quality dashboard** so everyone can track quality.
+
+**Proposed Model:**
+- **PR checks (fast):** Build, unit tests, formatting, and basic lint.
```
ahmed0mousa: What about having quality jobs (CodeQL, Clang-tidy, Coverage) on every PR but it got triggered (manually) only once after the review process is done so we don't need to run quality jobs after each commit on the same time we are sure that no PRs introduce new findings?

S-CORE_Quality_Maintenance_Proposal.md
```diff
@@ -0,0 +1,268 @@
...
+- Helps track improvement over time while preventing regressions
+- More robust than using only one approach
+
+**Cons**
+- Requires more setup effort
+- Needs agreement on which KPIs are monitored only and which are enforced in CI
+
+---
+
+## Pros
```
ahmed0mousa: under which option are then `pros` `cons` ?!

S-CORE_Quality_Maintenance_Proposal.md
```diff
@@ -0,0 +1,268 @@
...
+- Useful for reporting and management review
+
+**Cons**
+- Relies on manual team reaction
+- Does not prevent low-quality code from being merged
+- Loses effectiveness if the dashboard is not reviewed regularly
+
+### Option 2: CI Enforcement
+
+In this model, quality gates are enforced directly in the CI pipeline. If code does not meet the agreed thresholds, the build or merge process fails.
```
ahmed0mousa: Did you check How much time it cost to run (CodeQL, Clang-tidy, Coverage) ? What is the average time for each job?

PR #390, opened 2026-05-06
Add SubscriptionStateChangeHandler support

crimson11:
Although we had already an explicit requirement for the support of ProxyEvent/Field side
`SetSubscriptionStateChangeHandler()` and the related Unset-API implementation was still
missing.

This PR provides the implementation and updates the requirements and design accordingly.

### DIFF
--- a/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc
+++ b/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc
@@ -2365,7 +2365,7 @@ section "com" {
 
         /* broken_link_c/issue/21466236 */
         ScoreReq.CompReq BehaviourOfUnsetSubscriptionStateChangeHandler {
-            description = '''{{UnsetSubscriptionStateChangeHandler}} {{UnsetSubscriptionStateChangeHandler}}{{SubscriptionStateChangeHandler}}will unregister the {{SubscriptionStateChangeHandler}} (See SubscriptionStateChangeHandler) that was registered by the class instance with a call to {{SetSubscriptionStateChangeHandler}} After calling  the  will no longer be triggered when the {{SubscriptionState}} (See progress-cursor SubscriptionState changes.
+            description = '''{{UnsetSubscriptionStateChangeHandler}} will unregister the {{SubscriptionStateChangeHandler}} (See SubscriptionStateChangeHandler) that was registered by the class instance with a call to {{SetSubscriptionStateChangeHandler}} After calling  the  will no longer be triggered when the {{SubscriptionState}} (See progress-cursor SubscriptionState changes.
 
             {{UnsetSubscriptionStateChangeHandler}} shall fulfill the following requirement: Explicit Lifetime ending by API calls.'''
             safety = ScoreReq.Asil.B
--- a/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc
+++ b/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc
@@ -648,7 +648,7 @@ section "com" {
 
                 /* broken_link_c/issue/14137269 */
                 ScoreReq.CompReq ProxyEventCopySemantics {
-                    description = '''The ﻿{{ProxyEvent}} class shall neither be copy-constructable nor copy-assignable.'''
+                    description = '''The {{ProxyEvent}} class shall neither be copy-constructable nor copy-assignable.'''
                     safety = ScoreReq.Asil.B
                     derived_from = [Communication.EventType@1, Communication.DataLoss@1]
                     version = 1
@@ -1590,7 +1590,7 @@ section "com" {
 
                 The signature of an SubscriptionStateChangeHandler is:
 
-                {{{void(SubscriptionState)}}}
+                {{{bool(SubscriptionState)}}}
 
                 The type chosen for {{SubscriptionStateChangeHandler}} by the {{mw::com}} provider shall allow construction from any callable, with the given signature.'''
                 safety = ScoreReq.Asil.B
@@ -2307,8 +2307,7 @@ section "com" {
         /* broken_link_c/issue/21466369 */
         ScoreReq.CompReq BehaviourOfSetSubscriptionStateChangeHandler {
             description = '''The {{SubscriptionStateChangeHandler}} (See SubscriptionStateChangeHandler)
-            {{SetSubscriptionStateChangeHandler}}
-            that is passed to shall be called every time the {{SubscriptionState}} (See SubscriptionState) changes according to the requirements:
+            that is passed to {{SetSubscriptionStateChangeHandler}} shall be called every time the {{SubscriptionState}} (See SubscriptionState) changes according to the requirements:
 
             *  Calls to SubscriptionStateChangeHandler with kSubscriptionPending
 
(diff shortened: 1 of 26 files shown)


score/mw/com/impl/bindings/lola/proxy_event_common.cpp
```diff
@@ -64,20 +64,7 @@ void ProxyEventCommon::Unsubscribe()
...
-    {
-        return SubscriptionState::kSubscriptionPending;
-    }
-    else
-    {
-        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(current_state == SubscriptionStateMachineState::SUBSCRIBED_STATE,
-                                                    "Invalid subscription state machine state.");
-        return SubscriptionState::kSubscribed;
-    }
+    return SubscriptionStateMachineState2SubscriptionState(current_state);
```
KrishaDeshkool: i like this name but i think we do not follow this convention of using a digit in the Method names,
best to use `SubscriptionStateMachineStateToSubscriptionState`

score/mw/com/impl/bindings/lola/subscription_state_machine_events_test.cpp
```diff
@@ -391,5 +393,100 @@ TEST_F(StateMachineSubscribedStateFixture, CallingReOfferEventDoesNothing)
     EXPECT_EQ(state_machine_.GetCurrentState(), SubscriptionStateMachineState::SUBSCRIBED_STATE);
 }
 
+using StateMachineStateChangeFixture = StateMachineEventsFixture;
+TEST_F(StateMachineStateChangeFixture, CallingSetSubscriptionStateChangeHandlerSucceeds)
+{
+    auto subscription_state_change_handler = [](SubscriptionState new_subscription_state) {
+        return true;
+    };
+    state_machine_.SetSubscriptionStateChangeHandler(subscription_state_change_handler);
```
KrishaDeshkool, line 405: i dont see any explicit expectations or conditions in this test,
does having this really add value?

score/mw/com/impl/proxy_event_base.h
```diff
@@ -119,6 +120,22 @@ class ProxyEventBase
...
+     * \param handler
+     */
+    Result<void> SetSubscriptionStateChangeHandler(SubscriptionStateChangeHandler handler) noexcept;
+
+    /**
+     * \api
+     * \brief Unsets/Unregisters a SubscriptionStateChangeHandler for this event. After this method returns, it is
+     *        guaranteed, that the previously registered handler is neither active nor will be called anymore.
+     */
+    Result<void> UnSetSubscriptionStateChangeHandler() noexcept;
```
KrishaDeshkool: minor nitpick ;)
we have `UnSet` upper case `S` in the public API and other places have lowercase `S`,

score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc
```diff
@@ -2356,6 +2355,14 @@ section "com" {
...
+            description = '''In case the user provided handler returns true, the {{SubscriptionStateChangeHandler}} will be kept registered and will be called on the next {{SubscriptionState}} change again.
+            In case the user provided handler returns false, the {{SubscriptionStateChangeHandler}} will be unregistered and won't be called on the next {{SubscriptionState}} change again.'''
+            safety = ScoreReq.Asil.B
+            derived_from = [Communication.SupportForTimeBasedArchitecture@2]
+            version = 1
+        }
+
         /* broken_link_c/issue/21466236 */
         ScoreReq.CompReq BehaviourOfUnsetSubscriptionStateChangeHandler {
             description = '''{{UnsetSubscriptionStateChangeHandler}} {{UnsetSubscriptionStateChangeHandler}}{{SubscriptionStateChangeHandler}}will unregister the {{SubscriptionStateChangeHandler}} (See SubscriptionStateChangeHandler) that was registered by the class instance with a call to {{SetSubscriptionStateChangeHandler}} After calling  the  will no longer be triggered when the {{SubscriptionState}} (See progress-cursor SubscriptionState changes.
```
KrishaDeshkool: not changed in this PR,
but is here 
`{{UnsetSubscriptionStateChangeHandler}} {{UnsetSubscriptionStateChangeHandler}}{{SubscriptionStateChangeHandler}}`
 is expected to be like this ?
`{{UnsetSubscriptionStateChangeHandler}} {{SubscriptionStateChangeHandler}}`
?

score/mw/com/impl/proxy_field_base.h
```diff
@@ -86,6 +86,28 @@ class ProxyFieldBase
         proxy_event_base_dispatch_->Unsubscribe();
     }
 
+    /**
+     * \api
+     * \brief Sets/Registers a SubscriptionStateChangeHandler for this event. This handler will be called whenever the
```
KrishaDeshkool, line 91: Just a nitpick,
should this be called as `field` ?
or since it dispatches to the event_base it is still accurate to call this a event?
(from other methods we use field instead of event even if we dispatch to the event_base.

score/mw/com/impl/subscription_state_change_handler.h
```diff
@@ -0,0 +1,37 @@
...
+{
+
+/// \brief Callback for event/field subscription state change notifications on proxy side.
+/// \details This callback may be called under lock of the internal state-machine. This in general means, that the user
+/// shall not do any long-running activities within this handler as it will prolonge this lock.
+/// The user also shall not call any methods on the same event/field instance for which the handler was set, as they
+/// might also require the same lock, which is not guaranteed to be recursive, thus leading to a deadlock. If the user
+/// intends to do such activities from this callback, he shall dispatch it to a separate thread.
+/// However, unsetting the handler within its call is a reasonable/supported use-case. Instead of calling
+/// UnsetSubscriptionStateChangeHandler, the user shall return false from the handler. See return-value description.
```
KrishaDeshkool, line 30: eventhough we warn users about the consequence of deadlock,
is it better or worth doing something like we do with "is_in_receive_handler_context" if we are in the same ctx then we dont acquire the lock to prevent deadlock?

score/mw/com/impl/bindings/lola/subscription_state_machine_events_test.cpp
crimson11: Added comment, which clarifies the need/idea.

score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc
crimson11: This is broken anyhow :) because this is copied over from AUTOSAR requirements - I do not even get all these curly braces! But I cleaned up the mess a bit ;)

score/mw/com/impl/subscription_state_change_handler.h
crimson11: Checking/verifying it comes at additional costs! We need to manage a thread-local variable to monitor the mis-use.
Here we are basically adding an AoU - like many other AoUs we have and where we expect the user to adhere to. I.e.: We disallow, that a user calls APIs of a mw::com proxy/skeleton/event/field instance concurrently. We do NOT supervise this ... although we could do this! So - it is the same here! We don't supervise, that the user follows our AoU.

SetReceiveHandler is a "different animal". Because there we explicitly allow, that the user calls `UnsetReceiveHandler` from within the handler-context! Thus, we have to detect it, because we then have different code paths for `UnsetReceiveHandler` depending on the context. And the only reason is: In case of `ReceiveHandler` I wasn't smart enough to come up with the same idea as in `SubscriptionStateHandler` -> if you want "unset" it, just tell it via return-value. The underlying code is so much more effective! I'm tempted to change the signature/idea for `ReceiveHandler` maybe the same at some time ...

KrishaDeshkool (approved):

castler (approved):

PR #394, opened 2026-05-08
Fix typed proxy access to generic skeleton event storage

rudresh-systream:
Summary:
- Fixed typed ProxyEvent sample access so it no longer depends on interpreting shared memory as EventDataStorage<T>.
- Added raw-slot based access using sample size/alignment, making typed proxies compatible with GenericSkeleton-created event storage.
- Kept existing typed skeleton behavior compatible.
- Updated ProxyEvent and event storage related tests/resources.

Implemented a compatibility fix for the GenericSkeleton ↔ typed ProxyEvent shared-memory interaction described in #311.

### Root cause

The issue originates from the fact that GenericSkeleton and typed skeletons create/interprete `EventDataStorage` differently.

For typed skeletons, the storage is created through:

```cpp
EventDataStorage<SampleType>
```

which internally creates a typed `DynamicArray<SampleType>`.

For GenericSkeleton/GenericSkeletonEvent, the storage is instead created using:

```cpp
EventDataStorage<std::max_align_t>
```

with the storage size being calculated in units of `std::max_align_t`.

As a result, the underlying raw memory region may be large enough, but the metadata/layout interpretation of the `DynamicArray<T>` becomes incompatible between the producer and consumer sides.

The typed proxy path was still assuming a typed `EventDataStorage<T>` representation and therefore interpreted slot count/layout using the wrong type information. This creates an incompatibility whenever:

* a GenericSkeleton publishes samples
* and a typed proxy consumes them.

This issue is especially visible because the DynamicArray element count depends on the template type `T`, which differs between the generic and typed paths.

---

### Investigated solution approaches

Several approaches were considered:

#### 1. Extending `DynamicArray<T>`

One option was to add a constructor allowing externally managed/preallocated storage while preserving a typed `DynamicArray<T>` interface.

This was not chosen because:

* it would introduce LoLa/event-storage specific semantics into a generic container abstraction,
* it increases complexity in `DynamicArray`,
* and it still keeps the shared-memory interpretation tightly coupled to template typing.

#### 2. Replacing all event storage with `DynamicArray<std::byte>`

Another option was to fully migrate all event storage handling to raw byte storage.

While architecturally clean, this would require significantly broader refactoring across:

* typed skeletons,
* generic skeletons,
* proxies,
* allocation logic,
* and existing tests.

Given the scope/risk, this approach was considered too invasive for the current issue.

#### 3. Fixing typed proxy access to use raw slot storage (chosen solution)

The implemented solution changes typed proxy sample access so it no longer depends on interpreting the shared memory as `EventDataStorage<T>`.

Instead:

* the proxy accesses the shared-memory event region through raw slot memory/meta information,
* and calculates slot/sample access using:

  * `sizeof(T)`
  * `alignof(T)`

This keeps the existing shared-memory layout compatible while allowing typed proxies to correctly consume data produced by GenericSkeleton.

This approach was selected because it:

* directly resolves the interoperability problem,
* minimizes architectural disruption,
* preserves existing typed skeleton behavior,
* avoids changing generic container semantics,
* and provides a more robust producer/consumer compatibility model going forward.

---

#### `score/mw/com/impl/bindings/lola/proxy_event.h`

Main functional fix.

Updated typed `ProxyEvent<T>` sample access logic to avoid relying on interpreting the shared-memory region as `EventDataStorage<T>`.

Instead, access is now performed through raw event slot metadata and pointer arithmetic based on the actual sample type size/alignment.

This makes typed proxies compatible with GenericSkeleton-created storage.

---

#### `score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp`

Updated/supporting changes around event storage creation and raw slot metadata handling.

These changes ensure both generic and typed paths expose compatible storage information to consumers.

---

#### `score/mw/com/impl/bindings/lola/skeleton.cpp`

Adjusted skeleton-side integration to align with the updated event storage access model and metadata usage.

---

#### `score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h`

Updated and extended test resources/coverage to validate:

* typed proxy interaction with GenericSkeleton-created event storage,
* slot access correctness,
* and compatibility of the updated shared-memory interpretation model.

---

### Validation performed

The following builds/tests were executed successfully after the changes:

```bash
bazel build //score/mw/com/impl/bindings/lola:proxy
bazel build //score/mw/com/impl/bindings/lola:skeleton

bazel test //score/mw/com/impl/bindings/lola:proxy_event_test
bazel test //score/mw/com/impl/bindings/lola:generic_proxy_event_test
bazel test //score/mw/com/impl/bindings/lola:skeleton_test
bazel test //score/mw/com/impl/bindings/lola:event_data_storage_test
```

All relevant LoLa event/proxy/skeleton tests passed successfully after the fix.

### DIFF
--- a/score/mw/com/impl/bindings/lola/proxy_event.h
+++ b/score/mw/com/impl/bindings/lola/proxy_event.h
@@ -14,7 +14,11 @@
 #define SCORE_MW_COM_IMPL_BINDINGS_LOLA_PROXY_EVENT_H
 
 #include "score/mw/com/impl/bindings/lola/event_data_storage.h"
+#include "score/mw/com/impl/bindings/lola/event_meta_info.h"
 #include "score/mw/com/impl/bindings/lola/proxy_event_common.h"
+
+#include "score/language/safecpp/safe_math/safe_math.h"
+#include "score/memory/shared/pointer_arithmetic_util.h"
 #include "score/mw/com/impl/proxy_event_binding.h"
 #include "score/mw/com/impl/sample_reference_tracker.h"
 #include "score/mw/com/impl/subscription_state.h"
@@ -26,6 +30,7 @@
 #include <score/assert.hpp>
 #include <score/optional.hpp>
 
+#include <cstdint>
 #include <exception>
 #include <iostream>
 #include <limits>
@@ -64,7 +69,9 @@ class ProxyEvent final : public ProxyEventBinding<SampleType>
     ProxyEvent(Proxy& parent, const ElementFqId element_fq_id, const std::string_view event_name)
         : ProxyEventBinding<SampleType>{},
           proxy_event_common_{parent, element_fq_id, event_name},
-          samples_{parent.GetEventDataStorage<SampleType>(element_fq_id)}
+          meta_info_{parent.GetEventMetaInfo(element_fq_id)},
+          aligned_sample_size_{memory::shared::CalculateAlignedSize(sizeof(SampleType), alignof(SampleType))},
+          event_slots_raw_array_{InitialiseEventSlotsRawArray()}
     {
     }
 
@@ -130,12 +137,35 @@ class ProxyEvent final : public ProxyEventBinding<SampleType>
     };
 
   private:
+    const std::uint8_t* InitialiseEventSlotsRawArray();
+
     Result<std::size_t> GetNewSamplesImpl(Callback&& receiver, TrackerGuardFactory& tracker) noexcept;
     Result<std::size_t> GetNumNewSamplesAvailableImpl() const noexcept;
 
     ProxyEventCommon proxy_event_common_;
-    const EventDataStorage<SampleType>& samples_;
+    const EventMetaInfo& meta_info_;
+    const std::size_t aligned_sample_size_;
+    const std::uint8_t* event_slots_raw_array_;
 };
+
+template <typename SampleType>
+inline const std::uint8_t* ProxyEvent<SampleType>::InitialiseEventSlotsRawArray()
+{
+    auto& event_data_control_local = proxy_event_common_.GetConsumerEventDataControlLocal();
+
+    const auto event_slots_raw_array_size = safe_math::Multiply<safe_math::ReturnMode::kAbortOnError>(
+        aligned_sample_size_, event_data_control_local.GetMaxSampleSlots());
+
+    const void* const event_slots_raw_array = meta_info_.event_slots_raw_array_.get(event_slots_raw_array_size);
+
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(nullptr != event_slots_raw_array, "Null event slot array");
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(meta_info_.data_type_info_.size == sizeof(SampleType),
+                                                      "Event sample size mismatch");
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(meta_info_.data_type_info_.alignment == alignof(SampleType),
+                                                      "Event sample alignment mismatch");
+
+    return static_cast<const std::uint8_t*>(event_slots_raw_array);
+}
 
 template <typename SampleType>
 inline Result<std::size_t> ProxyEvent<SampleType>::GetNumNewSamplesAvailable() const noexcept
@@ -191,10 +221,26 @@ inline Result<std::size_t> ProxyEvent<SampleType>::GetNewSamplesImpl(Callback&& 
     const auto slot_indices = proxy_event_common_.GetNewSamplesSlotIndices(max_sample_count);
 
     auto& event_data_control_local = proxy_event_common_.GetConsumerEventDataControlLocal();
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(nullptr != event_slots_raw_array_, "Null event slot array");
 
     for (auto slot_index_it = slot_indices.begin; slot_index_it != slot_indices.end; ++slot_index_it)
     {
-        const SampleType& sample_data{samples_.at(static_cast<std::size_t>(*slot_index_it))};
+        // TODO: Replace this temporary raw-slot access when the LoLa binding layer is type-erased.
+        // The current fix avoids interpreting GenericSkeleton-created storage as EventDataStorage<SampleType>, since
+        // the DynamicArray element count may not match the typed proxy sample type.
+        // NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic) The pointer event_slots_raw_array_ points to
+        // the first byte of the type-erased event sample storage in shared memory. Samples may originate from either a
+        // typed SkeletonEvent or a GenericSkeletonEvent, therefore slot lookup must use the stable EventMetaInfo raw
+        // storage address and SampleType stride instead of interpreting the shared-memory DynamicArray object type.
+        const auto* const object_start_address = &event_slots_raw_array_[aligned_sample_size_ * (*slot_index_it)];
+        // NOLINTEND(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+
+        // Suppress "AUTOSAR C++14 M5-2-8" rule finding: "An object with integer type or pointer to void type shall
+        // not be converted to an object with pointer type.".
+        // The raw storage address is provided through EventMetaInfo. The regular typed proxy validates the expected
+        // type at construction time and calculates the slot offset with sizeof(SampleType)/alignof(SampleType).
+        // coverity[autosar_cpp14_m5_2_8_violation]
+        const SampleType& sample_data{*reinterpret_cast<const SampleType*>(object_start_address)};
         const EventSlotStatus event_slot_status{event_data_control_local[*slot_index_it]};
         const EventSlotStatus::EventTimeStamp sample_timestamp{event_slot_status.GetTimeStamp()};
 
--- a/score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
@@ -27,7 +27,11 @@ namespace score::mw::com::impl::lola
 namespace
 {
 
-using TestSampleType = std::uint16_t;
+// TestSampleType must have the same size and alignment as lola::ProxyMockedMemoryFixture::SampleType.
+// The fixture derives EventMetaInfo (element size and alignment) from its EventDataStorage<SampleType>,
+// and ProxyEvent<T> validates sizeof(T) and alignof(T) against that EventMetaInfo at construction time.
+// Using a type with different size or alignment would trigger a precondition failure.
+using TestSampleType = ProxyMockedMemoryFixture::SampleType;
 
 using namespace ::score::memory::shared;
 using ::testing::Return;
--- a/score/mw/com/impl/bindings/lola/proxy_event_test.cpp
+++ b/score/mw/com/impl/bindings/lola/proxy_event_test.cpp
@@ -703,6 +703,31 @@ TYPED_TEST(LolaProxyEventDeathTest, FailOnEventNotFound)
 }
 
 using LoLaTypedProxyEventTestFixture = LolaProxyEventFixture<ProxyEventStruct>;
+
+TEST_F(LoLaTypedProxyEventTestFixture, GetNewSamplesReturnsTypedSampleFromProviderStorage)
+{
+    // Given a typed LoLa ProxyEvent that is subscribed to a provider event containing one sample
+    const std::size_t max_sample_count_subscription{5U};
+    this->GivenAProxyEvent(this->element_fq_id_, this->event_name_)
+        .ThatIsSubscribedWithMaxSamples(max_sample_count_subscription)
+        .WithSkeletonEventData({{kDummySampleValue, kDummyInputTimestamp}});
+
+    // When GetNewSamples is called
+    const std::size_t max_samples{1U};
+    TestSampleType received_sample{0U};
+    const auto receiver = [&received_sample](impl::SamplePtr<TestSampleType> sample,
+                                             const tracing::ITracingRuntime::TracePointDataId) {
+        received_sample = *sample;
+    };
+
+    const auto num_callbacks_result = this->GetNewSamples(receiver, max_samples);
+
+    // Then one sample is returned to the typed proxy with the data written by the provider
+    ASSERT_TRUE(num_callbacks_result.has_value());
+    EXPECT_EQ(num_callbacks_result.value(), 1U);
+    EXPECT_EQ(received_sample, kDummySampleValue);
+}
+
 TEST_F(LoLaTypedProxyEventTestFixture, SampleConstness)
 {
     RecordProperty("Verifies", "SCR-6340729");
@@ -710,11 +735,21 @@ TEST_F(LoLaTypedProxyEventTestFixture, SampleConstness)
     RecordProperty("TestType", "Requirements-based test");
     RecordProperty("DerivationTechnique", "Analysis of requirements");
 
+    using SamplePtrDataType = std::remove_pointer_t<decltype(std::declval<impl::SamplePtr<TestSampleType>>().get())>;
+    static_assert(std::is_const<SamplePtrDataType>::value, "Proxy should expose const sample data.");
+}
+
+TEST_F(LoLaTypedProxyEventTestFixture, HoldsEventMetaInfoAsConstReference)
+{
+    // Given a typed LoLa ProxyEvent
     this->GivenAProxyEvent(this->element_fq_id_, this->event_name_);
 
+    // When accessing the proxy event through the test attorney
     ProxyEventAttorney<TestSampleType> proxy_event_attorney{*test_proxy_event_};
-    using SamplesMemberType = typename std::remove_reference<decltype(proxy_event_attorney.GetSamplesMember())>::type;
-    static_assert(std::is_const<SamplesMemberType>::value, "Proxy should hold const slot data.");
+
+    // Then the proxy stores EventMetaInfo as const data
+    using MetaInfoMemberType = typename std::remove_reference<decltype(proxy_event_attorney.GetMetaInfoMember())>::type;
+    static_assert(std::is_const<MetaInfoMemberType>::value, "Proxy should hold const event meta info.");
 }
 
 }  // namespace
--- a/score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
+++ b/score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
@@ -22,6 +22,7 @@
 #include "score/mw/com/impl/runtime.h"
 #include "score/mw/com/impl/skeleton_event_binding.h"
 
+#include "score/language/safecpp/safe_math/safe_math.h"
 #include "score/memory/shared/managed_memory_resource.h"
 #include "score/memory/shared/new_delete_delegate_resource.h"
 #include "score/memory/shared/shared_memory_factory.h"
@@ -232,7 +233,30 @@ void* SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory(
     SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(inserted_meta_info.second,
                                                 "Couldn't register/emplace event-meta-info in data-section.");
 
-    return data_storage;
+    return event_data_raw_array;
+}
+
+void* SkeletonMemoryManager::RetrieveGenericEventDataFromOpenedSharedMemory(
+    const ElementFqId element_fq_id,
+    const SkeletonEventProperties& element_properties) noexcept
+{
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(storage_ != nullptr, "Service data storage is not available.");
+
+    const auto event_meta_info_it = storage_->events_metainfo_.find(element_fq_id);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(event_meta_info_it != storage_->events_metainfo_.cend(),
+                                                "Could not find element fq id in meta info map");
+
+    const auto sample_size = event_meta_info_it->second.data_type_info_.size;
+    const auto sample_alignment = event_meta_info_it->second.data_type_info_.alignment;
+    const auto aligned_sample_size =
+        memory::shared::CalculateAlignedSize(sample_size, static_cast<std::size_t>(sample_alignment));
+    const auto total_event_slots_size = safe_math::Multiply<safe_math::ReturnMode::kAbortOnError>(
+        aligned_sample_size, element_properties.number_of_slots);
+
+    void* const event_slots_raw_array = event_meta_info_it->second.event_slots_raw_array_.get(total_event_slots_size);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(event_slots_raw_array != nullptr,
+                                                "Could not get generic EventDataStorage raw array");
+    return event_slots_raw_array;
 }
 
 auto SkeletonMemoryManager::RetrieveEventControlsFromOpenedSharedMemory(const ElementFqId element_fq_id)
--- a/score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
+++ b/score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
@@ -106,7 +106,7 @@ class SkeletonMemoryManager final
     /// \param element_properties Properties of the event.
     /// \param sample_size The size of a single data sample.
     /// \param sample_alignment The alignment of the data sample.
-    /// \return A pair containing the data storage pointer (void*) and the control composite.
+    /// \return A raw pointer to the first byte of the generic event sample storage.
     auto CreateGenericEventDataInCreatedSharedMemory(const ElementFqId element_fq_id,
                                                      const SkeletonEventProperties& element_properties,
                                                      size_t sample_size,
@@ -126,6 +126,14 @@ class SkeletonMemoryManager final
     /// OpenExistingSharedMemory.
     template <typename SampleType>
     auto RetrieveEventDataFromOpenedSharedMemory(const ElementFqId element_fq_id) -> EventDataStorage<SampleType>&;
+
+    /// \brief Retrieves the raw event sample storage pointer for a generic event from opened shared memory.
+    ///
+    /// Generic events use EventMetaInfo as the stable type-erased contract. No interpretation to a
+    /// DynamicArray<SampleType> takes place in this case.
+    auto RetrieveGenericEventDataFromOpenedSharedMemory(const ElementFqId element_fq_id,
+                                                        const SkeletonEventProperties& element_properties) noexcept
+        -> void*;
 
     /// \brief Rolls back any existing operations in the TransactionLog corresponding to a SkeletonEvent
     ///
--- a/score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
+++ b/score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
@@ -156,6 +156,9 @@ void ProxyMockedMemoryFixture::InitialiseDummySkeletonEvent(const ElementFqId el
     std::tie(event_control_, event_data_storage_) =
         fake_data_->AddEvent<SampleType>(element_fq_id, skeleton_event_properties);
     SCORE_LANGUAGE_FUTURECPP_ASSERT(event_control_ != nullptr);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT(event_data_storage_ != nullptr);
+    event_slots_raw_array_ = event_data_storage_->data();
+    SCORE_LANGUAGE_FUTURECPP_ASSERT(event_slots_raw_array_ != nullptr);
     consumer_event_data_control_local_.emplace(event_control_->data_control);
     provider_event_data_control_local_.emplace(event_control_->data_control);
 }
--- a/score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
+++ b/score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
@@ -13,6 +13,7 @@
 #ifndef SCORE_MW_COM_IMPL_BINDINGS_LOLA_TEST_PROXY_EVENT_TEST_RESOURCES_H
 #define SCORE_MW_COM_IMPL_BINDINGS_LOLA_TEST_PROXY_EVENT_TEST_RESOURCES_H
 
+#include "score/mw/com/impl/bindings/lola/event_data_storage.h"
 #include "score/mw/com/impl/bindings/lola/event_subscription_control.h"
 #include "score/mw/com/impl/bindings/lola/generic_proxy_event.h"
 #include "score/mw/com/impl/bindings/lola/i_runtime.h"
@@ -94,9 +95,9 @@ class ProxyEventAttorney
 
     ProxyEventAttorney(ProxyEvent<T>& proxy_event) noexcept : proxy_event_{proxy_event} {}
 
-    auto& GetSamplesMember()
+    auto& GetMetaInfoMember()
     {
-        return proxy_event_.samples_;
+        return proxy_event_.meta_info_;
     }
 
   private:
@@ -161,8 +162,10 @@ class SharedMemoryFactoryGuard
 
 class ProxyMockedMemoryFixture : public ::testing::Test
 {
-  protected:
+  public:
     using SampleType = std::uint32_t;
+
+  protected:
     static constexpr pid_t kDummyPid{123456};
     static constexpr uid_t kDummyUid{2543};
     static constexpr GlobalConfiguration::ApplicationId kDummyApplicationId{6543};
@@ -215,6 +218,7 @@ class ProxyMockedMemoryFixture : public ::testing::Test
     std::optional<ProviderEventDataControlLocalView<>> provider_event_data_control_local_{};
     std::optional<ConsumerEventDataControlLocalView<>> consumer_event_data_control_local_{};
     EventDataStorage<SampleType>* event_data_storage_{nullptr};
+    void* event_slots_raw_array_{nullptr};
     RollbackSynchronization rollback_synchronization_{};
 
     std::shared_ptr<MessagePassingServiceMock> mock_service_{std::make_shared<MessagePassingServiceMock>()};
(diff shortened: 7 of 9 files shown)


crimson11:
@rudresh-systream - thank you for your effort/fix proposal. @bemerybmw and me hat a quick 1st look. Looks promising as a short-term bugfix-solution! Will do a deeper review soon.

ShoroukRamzy:
Hi @rudresh-systream and @crimson11, I quickly reviewed this fix and also run our integration test https://github.com/eclipse-score/communication/pull/379 and all test cases passed for generic-typed/generic use cases.
Thanks for the PR!

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
crimson11: can you fix the return type ..?

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -192,9 +196,40 @@ inline Result<std::size_t> ProxyEvent<SampleType>::GetNewSamplesImpl(Callback&&
 
     auto& event_data_control_local = proxy_event_common_.GetConsumerEventDataControlLocal();
 
+    const std::size_t aligned_sample_size =
```
crimson11: The whole code-block/extension (line 199 - 215) you added here, now happens every time in `GetNewSamplesImpl`, which is the absolute HOT-PATH (called with extremly high frequency!). This is unnecessary, because the code is completely independent from the call arguments.
You should move this code to the Subscribe call (event better would be the `ctor` if possible)

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -192,9 +196,40 @@ inline Result<std::size_t> ProxyEvent<SampleType>::GetNewSamplesImpl(Callback&&
...
+        const auto* const object_start_address = &event_slots_array[aligned_sample_size * (*slot_index_it)];
+        // NOLINTEND(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+
+        // Suppress "AUTOSAR C++14 M5-2-8" rule finding: "An object with integer type or pointer to void type shall
+        // not be converted to an object with pointer type.".
+        // The raw storage address is provided through EventMetaInfo. The regular typed proxy validates the expected
+        // type at compile time and calculates the slot offset with sizeof(SampleType)/alignof(SampleType).
+        // coverity[autosar_cpp14_m5_2_8_violation]
+        const auto* const typed_sample_data = reinterpret_cast<const SampleType*>(object_start_address);
+        const SampleType& sample_data{*typed_sample_data};
```
crimson11: why this temp var?

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -127,6 +127,14 @@ class SkeletonMemoryManager final
     template <typename SampleType>
     auto RetrieveEventDataFromOpenedSharedMemory(const ElementFqId element_fq_id) -> EventDataStorage<SampleType>&;
 
+    /// \brief Retrieves the raw event sample storage pointer for a generic event from opened shared memory.
+    ///
+    /// Generic events use EventMetaInfo as the stable type-erased contract. This keeps generic skeleton restarts
```
crimson11: This comment is slightly "confusing" to me ;)
Imho, a given event-storage residing in shared-memory is always only opened/used by a (typed) SkeletonEvent or GenericSkeletonEvent. 
I don't see any use case, where an application providing a specific service-instance, once starts with using a typed Skeleton ... and next time it starts using a GenericSkeleton. So "interoperability" issues between Generic/typed Skeletons do not exist. You are basically fixing now the GenericSkeleton/typed Proxy interaction by changing the typed Proxy(event) in a way, that it basically does the same data-access as the GenericProxy (using meta-info, directly accessing the underlying storage of the `DynamicArray`)

But then this question comes up (I guess): If I now told you, that TypedSkeleton-GenericSkeleton interoperability on the same shm-object is NO supported use-case ... and in the light of the changes you did to the (typed) proxy event: Why are we STILL constructing a `DynamicArray<T>` in case of a `GenericSkeleton`? Why not generate the raw-array?
Regarding the comment - just change to 
> Generic events use EventMetaInfo as the stable type-erased contract. No interpretation to a `DynamicArray<SampleType>` takes place in this case.

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
rudresh-systream: Fixed the Doxygen return description for CreateGenericEventDataInCreatedSharedMemory() so it now matches the actual void* return type.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Moved the raw event-slot address calculation and metadata validation out of GetNewSamplesImpl(). This is now initialized once when the ProxyEvent is constructed, so the hot path no longer recalculates sample alignment, raw-array size, raw-array pointer, or type metadata on every GetNewSamplesImpl() call.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Removed the extra temporary pointer variable and now bind sample_data directly from the reinterpret-casted slot address.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
rudresh-systream: Updated the GenericSkeleton comment as suggested:
Generic events use EventMetaInfo as the stable type-erased contract. No interpretation to a DynamicArray<SampleType> takes place in this case.

Regarding the raw-array question: I kept the broader storage allocation model unchanged in this PR to avoid expanding the scope. The GenericSkeleton path still owns/reserves the backing memory through the existing storage abstraction, but consumers now use the EventMetaInfo raw-slot contract and do not interpret it as DynamicArray<SampleType>. A full replacement with a true raw-array allocation can be handled as a follow-up refactor if preferred.

crimson11: This is fine for me! As discussed - we will come up with a bigger rework later - where we switch over to completely type-erased storage on binding level... but this might take some time. So for this short-term fix it is wise to not change "everything" ;)

rudresh-systream: ok sure, can you please approve this PR

rudresh-systream:
@crimson11 
 I removed the duplicate GenericSkeleton typed-proxy integration test/application from this PR and force-pushed the branch as a single squashed commit.

This PR is now focused on the #311 production fix and the remaining focused unit/regression coverage. I will check PR #379 for the GenericSkeleton integration-test coverage and add review feedback there if I find any missing Generic-Typed coverage relevant to this issue.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
```diff
@@ -207,6 +208,7 @@ class ProxyMockedMemoryFixture : public ::testing::Test
     std::optional<ProviderEventDataControlLocalView<>> provider_event_data_control_local_{};
     std::optional<ConsumerEventDataControlLocalView<>> consumer_event_data_control_local_{};
     EventDataStorage<SampleType>* event_data_storage_{nullptr};
```
crimson11, line 220: shouldn't you remove this now!? After your change in `lola::ProxyEvent` ... there is no code-path anymore, where a `lola::ProxyEvent` should use/need this "typed" `event_data_storage_` ...

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
```diff
@@ -226,6 +228,8 @@ class LolaProxyEventResources : public ProxyMockedMemoryFixture
     void ExpectReregisterEventNotification(score::cpp::optional<pid_t> pid = {});
     void ExpectUnregisterEventNotification(score::cpp::optional<pid_t> pid = {});
 
+    void ReplaceDummySkeletonEventStorageWithGenericSkeletonLayout();
```
crimson11: You don't need this as an explicit func ... see above.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
```diff
@@ -129,6 +131,27 @@ void ProxyMockedMemoryFixture::InitialiseDummySkeletonEvent(const ElementFqId el
     provider_event_data_control_local_.emplace(event_control_->data_control);
 }
 
+void LolaProxyEventResources::ReplaceDummySkeletonEventStorageWithGenericSkeletonLayout()
```
crimson11: This is now the "default" case! I.e. you always need in EVERY test the `event_meta_info_it->second.event_slots_raw_array_ ` to be set up as now our `lola::ProxyEvent` will always and only access the data via "event-meta-info" ... or am I wrong??

So this func should not be called by any test-case explicitly! Instead this code needs to be shifted to `InitialiseDummySkeletonEvent`. 

If you do this, you have to check, whether `std::tuple<EventControl*, EventDataStorage<SampleType>*> FakeServiceData::AddEvent` needs to stay with this signature at all!? I guess you could/should then change it to `std::tuple<EventControl*, `void*> FakeServiceData::AddEvent` ....?

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
```diff
@@ -187,7 +210,17 @@ SlotIndexType LolaProxyEventResources::PutData(const std::uint32_t value,
     auto slot_result = provider_event_data_control_local_->AllocateNextSlot();
     EXPECT_TRUE(slot_result.has_value());
     auto slot_index = slot_result.value();
-    event_data_storage_->at(slot_index) = value;
+    if (generic_event_data_storage_ != nullptr)
```
crimson11: see above! No if/else needed. With your change a `lola::ProxyEvent` now only/solely works on `generic_event_data_storage_` aken from the meta-info ...

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
rudresh-systream: Done. I removed the typed `EventDataStorage<SampleType>* event_data_storage_` from the proxy-event test fixture.

After the `lola::ProxyEvent` change, the test fixture no longer needs a typed event-data-storage pointer. It now stores and uses the raw event slot pointer from `EventMetaInfo`, matching the production `ProxyEvent` access path.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
rudresh-systream: Done. I removed `ReplaceDummySkeletonEventStorageWithGenericSkeletonLayout()` from the fixture interface.

The raw-slot/EventMetaInfo setup is now part of the default dummy skeleton event initialization, so no test case needs to explicitly switch the fixture into a GenericSkeleton-style layout anymore.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
rudresh-systream: Done. I removed this function implementation and moved the raw event slot setup into the normal `InitialiseDummySkeletonEvent()` flow.

Now every proxy-event test initializes `EventMetaInfo::event_slots_raw_array_` by default, because the updated `lola::ProxyEvent` always accesses samples through the raw slot pointer from `EventMetaInfo`.

I also adjusted the fake service data helper so it returns the raw event sample storage pointer instead of exposing typed `EventDataStorage<SampleType>*`.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
rudresh-systream: Done. I simplified `PutData()` so it no longer has a typed/generic if-else path.

It now always writes sample data through the raw event slot pointer registered in `EventMetaInfo`, using the aligned sample size and slot index. This matches the updated production behavior where `lola::ProxyEvent` reads samples only through the raw-slot/EventMetaInfo path.

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
```diff
@@ -693,6 +693,33 @@ TYPED_TEST(LolaProxyEventDeathTest, FailOnEventNotFound)
 }
 
 using LoLaTypedProxyEventTestFixture = LolaProxyEventFixture<ProxyEventStruct>;
+TEST_F(LoLaTypedProxyEventTestFixture, ReadsSamplesFromEventMetaInfoRawStorage)
+{
+    RecordProperty("Verifies", "SCR-6225206");
```
bemerybmw: Why is this RecordProperty added?

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
```diff
@@ -693,6 +693,33 @@ TYPED_TEST(LolaProxyEventDeathTest, FailOnEventNotFound)
...
+    RecordProperty("Description",
+                   "Checks that a typed ProxyEvent reads samples through EventMetaInfo raw storage.");
+    RecordProperty("TestType", "Regression test");
+    RecordProperty("DerivationTechnique", "Analysis of bug 311");
+
+    const std::size_t max_sample_count_subscription{5U};
```
bemerybmw, line 710: All our tests should follow the given, when, then structure (e.g. https://martinfowler.com/bliki/GivenWhenThen.html)

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
```diff
@@ -693,6 +693,33 @@ TYPED_TEST(LolaProxyEventDeathTest, FailOnEventNotFound)
 }
 
 using LoLaTypedProxyEventTestFixture = LolaProxyEventFixture<ProxyEventStruct>;
+TEST_F(LoLaTypedProxyEventTestFixture, ReadsSamplesFromEventMetaInfoRawStorage)
```
bemerybmw: The test name should clearly state what is the expectation of the test. e.g. In the title you say it should read samples from EventMetaInfoRawStorage but I don't see that tested in the test. Most likely that's just an implementation detail and we simply need to check that GetNewSamples works?

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
```diff
@@ -703,8 +730,8 @@ TEST_F(LoLaTypedProxyEventTestFixture, SampleConstness)
     this->GivenAProxyEvent(this->element_fq_id_, this->event_name_);
 
     ProxyEventAttorney<TestSampleType> proxy_event_attorney{*test_proxy_event_};
-    using SamplesMemberType = typename std::remove_reference<decltype(proxy_event_attorney.GetSamplesMember())>::type;
-    static_assert(std::is_const<SamplesMemberType>::value, "Proxy should hold const slot data.");
+    using MetaInfoMemberType = typename std::remove_reference<decltype(proxy_event_attorney.GetMetaInfoMember())>::type;
```
bemerybmw, line 751: I would add in a new test instead of modifying the current one.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -232,7 +233,35 @@ void* SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory(
...
+                                                "Could not find element fq id in meta info map");
+
+    const auto sample_size = event_meta_info_it->second.data_type_info_.size;
+    const auto sample_alignment = event_meta_info_it->second.data_type_info_.alignment;
+    const auto aligned_sample_size =
+        memory::shared::CalculateAlignedSize(sample_size, static_cast<std::size_t>(sample_alignment));
+    const auto total_event_slots_size = safe_math::Multiply(aligned_sample_size, element_properties.number_of_slots);
+    if (!total_event_slots_size.has_value())
+    {
+        score::mw::log::LogFatal("lola") << "Could not calculate the event slots raw array size. Terminating.";
```
bemerybmw: Why are the others assertions and this is a terminate? You could also use the kAbortOnError return mode in safe_math

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -191,10 +227,23 @@ inline Result<std::size_t> ProxyEvent<SampleType>::GetNewSamplesImpl(Callback&&
     const auto slot_indices = proxy_event_common_.GetNewSamplesSlotIndices(max_sample_count);
 
     auto& event_data_control_local = proxy_event_common_.GetConsumerEventDataControlLocal();
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(nullptr != event_slots_raw_array_, "Null event slot array");
 
     for (auto slot_index_it = slot_indices.begin; slot_index_it != slot_indices.end; ++slot_index_it)
     {
-        const SampleType& sample_data{samples_.at(static_cast<std::size_t>(*slot_index_it))};
+        // NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic) The pointer event_slots_raw_array_ points to
```
bemerybmw, line 231: Please add a todo here stating that this is just a temporary fix since the size of EventDataStorage may incorrect and the long term solution will be to type erase the binding layer.

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -14,7 +14,11 @@
 #define SCORE_MW_COM_IMPL_BINDINGS_LOLA_PROXY_EVENT_H
 
 #include "score/mw/com/impl/bindings/lola/event_data_storage.h"
+#include "score/mw/com/impl/bindings/lola/event_meta_info.h"
 #include "score/mw/com/impl/bindings/lola/proxy_event_common.h"
+
+#include "score/language/safecpp/safe_math/safe_math.h"
```
bemerybmw, line 20: These should also be added as deps in the build file if they're not already there.

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -130,13 +138,41 @@ class ProxyEvent final : public ProxyEventBinding<SampleType>
...
+    const std::uint8_t* event_slots_raw_array_;
 };
 
+template <typename SampleType>
+inline void ProxyEvent<SampleType>::InitialiseEventSlotsRawArray() noexcept
+{
+    auto& event_data_control_local = proxy_event_common_.GetConsumerEventDataControlLocal();
+
+    const auto event_slots_raw_array_size =
+        safe_math::Multiply(aligned_sample_size_, event_data_control_local.GetMaxSampleSlots());
```
bemerybmw: You could also use the kAbortOnError return mode in safe_math

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -64,8 +69,11 @@ class ProxyEvent final : public ProxyEventBinding<SampleType>
     ProxyEvent(Proxy& parent, const ElementFqId element_fq_id, const std::string_view event_name)
         : ProxyEventBinding<SampleType>{},
           proxy_event_common_{parent, element_fq_id, event_name},
-          samples_{parent.GetEventDataStorage<SampleType>(element_fq_id)}
+          meta_info_{parent.GetEventMetaInfo(element_fq_id)},
+          aligned_sample_size_{memory::shared::CalculateAlignedSize(sizeof(SampleType), alignof(SampleType))},
+          event_slots_raw_array_{nullptr}
     {
+        InitialiseEventSlotsRawArray();
```
bemerybmw: This could return const std::uint8_t* and be used to directly initialize event_slots_raw_array_.

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -130,13 +138,41 @@ class ProxyEvent final : public ProxyEventBinding<SampleType>
     };
 
   private:
+    void InitialiseEventSlotsRawArray() noexcept;
```
bemerybmw: We should only mark destructors / move ops as noexcept

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
```diff
@@ -206,7 +206,7 @@ class ProxyMockedMemoryFixture : public ::testing::Test
     EventControl* event_control_{nullptr};
     std::optional<ProviderEventDataControlLocalView<>> provider_event_data_control_local_{};
     std::optional<ConsumerEventDataControlLocalView<>> consumer_event_data_control_local_{};
-    EventDataStorage<SampleType>* event_data_storage_{nullptr};
+    void* event_slots_raw_array_{nullptr};
```
bemerybmw, line 221: We shouldn't remove event_data_storage_. We should rather emulate the production code in which event_slots_raw_array_ is a type erased view into the event_data_storage_

score/mw/com/impl/bindings/lola/test_doubles/fake_mocked_service_data.h
```diff
@@ -48,14 +51,14 @@ struct FakeMockedServiceData
...
+    /// EventMetaInfo.
     template <typename SampleType>
-    std::tuple<EventControl*, EventDataStorage<SampleType>*> AddEvent(ElementFqId id,
-                                                                      SkeletonEventProperties event_properties);
+    std::tuple<EventControl*, void*> AddEvent(ElementFqId id, SkeletonEventProperties event_properties);
 };
 
 template <typename SampleType>
-inline std::tuple<EventControl*, EventDataStorage<SampleType>*> FakeMockedServiceData::AddEvent(
+inline std::tuple<EventControl*, void*> FakeMockedServiceData::AddEvent(
```
bemerybmw: I don't think this should be changed. Same point as: https://github.com/eclipse-score/communication/pull/394/changes#r3288235207

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
bemerybmw, line 1: Please add a descriptive but succinct summary of the changes to the commit message and remove "Signed-off-by"...

We also want to avoid merge commits in our prs. You can do the rebase locally or via github. e.g.

Image

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
rudresh-systream: Removed. This was only added as regression metadata for the temporary test and was not needed here. The added regression test now avoids extra RecordProperty entries and focuses only on the behavior being verified.

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
rudresh-systream: Done. I reworked the added regression test to follow the Given / When / Then structure explicitly.

The test now separates:
- Given: a typed LoLa ProxyEvent subscribed to provider storage containing one sample
- When: GetNewSamples is called
- Then: the typed proxy receives the expected sample value

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
rudresh-systream: Done. I renamed the test from the implementation-detail based name to:

GetNewSamplesReturnsTypedSampleFromProviderStorage

The test now describes the expected behavior instead of mentioning EventMetaInfo raw storage directly.

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
rudresh-systream: Done. I restored the existing SampleConstness test to stay focused on sample constness and added a separate test for the new EventMetaInfo const-member check.

The new test is named:

HoldsEventMetaInfoAsConstReference

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
rudresh-systream: Done. I replaced the manual has_value() / LogFatal / terminate handling with safe_math::ReturnMode::kAbortOnError for the event-slot size calculation.

This keeps the behavior fail-fast while matching the surrounding assertion-style handling more closely.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Done. I added a TODO near the raw-slot access in ProxyEvent.

The comment now states that this is a temporary compatibility fix and that the long-term solution is to type-erase the LoLa binding layer, because the EventDataStorage DynamicArray element count may not match the typed proxy sample type.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Checked. No BUILD file change was required.

The existing LoLa proxy target already has the relevant dependencies available for safe_math and pointer_arithmetic_util, so the PR builds successfully without adding new BUILD deps.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Done. I updated the ProxyEvent raw-slot size calculation to use safe_math::ReturnMode::kAbortOnError.

This removes the explicit result handling and keeps the constructor-time validation path simpler.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Done. InitialiseEventSlotsRawArray() now returns const std::uint8_t* and event_slots_raw_array_ is initialized directly in the constructor initializer list.

This removes the initial nullptr assignment plus later mutation pattern.

score/mw/com/impl/bindings/lola/proxy_event.h
rudresh-systream: Done. I removed noexcept from InitialiseEventSlotsRawArray().

The destructor and move operations remain noexcept as before.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.h
rudresh-systream: Done. I restored `event_data_storage_` in the proxy-event test fixture.

The fixture now emulates the production layout more closely: `event_data_storage_` remains the typed backing storage, and `event_slots_raw_array_` is only a type-erased view into `event_data_storage_->data()`.

So the provider/test setup still behaves like typed skeleton storage, while the updated `ProxyEvent` consumes through the raw slot pointer exposed via `EventMetaInfo`.

score/mw/com/impl/bindings/lola/test_doubles/fake_mocked_service_data.h
rudresh-systream: Done. I reverted this part.

`FakeMockedServiceData::AddEvent()` now keeps its original production-like signature and returns `EventDataStorage<SampleType>*` again instead of `void*`.

The raw slot pointer used by the proxy-event fixture is now derived from `event_data_storage_->data()`, so the type-erased view is created in the fixture without changing the fake service data API.

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
rudresh-systream: Done. I rebased the branch locally onto latest `origin/main`, squashed the PR to a single commit, removed the merge commits, and removed the `Signed-off-by` trailer.

The final commit message is now descriptive and concise:

`Fix typed proxy access to GenericSkeleton event storage`

It includes a short summary explaining that typed `ProxyEvent` now accesses samples through `EventMetaInfo` raw slot storage instead of interpreting provider storage as `EventDataStorage<SampleType>`.

rudresh-systream:
@bemerybmw 

Summary of changes:
- Reworked the added regression test to use a behavior-based name and Given / When / Then structure.
- Removed unnecessary RecordProperty metadata from the added regression test.
- Kept SampleConstness focused on sample constness and added a separate EventMetaInfo const-reference test.
- Changed raw-slot size calculations to use safe_math::ReturnMode::kAbortOnError.
- Changed InitialiseEventSlotsRawArray() to return const std::uint8_t* and initialize event_slots_raw_array_ directly.
- Removed noexcept from InitialiseEventSlotsRawArray().
- Added the requested TODO explaining that raw-slot access is a temporary compatibility fix until the LoLa binding layer is type-erased.
- Restored FakeMockedServiceData::AddEvent() to return EventDataStorage<SampleType>*.
- Restored event_data_storage_ in the proxy-event fixture and made event_slots_raw_array_ a type-erased view into event_data_storage_->data().
- Rebased onto latest origin/main, squashed to one commit, removed merge commits, and removed the Signed-off-by trailer.

Validation:
- bazel test //:format_test_C++_with_clang-format --test_output=all
- bazel build //score/mw/com/impl/bindings/lola:proxy
- bazel build //score/mw/com/impl/bindings/lola:skeleton
- bazel test //score/mw/com/impl/bindings/lola:proxy_event_test
- bazel test //score/mw/com/impl/bindings/lola:generic_proxy_event_test
- bazel test //score/mw/com/impl/bindings/lola:skeleton_test
- bazel test //score/mw/com/impl/bindings/lola:event_data_storage_test

Final branch state:
- one commit on top of origin/main
- no merge commits
- no Signed-off-by trailer

bemerybmw:
@rudresh-systream Please check the failing tests. Once the CI is green, I'll re-review. Also, please don't resolve reviewer comments. You can write a comment saying that you've addressed the comment / add a thumbs up on the comment but it's up to the reviewer to resolve the comment once they've checked that you addressed it correctly.

rudresh-systream:
> @rudresh-systream Please check the failing tests. Once the CI is green, I'll re-review. Also, please don't resolve reviewer comments. You can write a comment saying that you've addressed the comment / add a thumbs up on the comment but it's up to the reviewer to resolve the comment once they've checked that you addressed it correctly.

@bemerybmw 
Updated and force-pushed a CI fix.

The failing CI tests were constructing typed `ProxyEvent` instances with sample types that did not match the dummy skeleton storage created by `ProxyMockedMemoryFixture`. Since `ProxyEvent` now validates the `EventMetaInfo` sample size/alignment during construction, those tests aborted when the fixture-created provider metadata used `std::uint32_t` but the tests instantiated `ProxyEvent<std::uint16_t>` / `ProxyEvent<std::uint8_t>`.

I aligned the two affected construction tests with the fixture sample type:

* `proxy_event_common_test.cpp`
* `proxy_event_field_binding_factory_test.cpp`

Validation to run locally:

* `bazel test //score/mw/com/impl/bindings/lola:proxy_event_common_test --test_output=all`
* `bazel test //score/mw/com/impl/plumbing:proxy_event_field_binding_factory_test --test_output=all`
* `bazel test //... --build_tests_only`

The branch is squashed back to one commit with no merge commits and no `Signed-off-by` trailer.

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
bemerybmw: you've still modified the original test.

score/mw/com/impl/plumbing/proxy_event_field_binding_factory_test.cpp
```diff
@@ -37,7 +37,9 @@ namespace score::mw::com::impl
 
 using namespace ::testing;
 
-using TestSampleType = std::uint8_t;
+// Keep this aligned with lola::ProxyMockedMemoryFixture::SampleType because the fixture initializes
```
bemerybmw: In that case, please make this an alias to ProxyMockedMemoryFixture::SampleType

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
```diff
@@ -27,7 +27,9 @@ namespace score::mw::com::impl::lola
 namespace
 {
 
-using TestSampleType = std::uint16_t;
+// Keep this aligned with ProxyMockedMemoryFixture::SampleType because the fixture initializes
+// EventMetaInfo from its typed EventDataStorage backing storage.
+using TestSampleType = std::uint32_t;
```
bemerybmw: In that case, please make this an alias to ProxyMockedMemoryFixture::SampleType

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
```diff
@@ -27,7 +27,9 @@ namespace score::mw::com::impl::lola
 namespace
 {
 
-using TestSampleType = std::uint16_t;
+// Keep this aligned with ProxyMockedMemoryFixture::SampleType because the fixture initializes
+// EventMetaInfo from its typed EventDataStorage backing storage.
```
bemerybmw: this comment isn't really clear. It doesn't explain why ProxyMockedMemoryFixture::SampleType must be used. "because the fixture initializes EventMetaInfo from its typed EventDataStorage backing storage." is true whether the tests are using std::uint16_t, ProxyMockedMemoryFixture::SampleType or any other type.

score/mw/com/impl/bindings/lola/proxy_event.h
bemerybmw: please also mention that this is a temporary fix

mariuswbr:
FYI @rudresh-systream, @bemerybmw, @crimson11, @LittleHuba, @NEOatNHNG 
Just gave the current state of the fix a try within the SOME/IP gateway, where the GenericSkeleton is used. 
Overall it seems to be working. Data transfer can be seen within the benchmark application and on the network using wireshark, which is already great, as it was broken completely.

The only caveat I faced: At the teardown of the benchmark application it terminates due to a std::bad_variant_access, which we did not have in the past with a typed Skeleton. Not sure if that relates somehow to the changes. Also, not sure if it's worth the investigation effort right now, or if this fix should be merged first to improve the situation at least

Termination log message:

grafik

score/mw/com/impl/bindings/lola/proxy_event_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/proxy_event.h
Tejveerpratap2803: Done

score/mw/com/impl/plumbing/proxy_event_field_binding_factory_test.cpp
Tejveerpratap2803: As SampleType is protected member of ProxyMockedMemoryFixture.
we cannot change to like this - 
`using TestSampleType = ProxyMockedMemoryFixture::SampleType;`

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
Tejveerpratap2803: As SampleType is protected member of ProxyMockedMemoryFixture.
we cannot change to like this - 
`using TestSampleType = ProxyMockedMemoryFixture::SampleType;`

score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
Tejveerpratap2803: Done

bemerybmw (approved):

PR #404, opened 2026-05-11
Remove redundant .clang-tidy-extra file

sahithi-nukala:
- Remove duplicate **modernize-concat-nested-namespaces** check from .clang-tidy-extra as it already exists in the .clang-tidy configuration


limdor (commented):
I'm a bit concerned that checks are passing. We should have some check failing because now this target is inbalid and it should also be removed
https://github.com/sahithi-nukala/communication/blob/1e03b3110c120ae3f9aff07a366ba8c99cf267d3/score/mw/com/BUILD#L176

limdor:
The loading statement should also be deleted
https://github.com/eclipse-score/communication/pull/404/changes#diff-ddfd00285269d0a8d1275f566736ac678491641b79cf6451b463984571b9cf34L17

limdor (approved):

PR #405, opened 2026-05-11
Add get and set and notification support to the fields in the configuration

sahithi-nukala:
Add useGetIfAvailable/useSetIfAvailable to LoLa field deployment

- Introduce dedicated LolaFieldInstanceDeployment class replacing the former LolaEventInstanceDeployment type alias, adding field-specific members use_get_if_available_ and use_set_if_available_
- Fix mw_com_config_schema.json: move useGetIfAvailable and useSetIfAvailable into the field "properties" object so they are correctly validated when additionalProperties is false.
- Adapt existing test and add new tests for LolaFieldInstanceDeployment
- Update PUML diagram and readme document

Issue: SWP-250429


score/mw/com/impl/configuration/lola_field_instance_deployment.h
```diff
@@ -13,12 +13,59 @@
...
+#include <optional>
+
 namespace score::mw::com::impl
 {
 
-using LolaFieldInstanceDeployment = LolaEventInstanceDeployment;
+class LolaFieldInstanceDeployment
+{
+  public:
+    using SampleSlotCountType = std::uint16_t;
```
bemerybmw: These aliases are part of LolaEventInstanceDeployment so should be removed. If you need them here, then you should access them via LolaEventInstanceDeployment.

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -173,7 +173,7 @@
...
...esenting a port prototype of a software component instance). For comfort reasons, we allow also the usage of abbreviated short-name-paths (short-name-paths with a number of components starting from root being removed) in case they are still unique within an executable (this json/configuration is a per executable config!). From AUTOSAR concept perspective an executable contains a 'root sw-component', which itself can be a nested composition of AdaptiveApplicationSwComponentTypes). Because of this 'compositing feature' it could be the case, that the direct short-name of a port prototype isn\u2019t unique within the composition making up the executable. In reality however the port names are 95% unique and forcing the usage of lengthy short-name-paths can be annoying! So this translates to the following rules:\n1. A config file with two \u201cinstanceSpecifiers\u201d, where one is only a trailing substring of the other is INVALID!\n2. If user code makes a resolve/lookup providing an 'instanceSpecifier' in the form of an abbreviated short-name-path, which turns out NOT being unique with respect to the config, this is a hard programming/configuration fault, which shall lead to an abort."
```
bemerybmw: Why did this change?

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
```diff
@@ -50,6 +63,86 @@ TEST_F(LolaFieldInstanceDeploymentFixture, CanCreateFromSerializedObjectWithoutO
     ExpectLolaFieldInstanceDeploymentObjectsEqual(reconstructed_unit, unit);
 }
 
+TEST_F(LolaFieldInstanceDeploymentFixture, UseGetIfAvailableIsTrueAfterRoundTripSerialisation)
```
bemerybmw: You can use a paramaterised test for this. You can search for TEST_P in the code base for examples.

score/mw/com/impl/tracing/configuration/tracing_filter_config.cpp
```diff
@@ -228,34 +228,41 @@ std::size_t FindNumberOfTracingSlots(
                 std::terminate();
             }
 
-            const auto& service_instance_map = [service_element_type, &lola_service_instance_deployment]() {
-                if (service_element_type == ServiceElementType::EVENT)
+            const auto service_element_name = service_element.service_element_name.data();
```
bemerybmw: To minimize the duplicated code, maybe we can change the lambda to be get_number_of_tracing_slots which accepts auto service_elements_map (which can be events_ or fields_) and do the map lookup within the lambda?

score/mw/com/impl/configuration/config_parser_test.cpp
```diff
@@ -113,9 +113,17 @@ TEST_F(ConfigParserFixture, ParseExampleJson)
...
+    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft")
+                  .lola_event_instance_deployment_.max_subscribers_.value(),
+              6);
+    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft")
+                  .lola_event_instance_deployment_.enforce_max_samples_,
+              true);
+    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft")
+                  .lola_event_instance_deployment_.max_concurrent_allocations_.value(),
+              1);
+    EXPECT_FALSE(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_get_if_available_);
```
bemerybmw: The values should be changed so that they're true. Makes it more likely to catch if we're not correctly parsing the value (since defualt is false).

score/mw/com/design/configuration/structural_view.puml
```diff
@@ -54,11 +54,22 @@ package "configuration" {
...
+    + enforce_max_samples_ : bool
+    + use_get_if_available_ : bool
+    + use_set_if_available_ : bool
+    - number_of_sample_slots_ : std::optional<SampleSlotCountType>
+    - number_of_tracing_slots_ : TracingSlotSizeType
+    __
+    {static} + CreateFromJson(json_object : const score::json::Object&) : LolaFieldInstanceDeployment
+    + Serialize() const : score::json::Object
+    --
+    <u>Type Aliases:</u>
```
bemerybmw: Aliases should be removed since they're part of the event deployment.

score/mw/com/design/configuration/structural_view.puml
```diff
@@ -54,11 +54,22 @@ package "configuration" {
   }
 
   class "score::mw::com::impl::LolaFieldInstanceDeployment" {
-    + max_samples_: std::uint16_t
-    + max_subscribers_: std::uint8_t
-    + enforce_max_samples_ : score::cpp::optional<bool>
-    + is_set_configured: bool
-    + is_get_configured: bool
+    {static} + constexpr serializationVersion : std::uint32_t
+    + max_subscribers_ : std::optional<SubscriberCountType>
```
bemerybmw: We now compose a LolaEventInstanceDeployment.

bemerybmw (changes requested):

score/mw/com/impl/configuration/lola_field_instance_deployment.h
```diff
@@ -13,12 +13,59 @@
...
+
+    [[nodiscard]] TracingSlotSizeType GetNumberOfTracingSlots() const noexcept;
+
+    // Note the struct is not compliant to POD type containing non-POD member.
+    // The struct is used as a config storage obtained by performing the parsing json object.
+    // Public access is more convenient to reach the following members of the struct.
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    LolaEventInstanceDeployment lola_event_instance_deployment_;
+    // coverity[autosar_cpp14_m11_0_1_violation]
+    bool use_get_if_available_;
```
bemerybmw: Should be an optional because it's mandatory on proxy side and not used on skeleton side. Doesn't make sense to have a value on skeleton side.

score/mw/com/impl/configuration/config_parser_test.cpp
sahithi-nukala: Done. And updated the mw_com_config.json file also accordingly

score/mw/com/impl/configuration/lola_field_instance_deployment.h
sahithi-nukala: 1. I have removed -
`using SampleSlotCountType = std::uint16_t;
using SubscriberCountType = std::uint8_t;
using TracingSlotSizeType = std::uint8_t;`
and updated with the "**LolaEventInstanceDeployment**". example like below -
`void SetNumberOfSampleSlots(LolaEventInstanceDeployment::SampleSlotCountType number_of_sample_slots) noexcept;`

score/mw/com/impl/configuration/lola_field_instance_deployment.h
sahithi-nukala: Done

score/mw/com/design/configuration/structural_view.puml
sahithi-nukala: Done. I have removed the max_subscribers_, max_concurrent_allocations_, enforce_max_samples_ 
and updated with  LolaEventInstanceDeployment.

score/mw/com/design/configuration/structural_view.puml
sahithi-nukala: Done

score/mw/com/impl/tracing/configuration/tracing_filter_config.cpp
sahithi-nukala: Updated to lambda that takes service_elements_map (which can be events_ or fields_) to avoid the duplicated code.

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
sahithi-nukala: done.

score/mw/com/impl/bindings/lola/test/skeleton_test_resources.h
```diff
@@ -239,7 +239,8 @@ static const ServiceInstanceDeployment kValidInstanceDeploymentWithField{
     CreateLolaServiceInstanceDeployment(
         kDefaultLolaInstanceId,
         {},
-        {{test::kFooEventName, LolaFieldInstanceDeployment{test::kMaxSlots, 10U, 1U, true, 0}}},
+        {{test::kFooEventName,
```
crimson11: hm - unexpected to me ... why don't you use `kFooFieldName` here?

score/mw/com/impl/bindings/lola/test/skeleton_test_resources.h
```diff
@@ -284,7 +285,8 @@ static const ServiceInstanceDeployment kValidAsilInstanceDeploymentWithField{
...
```
crimson11: see above ... why not `kFooFieldName`?

score/mw/com/impl/bindings/lola/skeleton_test.cpp
```diff
@@ -1167,9 +1167,13 @@ TEST_P(SkeletonRegisterParamaterisedFixture, ValidEventMetaInfoExistAfterEventIs
         fields_.emplace(test::kDumbEventName, dumb_event);
 
         lola_field_inst_depls.push_back(
-            {test::kFooEventName, LolaFieldInstanceDeployment{number_of_slots, 10U, 1U, true, 0}});
+            {test::kFooEventName,
```
crimson11: see above ... `kFooFieldName` ?

score/mw/com/impl/bindings/lola/skeleton_test.cpp
```diff
@@ -1167,9 +1167,13 @@ TEST_P(SkeletonRegisterParamaterisedFixture, ValidEventMetaInfoExistAfterEventIs
...
+             LolaFieldInstanceDeployment{
+                 LolaEventInstanceDeployment{number_of_slots, 10U, 1U, true, 0}, false, false}});
         lola_field_inst_depls.push_back(
-            {test::kDumbEventName, LolaFieldInstanceDeployment{number_of_slots, 10U, 1U, true, 0}});
+            {test::kDumbEventName,
```
crimson11: see above ... `kDumbFieldName` ?

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
```diff
@@ -81,15 +91,19 @@ TEST_F(LolaFieldInstanceDeploymentFixture, UseGetIfAvailableIsTrueAfterRoundTrip
...
-    EXPECT_EQ(reconstructed_unit.use_get_if_available_, kUseGetIfAvailable);
-    EXPECT_EQ(reconstructed_unit.use_set_if_available_, kUseSetIfAvailable);
+    ASSERT_TRUE(reconstructed_unit.use_get_if_available_.has_value());
+    ASSERT_TRUE(reconstructed_unit.use_set_if_available_.has_value());
+    EXPECT_EQ(reconstructed_unit.use_get_if_available_.value(), kUseGetIfAvailable);
+    EXPECT_EQ(reconstructed_unit.use_set_if_available_.value(), kUseSetIfAvailable);
 }
 
-TEST_F(LolaFieldInstanceDeploymentFixture, UseSetIfAvailableIsTrueAfterRoundTripSerialisation)
+INSTANTIATE_TEST_SUITE_P(UseGetIfAvailable, LolaFieldInstanceDeploymentUseGetParamFixture, ::testing::Values(true));
```
crimson11: Hm? You are only instantiating the parametrized test with `true` here? Shouldn't you do:
`INSTANTIATE_TEST_SUITE_P(UseGetIfAvailable, LolaFieldInstanceDeploymentUseGetParamFixture, ::testing::Values(true, false));` or 
`INSTANTIATE_TEST_SUITE_P(UseGetIfAvailable, LolaFieldInstanceDeploymentUseGetParamFixture, ::testing::Bool());`

... and I think you still have a lot duplication and don't exploit parametrized-tests fully!
Your parametrization should consist of bool-pairs (1st element is enableGet, 2nd element is enableSet).
So your testing-values should be:

```
std::pair<bool, bool>{true, true}
std::pair<bool, bool>{true, false}
std::pair<bool, bool>{false, true}
std::pair<bool, bool>{false, false}
```

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
```diff
@@ -43,6 +43,16 @@ namespace score::mw::com::impl
 namespace detail
 {
 
+template <typename LolaServiceElementInstanceDeployment>
```
crimson11: This pre-decl seems unneeded? Just shift the definition in line 50 below the `GetSkeletonEventProperties()` func template beginning in line 57?

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
crimson11: OK. I guess our plumbing code now is "a mess"!?
`CreateSkeletonServiceElement` and the template arg `element_type` imply, that this func-template can deal with all the element types (Event/Field/Method). But this is dead-wrong! If you look at the code (namely line 119), where explicitly event-properties are set up ... it is obvious, that we woud get compile-time errors, when trying to instantiate this func-template for a METHOD. (Right now we are obviously NOT doing it - the plumbing code for methods doesn't use this.)

Bottom line:

- at least rename the func-template to `CreateSkeletonEventOrField()`
- since we still might need the template arg `element_type`, we should `static_assert` here, that it is either FIELD or EVENT.
- eventually shift the whole code from `skeleton_service_element_binding_factory_impl.h` to `skeleton_event_binding_factory_impl.h` ... and "re-use" the functionality from there in case of a FIELD. `skeleton_service_element_binding_factory_impl.h` can then be removed ...

crimson11: @bemerybmw : Please add your view on this!

score/mw/com/impl/tracing/configuration/tracing_filter_config.cpp
```diff
@@ -228,34 +228,38 @@ std::size_t FindNumberOfTracingSlots(
...
 
-            const auto& service_instance_map = [service_element_type, &lola_service_instance_deployment]() {
-                if (service_element_type == ServiceElementType::EVENT)
-                {
-                    return lola_service_instance_deployment->events_;
-                }
-                if (service_element_type == ServiceElementType::FIELD)
+            const auto service_element_name = service_element.service_element_name.data();
+            const auto get_number_of_tracing_slots =
+                [&](const auto& service_elements_map) noexcept -> LolaEventInstanceDeployment::TracingSlotSizeType {
```
crimson11: Nitpicky ;) - You don't need to (globally) capture here (this you should typically avoid) ... why not hand over `service_element_name` as a call-arg to the lambda?

crimson11 (dismissed):

score/mw/com/impl/bindings/lola/test/skeleton_test_resources.h
sahithi-nukala: Updated with kFooFieldName

score/mw/com/impl/bindings/lola/test/skeleton_test_resources.h
sahithi-nukala: Done

score/mw/com/impl/bindings/lola/skeleton_test.cpp
sahithi-nukala: done

score/mw/com/impl/bindings/lola/skeleton_test.cpp
sahithi-nukala: done

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
sahithi-nukala: Done

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
sahithi-nukala: done

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
```diff
@@ -67,6 +67,12 @@ lola::SkeletonEventProperties GetSkeletonEventProperties(
                                          lola_service_element_instance_deployment.enforce_max_samples_};
 }
 
+inline lola::SkeletonEventProperties GetSkeletonEventProperties(
```
bemerybmw, line 69: Why have this intermediate function?

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
bemerybmw: From memory, Krishna has a pr somewhere making a similar change. But I agree that it should be CreateSkeletonEventOrField for the moment. In the future, I don't have a strong preference about having skeleton_service_element_binding_factory_impl (although renamed) or calling the event function directly from the field factory.

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
bemerybmw: I would even go one step further and make it something like:
```
enum class Getter
{
    ENABLED,
    DISABLED
};

std::pair<Getter, Setter>{Getter::ENABLED, Setter::ENABLED}; 
std::pair<Getter, Setter>{Getter::ENABLED, Setter::DISABLED};

Comment (sahithi-nukala) [reply score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp]: Done

File: score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h (hunk 16 above)
Comment (sahithi-nukala) [reply score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h:69]: removed the inline function and directly updated the GetSkeletonEventProperties

File: score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp [hunk 17]
```diff
@@ -16,12 +16,45 @@
...
+
+enum class Setter
+{
+    ENABLED,
+    DISABLED
+};
+
+constexpr bool IsEnabled(const Getter getter) noexcept
+{
+    return getter == Getter::ENABLED;
```
bemerybmw: I don't think `IsEnabled(getter)` is any more readable than `getter == Getter::ENABLED`  so I think we should remove the extra indirection i.e. remove is function.

score/mw/com/impl/configuration/lola_field_instance_deployment.h
```diff
@@ -13,12 +13,57 @@
...
+
+    explicit LolaFieldInstanceDeployment(const score::json::Object& json_object) noexcept;
+
+    static LolaFieldInstanceDeployment CreateFromJson(const score::json::Object& json_object) noexcept;
+
+    score::json::Object Serialize() const noexcept;
+
+    void SetNumberOfSampleSlots(LolaEventInstanceDeployment::SampleSlotCountType number_of_sample_slots) noexcept;
+
+    [[nodiscard]] std::optional<LolaEventInstanceDeployment::SampleSlotCountType> GetNumberOfSampleSlots()
```
bemerybmw: These setter / getter functions can be called on the LolaEventInstanceDeployment directly e.g. : field_deployment.lola_event_instance_deployment_.GetNumberOfSampleSlots()? So I think we can just remove them instead of having to maintain them in 2 places?

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
```diff
@@ -106,8 +106,17 @@ auto CreateSkeletonServiceElement(const InstanceIdentifier& identifier,
             const std::string service_element_name_str{service_element_name};
             const auto& lola_service_element_instance_deployment = GetServiceElementInstanceDeployment<element_type>(
                 lola_service_instance_deployment, service_element_name_str);
-            const auto skeleton_event_properties =
-                detail::GetSkeletonEventProperties(lola_service_element_instance_deployment);
+            lola::SkeletonEventProperties skeleton_event_properties{};
+            if constexpr (element_type == ServiceElementType::FIELD)
+            {
+                skeleton_event_properties = detail::GetSkeletonEventProperties(
```
bemerybmw: Instead of having GetSkeletonEventProperties be templated and adding this additional branch here, why not have 2 overloads for GetSkeletonEventProperties (and make it non templated). The first overload can take LolaEventInstanceDeployment and the second can take LolaFieldInstanceDeployment (and this one can call the first one with field_deployment.lola_event_instance_deployment_? This might be actually similar to what you originally had...

score/mw/com/impl/configuration/lola_field_instance_deployment_test.cpp
sahithi-nukala: Done

score/mw/com/impl/configuration/lola_field_instance_deployment.h
sahithi-nukala: done

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
sahithi-nukala: done

bemerybmw (approved):

PR #409, opened 2026-05-12
Extend GenericSkeletonEvent to trigger EventUpdate notification

Rahul-Sutariya:


score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,459 @@
...
+                                    const std::uint8_t max_subscribers,
+                                    const bool enforce_max_samples)
+    {
+        const SkeletonEventProperties properties{max_samples, max_subscribers, enforce_max_samples};
+        generic_skeleton_event_ = std::make_unique<GenericSkeletonEvent>(
+            *skeleton_, event_name, properties, element_fq_id, size_info_, impl::tracing::SkeletonEventTracingData{});
+    }
+
+    std::unique_ptr<GenericSkeletonEvent> generic_skeleton_event_;
+    const DataTypeMetaInfo size_info_{10U, 16U};
```
crimson11: Alignment can't be larger than size ... see comment in `CreateDataTypeSizeInfoFromTypes()` in `type_erased_storage.h`

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,459 @@
...
+};
+
+// Test: Construction
+TEST_F(GenericSkeletonEventFixture, CanConstructAGenericSkeletonEvent)
+{
+    RecordProperty("Verifies", "SCR-14035184");
```
crimson11, line 65: 1st: I much appreciate, that you now did invest in all these missing unit-tests for `GenericSkeletonEvent`! GOOD!
The issue with the "linked" requirements (like SCR-14035184) is: These are IDs from the BMW internal codebeamer requirement system. They aren't visible in S-CORE, where we are now working. So we should rather link to the new requirements... see here: https://github.com/eclipse-score/communication/blob/main/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc.trlc#L1487
This is the same req. in its new form (in the broken link comment you see the relation to the old codebeamer req.)
So - please get in contact with Jochen Hoenle and ask him, how you should proceed (what you should link to in your test -> i.e. if instead of writing `SCR-14035184` you should write `ScoreReq.CompReq GenericProxyEventGetSampleSize`

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+    // And notifying consumers
+    auto notify_result = generic_skeleton_event_->Notify();
+    EXPECT_TRUE(notify_result.has_value());
+
+    // And stopping the offer
+    generic_skeleton_event_->PrepareStopOffer();
+
+    // Then the complete lifecycle succeeds
+    EXPECT_TRUE(true);
```
crimson11: What does this "add"

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,459 @@
...
+
+    // Given a GenericSkeletonEvent
+    CreateGenericSkeletonEvent(
+        fake_element_fq_id_, fake_event_name_, max_samples, max_subscribers, enforce_max_samples);
+
+    // When calling PrepareOffer
+    std::ignore = generic_skeleton_event_->PrepareOffer();
+
+    // When calling Notify (without handlers registered, this should succeed)
+    auto result = generic_skeleton_event_->Notify();
```
crimson11, line 355: I guess we also need a test, which shows, that in case of registered receive-handlers `Notify` calls to message-passing?

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Fixed invalid DataTypeMetaInfo alignment

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Placeholder for the for the //Then section, Removed redundant always-true assertion from the full lifecycle test.

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Added a new test that proves Notify calls message-passing when receive-handlers are registered

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: On score linking attributes not finalized yet added Todo.
// TODO: Fix requirement linkage as soon as requirements are matured in S-CORE.

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+    const bool enforce_max_samples{true};
+    const std::size_t max_samples{5U};
+    const std::uint8_t max_subscribers{3U};
+
+    // Given a GenericSkeletonEvent
+    CreateGenericSkeletonEvent(
+        fake_element_fq_id_, fake_event_name_, max_samples, max_subscribers, enforce_max_samples);
+
+    // And receive handlers reported as registered for both quality levels
```
crimson11: OK. Now I'm a bit nitpicky here ;)
The comment above/before the two EXPECT_CALL should be
> expect, that EventNotificationExistenceChangedCallbacks are registered for both ASIL levels at message_passing

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,504 @@
...
+            }));
+    EXPECT_CALL(
+        message_passing_mock_,
+        RegisterEventNotificationExistenceChangedCallback(QualityType::kASIL_B, fake_element_fq_id_, testing::_))
+        .WillOnce(testing::Invoke(
+            [](QualityType, ElementFqId, IMessagePassingService::HandlerStatusChangeCallback has_handlers_callback) {
+                has_handlers_callback(true);
+            }));
+
+    // When offering the event and calling Notify
```
crimson11: this should be just
> when offering the event

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,504 @@
...
+    std::ignore = generic_skeleton_event_->PrepareOffer();
+    EXPECT_CALL(message_passing_mock_, NotifyEvent(QualityType::kASIL_QM, fake_element_fq_id_));
```
crimson11, line 398: then comment before these two EXPECT_CALL
> expect, that NotifyEvent gets called at message-passing for both ASIL-levels

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,504 @@
...
+    EXPECT_CALL(message_passing_mock_, NotifyEvent(QualityType::kASIL_B, fake_element_fq_id_));
+    auto result = generic_skeleton_event_->Notify();
```
crimson11, line 401: comment above
> when Notify() gets called on the GenericSkeletonEvent

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
```diff
@@ -0,0 +1,504 @@
...
+
+    // Then message passing is notified for both quality levels
```
crimson11: here the comment should then rather be
> ... and that the Notify() call had no errors.

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Updated, Thanks

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Updated, Thanks

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Updated, Thanks

score/mw/com/impl/bindings/lola/generic_skeleton_event_test.cpp
Rahul-Sutariya: Updated, Thanks

crimson11 (approved):

PR #414, opened 2026-05-12
mw/com: Make ProxyBase::AreBindingsValid symmetrical with Skeleton side

bemerybmw:
Previously, ProxyBase::AreBindingsValid was implemented differently to avoid have to store a map of service elements (which also requires updating the map when moving a Proxy) in ProxyBase. However, we have since added these maps so there's no longer any reason to have the asymmetry in the implementations of AreBindingsValid.

### DIFF
--- a/score/mw/com/design/skeleton_proxy/README.md
+++ b/score/mw/com/design/skeleton_proxy/README.md
@@ -188,34 +188,23 @@ independent level.
 
 #### Binding level Registration of proxy events/fields at their parent proxy
 
-On the binding **specific** level things look different!
-I.e. the binding specific proxy needs to interact with its dependent/child proxy events of type `ProxyEventBindingBase`
+On the binding **specific** level things look similar:
+the binding specific proxy needs to interact with its dependent/child proxy events of type `ProxyEventBindingBase`
 (at least the `LoLa`/shared-memory specific does, so we introduced it on the `ProxyBinding` interface level) .
 
-Therefore, `impl::ProxyBinding` has a method `RegisterEventBinding` (and `UnregisterEventBinding`) and our `LoLa`
-binding specific proxy `lola::Proxy`, which implements `impl::ProxyBinding`, has a member `event_bindings_` (a map
-containing its child proxy events), which it populates in its implementation of `RegisterEventBinding` with its
-dependent proxy events.
-
-The call to `RegisterEventBinding` happens during creation of a binding independent `impl::ProxyEvent` via an `RAII`
-style member `event_binding_registration_guard_` of type `EventBindingRegistrationGuard`, which the `impl::ProxyEvent`
-owns.
-This `EventBindingRegistrationGuard` takes over two functionalities:
-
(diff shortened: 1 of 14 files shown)


KrishaDeshkool:
The removal of EventBindingRegistrationGuard would happen later?
other than the changes are fine as we discussed.

bemerybmw:
> The removal of EventBindingRegistrationGuard would happen later? other than the changes are fine as we discussed.

Yep, that's happening in the next change that I'm working on.

KrishaDeshkool (approved):

score/mw/com/design/skeleton_proxy/README.md
```diff
@@ -168,54 +174,30 @@ On `ProxyBase` level we have two types of methods:
...
-the proxy side factories.
-
-So, the binding independent `impl::ProxyBase` doesn't "know" its semantically aggregated `ProxyEvent`s.
-Although the `ProxyEvent`s get a reference to their parent `ProxyBase` during construction (and therefore "know" their
-parent in the context of their `ctor`), they don't store this reference (as it isn't really needed currently and
-would require additional logic to update this reference if the parent `impl::ProxyBase` gets moved).
-Instead, they only set once (see above) the `ProxyBase::are_service_element_bindings_valid_` member variable of their
-parent during construction as this is currently the only feedback needed between proxy and its proxy events on binding
-independent level.
+Similarly to the skeleton side, the `impl::SkeletonBase` doesn't "know" its event/field/method children. Therefore,
```
crimson11: > impl::SkeletonBase

impl::ProxyBase

score/mw/com/design/skeleton_proxy/README.md
```diff
@@ -168,54 +174,30 @@ On `ProxyBase` level we have two types of methods:
...
+the `impl::ProxyEventBase` registers itself with the `impl::ProxyBase` in the same way as `impl::SkeletonEventBase`
+(as explained
+[here on the skeleton side](#binding-independent-level-registration-of-skeleton-eventsfields-at-their-parent-skeleton). 
 
 #### Binding level Registration of proxy events/fields at their parent proxy
 
 On the binding **specific** level things look different!
```
crimson11: Does this sentence now make sense anymore? So there is no difference between binding-specific/binding-unspecific level anymore? In both cases the proxy knows its childs, because on both levels this registration process takes place ...

score/mw/com/impl/bindings/lola/proxy_test.cpp
crimson11, line 460: Nitpicky:

`.WillOnce(InvokeArgument<0>([valid_find_service_handle, this](auto find_service_handler) noexcept {`
is a lot clearer ;)

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/proxy_test.cpp
bemerybmw: I don't think this works. I think this is intended to call the argument directly but we rather need to do some prep, call the callable and then return a value (via a lambda). From the docs, I think InvokeArgument was added before c++ had lambdas.

crimson11 (approved):

PR #415, opened 2026-05-13
Execute Integration Tests on QNX in CI

castler:


MODULE.bazel
```diff
@@ -63,10 +63,10 @@ gcc_toolchains.toolchain(
 )
 use_repo(gcc_toolchains, "gcc_toolchain_x86_64")
 
-bazel_dep(name = "score_bazel_cpp_toolchains", version = "0.3.0", dev_dependency = True)
+bazel_dep(name = "score_bazel_cpp_toolchains", version = "0.5.1", dev_dependency = True)
 git_override(
     module_name = "score_bazel_cpp_toolchains",
-    commit = "79982259df6dfa9d82cc458163a6d9b08bd24471",
+    commit = "30d9f4151db55af066e5bbf8a0a3672469ec0293",
```
limdor: Why do we need this override? 0.5.1 is already having the commit that we are interested. In this case with the override we are removing 3 commits from the released version. Do we have an issue with any of these removed commits? If yes, was this reported to score_bazel_cpp_toolchains?

MODULE.bazel
```diff
@@ -23,7 +23,16 @@ bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "rules_python", version = "1.8.5")
 bazel_dep(name = "rules_rust", version = "0.68.1-score")
+
+# Patch baselibs with a QNX8 workaround for poll
+# until a fix is provided by the QNX SDP
```
limdor, line 28: AFAIK this fix should be available in 8.0.4, if we add this information here it will be easier to remember when to remove the patch.

score/mw/com/test/api/BUILD
```diff
@@ -30,6 +30,7 @@ cc_gtest_unit_test(
     name = "generated_api_test",
     srcs = ["generated_api_test.cpp"],
     features = COMPILER_WARNING_FEATURES,
+    target_compatible_with = ["@platforms//os:linux"],
```
limdor, line 33: Do we want to add a comment with the ticket to have QNX unit tests enabled? In this way it will be harder to forget about removing this once we have the ability to run them on QNX.

score/mw/com/test/common_test_resources/BUILD
```diff
@@ -168,6 +168,7 @@ cc_gtest_unit_test(
     srcs = [
         "command_line_parser_test.cpp",
     ],
+    target_compatible_with = ["@platforms//os:linux"],
```
limdor, line 171: Do we want to add a comment with the ticket to have QNX unit tests enabled? In this way it will be harder to forget about removing this once we have the ability to run them on QNX.

MODULE.bazel
limdor: I discussed in offline with @castler and this should not be needed. We agreed to do it in a follow up PR in this way we can have the failing tests fixed asap.

limdor (dismissed):
I discussed in offline with @castler the comments and we agreed to do them in a follow up PR in this way we can have the failing tests fixed asap.

crimson11 (approved):

PR #419, opened 2026-05-13
Add extra checks to clang-tidy config

sahithi-nukala:
- Configure CheckOptions for bugprone, modernize and performance
- checks with std:: types, ::score::cpp::pmr::vector; SCORE_LANGUAGE_FUTURECPP assert macros.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -87,7 +87,55 @@ Checks: >
   # --- qnx-checks ---
   bugprone-reserved-identifier,
   # --- extra ---
+  bugprone-assert-side-effect,
+  bugprone-copy-constructor-init,
+  bugprone-dangling-handle,
+  bugprone-suspicious-stringview-data-usage,
+  misc-include-cleaner,
+  misc-use-anonymous-namespace,
+  misc-use-internal-linkage,
+  modernize-avoid-bind,
   modernize-concat-nested-namespaces,
+  modernize-loop-convert,
+  modernize-make-unique,
+  modernize-min-max-use-initializer-list,
+  modernize-pass-by-value,
+  modernize-raw-string-literal,
+  modernize-redundant-void-arg,
+  modernize-return-braced-init-list,
+  modernize-shrink-to-fit,
+  modernize-type-traits,
+  modernize-unary-static-assert,
+  modernize-use-bool-literals,
+  modernize-use-emplace,
+  modernize-use-equals-delete,
+  modernize-use-integer-sign-comparison,
+  modernize-use-nodiscard,
+  modernize-use-nullptr,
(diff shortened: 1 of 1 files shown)


.clang-tidy
```diff
@@ -111,3 +159,72 @@ CheckOptions:
...
+  - key: modernize-loop-convert.IncludeStyle
+    value: "llvm"
+  - key: modernize-make-unique.IncludeStyle
+    value: "llvm"
+  - key: modernize-min-max-use-initializer-list.IncludeStyle
+    value: "llvm"
+  - key: modernize-pass-by-value.IncludeStyle
+    value: "llvm"
+  - key: modernize-use-emplace.ContainersWithPush
+    value: "::std::pmr::deque;::std::stack;::std::queue;::std::priority_queue"
```
limdor:
```suggestion
```
`::std::pmr::deque` has no `push` only `push_back`. Because the rest are the default you can remove the full config. In this way we do not have to add new types manually in the future if they appear in newer standards.

.clang-tidy
```diff
@@ -111,3 +159,72 @@ CheckOptions:
...
+  - key: modernize-use-emplace.ContainersWithPushBack
+    value: "::std::pmr::deque;::score::cpp::pmr::vector;::std::vector;::std::list;::std::deque"
```
limdor:
```suggestion
    value: "::score::cpp::inplace_vector;::score::cpp::pmr::deque;::score::cpp::pmr::list;::score::cpp::pmr::vector;::std::vector;::std::list;::std::deque"
```
come custom tzpes are missing

.clang-tidy
```diff
@@ -111,3 +159,72 @@ CheckOptions:
...
+  - key: modernize-use-emplace.ContainersWithPushFront
+    value: "::std::forward_list;::std::list;::std::deque"
```
limdor:
```suggestion
```
this is the same like the default, we can leave it out.  In this way we do not have to add new types manually in the future if they appear in newer standards.

.clang-tidy
```diff
@@ -111,3 +159,72 @@ CheckOptions:
...
+  - key: modernize-use-integer-sign-comparison.IncludeStyle
+    value: "llvm"
+  - key: modernize-use-nodiscard.ReplacementString
+    value: "[[nodiscard]]"
+  - key: modernize-use-transparent-functors.SafeMode
+    value: true
+  - key: performance-faster-string-find.StringLikeClasses
+    value: >
+      ::score::cpp::basic_string_view;
+      ::score::cpp::span;
```
limdor:
```suggestion
```
We should not consider span as a string like class

limdor (approved):

PR #430, opened 2026-05-14
Docs(sphinx): configure GitHub Pages

AAmbuj:
added DOCS_BASE_URL-based GitHub Pages URL configuration include version flyout CSS and JS assets in Sphinx output

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -1,0 +1,211 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy_docs.yml
```diff
@@ -0,0 +1,193 @@
...
+
+jobs:
+  build-docs:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@eef61447b9ff4aafe5dcd4e0bbf5d482be7e7871 # v4.2.1
+
+      - name: Set up Bazel
+        uses: bazel-contrib/setup-bazel@0.14.0
```
castler, line 49: Please use the same approach as in the other jobs.

castler (approved):
I still think there are major reworks needed - but lets start with this for the time being.

PR #438, opened 2026-05-18
Add SkeletonField::Get support (EnableGet flag)

Tejveerpratap2803:
mw/com/impl: introduces and implements SkeletonField::Get

- SkeletonField now includes a new template parameter, EnableGet.
- When EnableGet is true, a skeleton Get method attribute is added to SkeletonField.
- Constructors were updated to support all relevant EnableSet and EnableGet combinations.
- Automatically register a Get handler for fields with EnableGet.
- Serve Get requests by reading the latest field value and returning it through the Get method.
- Pass additional_slots through field binding creation.
- Compute additional_slots from EnableSet || EnableGet.
- Use configured_slots + additional_slots in event properties.
- Add unit tests for +1 and default (unchanged) slot count.


bemerybmw:
Taken over in: https://github.com/eclipse-score/communication/pull/758

PR #448, opened 2026-05-18
Create a KPI dashboard for quality numbers

ahmed0mousa:
Add a nightly CI pipeline that runs three quality jobs in parallel (coverage, CodeQL, and clang-tidy) and publishes all results to GitHub Pages under `latest/quality/`. A Jinja2-based dashboard aggregates the findings into a single page with KPI trend tracking across runs. The Sphinx documentation is extended with a dedicated quality reports page and a version switcher navbar, and on every push to main the docs automatically pull the latest nightly KPI numbers so they stay current without waiting for another nightly run.

### DIFF
--- a/.github/workflows/automated_release.yml
+++ b/.github/workflows/automated_release.yml
@@ -27,6 +27,7 @@ on:
 env:
   ANDROID_HOME: ""
   ANDROID_SDK_ROOT: ""
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 jobs:
   create-draft-release:
--- a/.github/workflows/automated_release.yml
+++ b/.github/workflows/automated_release.yml
@@ -107,6 +107,16 @@ jobs:
       - name: Checkout repository
         uses: actions/checkout@v6.0.2
 
+      - name: Abort if coverage failed
+        # coverage_report.yml uses continue-on-error so its job status is always
+        # "success" — the real outcome is exposed via the conclusion output.
+        # Failing here propagates the failure to finalize-release (skipped) and
+        # delete-release-on-failure (triggered), aborting the release correctly.
+        if: needs.run-coverage-report.outputs.conclusion != 'success'
+        run: |
+          echo "::error::Coverage failed — aborting release. Draft will be deleted."
+          exit 1
+
       - name: Download coverage report artifact
         uses: actions/download-artifact@v4
         with:
--- a/.github/workflows/coverage_report.yml
+++ b/.github/workflows/coverage_report.yml
@@ -48,11 +48,7 @@ jobs:
         with:
           level: 4
 
-      - name: Install lcov
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y lcov
-
+      # TODO(follow-up PR): align setup-bazel version with the other workflows.
       - name: Setup Bazel with shared caching
         uses: bazel-contrib/setup-bazel@0.18.0
         with:
@@ -70,12 +66,12 @@ jobs:
         run: |
           bazel coverage //... --build_tests_only
 
+      # TODO(follow-up PR): merge generate_coverage_html and create_coverage_archive
+      # into a single sh_binary with a --archive flag.
       - name: Generate HTML Coverage Report
         if: steps.run-coverage.outcome == 'success'
         run: |
-          bash quality/scripts/generate_coverage_html.sh \
-            "$(bazel info output_path)/_coverage/_coverage_report.dat" \
-            cpp_coverage
+          bazel run //quality/coverage:generate_coverage_html -- cpp_coverage
 
       - name: Create archive of test report
         if: steps.run-coverage.outcome == 'success'
@@ -94,10 +90,12 @@ jobs:
 
       - name: Set artifact name
         id: set-artifact-name
+        if: steps.run-coverage.outcome == 'success'
         run: |
           echo "artifact-name=${{ github.event.repository.name }}_coverage_report_${{ github.sha }}" >> $GITHUB_OUTPUT
 
       - name: Upload coverage artifacts
+        if: steps.run-coverage.outcome == 'success'
         uses: actions/upload-artifact@v4
         with:
           name: ${{ steps.set-artifact-name.outputs.artifact-name }}
--- a/.github/workflows/coverage_report.yml
+++ b/.github/workflows/coverage_report.yml
@@ -18,6 +18,9 @@ on:
       artifact-name:
         description: 'Name of the coverage report artifact'
         value: ${{ jobs.coverage_report.outputs.artifact-name }}
+      conclusion:
+        description: 'Job conclusion: success or failure'
+        value: ${{ jobs.coverage_report.outputs.conclusion }}
 
 permissions:
   contents: read
@@ -28,11 +31,13 @@ concurrency:
 env:
   ANDROID_HOME: ""
   ANDROID_SDK_ROOT: ""
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 jobs:
   coverage_report:
     runs-on: ubuntu-24.04
     outputs:
       artifact-name: ${{ steps.set-artifact-name.outputs.artifact-name }}
+      conclusion: ${{ steps.set-conclusion.outputs.conclusion }}
 
     steps:
       - name: Checkout Repository
@@ -54,33 +59,38 @@ jobs:
           bazelisk-cache: true
           disk-cache: "coverage_report"
           repository-cache: true
-          cache-save: ${{ github.event_name == 'merge_group' }}
+          cache-save: true
 
       - name: Allow linux-sandbox
         uses: ./actions/unblock_user_namespace_for_linux_sandbox
 
       - name: Run Unit Test with Coverage for C++
+        id: run-coverage
+        continue-on-error: true
         run: |
           bazel coverage //... --build_tests_only
 
       - name: Generate HTML Coverage Report
-        # FIXME: "--ignore-errors category,inconsistent" is a workaround to cope with gcov messing up hit counts because of internal data races
+        if: steps.run-coverage.outcome == 'success'
         run: |
-          genhtml "$(bazel info output_path)/_coverage/_coverage_report.dat" \
-            -o=cpp_coverage \
-            --show-details \
-            --legend \
-            --function-coverage \
-            --branch-coverage \
-            --ignore-errors category,inconsistent
+          bash quality/scripts/generate_coverage_html.sh \
+            "$(bazel info output_path)/_coverage/_coverage_report.dat" \
+            cpp_coverage
 
       - name: Create archive of test report
+        if: steps.run-coverage.outcome == 'success'
         run: |
-          mkdir -p artifacts
-          find bazel-testlogs/score/ -name 'test.xml' -print0 | xargs -0 -I{} cp --parents {} artifacts/
-          cp -r cpp_coverage artifacts/
-          zip -r ${{ github.event.repository.name }}_coverage_report_${{ github.sha }}.zip artifacts/
-        shell: bash
+          bash quality/scripts/create_coverage_archive.sh \
+            "${{ github.event.repository.name }}_coverage_report_${{ github.sha }}"
+
+      - name: Set conclusion
+        id: set-conclusion
+        run: |
+          if [[ "${{ steps.run-coverage.outcome }}" == "success" ]]; then
+            echo "conclusion=success" >> $GITHUB_OUTPUT
+          else
+            echo "conclusion=failure" >> $GITHUB_OUTPUT
+          fi
 
       - name: Set artifact name
         id: set-artifact-name
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -24,11 +24,15 @@ on:
       - 'docs/**'
       - '.github/workflows/deploy_docs.yml'
   workflow_dispatch:
+  workflow_run:
+    workflows: ["Nightly Quality Jobs"]
+    types: [completed]
 
 permissions:
   contents: write  # needed for uploading release assets
   pages: write
   id-token: write
+  actions: read    # needed to download artifacts from the nightly workflow_run
 
 concurrency:
   group: "pages"
@@ -37,6 +41,7 @@ concurrency:
 env:
   ANDROID_HOME: ""
   ANDROID_SDK_ROOT: ""
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
 
 jobs:
   build-docs:
@@ -74,9 +79,6 @@ jobs:
           fi
 
       - name: Build Sphinx documentation
-        env:
-          DOCS_BASE_URL: "https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}"
-          DOCS_VERSION: ${{ steps.version.outputs.version }}
         run: |
           bazel build //docs/sphinx:sphinx_doc
 
@@ -119,22 +121,13 @@ jobs:
           gh release create "${VERSION}" --draft=false --notes "Release ${VERSION}" 2>/dev/null || true
           gh release upload "${VERSION}" "docs-${VERSION}.tar.gz" --clobber
 
-      - name: Assemble publish tree
+      - name: Download previously released versions
         if: github.event_name != 'pull_request'
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           VERSION="${{ steps.version.outputs.version }}"
-          IS_TAG="${{ steps.version.outputs.is_tag }}"
-          REPO_URL="https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}"
-
           mkdir -p publish
-
-          # Place current build
-          mkdir -p "publish/${VERSION}"
-          cp -a docs_output/. "publish/${VERSION}/"
-
-          # Download previously released versions
           for tag in $(gh api "repos/${{ github.repository }}/releases" --paginate --jq '.[].tag_name' | grep '^v'); do
             if [ "${tag}" = "${VERSION}" ]; then
               continue
@@ -149,42 +142,28 @@ jobs:
             fi
           done
 
-          # stable = newest tagged version that has docs
-          if [[ "${IS_TAG}" == "true" ]]; then
-            rm -rf publish/stable
-            cp -a docs_output/. publish/stable/
-          else
-            NEWEST_TAG=$(find publish -maxdepth 1 -mindepth 1 -type d -name "v*" -printf '%f\n' | sort -rV | head -1)
-            if [ -n "${NEWEST_TAG}" ]; then
-              rm -rf publish/stable
-              cp -a "publish/${NEWEST_TAG}/." publish/stable/
-            fi
-          fi
+      - name: Assemble publish tree
+        if: github.event_name != 'pull_request'
+        run: |
+          bazel run //docs/sphinx/utils:assemble_publish_tree -- \
+            --version     "${{ steps.version.outputs.version }}" \
+            --is-tag      "${{ steps.version.outputs.is_tag }}" \
+            --docs-output docs_output \
+            --publish-dir publish \
+            --repo-url    "https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}" \
+            --shared-css  docs/sphinx/_static/css/version_flyout.css \
+            --shared-js   docs/sphinx/_static/js/version_flyout.js \
+            --root-index  docs/sphinx/_gh_pages/index.html
 
-          # Shared assets
-          mkdir -p publish/_shared/css publish/_shared/js
-          cp docs/sphinx/_static/css/version_flyout.css publish/_shared/css/
-          cp docs/sphinx/_static/js/version_flyout.js publish/_shared/js/
-
-          # Root index and Jekyll bypass
-          cp docs/sphinx/_gh_pages/index.html publish/index.html
-          touch publish/.nojekyll
-
-          # Generate switcher.json from actual directory contents
-          echo "[" > publish/switcher.json
-          echo "  {\"name\": \"latest\", \"version\": \"latest\", \"url\": \"${REPO_URL}/latest/\"}," >> publish/switcher.json
-
-          if [ -d "publish/stable" ]; then
-            echo "  {\"name\": \"stable\", \"version\": \"stable\", \"url\": \"${REPO_URL}/stable/\", \"preferred\": true}," >> publish/switcher.json
-          fi
-
-          for dir in $(find publish -maxdepth 1 -mindepth 1 -type d -name "v*" | sort -rV); do
-            ver=$(basename "$dir")
-            echo "  {\"name\": \"${ver}\", \"version\": \"${ver}\", \"url\": \"${REPO_URL}/${ver}/\"}," >> publish/switcher.json
-          done
-
-          sed -i '$ s/,$//' publish/switcher.json
-          echo "]" >> publish/switcher.json
+      - name: Bundle quality reports into latest/
+        if: github.event_name != 'pull_request'
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          bash quality/scripts/bundle_quality_reports.sh \
+            "${{ github.event_name }}" \
+            "${{ github.event.workflow_run.id }}" \
+            "${{ github.repository }}"
 
       - name: Upload Pages artifact
         if: github.event_name != 'pull_request'
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -151,8 +151,6 @@ jobs:
             --docs-output docs_output \
             --publish-dir publish \
             --repo-url    "https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}" \
-            --shared-css  docs/sphinx/_static/css/version_flyout.css \
-            --shared-js   docs/sphinx/_static/js/version_flyout.js \
             --root-index  docs/sphinx/_gh_pages/index.html
 
       - name: Bundle quality reports into latest/
--- a/.github/workflows/nightly_quality.yml
+++ b/.github/workflows/nightly_quality.yml
@@ -69,9 +69,8 @@ jobs:
         uses: bazel-contrib/setup-bazel@0.18.0
         with:
           bazelisk-cache: true
-          disk-cache: "quality_dashboard"
           repository-cache: true
-          cache-save: ${{ github.event_name == 'schedule' }}
+          cache-save: true
 
       - name: Allow linux-sandbox
         uses: ./actions/unblock_user_namespace_for_linux_sandbox
@@ -92,11 +91,10 @@ jobs:
           echo "coverage=$(map_conclusion '${{ needs.run-coverage.result }}')" >> $GITHUB_OUTPUT
 
       # ------------------------------------------------------------------
-      # Download artifacts (continue-on-error: artifacts may not exist if job failed)
+      # Download artifacts (only if the upstream job succeeded)
       # ------------------------------------------------------------------
       - name: Download coverage artifact
         if: needs.run-coverage.result == 'success'
-        continue-on-error: true
         uses: actions/download-artifact@v4
         with:
           name: ${{ needs.run-coverage.outputs.artifact-name }}
@@ -104,7 +102,6 @@ jobs:
 
       - name: Extract coverage HTML report
         if: needs.run-coverage.result == 'success'
-        continue-on-error: true
         run: |
           bash quality/scripts/extract_coverage_artifact.sh \
             /tmp/coverage_zip \
@@ -123,7 +120,6 @@ jobs:
           bazel run //quality/dashboard:generate_dashboard -- \
             --lcov     /tmp/coverage_zip/extracted/artifacts/coverage_report.dat \
             --html     _quality/index.html \
-            --history  _quality/data/quality_history.json \
             --github-summary
 
           echo "Dashboard generated. Contents of _quality/:"
@@ -142,6 +138,7 @@ jobs:
           retention-days: 7
 
       - name: Print dashboard URL
+        if: always()
         run: |
           OWNER="${{ github.repository_owner }}"
           REPO="${{ github.event.repository.name }}"
--- a/.github/workflows/nightly_quality.yml
+++ b/.github/workflows/nightly_quality.yml
@@ -1,0 +1,148 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+# Workflow: Run nightly coverage job and publish results to GitHub Pages.
+#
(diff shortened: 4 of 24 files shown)


.github/workflows/automated_release.yml
```diff
@@ -27,6 +27,7 @@ on:
 env:
   ANDROID_HOME: ""
   ANDROID_SDK_ROOT: ""
+  FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: true
```
castler, line 30: We should not have things different in the release workflow then in others. So either, we add this everywhere or nowhere.

Can you please also state in the commit message _why_ this change is needed?

quality/static_analysis/codeql_lint.py
```diff
@@ -60,13 +60,25 @@ def main():
...
+            # Finalize: compile tracing data into the queryable database.
+            subprocess.run(
+                f"{code_ql_path} database finalize -j=0 -- {database_location}",
+                shell=True, check=False)
 
             output_base = _get_bazel_info(source_root).get('output_path')
-            os.system(
-                f"{code_ql_path} database analyze -j=0 {database_location} --format=sarifv2.1.0 --output={output_base}/codeql.sarif")
-            os.system(
-                f"{code_ql_path} database analyze -j=0 {database_location} --format=csv --output={output_base}/codeql.csv")
```
castler, line 69: We should keep the CSV output for direct human readibility

quality/static_analysis/codeql_lint.py
```diff
@@ -60,13 +60,25 @@ def main():
...
+
+            # Analyze: run MISRA/AUTOSAR queries and produce SARIF.
+            # --ram:     cap at 5 GB to prevent swap thrashing on GitHub runners (7 GB total)
```
castler: We should not make this by default. We should add an extra option for github runners that we then only enable in the CI where these parameters are changed.

.github/workflows/coverage_report.yml
```diff
@@ -75,13 +83,27 @@ jobs:
...
 
+      - name: Set conclusion
+        id: set-conclusion
+        run: |
+          if [[ "${{ steps.run-coverage.outcome }}" == "success" ]]; then
+            echo "conclusion=success" >> $GITHUB_OUTPUT
+          else
+            echo "conclusion=failure" >> $GITHUB_OUTPUT
+          fi
+
```
castler, line 90: Why is this needed, can we try to remove this again please?

.github/workflows/clang_tidy.yml
```diff
@@ -0,0 +1,179 @@
+# *******************************************************************************
```
castler: Can we first just take care of the code coverage please to reduce the scope of the PR.

.github/workflows/nightly_quality.yml
```diff
@@ -0,0 +1,195 @@
...
+
+      # Install Jinja2 (already in requirements_lock.txt; just ensure it is
+      # available on the runner before running the dashboard script).
+      - name: Install dashboard dependencies
+        if: always() && steps.conclusions.outcome == 'success'
+        run: pip install jinja2 markupsafe
+
+      # Restore KPI history from the previous gh-pages deployment so the
+      # dashboard can show delta badges and trend sparklines across runs.
+      - name: Restore KPI history
```
castler: I thought we agreed that we do not want history at the moment?

.github/workflows/nightly_quality.yml
```diff
@@ -0,0 +1,195 @@
...
+            --github-summary
+
+          echo "Dashboard generated. Contents of _quality/:"
+          find _quality -type f | sort
+
+      # ------------------------------------------------------------------
+      # Deploy to GitHub Pages
+      # ------------------------------------------------------------------
+      - name: Deploy quality reports to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v4
```
castler: This way it is not integrated into our Sphinx build, maybe you can talk with Jochen about that

castler (commented):
Not yet fully through

.github/workflows/automated_release.yml
ahmed0mousa: Node 20 will be deprecated next month on GitHub Actions runners, I can add this to the commit message
https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

quality/static_analysis/codeql_lint.py
ahmed0mousa: reverted

quality/static_analysis/codeql_lint.py
ahmed0mousa: applied

.github/workflows/coverage_report.yml
ahmed0mousa: Because the coverage step has `continue-on-error: true`  meaning if `bazel coverage` fails, the job doesn't stop, it keeps running. Without the "Set conclusion" step, the caller `nightly_quality.yml` has no way to know whether coverage actually passed or failed; it only sees the job as success because `continue-on-error` suppresses the failure. So In `continue-on-error` hides the failure from GitHub's job status, "Set conclusion" exists to un-hide it for the dashboard. For a nightly quality pipeline, partial data is actually useful, if coverage fails at night, you want something to look at the next morning rather than an empty artifact.

.github/workflows/nightly_quality.yml
ahmed0mousa: yes that was deploys quality reports directly to `gh-pages` as a separate, uncoordinated publish. I made `nightly_quality.yml` upload quality reports as an artifact instead of deploying, then have `docs.yml` trigger on its completion and deploy everything in one shot.

.github/workflows/nightly_quality.yml
ahmed0mousa: removed

.github/workflows/clang_tidy.yml
ahmed0mousa: during the discussion we were always talking about three jobs not only coverage, and I already implemented that.
if you insist I can add a commit on top to remove them so I can just apply it in reverse on another PR

quality/dashboard/dashboard.html.j2
hoe-jo, line 1: Instead of this Jinja Template maybe generate a RST File which we can include in the sphinx build?

quality/static_analysis/codeql_lint.py
```diff
@@ -60,13 +89,32 @@ def main():
...
+            # Analyze: run MISRA/AUTOSAR queries and produce SARIF and CSV.
+            # Resource-cap flags (--ram, --timeout, -j) are injected via CLI
+            # args so they apply only when explicitly set (e.g. in CI).
+            # SARIF is consumed by the CI quality dashboard; CSV is kept for
+            # direct human readability.
+            subprocess.run(
+                f"{code_ql_path} database analyze"
+                f"{_analyze_flags}"
+                f" {database_location}"
+                f" --format=sarifv2.1.0 --output={output_base}/codeql.sarif",
```
hoe-jo: Should we maybe upload the sarif file if we have it already? Could be used for local development

.github/workflows/docs.yml
```diff
@@ -0,0 +1,221 @@
...
+        run: pip install -r requirements_lock.txt
+
+      # ------------------------------------------------------------------
+      # Build Sphinx HTML
+      # ------------------------------------------------------------------
+      - name: Build Sphinx documentation
+        env:
+          DOCS_VERSION: ${{ steps.vars.outputs.version }}
+          DOCS_BASE_URL: ${{ steps.vars.outputs.base_url }}
+        run: sphinx-build docs/sphinx _sphinx_output
```
hoe-jo: Is there a reason why you run without bazel?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,221 @@
...
+            <p>Redirecting to <a href="latest/">latest documentation</a>...</p>
+          </body>
+          </html>
+          REDIRECT
+
+          # Prevent Jekyll from processing the site
+          touch _deploy/.nojekyll
+
+          # Generate versions.json for the pydata-sphinx-theme version switcher
+          python3 docs/sphinx/utils/update_versions_json.py \
```
hoe-jo: mixing here again python and bazel

quality/dashboard/dashboard.html.j2
ahmed0mousa: That is technically possible but we will lose the visual dashboard (we agreed during the discussions to have a dashboard), all what we get is a summary tables like what we have here https://github.com/ahmed0mousa/communication/actions/runs/26016646593

.github/workflows/docs.yml
ahmed0mousa: Yes, the `sphinx_doc` Bazel target is marked `testonly = True`, meaning it's for CI testing (lint, link-checking), not for producing the deploy artifact. There are also two practical reasons for running sphinx-build directly:

Environment variables: `conf.py` reads `DOCS_VERSION` and `DOCS_BASE_URL` to configure the version switcher navbar. Passing env vars into Bazel's sandbox requires `--action_env` flags and is fragile in CI.

Output location: Bazel writes HTML under `bazel-out/…/sphinx_doc/…` making it awkward to reference in the subsequent deploy step. `sphinx-build … _sphinx_output` puts it exactly where we need it.

The two-phase approach in the workflow is intentional: Bazel handles the complex artifact generation (Doxygen XML, generated RST files), then sphinx-build is called directly for the final HTML output.

.github/workflows/docs.yml
ahmed0mousa: There's no Bazel target for `update_versions_json.py`, it's a plain utility script with no BUILD rule. So python3 is the correct call here. Also bazel's sandbox blocks all network access by default, so this can only run outside Bazel. It also needs `GH_TOKEN` injected from the runner environment, which Bazel would strip.

.github/workflows/docs.yml
```diff
@@ -0,0 +1,221 @@
+# *******************************************************************************
```
castler: What about the deploy_docs workflow?

quality/coverage/generate_coverage_html.sh
```diff
@@ -0,0 +1,53 @@
...
+# Arguments:
+#   lcov.dat    Path to the LCOV .dat coverage data file (required)
+#   output-dir  Directory to write the HTML report to (default: cpp_coverage)
+
+set -euo pipefail
+
+LCOV_DAT="${1:?Usage: $0 <lcov.dat> [output-dir]}"
+OUTPUT_DIR="${2:-cpp_coverage}"
+
+# NOTE: "--ignore-errors category,inconsistent"
```
ahmed0mousa, line 57: please share your thoughts here

quality/scripts/extract_coverage_artifact.sh
```diff
@@ -0,0 +1,33 @@
...
+# *******************************************************************************
+# Extracts the coverage zip artifact and copies HTML + LCOV data to the
+# quality output directory for the dashboard generator.
+#
+# Usage:
+#   bash quality/scripts/extract_coverage_artifact.sh [zip-dir] [output-dir]
+#
+# Arguments:
+#   zip-dir     Directory containing the downloaded coverage zip (default: /tmp/coverage_zip)
+#   output-dir  Directory to copy the HTML report into (default: ${GITHUB_WORKSPACE}/_quality/coverage)
```
hoe-jo, line 22: Could we maybe link it via the static part of the sphinx build? Meaning if we move it here:
docs/sphinx/_static/... 
then sphinx should automatically be able to pick it up

.github/workflows/coverage_report.yml
castler, line 53: We should use the same setup-bazel as in the other workflows (can be done in a follow up PR).

.github/workflows/coverage_report.yml
```diff
@@ -54,33 +59,38 @@ jobs:
...
       - name: Create archive of test report
+        if: steps.run-coverage.outcome == 'success'
         run: |
-          mkdir -p artifacts
-          find bazel-testlogs/score/ -name 'test.xml' -print0 | xargs -0 -I{} cp --parents {} artifacts/
-          cp -r cpp_coverage artifacts/
-          zip -r ${{ github.event.repository.name }}_coverage_report_${{ github.sha }}.zip artifacts/
-        shell: bash
+          bash quality/scripts/create_coverage_archive.sh \
+            "${{ github.event.repository.name }}_coverage_report_${{ github.sha }}"
```
castler: We should merge these into one script, and just have an cli option to create the archive. This will reduce the steps here in the CI.

Besides that, we should execute this via `bazel run`.

(can be done in follow up PR).

.github/workflows/coverage_report.yml
castler: Agreed, but then we have to also change this in the release workflow. Because there we expect to abort the release, if the coverage fails.

.github/workflows/coverage_report.yml
castler: We should remove this (in a follow up PR). As this is only needed to have `genhtml` in the system. The right way is to follow this approach:

https://github.com/eclipse-score/tooling/pull/238/changes#diff-4f29656c7bf17acdd9401fd9a35a68d19a163b1b1e146b8251003264b21ae5c5R43

https://github.com/eclipse-score/tooling/pull/238/changes#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR221

docs/sphinx/quality_reports.rst
```diff
@@ -0,0 +1,20 @@
...
+.. list-table::
+   :widths: 20 45 35
+   :header-rows: 1
+
+   * - Job
+     - Description
+     - Report
+   * - Coverage
+     - Line, function, and branch coverage from C++ unit tests (gcov/lcov)
+     - `Coverage report <quality/coverage/index.html>`_
```
castler, line 16: The problem will be that in local builds, this link will be empty IMHO

quality/dashboard/BUILD
```diff
@@ -0,0 +1,27 @@
...
+
+load("@rules_python//python:defs.bzl", "py_binary")
+load("@score_communication_pip//:requirements.bzl", "requirement")
+
+py_binary(
+    name = "generate_dashboard",
+    srcs = ["generate_dashboard.py"],
+    data = ["dashboard.html.j2"],
+    main = "generate_dashboard.py",
+    visibility = ["//visibility:public"],
```
castler: We should be clear about our public API, generating a dashboard should not be something that we expose publicly.

quality/dashboard/BUILD
```diff
@@ -0,0 +1,27 @@
...
+    deps = [
+        requirement("jinja2"),
+        requirement("markupsafe"),
```
castler: Please change `@`-Syntax.

quality/scripts/generate_coverage_html.sh
```diff
@@ -0,0 +1,53 @@
...
+# Known alternatives (not yet adopted):
+#   A) Add --local_test_jobs=1 to the bazel coverage command.
+#      Tests execute sequentially → no races, exact counts.
+#      Trade-off: longer nightly runtime (build parallelism is unaffected;
+#      only test execution is serialised).
+#   B) Switch to LLVM coverage (bazel coverage --experimental_generate_llvm_lcov).
+#      llvm-cov writes per-process .profraw files → no shared-file races by
+#      design, accurate counts, no slowdown.
+#      Trade-off: requires building with Clang — a broader toolchain change.
+genhtml "${LCOV_DAT}" \
```
castler: This will require a local installation of `genhtml`, please use the same approach as described here: 

https://github.com/eclipse-score/tooling/pull/238/changes#diff-4f29656c7bf17acdd9401fd9a35a68d19a163b1b1e146b8251003264b21ae5c5R43

https://github.com/eclipse-score/tooling/pull/238/changes#diff-6136fc12446089c3db7360e923203dd114b6a1466252e71667c6791c20fe6bdcR221

quality/coverage/generate_coverage_html.sh
castler: We keep it for now - but we should definitely try out the `--local_test_jobs=1`

docs/sphinx/conf.py
```diff
@@ -87,7 +81,7 @@
     'navbar_align': 'left',
     'navbar_start': ['navbar-logo'],
     'navbar_center': ['navbar-nav'],
-    'navbar_end': ['version-switcher', 'navbar-icon-links', 'theme-switcher'],
```
castler, line 90: Why is this removed? We still want to have the version switcher? Also the changes below should stay I guess?

docs/sphinx/conf.py
```diff
@@ -28,12 +28,6 @@
 author = 'Eclipse Foundation'
 release = '1.0.0'
 
-# GitHub Pages base URL (update org/repo as needed)
-GITHUB_PAGES_URL = os.environ.get(
-    'DOCS_BASE_URL',
-    'https://eclipse-score.github.io/communication'
-)
```
castler, line 35: I also see no argumentation and reason why this shall be removed?

quality/coverage/generate_coverage_html.sh
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env bash
```
castler, line 1: I would move this script into a folder `quality/coverage`.

quality/scripts/generate_coverage_html.sh
```diff
@@ -0,0 +1,53 @@
...
+# Usage:
+#   bash quality/scripts/generate_coverage_html.sh <lcov.dat> [output-dir]
+#
+# Arguments:
+#   lcov.dat    Path to the LCOV .dat coverage data file (required)
+#   output-dir  Directory to write the HTML report to (default: cpp_coverage)
+
+set -euo pipefail
+
+LCOV_DAT="${1:?Usage: $0 <lcov.dat> [output-dir]}"
```
castler: Since the dat file is always at the same location in our setup - I would not make this configurable but "hard-code" the path, the similar way how it is hard-coded in the action right now.

This will make it easier for people to use this

quality/scripts/generate_coverage_html.sh
```diff
@@ -0,0 +1,53 @@
...
+OUTPUT_DIR="${2:-cpp_coverage}"
```
castler: Maybe as default we should use some tmp-directory? like `/tmp/ß13248ß/cpp_coverage` via mktmpdir?

.github/workflows/deploy_docs.yml
```diff
@@ -74,9 +79,6 @@ jobs:
           fi
 
       - name: Build Sphinx documentation
-        env:
```
castler, line 77: I see no reason to change this?

.github/workflows/coverage_report.yml
castler, line 99: This will try to upload the created zip, but if line 71 fails, then the zip will not be created. So either we are missing in this part of the workflow the "if-successful" OR we have to remove the continue on build error

.github/workflows/coverage_report.yml
```diff
@@ -54,33 +59,38 @@ jobs:
...
+
+      - name: Set conclusion
+        id: set-conclusion
```
castler, line 83: We also need to adjust the release workflow to look into this.

.github/workflows/nightly_quality.yml
```diff
@@ -0,0 +1,148 @@
...
+      contents: read
+
+  # --------------------------------------------------------------------
+  # Collect results, build the dashboard, upload artifact for docs.yml
+  # --------------------------------------------------------------------
+  deploy-quality-reports:
+    needs: [run-coverage]
+    # Always run even if individual quality jobs fail, so the dashboard
+    # still reflects which jobs passed and which failed.
+    if: always()
```
castler, line 59: if we have this, do we then even need the "if"-statements and the continue on error in the single jobs?

.github/workflows/nightly_quality.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v6.0.2
+
+      - name: Setup Bazel with shared caching
+        uses: bazel-contrib/setup-bazel@0.18.0
+        with:
+          bazelisk-cache: true
+          disk-cache: "quality_dashboard"
+          repository-cache: true
+          cache-save: ${{ github.event_name == 'schedule' }}
```
castler: Since we execute this at night, I would save the cache. Thus: disk-cache: no, cache-save: no

.github/workflows/nightly_quality.yml
```diff
@@ -0,0 +1,148 @@
...
+      - name: Generate quality dashboard
+        if: always() && steps.conclusions.outcome == 'success'
+        run: |
+          # The LCOV .dat file is inside the extracted coverage zip at a known
+          # path; pass it unconditionally — generate_dashboard handles the
+          # case where the file is absent (returns empty coverage data).
+          bazel run //quality/dashboard:generate_dashboard -- \
+            --lcov     /tmp/coverage_zip/extracted/artifacts/coverage_report.dat \
+            --html     _quality/index.html \
+            --history  _quality/data/quality_history.json \
```
castler: I think you removed on other parts the quality_history.json, so this will not exist?

.github/workflows/nightly_quality.yml
```diff
@@ -0,0 +1,148 @@
...
+            --github-summary
+
+          echo "Dashboard generated. Contents of _quality/:"
+          find _quality -type f | sort
+
+      # ------------------------------------------------------------------
+      # Upload all quality reports as a single artifact for docs.yml to
+      # pick up and deploy as part of the unified Sphinx site.
+      # ------------------------------------------------------------------
```
castler, line 131: Do I understand that correctly, that this job here will be executed nightly, but it will not update the dashboard directly, it will rather "just" upload the quality part of the docs site to its job results and then the docs workflow will download it form there?

quality/scripts/bundle_quality_reports.sh
```diff
@@ -0,0 +1,54 @@
...
+
+mkdir -p "${PUBLISH_DIR}"
+
+if [[ "${EVENT_NAME}" == "workflow_run" ]]; then
+    # Triggered by nightly — download the fresh artifact directly.
+    RUN_ID="${WORKFLOW_RUN_ID}"
+else
+    # Other triggers — restore from the latest successful nightly run so
+    # quality reports are preserved in every Pages deployment.
+    RUN_ID=$(gh api \
```
castler, line 41: ok. this is good. I was concerned that we would lose the data in the next doc build otherwise.

.github/workflows/deploy_docs.yml
```diff
@@ -149,42 +142,28 @@ jobs:
...
+        run: |
+          bazel run //docs/sphinx/utils:assemble_publish_tree -- \
+            --version     "${{ steps.version.outputs.version }}" \
+            --is-tag      "${{ steps.version.outputs.is_tag }}" \
+            --docs-output docs_output \
+            --publish-dir publish \
+            --repo-url    "https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}" \
+            --shared-css  docs/sphinx/_static/css/version_flyout.css \
+            --shared-js   docs/sphinx/_static/js/version_flyout.js \
+            --root-index  docs/sphinx/_gh_pages/index.html
```
castler: Can we not inject this via bazel data dependencies and thus make this here way easier?

docs/sphinx/utils/BUILD
```diff
@@ -24,6 +24,13 @@ py_library(
     visibility = ["//docs/sphinx:__pkg__"],
 )
 
+py_binary(
+    name = "assemble_publish_tree",
+    srcs = ["assemble_publish_tree.py"],
+    main = "assemble_publish_tree.py",
+    visibility = ["//visibility:public"],
```
castler: This should not be publicly visible.

docs/sphinx/quality_reports.rst
castler, line 16: This will be a broken link on local builds, is it possible that we make here some kind of "if-statements" that we show another text in local builds? Or is this a tradeoff that we need. The problem is that we upload the docs also for each release and this means that this will be there broken also.

.github/workflows/deploy_docs.yml
```diff
@@ -74,9 +79,6 @@ jobs:
           fi
 
       - name: Build Sphinx documentation
-        env:
-          DOCS_BASE_URL: "https://${{ github.repository_owner }}.github.io/${{ github.event.repository.name }}"
-          DOCS_VERSION: ${{ steps.version.outputs.version }}
```
castler, line 79: I see no reason why this is removed?

castler (commented):
A very huge PR....I hope I figured out everyhting....

quality/scripts/extract_coverage_artifact.sh
ahmed0mousa: Short answer: not with Bazel's hermetic build model.

`_static` is evaluated by Bazel at analysis time via `glob(["_static/**/*"])` in the BUILD file. `glob()` runs when Bazel reads the BUILD file before any execution. Coverage HTML is generated at CI runtime, so it doesn't exist when Bazel analyses the graph. Bazel would simply not see it and not include it in the sandbox.

Even if you copied the coverage files into `_static/` before calling bazel build, the sandbox only sees declared inputs. Files copied in after `glob()` evaluated are invisible to it.

.github/workflows/docs.yml
ahmed0mousa: now we only have docs.yml

.github/workflows/coverage_report.yml
ahmed0mousa: to be done in a follow up PR

.github/workflows/coverage_report.yml
ahmed0mousa: to be done in a follow up PR

docs/sphinx/quality_reports.rst
ahmed0mousa: The relative links `quality/coverage/index.html` only exist after the CI deploy step copies them alongside the Sphinx output. Locally, sphinx-build won't have those files. A note makes this explicit.

docs/sphinx/conf.py
ahmed0mousa: The version switcher is still there, it's just a different implementation. The `pydata-sphinx-theme` built-in version-switcher widget was removed from `conf.py`, but (since we now only have) `deploy_docs.yml` injects a custom `version_flyout.js` into every HTML page post-build via sed: `JS_TAG="<script src=\"/${REPO_NAME}/_shared/js/version_flyout.js\"></script>"
find docs_output -name '*.html' -exec sed -i "s|</body>|${JS_TAG}\n</body>|" {} +`

Why this is actually better than the pydata built-in:

It's injected after the Sphinx build, so it works identically across all deployed versions, even old ones built before the switcher existed
The pydata built-in is baked into the HTML at build time and requires `DOCS_VERSION/DOCS_BASE_URL` to be correct old releases wouldn't show new versions

docs/sphinx/conf.py
ahmed0mousa: same https://github.com/eclipse-score/communication/pull/448#discussion_r3287434260

.github/workflows/deploy_docs.yml
ahmed0mousa: Since `conf.py` is the only place that would use them, passing them as env vars to the build step has no effect.
 https://github.com/eclipse-score/communication/pull/448#discussion_r3287434260

.github/workflows/coverage_report.yml
ahmed0mousa: The `Upload coverage artifacts` step is missing the same guard. added

.github/workflows/coverage_report.yml
ahmed0mousa: The release workflow is already correctly handling this. Tracing through the failure path:

`coverage_report.yml`: `run-coverage` fails -> Generate/Archive/Upload steps all skipped (guarded by `if`:) -> `Set conclusion` still runs -> `conclusion=failure` output set -> job exits with success status (from GitHub's perspective, `due to continue-on-error`)

`automated_release.yml` → `upload-coverage-report` starts:

- "Abort if coverage failed" gate step: `conclusion != 'success'` -> `exit 1` -> job status becomes failure
- "Download artifact" and "Upload to release" steps: automatically skipped

`delete-release-on-failure` has `if: ${{ failure() || cancelled() }}` and `upload-coverage-report` in its needs list -> triggers correctly and deletes the draft.

The gate step is intentionally a hard failure (not a job-level `if`: skip) because a skipped job wouldn't satisfy `failure()` in `delete-release-on-failure`. No changes needed.

.github/workflows/nightly_quality.yml
ahmed0mousa: Yes, that's exactly right. The flow is:

Nightly (`nightly_quality.yml`): runs coverage -> generates dashboard HTML -> uploads everything as the `nightly-quality-reports` artifact (a GitHub Actions artifact, not yet published anywhere)

Deploy docs (`deploy_docs.yml`): triggered by `workflow_run` after nightly completes -> downloads the `nightly-quality-reports` artifact -> bundles it into the Sphinx publish tree alongside the versioned docs -> deploys the whole thing to GitHub Pages

So the quality reports only become publicly visible after `deploy_docs.yml` finishes. That's also why the "Print dashboard URL" step at the bottom says "once `docs.yml` completes".

docs/sphinx/quality_reports.rst
castler: The question is, if we have a way to inject this either via a flag or via a environtaml variable or something like this. That in a local build it says: this is a local build execute the following command to get the coverage...and in a release build it says: the coverage was uploaded here...Something like this.

ahmed0mousa: This is genuinely a tradeoff. Three realistic options:

Option A : Accept it, strengthen the note (simplest)
Keep the relative link for the deployed site, update the note to explicitly mention it also doesn't work in versioned release docs - only on `latest`.

Option B : Absolute URL
Replace the relative link with `https://eclipse-score.github.io/communication/latest/quality/coverage/index.html`. Works from anywhere (local build, release archive, deployed site), but always points to `latest` regardless of which version of the docs you're reading, and hardcodes the repo URL.

Option C : `.. only::` directive
Set a Sphinx tag (e.g. `deployed`) via `conf.py` reading an env var set in `deploy_docs.yml`. Show the relative link only when the tag is present; show a note otherwise. Cleanest, but requires wiring an env var through the build.

Given that quality reports are only ever published under `latest/` (not under versioned tags), option B's "always points to latest" is actually correct behaviour. Option C adds build complexity for the same end result.

Which would you prefer? change could be done in a follow up

.github/workflows/deploy_docs.yml
ahmed0mousa: same https://github.com/eclipse-score/communication/pull/448#discussion_r3269653582

castler (approved):

quality/coverage/generate_coverage_html.sh
ahmed0mousa: I left for now

docs/sphinx/quality_reports.rst
ahmed0mousa: I find a simpler solution which is a binary flag with DOCS_VERSION + DOCS_BASE_URL so the template can handle all three cases (latest/, Versioned release, and local build) differently. applied in https://github.com/eclipse-score/communication/pull/472

.github/workflows/coverage_report.yml
ahmed0mousa: fixed in https://github.com/eclipse-score/communication/pull/472

.github/workflows/coverage_report.yml
ahmed0mousa: why? its still the same behaviour:
1. run-coverage has continue-on-error: true -> job never fails, so set-conclusion always runs
2. set-conclusion sets conclusion=failure when coverage failed
3. upload-coverage-report gate step checks needs.run-coverage-report.outputs.conclusion != 'success' -> exits 1 -> job fails
4. delete-release-on-failure triggers on failure() -> draft deleted

PR #454, opened 2026-05-20
mw/com: Use pointer instead of copy for EventDataControlComposite

ahmed0mousa:
Changed SampleAllocateePtr to hold pointer to SkeletonEvent's EventDataControlComposite
Added SampleAllocateeTracker/Guard to track outstanding allocations
SkeletonEvent terminates if destroyed while SampleAllocateePtr instances exist
Update unit tests

Fixes state synchronization issues and prevents dangling pointer access.

### DIFF
--- a/score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
+++ b/score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
@@ -20,7 +20,6 @@
 #include <functional>
 #include <limits>
 #include <memory>
-#include <optional>
 #include <utility>
 
 namespace score::mw::com::impl::lola
@@ -36,7 +35,7 @@ namespace score::mw::com::impl::lola
 template <typename SampleType>
 class SampleAllocateePtr
 {
-    // friends to the View wrapper; is used to acces managed obj and event_data_control_
+    // friends to the View wrapper; is used to access managed obj and event_data_control_ptr_
     template <typename T>
     // coverity[autosar_cpp14_a11_3_1_violation] see above
     friend class SampleAllocateePtrView;
@@ -71,7 +70,7 @@ class SampleAllocateePtr
     explicit SampleAllocateePtr(std::nullptr_t /* ptr */) noexcept
         : managed_object_{nullptr},
           event_slot_index_{kUninitialisedEventSlotIndex},
-          event_data_control_{std::nullopt},
+          event_data_control_ptr_{nullptr},
           consumer_event_data_control_local_view_{nullptr}
     {
     }
@@ -81,12 +80,12 @@ class SampleAllocateePtr
     /// \param event_data_ctrl event data control structure, which manages the underlying event/sample in shmem.
     /// \param slot_index index of event slot
     SampleAllocateePtr(pointer ptr,
-                       const EventDataControlComposite<>& event_data_ctrl,
+                       EventDataControlComposite<>& event_data_ctrl,
                        ConsumerEventDataControlLocalView<>& consumer_event_data_control_local_view,
                        const SlotIndexType slot_index) noexcept
         : managed_object_{ptr},
           event_slot_index_{slot_index},
-          event_data_control_{event_data_ctrl},
+          event_data_control_ptr_{&event_data_ctrl},
           consumer_event_data_control_local_view_{&consumer_event_data_control_local_view}
     {
     }
@@ -128,7 +127,7 @@ class SampleAllocateePtr
 
         swap(this->managed_object_, other.managed_object_);
         swap(this->event_slot_index_, other.event_slot_index_);
-        swap(this->event_data_control_, other.event_data_control_);
+        swap(this->event_data_control_ptr_, other.event_data_control_ptr_);
         swap(this->consumer_event_data_control_local_view_, other.consumer_event_data_control_local_view_);
     }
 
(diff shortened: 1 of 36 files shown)


score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
```diff
@@ -185,18 +184,23 @@ class SampleAllocateePtr
...
 
     pointer managed_object_;
     SlotIndexType event_slot_index_;
-    std::optional<EventDataControlComposite<>> event_data_control_;
+    /// \brief Pointer to EventDataControlComposite owned by the SkeletonEvent.
+    /// \details This is a non-owning pointer. The SkeletonEvent owns the EventDataControlComposite and must outlive
+    /// any SampleAllocateePtr instances created from it. The SampleAllocateeTracker ensures this by terminating
+    /// if SampleAllocateePtr instances outlive the SkeletonEvent. Can only be a nullptr if SampleAllocateePtr is
+    /// default constructed.
+    EventDataControlComposite<>* event_data_control_ptr_;
```
bemerybmw, line 204: should have static asserts here that EventDataControlComposite is not movable / copyable.

score/mw/com/impl/sample_allocatee_guard.h
```diff
@@ -0,0 +1,68 @@
...
+/// This class is move-only to ensure each allocation is tracked exactly once.
+class SampleAllocateeGuard final
+{
+  public:
+    /// \brief Default constructs a null guard that does not track any allocation.
+    ///
+    /// \details In production code, guards must be obtained via SampleAllocateeTracker::Allocate() so that
+    /// every allocation is tracked. Direct default construction produces a no-op guard and is therefore
+    /// only intended for use in tests, where a real SampleAllocateeTracker is not available.
+    SampleAllocateeGuard() = default;
```
bemerybmw, line 33: This class should use "score/scope_exit/scope_exit.h" for handling deallocation on destruction so that we're not reimplementing the behaviour here. You can also then remove the move tests if you add a static assert that ScopeExit is used. See "score/mw/com/impl/bindings/lola/transaction_log_registration_guard.h" and the test in this pr as an example: https://github.com/eclipse-score/communication/pull/268

score/mw/com/impl/plumbing/rust/BUILD
```diff
@@ -31,7 +31,10 @@ rust_library(
     visibility = [
         "//score/mw/com:__subpackages__",
     ],
-    deps = [":common_rs"],
+    deps = [
```
bemerybmw: Please ask @bharatGoswami8 or @darkwisebear to double check the rust changes

score/mw/com/impl/plumbing/sample_allocatee_ptr.h
```diff
@@ -346,13 +369,23 @@ void swap(SampleAllocateePtr<T>& lhs, SampleAllocateePtr<T>& rhs) noexcept
     lhs.Swap(rhs);
 }
 
-/// \brief Helper function to create a SampleAllocateePtr within the middleware (not to be used by the user)
+/// \brief Helper function to create a SampleAllocateePtr for use in tests (not to be used in production code).
+/// In production, guards must be obtained from SampleAllocateeTracker::Allocate(); use MakeSampleAllocateePtr
+/// with a guard instead.
 template <typename T>
 auto MakeSampleAllocateePtr(T ptr) noexcept -> SampleAllocateePtr<typename T::element_type>
```
bemerybmw, line 376: Can this now be renamed to MakeTestSampleAllocateePtr?

score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
```diff
@@ -185,18 +184,23 @@ class SampleAllocateePtr
...
     }
 
     pointer managed_object_;
     SlotIndexType event_slot_index_;
-    std::optional<EventDataControlComposite<>> event_data_control_;
+    /// \brief Pointer to EventDataControlComposite owned by the SkeletonEvent.
+    /// \details This is a non-owning pointer. The SkeletonEvent owns the EventDataControlComposite and must outlive
+    /// any SampleAllocateePtr instances created from it. The SampleAllocateeTracker ensures this by terminating
+    /// if SampleAllocateePtr instances outlive the SkeletonEvent. Can only be a nullptr if SampleAllocateePtr is
+    /// default constructed.
```
bemerybmw: this comment is different to the latest state of PR-17115. Can you please make sure that you ported the latest version?

bemerybmw (changes requested):

score/mw/com/impl/plumbing/rust/BUILD
```diff
@@ -31,7 +31,10 @@ rust_library(
...
+        ":common_rs",
+        "@score_communication_crate_index//:libc",
```
bharatGoswami8: Why `libc` , i did not see any dependency related to this on this target.

score/mw/com/impl/plumbing/rust/BUILD
```diff
@@ -56,6 +59,7 @@ rust_unit_test(
         ":common_rs",
         "//score/mw/com/impl/plumbing/rust/test_support:test_helper_size_ffi_rs",
         "//score/mw/com/impl/plumbing/rust/test_support:test_utils_rs",
+        "@score_communication_crate_index//:libc",
```
bharatGoswami8: same here ?

score/mw/com/impl/plumbing/rust/sample_allocatee_ptr.rs
bharatGoswami8, line 1: This change looks fine with current changes in c++ side.

score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
ahmed0mousa: we already have these asserts at the end of the file, in the current case the asserts evaluated whenever the header is included - guaranteed, regardless of whether `SampleAllocateePtr<T>` is ever instantiated.

score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
ahmed0mousa: I added the missed sentence.

score/mw/com/impl/plumbing/sample_allocatee_ptr.h
ahmed0mousa: Let's not add more changes to this PR please and I can check that in a follow up PR

score/mw/com/impl/sample_allocatee_guard.h
ahmed0mousa: `ScopeExit<score::cpp::callback<void()>>` is not bitwise-moveable (not trivially relocatable). it has this layout:
storage_  : which is callable stored inline here
wrapper_  : points INTO storage_ above

`wrapper_` is a self-referential pointer it points into the same struct's own `storage_`. `ScopeExit` wraps this in a move constructor that calls `relocate()` to fix up this pointer after a C++ move. That works correctly in C++.
The problem is that Rust does bitwise memcpy when it moves structs. When Rust moves a `SampleAllocateePtr<T>` across an FFI boundary (e.g. returning it from create_allocatee_ptr, assigning it into a MaybeUninit, etc.), it copies the raw bytes `wrapper_` still points to the old location in memory. When the C++ destructor later runs through that dangling pointer -> SIGSEGV.

the safe path is to make `SampleAllocateePtr<T>` heap-allocated at the FFI boundary. but anyway I would do that in another PR

score/mw/com/impl/plumbing/rust/BUILD
ahmed0mousa: I see Bharat already did

score/mw/com/impl/plumbing/rust/BUILD
ahmed0mousa: removed

score/mw/com/impl/plumbing/rust/BUILD
ahmed0mousa: removed

score/mw/com/impl/plumbing/rust/sample_allocatee_ptr.rs
```diff
@@ -1,5 +1,5 @@
 /********************************************************************************
- * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
```
bharatGoswami8, line 2: I don't think this need to be change ?

ahmed0mousa: as discussed in chat. I'll double check that. but this is not a blocker

bemerybmw (dismissed):

bemerybmw (approved):

PR #456, opened 2026-05-20
actions: Introduce action to close stale PRs

castler:
In order to keep a better overview, we try to close stale PRs. This way we know which things are still relevant and which are not.


limdor (approved):
LGTM, it will help to keep a better overview of still relevant PRs

PR #461, opened 2026-05-21
Use Graphviz dot for PlantUML rendering

AAmbuj:
Install graphviz in the docs deployment workflow and configure the Sphinx PlantUML setup to use dot when available. Keep a fallback when dot is missing and log how to install graphviz locally.

### DIFF
--- a/.github/workflows/deploy_docs.yml
+++ b/.github/workflows/deploy_docs.yml
@@ -73,6 +73,10 @@ jobs:
             echo "should_deploy=true" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Install Graphviz
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy_docs.yml
```diff
@@ -73,6 +73,10 @@ jobs:
             echo "should_deploy=true" >> "$GITHUB_OUTPUT"
           fi
 
+      - name: Install Graphviz
+        run: |
+          sudo apt-get update && sudo apt-get install -y graphviz
```
castler, line 78: IMHO this should be included in bazel no?

AAmbuj:
used another approch to done a job

PR #463, opened 2026-05-21
Implement integration tests for fields with set and notifier

sahithi-nukala:
- Update traits and traits tests for field get/set/notifier parameters.
- Restructure field initial value integration tests and add set and notifier scenarios

### DIFF
--- a/score/mw/com/test/fields/set_and_notifier/config/logging.json
+++ b/score/mw/com/test/fields/set_and_notifier/config/logging.json
@@ -1,0 +1,7 @@
+{
+  "appId": "FSAN",
+  "appDesc": "field_set_and_notifier",
+  "logLevel": "kDebug",
+  "logLevelThresholdConsole": "kDebug",
+  "logMode": "kRemote|kConsole"
+}
--- a/score/mw/com/test/fields/set_and_notifier/consumer.cpp
+++ b/score/mw/com/test/fields/set_and_notifier/consumer.cpp
@@ -1,0 +1,247 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#include "score/mw/com/test/fields/set_and_notifier/consumer.h"
+
+#include "score/mw/com/test/common_test_resources/fail_test.h"
+#include "score/mw/com/test/common_test_resources/process_synchronizer.h"
+#include "score/mw/com/test/common_test_resources/proxy_container.h"
+#include "score/mw/com/test/common_test_resources/proxy_event_receiver.h"
+#include "score/mw/com/test/common_test_resources/proxy_event_state_change_notifier.h"
+#include "score/mw/com/test/fields/set_and_notifier/initial_only_field.h"
+#include "score/mw/com/test/fields/set_and_notifier/set_and_notifier_enabled_field.h"
+#include "score/mw/com/test/fields/set_and_notifier/test_constants.h"
+#include "score/mw/com/types.h"
+
+#include <score/stop_token.hpp>
+
+#include "score/mw/com/test/common_test_resources/command_line_parser.h"
+
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <optional>
+#include <string_view>
+#include <utility>
+#include <vector>
+
+namespace score::mw::com::test
+{
+namespace
+{
+
+const std::string kNotifierConsumerDoneShmPath{"/fields_notifier_consumer_done"};
+const std::string kSetAndNotifierConsumerDoneShmPath{"/fields_set_and_notifier_consumer_done"};
+
+// InstanceSpecifier::Create can only fail if the provided string is invalid.
+// Verified once here; all test functions reuse this constant.
+const InstanceSpecifier kInstanceSpecifier = InstanceSpecifier::Create(std::string{kInstanceSpecifierString}).value();
+
+}  // namespace
+
+void run_notifier_consumer(const score::cpp::stop_token& stop_token)
+{
+    auto done_synchronizer_result = ProcessSynchronizer::Create(kNotifierConsumerDoneShmPath);
+    if (!done_synchronizer_result.has_value())
+    {
+        FailTest("Consumer: Could not create done ProcessSynchronizer");
+    }
+    ExitFunctionGuard exit_guard{[&done_synchronizer_result]() {
+        done_synchronizer_result->Notify();
+    }};
+
+    // Step 1. Find service and create proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    ProxyContainer<InitialOnlyProxy> proxy_container{};
+    proxy_container.CreateProxy(kInstanceSpecifier, "notifier");
+    auto& proxy = proxy_container.GetProxy();
+
+    // Step 2. Register unified receive handler that verifies samples arrive in the expected order
+    std::cout << "\nConsumer: Step 2 - Register unified receive handler" << std::endl;
+    constexpr auto kMaxNumSamples{2U};
+    std::size_t sample_index{0U};
+    auto value_callback = [&sample_index](const auto& sample_ptr) noexcept {
+        if (sample_index == 0U && *sample_ptr != kInitialValue)
+        {
+            FailTest("Consumer: Did not receive expected initial value ", kInitialValue, " in notifier scenario");
+        }
+        else if (sample_index == 1U && *sample_ptr != kUpdatedValue)
+        {
+            FailTest("Consumer: Did not receive expected updated value ", kUpdatedValue, " in notifier scenario");
+        }
+        ++sample_index;
+    };
+    ProxyEventReceiver field_receiver{proxy.initial_only_field, std::move(value_callback)};
+
+    // Step 3. Register state change handler
+    std::cout << "\nConsumer: Step 3 - Register state change handler" << std::endl;
+    ProxyEventStateChangeNotifier subscription_notifier{proxy.initial_only_field};
+
+    // Step 4. Subscribe to field with enough buffer for all samples the provider will send
+    std::cout << "\nConsumer: Step 4 - Subscribe to field" << std::endl;
+    std::ignore = proxy.initial_only_field.Subscribe(kMaxNumSamples);
+
+    // Step 5. Wait for subscription
+    std::cout << "\nConsumer: Step 5 - Wait for subscription" << std::endl;
+    if (!subscription_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    {
+        FailTest("Consumer: Subscription failed in notifier scenario");
+    }
+
+    // Step 6. Wait for all expected samples
+    std::cout << "\nConsumer: Step 6 - Wait for all expected samples" << std::endl;
+    if (!field_receiver.WaitForSamples(stop_token, kMaxNumSamples))
+    {
+        FailTest("Consumer: Did not receive all expected samples in notifier scenario");
+    }
+
+    proxy.initial_only_field.Unsubscribe();
+}
+
+void run_set_and_notifier_consumer(const score::cpp::stop_token& stop_token)
+{
+    auto process_synchronizer_result = ProcessSynchronizer::Create(kSetAndNotifierConsumerDoneShmPath);
+    if (!process_synchronizer_result.has_value())
+    {
+        FailTest("Consumer: Could not create ProcessSynchronizer");
+    }
+    ExitFunctionGuard process_synchronizer_guard{[&process_synchronizer_result]() {
+        process_synchronizer_result->Notify();
+    }};
+
+    // Step 1. Find service and create proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    ProxyContainer<SetAndNotifierEnabledProxy> proxy_container{};
+    proxy_container.CreateProxy(kInstanceSpecifier, "set_and_notifier");
+    auto& proxy = proxy_container.GetProxy();
+
+    // Step 2. Register unified receive handler that verifies samples arrive in the expected order
+    std::cout << "\nConsumer: Step 2 - Register unified receive handler" << std::endl;
+    constexpr auto kMaxNumSamples{1U};
+    std::size_t sample_index{0U};
+    auto value_callback = [&sample_index](const auto& sample_ptr) noexcept {
+        if (sample_index == 0U && *sample_ptr != kInitialValue)
+        {
+            FailTest("Consumer: Did not receive initial value ", kInitialValue, " in set scenario");
+        }
+        else if (sample_index == 1U && *sample_ptr != kSetRequestValue * 2 + 1)
+        {
+            FailTest("Consumer: Did not receive transformed value ", kSetRequestValue * 2 + 1, " after Set call");
+        }
+        ++sample_index;
+    };
+    ProxyEventReceiver field_receiver{proxy.set_and_notifier_enabled_field, std::move(value_callback)};
+
+    // Step 3. Register state change handler
+    std::cout << "\nConsumer: Step 3 - Register state change handler" << std::endl;
+    ProxyEventStateChangeNotifier subscription_notifier{proxy.set_and_notifier_enabled_field};
+
+    // Step 4. Subscribe to field with enough buffer for all samples the provider will send
+    std::cout << "\nConsumer: Step 4 - Subscribe to field" << std::endl;
+    std::ignore = proxy.set_and_notifier_enabled_field.Subscribe(kMaxNumSamples);
+
+    // Step 5. Wait for subscription and verify initial value
+    std::cout << "\nConsumer: Step 5 - Wait for subscription and verify initial value" << std::endl;
+    if (!subscription_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    {
+        FailTest("Consumer: Subscription failed in set scenario");
+    }
+    if (!field_receiver.WaitForSamples(stop_token, 1U))
+    {
+        FailTest("Consumer: Did not receive initial value ", kInitialValue, " in set scenario");
+    }
+
+    // Step 6. Set new field value and verify accepted value matches expected transformed value
+    std::cout << "\nConsumer: Step 6 - Set field value and verify accepted value" << std::endl;
+    const auto set_result = proxy.set_and_notifier_enabled_field.Set(kSetRequestValue);
+    if (!set_result.has_value())
+    {
+        FailTest("Consumer: Set call failed: ", set_result.error());
+    }
+    const std::int32_t accepted_value = *(set_result.value());
+
+    if (accepted_value != kSetRequestValue * 2 + 1)
+    {
+        FailTest(
+            "Consumer: Set accepted value mismatch. Expected ", kSetRequestValue * 2 + 1, " but got ", accepted_value);
+    }
+
+    // Step 7. Verify transformed value received via field notification
+    std::cout << "\nConsumer: Step 7 - Verify transformed value via field notification" << std::endl;
+    if (!field_receiver.WaitForSamples(stop_token, 1U))
+    {
+        FailTest("Consumer: Did not receive transformed value ", kSetRequestValue * 2 + 1, " after Set call");
+    }
+    proxy.set_and_notifier_enabled_field.Unsubscribe();
+}
+
+std::optional<ConsumerMode> ParseConsumerMode(std::string_view mode)
+{
+    if (mode == "notifier")
+    {
+        return ConsumerMode::kNotifier;
+    }
+    if (mode == "set_and_notifier")
+    {
+        return ConsumerMode::kSetAndNotifier;
+    }
+    // TODO: Add "get" mode consumer scenario coverage once getter-enabled field variant is introduced.
+    return std::nullopt;
+}
+
+ConsumerConfig ParseConsumerConfig(int argc, const char** argv)
+{
+    constexpr auto kModeArg = "mode";
+    constexpr auto kServiceInstanceManifestArg = "service-instance-manifest";
+
+    const std::vector<std::pair<std::string, std::string>> parameter_description_pairs{
+        {kModeArg, "Consumer mode: notifier or set_and_notifier"},
+        {kServiceInstanceManifestArg, "Path to the service instance manifest"},
+    };
+
+    const auto args = ParseCommandLineArguments(argc, argv, parameter_description_pairs);
+
+    const auto mode_result = GetValueIfProvided<std::string>(args, kModeArg);
+    if (!mode_result.has_value())
+    {
+        FailTest("Consumer: missing or invalid --", kModeArg, " argument");
+    }
+
+    const auto mode = ParseConsumerMode(mode_result.value());
+    if (!mode.has_value())
+    {
+        FailTest("Consumer: unsupported --", kModeArg, " value: ", mode_result.value());
+    }
+
+    const auto manifest_result = GetValueIfProvided<std::string>(args, kServiceInstanceManifestArg);
+    if (!manifest_result.has_value())
+    {
+        FailTest("Consumer: missing or invalid --", kServiceInstanceManifestArg, " argument");
+    }
+
+    return ConsumerConfig{mode.value(), manifest_result.value()};
+}
+
+void run_consumer(const score::cpp::stop_token& stop_token, ConsumerMode mode)
+{
+    switch (mode)
+    {
+        case ConsumerMode::kNotifier:
+            run_notifier_consumer(stop_token);
+            return;
+        case ConsumerMode::kSetAndNotifier:
+            run_set_and_notifier_consumer(stop_token);
+            return;
+    }
+}
+}  // namespace score::mw::com::test
--- a/score/mw/com/test/fields/set_and_notifier/consumer.h
+++ b/score/mw/com/test/fields/set_and_notifier/consumer.h
@@ -1,0 +1,46 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#ifndef SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_CONSUMER_H
+#define SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_CONSUMER_H
+
+#include <score/stop_token.hpp>
+
+#include <optional>
+#include <string>
+#include <string_view>
+
+namespace score::mw::com::test
+{
+
+enum class ConsumerMode
+{
+    kNotifier,
+    kSetAndNotifier,
+};
+
+struct ConsumerConfig
+{
+    ConsumerMode mode;
+    std::string manifest;
+};
+
+std::optional<ConsumerMode> ParseConsumerMode(std::string_view mode);
+
+ConsumerConfig ParseConsumerConfig(int argc, const char** argv);
+
+void run_consumer(const score::cpp::stop_token& stop_token, ConsumerMode mode);
+
+}  // namespace score::mw::com::test
+
+#endif  // SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_CONSUMER_H
--- a/score/mw/com/test/fields/set_and_notifier/initial_only_field.h
+++ b/score/mw/com/test/fields/set_and_notifier/initial_only_field.h
@@ -1,0 +1,37 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#ifndef SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_INITIAL_ONLY_FIELD_H
+#define SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_INITIAL_ONLY_FIELD_H
+
+#include "score/mw/com/types.h"
+#include <cstdint>
+
+namespace score::mw::com::test
+{
+
+template <typename T>
+class InitialOnlyInterface : public T::Base
+{
+  public:
+    using T::Base::Base;
+
+    typename T::template Field<std::int32_t, WithNotifier> initial_only_field{*this, "initial_only_field"};
+};
+
+using InitialOnlyProxy = score::mw::com::AsProxy<InitialOnlyInterface>;
+using InitialOnlySkeleton = score::mw::com::AsSkeleton<InitialOnlyInterface>;
+
+}  // namespace score::mw::com::test
+
+#endif  // SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_INITIAL_ONLY_FIELD_H
--- a/score/mw/com/test/fields/set_and_notifier/integration_test/test_set_and_notifier.py
+++ b/score/mw/com/test/fields/set_and_notifier/integration_test/test_set_and_notifier.py
@@ -1,0 +1,24 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+from test_fixture import consumer, provider, FieldScenario
+
+
+def test_field_set_and_notifier_value(target):
+    """Test field set exchange and accepted value propagation between provider and consumer."""
+    with provider(target, FieldScenario.SET_AND_NOTIFIER, "mw_com_config.json"):
+        with consumer(target, FieldScenario.SET_AND_NOTIFIER, "mw_com_config.json"):
+            pass
+
+
+# TODO: Add a dedicated get scenario test once getter-enabled field mode is available.
--- a/score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
+++ b/score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
@@ -1,0 +1,36 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#include "score/mw/com/runtime.h"
+#include "score/mw/com/test/common_test_resources/stop_token_sig_term_handler.h"
+#include "score/mw/com/test/fields/set_and_notifier/consumer.h"
+
+#include <cstdlib>
+#include <iostream>
+
+int main(int argc, const char** argv)
+{
+    const auto config = score::mw::com::test::ParseConsumerConfig(argc, argv);
+
+    score::mw::com::runtime::InitializeRuntime(score::mw::com::runtime::RuntimeConfiguration{config.manifest});
+
+    score::cpp::stop_source stop_source{};
+    const bool sig_term_handler_setup_success = score::mw::com::SetupStopTokenSigTermHandler(stop_source);
+    if (!sig_term_handler_setup_success)
+    {
+        std::cerr << "Unable to set signal handler for SIGINT and/or SIGTERM, cautiously continuing\n";
+    }
+
+    score::mw::com::test::run_consumer(stop_source.get_token(), config.mode);
+    return EXIT_SUCCESS;
+}
--- a/score/mw/com/test/fields/set_and_notifier/provider.cpp
+++ b/score/mw/com/test/fields/set_and_notifier/provider.cpp
@@ -1,0 +1,175 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ *******************************************************************************/
+
+#include "score/mw/com/test/fields/set_and_notifier/provider.h"
+
+#include "score/mw/com/test/common_test_resources/fail_test.h"
+#include "score/mw/com/test/common_test_resources/process_synchronizer.h"
+#include "score/mw/com/test/common_test_resources/skeleton_container.h"
+#include "score/mw/com/test/fields/set_and_notifier/initial_only_field.h"
+#include "score/mw/com/test/fields/set_and_notifier/set_and_notifier_enabled_field.h"
+#include "score/mw/com/test/fields/set_and_notifier/test_constants.h"
+
+#include <chrono>
+#include <iostream>
+#include <optional>
+#include <string_view>
+#include <utility>
+
+namespace score::mw::com::test
+{
+namespace
+{
+
+const std::string kNotifierConsumerDoneShmPath{"/fields_notifier_consumer_done"};
+const std::string kSetAndNotifierConsumerDoneShmPath{"/fields_set_and_notifier_consumer_done"};
+
+// InstanceSpecifier::Create can only fail if the provided string is invalid.
+// Verified once here; all test functions reuse this constant.
+const InstanceSpecifier kInstanceSpecifier = InstanceSpecifier::Create(std::string{kInstanceSpecifierString}).value();
+
+void run_notifier_provider(const score::cpp::stop_token& stop_token)
+{
+    // Step 1. Create process synchronizer
+    std::cout << "\nProvider: Step 1 - Create process synchronizer" << std::endl;
+    auto done_synchronizer_result = ProcessSynchronizer::Create(kNotifierConsumerDoneShmPath);
+    if (!done_synchronizer_result.has_value())
+    {
+        FailTest("Provider: Could not create done ProcessSynchronizer");
+    }
+
+    // Step 2. Create skeleton
+    std::cout << "\nProvider: Step 2 - Create skeleton" << std::endl;
+    SkeletonContainer<InitialOnlySkeleton> skeleton_container{};
+    skeleton_container.CreateSkeleton(kInstanceSpecifier, "notifier");
+
+    auto& service = skeleton_container.GetSkeleton();
+
+    // Step 3. Update initial field value
+    std::cout << "\nProvider: Step 3 - Update initial field value" << std::endl;
+    {
+        const auto update_result = service.initial_only_field.Update(kInitialValue);
+        if (!update_result.has_value())
+        {
+            FailTest("Provider: Unable to update initial field value: ", update_result.error());
+        }
+    }
+
+    // Step 4. Offer service
+    std::cout << "\nProvider: Step 4 - Offer service" << std::endl;
+    skeleton_container.OfferService("notifier");
+
+    // Step 5. Update field with updated value
+    std::cout << "\nProvider: Step 5 - Update field with updated value" << std::endl;
+    {
+        const auto update_result = service.initial_only_field.Update(kUpdatedValue);
+        if (!update_result.has_value())
+        {
+            FailTest("Provider: Unable to update field with updated value: ", update_result.error());
+        }
+    }
+
+    // Step 6. Wait until consumer signals done
+    std::cout << "\nProvider: Step 6 - Wait for consumer done notification" << std::endl;
+    if (!done_synchronizer_result->WaitWithAbort(stop_token))
+    {
+        FailTest("Provider: WaitWithAbort (done) was stopped by stop_token instead of notification");
+    }
+
+    // Step 7. Stop offering service
+    std::cout << "\nProvider: Step 7 - Stop offering service" << std::endl;
+    service.StopOfferService();
+}
+
+void run_set_and_notifier_provider(const score::cpp::stop_token& stop_token)
+{
+    // Step 1. Create process synchronizer
+    std::cout << "\nProvider: Step 1 - Create process synchronizer" << std::endl;
+    auto process_synchronizer_result = ProcessSynchronizer::Create(kSetAndNotifierConsumerDoneShmPath);
+    if (!process_synchronizer_result.has_value())
+    {
+        FailTest("Provider: Could not create ProcessSynchronizer");
+    }
+
+    // Step 2. Create skeleton
+    std::cout << "\nProvider: Step 2 - Create skeleton" << std::endl;
+    SkeletonContainer<SetAndNotifierEnabledSkeleton> skeleton_container{};
+    skeleton_container.CreateSkeleton(kInstanceSpecifier, "set_and_notifier");
+
+    auto& service = skeleton_container.GetSkeleton();
+
+    // Step 3. Register set handler
+    std::cout << "\nProvider: Step 3 - Register set handler" << std::endl;
+    const auto register_handler_result =
+        service.set_and_notifier_enabled_field.RegisterSetHandler([](std::int32_t& value) noexcept {
+            value = value * 2 + 1;
+        });
+    if (!register_handler_result.has_value())
+    {
+        FailTest("Provider: Unable to register set handler: ", register_handler_result.error());
+    }
+
+    // Step 4. Update initial field value
+    std::cout << "\nProvider: Step 4 - Update initial field value" << std::endl;
+    const auto update_result = service.set_and_notifier_enabled_field.Update(kInitialValue);
+    if (!update_result.has_value())
+    {
+        FailTest("Provider: Unable to update initial field value: ", update_result.error());
+    }
+
+    // Step 5. Offer service
+    std::cout << "\nProvider: Step 5 - Offer service" << std::endl;
+    skeleton_container.OfferService("set_and_notifier");
+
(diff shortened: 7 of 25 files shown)


score/mw/com/impl/traits.h
```diff
@@ -141,8 +141,8 @@ class ProxyTrait
 
     // Note : at the moment the SkeletonField::Get implementation is not in the branch which means the skeleton and
     // proxy side does not have same template parameters.
-    template <typename SampleType, bool EnableSet = false, bool EnableGet = false, bool EnableNotifier = false>
```
bemerybmw: Why did you reorder these?

bemerybmw: Please put this change in a separate commit with a good description of why you're changing traits / traits_test.

score/mw/com/impl/traits_test.cpp
```diff
@@ -67,10 +68,13 @@ class MyInterface : public InterfaceTrait::Base
   public:
     using InterfaceTrait::Base::Base;
 
+    static constexpr bool kEnableSet{std::is_same_v<InterfaceTrait, SkeletonTrait>};
```
bemerybmw: Why did you add this?

score/mw/com/test/common_test_resources/BUILD
```diff
@@ -15,6 +15,28 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
 load("@score_baselibs//score/language/safecpp:toolchain_features.bzl", "COMPILER_WARNING_FEATURES")
 load("//quality/unit_testing:unit_testing.bzl", "cc_unit_test")
 
+cc_library(
```
bemerybmw, line 18: Moving these should be a separate commit. You can read my comment here to better understand how commits should look / why we split them up: https://github.com/eclipse-score/communication/pull/385#discussion_r3271513806

bemerybmw (changes requested):

score/mw/com/impl/traits.h
sahithi-nukala: done. updated the traits file changes into a separate commit

score/mw/com/test/common_test_resources/BUILD
sahithi-nukala: done. pushed these changes with separate commit.

score/mw/com/impl/traits_test.cpp
sahithi-nukala: because this uses same templated interface for both proxy and skeleton, but the field should not have the same capabilities on both sides. so for this set needs to stay enabled on the skeleton side, while the proxy side should remain without set.

score/mw/com/test/field_initial_value/basic_acceptance_test/BUILD
```diff
@@ -0,0 +1,92 @@
+# *******************************************************************************
```
bemerybmw, line 1: Why do you have basic_acceptance_test, non_trivial_constructors / signature variations as folders within field_initial_value? I think that we should have a directory called fields which contains all field tests (like we have for methods https://github.com/sahithi-nukala/communication/tree/main/score/mw/com/test/methods). That fields directory should contain field_initial_value test and also the new tests that you add. But the new tests that you add should be appropriately named, you shouldn't just use the same directory structure that we have in the methods tests.

score/mw/com/test/fields/BUILD
```diff
@@ -0,0 +1,21 @@
...
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+test_suite(
```
bemerybmw: We decided not to use these component_test suites but we will instead run tests using wildcards e.g. `bazel test //score/mw/com/test/...`. So you can remove this.

score/mw/com/test/fields/set_and_notifier/main_client.cpp
bemerybmw, line 1: We usually consumer / provider instead of client / service. Could you please follow the same naming convention?

score/mw/com/test/fields/set_and_notifier/config/logging.json
bemerybmw, line 1: Please use the tests in this pr as a guide for new integration tests: https://github.com/eclipse-score/communication/pull/485

score/mw/com/test/fields/set_and_notifier/main_service.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include "score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.h"
+
+#include <vector>
+
+int main(int argc, const char** argv)
+{
+    using Parameters = score::mw::com::test::SctfTestRunner::RunParameters::Parameters;
+
+    const std::vector<Parameters> allowed_parameters{Parameters::CYCLE_TIME, Parameters::MODE};
+    score::mw::com::test::SctfTestRunner test_runner(argc, argv, allowed_parameters);
```
bemerybmw: SctfTestRunner is very old and we don't use it for new tests. Please look at how we do config parsing here: https://github.com/eclipse-score/communication/blob/brem_methods_sctfs/score/mw/com/test/methods/multiple_proxies/main_provider.cpp

score/mw/com/test/fields/set_and_notifier/main_service.cpp
```diff
@@ -0,0 +1,29 @@
...
+    const auto& run_parameters = test_runner.GetRunParameters();
```
bemerybmw: Need to register a sig term handler like in https://github.com/eclipse-score/communication/blob/brem_methods_sctfs/score/mw/com/test/methods/multiple_proxies/main_provider.cpp

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
```diff
@@ -0,0 +1,149 @@
...
+    }
+
+    const auto offer_result = service.OfferService();
+    if (!offer_result.has_value())
+    {
+        std::cerr << "Unable to offer SetEnabledSkeleton: " << offer_result.error() << ", bailing!\n";
+        return -17;
+    }
+
+    while (!stop_token.stop_requested())
```
bemerybmw: Use ProcessSynchronizer instead of a loop on the stop token like in https://github.com/eclipse-score/communication/blob/f6a14ef6f69482b404ec9392067fdd67c5126d51/score/mw/com/test/methods/basic_acceptance_test/provider.cpp#L43

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
```diff
@@ -0,0 +1,149 @@
...
+{
+    auto instance_specifier_result = InstanceSpecifier::Create(std::string{kInstanceSpecifierString});
+    if (!instance_specifier_result.has_value())
+    {
+        std::cerr << "Unable to create instance specifier, terminating\n";
+        return -13;
+    }
+    auto instance_specifier = std::move(instance_specifier_result).value();
+
+    auto service_result = SetEnabledSkeleton::Create(std::move(instance_specifier));
```
bemerybmw: Use SkeletonContainer instead of doing this yourself

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
```diff
@@ -0,0 +1,149 @@
...
+    const auto offer_result = service.OfferService();
+    if (!offer_result.has_value())
+    {
+        std::cerr << "Unable to offer SetEnabledSkeleton: " << offer_result.error() << ", bailing!\n";
+        return -17;
+    }
+
+    while (!stop_token.stop_requested())
+    {
+        std::this_thread::sleep_for(cycle_time);
```
bemerybmw: We don't need cycle_time to be passed in from python, we only want to pass in params that need to be changed when calling the binary from python (e.g. mode). Anyway, this won't be needed when using ProcessSynchronizer.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
```diff
@@ -0,0 +1,149 @@
...
+namespace
+{
+
+int run_notifier_service(const std::chrono::milliseconds& cycle_time, const score::cpp::stop_token& stop_token)
+{
+    auto instance_specifier_result = InstanceSpecifier::Create(std::string{kInstanceSpecifierString});
+    if (!instance_specifier_result.has_value())
+    {
+        std::cerr << "Unable to create instance specifier, terminating\n";
+        return -3;
```
bemerybmw: Instead of returning error codes and propagating this up through all the layers, we should instead use FailTest to immediately fail the test e.g. https://github.com/eclipse-score/communication/blob/f6a14ef6f69482b404ec9392067fdd67c5126d51/score/mw/com/test/methods/basic_acceptance_test/provider.cpp#L64

bemerybmw: Same applies for the consumer

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,300 @@
...
+    auto instance_specifier_result = InstanceSpecifier::Create(std::string{kInstanceSpecifierString});
+    if (!instance_specifier_result.has_value())
+    {
+        std::cerr << "Unable to create instance specifier, terminating\n";
+        return -7;
+    }
+    auto instance_specifier = std::move(instance_specifier_result).value();
+
+    std::promise<std::vector<InitialOnlyProxy::HandleType>> service_discovery_promise{};
+    auto service_discovery_future = service_discovery_promise.get_future();
```
bemerybmw: Use ProxyContainer

bemerybmw (changes requested):

score/mw/com/test/fields/set_and_notifier/fields_test_resources/test_datatype.h
```diff
@@ -0,0 +1,59 @@
...
+class InitialOnlyInterface : public T::Base
+{
+  public:
+    using T::Base::Base;
+
+    typename T::template Field<std::int32_t> test_field{*this, "test_field"};
+};
+
+template <typename T>
+class SetEnabledInterface : public T::Base
```
bemerybmw: We should have one class per file.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/test_datatype.h
```diff
@@ -0,0 +1,59 @@
...
+#define SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_FIELDS_TEST_RESOURCES_TEST_DATATYPE_H
+
+#include "score/mw/com/types.h"
+
+#include <cstdint>
+
+namespace score::mw::com::test
+{
+
+constexpr const char* const kInstanceSpecifierString = "test/fields/set_and_notifier";
```
bemerybmw: The shared constants should be in a different file, not with the interface classes.

score/mw/com/test/fields/set_and_notifier/integration_test/test_field_set_and_notifier.py
```diff
@@ -0,0 +1,39 @@
...
+    args = ["--num-retries", "20", "--backoff-time", "50", "--mode", mode]
+    return target.wrap_exec("bin/main_client", args, cwd="/opt/MainClientApp", wait_on_exit=True, **kwargs)
+
+
+def service(target, mode, **kwargs):
+    args = ["--cycle-time", "250", "--mode", mode]
+    return target.wrap_exec("bin/main_service", args, cwd="/opt/MainServiceApp", **kwargs)
+
+
+def test_field_notifier_initial_value(target):
```
bemerybmw: We should have different python files per test configuration (e.g. https://github.com/eclipse-score/communication/tree/f6a14ef6f69482b404ec9392067fdd67c5126d51/score/mw/com/test/methods/mixed_criticality/integration_test). The configurations needed are mentioned in the ticket.

score/mw/com/test/fields/BUILD
sahithi-nukala: Done

score/mw/com/test/fields/set_and_notifier/main_client.cpp
sahithi-nukala: Done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
sahithi-nukala: Done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
sahithi-nukala: Done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
sahithi-nukala: Done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/main_service.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/main_service.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/integration_test/test_field_set_and_notifier.py
sahithi-nukala: Separate files per configuration have been added for 2 cases from the ticket. and for the 4 configurations requiring getter support (test_get.py, test_get_and_notifier.py, test_set_and_get.py, test_set_get_and_notifier.py)

score/mw/com/test/fields/set_and_notifier/fields_test_resources/test_datatype.h
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/test_datatype.h
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/initial_only_field.h
```diff
@@ -0,0 +1,33 @@
+/*******************************************************************************
```
bemerybmw, line 1: all header files should have a corresponding cpp file, even if it's empty.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/initial_only_field.h
```diff
@@ -0,0 +1,33 @@
...
+namespace score::mw::com::test
+{
+
+template <typename T>
+class InitialOnlyInterface : public T::Base
+{
+  public:
+    using T::Base::Base;
+
+    typename T::template Field<std::int32_t> test_field{*this, "test_field"};
```
bemerybmw: name it something like initial_only_field so it doesn't have the same name as the other field.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/set_enabled_field.h
```diff
@@ -0,0 +1,33 @@
...
+namespace score::mw::com::test
+{
+
+template <typename T>
+class SetEnabledInterface : public T::Base
+{
+  public:
+    using T::Base::Base;
+
+    typename T::template Field<std::int32_t, false, true> test_field{*this, "test_field"};
```
bemerybmw: Name it set_enabled_field

score/mw/com/test/fields/set_and_notifier/fields_test_resources/test_datatype.h
```diff
@@ -14,38 +14,13 @@
...
-template <typename T>
-class SetEnabledInterface : public T::Base
-{
-  public:
-    using T::Base::Base;
-
-    typename T::template Field<std::int32_t, false, true> test_field{*this, "test_field"};
-};
-
 using InitialOnlyProxy = score::mw::com::AsProxy<InitialOnlyInterface>;
```
bemerybmw: This file should be removed and these aliases should go in the relevant interface files

score/mw/com/test/fields/field_initial_value/client.cpp
```diff
@@ -31,10 +33,19 @@ namespace
 
 constexpr auto kMaxNumSamples{1U};
 
+const std::string kInterprocessNotificationShmPath{"/field_initial_value_interprocess_notification"};
+
 int run_client(const std::size_t num_retries, const std::chrono::milliseconds retry_backoff_time)
```
bemerybmw: Also use FailTest in this file and run_client should return void?

score/mw/com/test/fields/set_and_notifier/fields_test_resources/initial_only_field.h
bemerybmw: We then would also need a differenet configuration file per test.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+namespace score::mw::com::test
+{
+namespace
+{
+
+constexpr auto kMaxNumSamples{1U};
+const std::string kInterprocessNotificationShmPath{"/fields_set_and_notifier_interprocess_notification"};
+
+template <typename... Args>
+void NotifyAndFail(ProcessSynchronizer& process_synchronizer, Args&&... args)
```
bemerybmw: This is already handled by the Exit FunctionGuard: https://github.com/eclipse-score/communication/blob/400e4aa3b5c33ddf6745693a739bdc728f4812fe/score/mw/com/test/methods/basic_acceptance_test/consumer.cpp#L105

How you have it now doesn't have RAII semantics in run_notifier_consumer so if we forget to call NotifyAndFail in one of the exit branches, the test will hang. Also if a function that we call calls FailTest the test will hang.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        }
+
+        --retries;
+        std::this_thread::sleep_for(retry_backoff_time);
+    }
+}
+
+void run_notifier_consumer(const std::size_t num_retries, const std::chrono::milliseconds retry_backoff_time)
+{
+    auto process_synchronizer_result = ProcessSynchronizer::Create(kInterprocessNotificationShmPath);
```
bemerybmw: Please add steps to the test with comments like in the tests I already sent you. The tests have been merged now so you can find them on master.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    if (!process_synchronizer_result.has_value())
+    {
+        FailTest("Consumer: Could not create ProcessSynchronizer");
+    }
+    auto& process_synchronizer = process_synchronizer_result.value();
+
+    auto instance_specifier_result = InstanceSpecifier::Create(std::string{kInstanceSpecifierString});
```
bemerybmw: InstanceSpecifier::Create can only fail if the provided string is invalid. So we just have to check this once when writing the test. So please follow the test that I already sent you and create a global constant to minimise the clutter in the test sequence.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    if (!instance_specifier_result.has_value())
+    {
+        NotifyAndFail(process_synchronizer, "Consumer: Unable to create instance specifier");
+    }
+    auto instance_specifier = std::move(instance_specifier_result).value();
+
+    ProxyContainer<InitialOnlyProxy> proxy_container{};
+    if (!proxy_container.CreateProxy(std::move(instance_specifier)))
```
bemerybmw: Please resolve merge conflicts. FailTest is now called within CreateProxy

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    {
+        NotifyAndFail(process_synchronizer, "Consumer: Unable to create InitialOnlyProxy");
+    }
+
+    auto& proxy = proxy_container.GetProxy();
+    std::ignore = proxy.test_field.Subscribe(kMaxNumSamples);
+    if (!WaitForSubscription(proxy.test_field, num_retries, retry_backoff_time))
```
bemerybmw: You can use the subscription state change handler instead of polling: https://github.com/eclipse-score/communication/blob/400e4aa3b5c33ddf6745693a739bdc728f4812fe/score/mw/com/impl/proxy_field_base.h#L96

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+template <typename FieldType>
+bool PollForValue(FieldType& field,
+                  const std::int32_t expected_value,
+                  std::size_t retries,
+                  const std::chrono::milliseconds retry_backoff_time)
+{
+    while (retries > 0)
+    {
+        score::cpp::optional<std::int32_t> received_value;
```
bemerybmw: I think it would make more sense to register an event handler with SetReceiveHandler so that we're not dependent on timing. With this code, the test could fail if the provider is slow to publish due to system load for example.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+
+    auto& proxy = proxy_container.GetProxy();
+    std::ignore = proxy.test_field.Subscribe(kMaxNumSamples);
+    if (!WaitForSubscription(proxy.test_field, num_retries, retry_backoff_time))
+    {
+        NotifyAndFail(process_synchronizer, "Consumer: Subscription failed in notifier scenario");
+    }
+
+    const bool initial_value_received = PollForValue(proxy.test_field, kInitialValue, num_retries, retry_backoff_time);
```
bemerybmw: For the notifier case, we don't want to just check the initial value. We also want to check that it receives newly published values.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        NotifyAndFail(process_synchronizer,
+                      "Consumer: Did not receive expected initial value ",
+                      kInitialValue,
+                      " in notifier scenario");
+    }
+
+    process_synchronizer.Notify();
+}
+
+void run_set_consumer(const std::size_t num_retries, const std::chrono::milliseconds retry_backoff_time)
```
bemerybmw: set always has to be with get or notifier. In this case it's with notifier so that should be mentioned here.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+    if (!initial_value_received)
+    {
+        NotifyAndFail(
+            process_synchronizer, "Consumer: Did not receive initial value ", kInitialValue, " in set scenario");
+    }
+
+    std::int32_t valid_requested_value = kSetValidValue;
+    std::int32_t valid_accepted_value{0};
+    std::string valid_set_error;
+    if (!SetWithRetries(proxy.test_field,
```
bemerybmw: Why are we retrying? If the set call fails, then we should fail the test?

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+        std::this_thread::sleep_for(retry_backoff_time);
+        --retries;
+    }
+
+    return false;
+}
+
+template <typename FieldType>
+bool SetWithRetries(FieldType& field,
+                    std::int32_t& requested_value,
```
bemerybmw: When you use a reference here, it makes it seem like it's an out parameter. It should either be a const ref or since an int is trivially copyable, it can be passed by value.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+                    std::int32_t& accepted_value,
```
bemerybmw: In general, std::optional is used exactly for this case in which you want to return a flag indicating success and a value if it succeeded. But the point of using FailTest is that we don't have to propagate errors and risk missing error handling in one of the layers, so we should just call FailTest instead of returning false.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,281 @@
...
+void run_consumer(const std::size_t num_retries,
+                  const std::chrono::milliseconds retry_backoff_time,
+                  const std::string& mode)
+{
+    if (mode == "notifier")
+    {
+        run_notifier_consumer(num_retries, retry_backoff_time);
+        return;
+    }
+    if (mode == "set")
```
bemerybmw: This should be set_and_notifier?

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    SkeletonContainer<SetEnabledSkeleton> skeleton_container{kInstanceSpecifierString};
+    if (!skeleton_container.CreateSkeleton())
+    {
+        FailTest("Provider: Unable to construct SetEnabledSkeleton");
+    }
+
+    auto& service = skeleton_container.GetSkeleton();
+    const auto register_handler_result = service.test_field.RegisterSetHandler([](std::int32_t& value) noexcept {
+        if (value > kSetAcceptedValue)
```
bemerybmw: I think that instead of doing clamping, we can simply have the set handler do some constant transformation, e.g. value * a + b; We can then check on proxy side that the value was transformed as expected. It means that we don't have to do 2 method calls.

score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
bemerybmw, line 1: mains should have main_ prefix

score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <vector>
+
+int main(int argc, const char** argv)
+{
+    constexpr auto kNumRetriesArg = "num-retries";
+    constexpr auto kBackoffTimeArg = "backoff-time";
+    constexpr auto kModeArg = "mode";
+    constexpr auto kServiceInstanceManifestArg = "service-instance-manifest";
+
+    const std::vector<std::pair<std::string, std::string>> parameter_description_pairs{
```
bemerybmw: please extract config parsing into a function like in the tests I sent you

score/mw/com/test/methods/methods_test_resources/BUILD
bemerybmw: If this is now used not just in fields, it should be moved to common resources folder which is accessible by both. e.g. https://github.com/eclipse-score/communication/tree/main/score/mw/com/test/common_test_resources

bemerybmw (changes requested):

score/mw/com/test/methods/methods_test_resources/BUILD
sahithi-nukala: moved the process_synchronizer target to common_test_resources

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_provider.cpp
sahithi-nukala: i replaced the clamping logic with a transformation ([value * 2 + 1] in the set handler. this way the consumer only needs to make one set call - it sends [kSetRequestValue] and checks that the received notification equals [kSetRequestValue * 2 + 1].

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
sahithi-nukala: i added **SetWithRetries** defensively thinking set might fail transiently due to timing issues between the consumer and provider. But you're right -> since the consumer already polls for the initial field value before calling set, the provider is fully up, and the set handler is already registered by that point. so, retrying doesn't make sense here. Fixed by replacing it with a direct set call that fails immediately if it doesn't succeed.

score/mw/com/test/fields/set_and_notifier/initial_only_field.h
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/initial_only_field.h
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/fields_test_resources/test_datatype.h
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/integration_test/test_set_and_notifier.py
```diff
@@ -0,0 +1,24 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+from test_fixture import consumer, provider
+
+
+def test_field_set_and_notifier_value(target):
+    """Test field set exchange and accepted value propagation between provider and consumer."""
+    with provider(target, "set_and_notifier", "mw_com_config.json"):
```
bemerybmw: Please use an enum instead of a string for the test scenario (e.g. https://github.com/eclipse-score/communication/blob/503876d4fe730732b6acaa93c83c9f032cb2221c/score/mw/com/test/move_semantics/skeleton_event/integration_test/test_fixture.py#L16)

score/mw/com/test/fields/set_and_notifier/consumer.h
bemerybmw, line 1: test resources are helpers for tests (e.g. which reduce code duplication, which perform some specific task which is used in multiple places). field_consumer.h/cpp and field_provider.h/cpp etc contain the actual test code so should not be in a test resources directory. 

Having checked what's in this folder, I don't think any of these qualify as test resources.

score/mw/com/test/fields/set_and_notifier/consumer.h
```diff
@@ -0,0 +1,28 @@
+/*******************************************************************************
```
bemerybmw, line 1: We are already clearly in a test for fields (based on directory structure). I would not add a field_ prefix to the file name.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,347 @@
...
+constexpr auto kMaxNumSamples{1U};
+const std::string kInterprocessNotificationShmPath{"/fields_set_and_notifier_interprocess_notification"};
+const std::string kNotifierConsumerReadyShmPath{"/fields_notifier_consumer_ready"};
+
+// InstanceSpecifier::Create can only fail if the provided string is invalid.
+// Verified once here; all test functions reuse this constant.
+const InstanceSpecifier kInstanceSpecifier = InstanceSpecifier::Create(std::string{kInstanceSpecifierString}).value();
+
+template <typename FieldType>
+bool WaitForSubscription(FieldType& field, std::size_t retries, const std::chrono::milliseconds retry_backoff_time)
```
bemerybmw: You should use ProxyEventStateChangeNotifier here.

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,347 @@
...
+    });
+    if (!success)
+    {
+        std::ignore = field.UnsetSubscriptionStateChangeHandler();
+    }
+    return success;
+}
+
+template <typename FieldType>
+bool WaitForValue(FieldType& field,
```
bemerybmw: You should use ProxyEventReceiver here

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,347 @@
...
+    }
+    ExitFunctionGuard exit_guard{[&consumer_ready_synchronizer_result, &done_synchronizer_result]() {
+        consumer_ready_synchronizer_result->Notify();
+        done_synchronizer_result->Notify();
+    }};
+
+    // Step 1. Find service and create proxy
+    std::cout << "\nConsumer: Step 1" << std::endl;
+    ProxyContainer<InitialOnlyProxy> proxy_container{};
+    proxy_container.CreateProxy(kInstanceSpecifier, "set_and_notifier");
```
bemerybmw: proxy_container.CreateProxy(kInstanceSpecifier, "notifier");

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,347 @@
...
+    std::cout << "\nConsumer: Step 4" << std::endl;
+    struct SyncState
+    {
+        std::mutex mutex{};
+        std::condition_variable cv{};
+        bool value_received{false};
+    };
+    SyncState sync_state;
+
+    const auto handler_result = proxy.initial_only_field.SetReceiveHandler([&proxy, &sync_state]() noexcept {
```
bemerybmw: you registered a receive handler in WaitForValue. Why are you registering another one here?

score/mw/com/test/fields/set_and_notifier/fields_test_resources/field_consumer.cpp
```diff
@@ -0,0 +1,347 @@
...
+    }
+    std::ignore = field.UnsetReceiveHandler();
+    return success;
+}
+
+}  // namespace
+
+void run_notifier_consumer(const std::size_t num_retries, const std::chrono::milliseconds retry_backoff_time)
+{
+    auto consumer_ready_synchronizer_result = ProcessSynchronizer::Create(kNotifierConsumerReadyShmPath);
```
bemerybmw: I think this function look very similar to https://github.com/eclipse-score/communication/blob/main/score/mw/com/test/move_semantics/skeleton_event/consumer.cpp

score/mw/com/test/fields/set_and_notifier/provider.cpp
```diff
@@ -0,0 +1,144 @@
...
+    }
+    skeleton_container.OfferService("set_and_notifier");
+
+    if (!consumer_ready_synchronizer_result->WaitWithAbort(stop_token))
+    {
+        FailTest("Provider: WaitWithAbort (consumer ready) was stopped by stop_token instead of notification");
+    }
+
+    {
+        const auto update_result = service.initial_only_field.Update(kUpdatedValue);
```
bemerybmw, line 75: We should send multiple events. Very similar to https://github.com/eclipse-score/communication/blob/503876d4fe730732b6acaa93c83c9f032cb2221c/score/mw/com/test/move_semantics/skeleton_event/provider.cpp#L50

score/mw/com/test/fields/set_and_notifier/provider.cpp
```diff
@@ -0,0 +1,144 @@
...
+const std::string kInterprocessNotificationShmPath{"/fields_set_and_notifier_interprocess_notification"};
+const std::string kNotifierConsumerReadyShmPath{"/fields_notifier_consumer_ready"};
+
+// InstanceSpecifier::Create can only fail if the provided string is invalid.
+// Verified once here; all test functions reuse this constant.
+const InstanceSpecifier kInstanceSpecifier = InstanceSpecifier::Create(std::string{kInstanceSpecifierString}).value();
+
+void run_notifier_provider(const score::cpp::stop_token& stop_token)
+{
+    auto consumer_ready_synchronizer_result = ProcessSynchronizer::Create(kNotifierConsumerReadyShmPath);
```
bemerybmw: please add steps comments / cout so that it's easy to follow the test in the logs. e.g. https://github.com/eclipse-score/communication/blob/503876d4fe730732b6acaa93c83c9f032cb2221c/score/mw/com/test/move_semantics/skeleton_event/provider.cpp#L50

bemerybmw (changes requested):

score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
```diff
@@ -123,15 +123,38 @@ Result<std::unique_ptr<ProxyMethodBinding>> ProxyMethodBindingFactoryImpl<Return
             const auto& lola_service_instance_deployment =
                 GetServiceInstanceDeploymentBinding<LolaServiceInstanceDeployment>(service_instance_deployment);
 
-            const auto method_it = lola_service_instance_deployment.methods_.find(method_name_str);
-            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(method_it != lola_service_instance_deployment.methods_.end(),
-                                                        "Could not find method deployment information for method");
-            if (!method_it->second.enabled_)
+            if ((method_type == MethodType::kGet) || (method_type == MethodType::kSet))
```
bemerybmw: Changes like this are completely unrelated to the integration tests so should be done in a separate pr.

score/mw/com/impl/traits_test.cpp
```diff
@@ -30,6 +30,7 @@
 #include <gtest/gtest.h>
 #include <memory>
 #include <optional>
+#include <type_traits>
```
bemerybmw, line 33: Please go through the commits and clean them up. Some issues with the commits: 
* "Update traits and traits tests for field get/set/notifier parameters" which just adds an include.
* "refactor: move ProcessSynchronizer to common_test_resources", "mw/com: move field_initial_value test package under fields" etc are unrelated to this pr. In general, it would be fine to have these in small prs, but since this is a large pr with a lot of changes and review comments, it would make sense to extract them into a separate pr which can be merged quickly.
* "mw/com: use event-driven SetReceiveHandler in field_consumer" modifies a file which appears to be completely removed in a later commit?
* "mw/com: flatten fields test resources and support field set/get bindings" does so many different things in it with basically no explanation in the commit message to say what / why. It also contains production code changes which should be done in a separate pr (and also need tests etc).

As a general point, it's very difficult to review this pr because there are so many commits, and many of the later commits completely change the earlier commits, so I can't really review commit by commit. In general, it's good to break up changes into smaller, atomic commits. But IMO, this is mainly for production code changes in which each commit does one thing. However, in this case, you're writing a completely new test from scratch. So having commits which build on each other can make it harder to review, since the commits are touching the same code. 

If I had just given a few small comments, then it might make sense to introduce a few commits with the fixes so that I could just check those commits and then approve. But since I made a lot of comments which required substantial changes to the test, I think it would be better to simply have one commit like "mw/com: Implement integration test for Field setter". That way I don't have to try and figure out which commits are related to the tests and which ones are other changes (e.g. moving other files around etc).

There's no exact science to this and maybe what I've suggested isn't the best and only way to do it so I'm also open to feedback. But the pain points that I mentioned have to be addressed somehow to make the review possible.

score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace
+{
+
+struct ConsumerConfig
+{
+    std::string mode;
```
bemerybmw: We should have a c++ enum here instead of a std::string. You can still pass it in via a string from python, but it's better to be explicit about the possible modes.

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,219 @@
...
+        done_synchronizer_result->Notify();
+    }};
+
+    // Step 1. Find service and create proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    ProxyContainer<InitialOnlyProxy> proxy_container{};
+    proxy_container.CreateProxy(kInstanceSpecifier, "notifier");
+    auto& proxy = proxy_container.GetProxy();
+
+    // Step 2. Register receive handler for the initial value
```
bemerybmw: Why have 2 ProxyEventReceivers?

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,219 @@
...
+    std::cout << "\nConsumer: Step 2 - Register receive handler for the initial value" << std::endl;
+    std::int32_t initial_value{};
+    auto initial_value_callback = [&initial_value](const auto& sample_ptr) noexcept {
+        initial_value = *sample_ptr;
```
bemerybmw: Instead of storing the value in initial_value, why not check the value within the callback? You can subscribe with a number of samples which is larger than the number of samples that the provider will send. You can then be sure you won't lose any samples. You can then call WaitForSamples once with that number and check the value is expected (e.g. if the provider increments the value, then you can check that each sample is the expected value).

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,219 @@
...
+    // Step 6. Register receive handler for updated value before notifying provider, to avoid missing the
+    // notification if the provider calls Update() before the handler is registered.
+    std::cout << "\nConsumer: Step 6 - Register receive handler for updated value" << std::endl;
+    std::int32_t updated_value{};
+    auto updated_value_callback = [&updated_value](const auto& sample_ptr) noexcept {
+        updated_value = *sample_ptr;
+    };
+    ProxyEventReceiver updated_value_receiver{proxy.initial_only_field, std::move(updated_value_callback)};
+
+    consumer_ready_synchronizer.Notify();
```
bemerybmw: You shouldn't need this synchronisation point if you use my suggestion above. You can just call WaitForSamples with the total number of samples that the provider is going to send.

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    // Step 6. Register receive handler for the transformed value before calling Set, to avoid missing the
+    // notification triggered by the Set call.
+    std::cout << "\nConsumer: Step 6 - Register receive handler for the transformed value" << std::endl;
+    std::int32_t transformed_value{};
+    auto transformed_value_callback = [&transformed_value](const auto& sample_ptr) noexcept {
+        transformed_value = *sample_ptr;
+    };
+    ProxyEventReceiver transformed_value_receiver{proxy.set_enabled_field, std::move(transformed_value_callback)};
```
bemerybmw: You should be able to reuse the same ProxyEventReceiver. You can call WaitForSamples(1) to get the initial value, then call set, then call WaitForSamples(1) again. You might need to modify ProxyEventReceiver to reset the score::concurrency::Notification

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,219 @@
...
+    // Step 7. Set new field value and verify accepted value matches expected transformed value
+    std::cout << "\nConsumer: Step 7 - Set field value and verify accepted value" << std::endl;
+    const auto set_result = proxy.set_enabled_field.Set(kSetRequestValue);
+    if (!set_result.has_value())
+    {
+        FailTest("Consumer: Set call failed: ", set_result.error());
+    }
+    const std::int32_t accepted_value = *(set_result.value());
+
+    if (accepted_value != kSetTransformedValue)
```
bemerybmw: Instead of checking kSetTransformedValue, why not check that accepted_value == kSetRequestValue * 2 + 1 and remove kSetTransformedValue? It's currently more brittle and if you change the value of kSetRequestValue, the test will fail. With my suggested change, all you have to ensure is that the algorithm here matches the algorithm on the provider side.

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,219 @@
...
+    {
+        FailTest("Consumer: Set accepted value mismatch. Expected ", kSetTransformedValue, " but got ", accepted_value);
+    }
+
+    // Step 8. Verify transformed value received via field notification
+    std::cout << "\nConsumer: Step 8 - Verify transformed value via field notification" << std::endl;
+    const bool transformed_value_received = transformed_value_receiver.WaitForSamples(stop_token, kMaxNumSamples) &&
+                                            transformed_value == kSetTransformedValue;
```
bemerybmw: Same with my comment above, why not check the received value directly in the callback provided to ProxyEventReceiver?

score/mw/com/test/fields/set_and_notifier/set_enabled_field.h
```diff
@@ -0,0 +1,37 @@
...
+namespace score::mw::com::test
+{
+
+template <typename T>
+class SetEnabledInterface : public T::Base
+{
+  public:
+    using T::Base::Base;
+
+    typename T::template Field<std::int32_t, WithSetter, WithNotifier> set_enabled_field{*this, "set_enabled_field"};
```
bemerybmw: This should be set_and_notifier_enabled_field? And also the filename etc updated?

score/mw/com/test/fields/set_and_notifier/provider.cpp
```diff
@@ -0,0 +1,176 @@
...
+#include <iostream>
+#include <string>
+#include <utility>
+
+namespace score::mw::com::test
+{
+namespace
+{
+
+const std::string kInterprocessNotificationShmPath{"/fields_set_and_notifier_interprocess_notification"};
```
bemerybmw: Some of the naming is inconsistent. e.g. the directory is called set_and_notifier and here the name is fields_**set_and_notifier**_interprocess_notification. If this directory is going to have notifier, setter and getter tests, then the naming should be updated.

score/mw/com/test/fields/set_and_notifier/provider.cpp
```diff
@@ -0,0 +1,176 @@
...
+    auto done_synchronizer_result = ProcessSynchronizer::Create(kInterprocessNotificationShmPath);
+    if (!done_synchronizer_result.has_value())
+    {
+        FailTest("Provider: Could not create done ProcessSynchronizer");
+    }
+
+    // Step 2. Create skeleton
+    std::cout << "\nProvider: Step 2 - Create skeleton" << std::endl;
+    SkeletonContainer<InitialOnlySkeleton> skeleton_container{};
+    skeleton_container.CreateSkeleton(kInstanceSpecifier, "set_and_notifier");
```
bemerybmw:
```suggestion
    skeleton_container.CreateSkeleton(kInstanceSpecifier, "notifier");
```

bemerybmw (changes requested):

score/mw/com/test/fields/set_and_notifier/provider.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/consumer.cpp
sahithi-nukala: Updated the ProxyEventReceiver per function(for run_notifier_consumer & run_set_and_notifier_consumer)

score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
sahithi-nukala: updated the mode with c++ enum

score/mw/com/test/fields/set_and_notifier/provider.cpp
sahithi-nukala: done

score/mw/com/test/fields/set_and_notifier/main_consumer.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include "score/mw/com/runtime.h"
+#include "score/mw/com/test/common_test_resources/stop_token_sig_term_handler.h"
+#include "score/mw/com/test/fields/set_and_notifier/consumer.h"
+
+#include <cstdlib>
+#include <iostream>
+
+int main(int argc, const char** argv)
+{
+    const auto config = score::mw::com::test::ParseConsumerConfig(argc, argv);
```
bemerybmw, line 23: move parse consurer config to main

score/mw/com/test/fields/set_and_notifier/initial_only_field.h
```diff
@@ -0,0 +1,37 @@
...
+#define SCORE_MW_COM_TEST_FIELDS_SET_AND_NOTIFIER_INITIAL_ONLY_FIELD_H
+
+#include "score/mw/com/types.h"
+#include <cstdint>
+
+namespace score::mw::com::test
+{
+
+template <typename T>
+class InitialOnlyInterface : public T::Base
```
bemerybmw, line 24: rename to notifier only field

score/mw/com/test/fields/set_and_notifier/provider.cpp
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    // Step 4. Offer service
+    std::cout << "\nProvider: Step 4 - Offer service" << std::endl;
+    skeleton_container.OfferService("notifier");
+
+    // Step 5. Update field with updated value
+    std::cout << "\nProvider: Step 5 - Update field with updated value" << std::endl;
+    {
+        const auto update_result = service.initial_only_field.Update(kUpdatedValue);
```
bemerybmw, line 75: Should send multiple https://github.com/eclipse-score/communication/blob/503876d4fe730732b6acaa93c83c9f032cb2221c/score/mw/com/test/move_semantics/skeleton_event/provider.cpp#L50

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -0,0 +1,247 @@
...
+    std::cout << "\nConsumer: Step 4 - Subscribe to field" << std::endl;
+    std::ignore = proxy.set_and_notifier_enabled_field.Subscribe(kMaxNumSamples);
+
+    // Step 5. Wait for subscription and verify initial value
+    std::cout << "\nConsumer: Step 5 - Wait for subscription and verify initial value" << std::endl;
+    if (!subscription_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    {
+        FailTest("Consumer: Subscription failed in set scenario");
+    }
+    if (!field_receiver.WaitForSamples(stop_token, 1U))
```
bemerybmw, line 159: The callback should be passed to waitforsamples

bemerybmw (dismissed):

bemerybmw (dismissed):

bemerybmw (approved):

Issue #242, opened 2026-03-26
Improvement: Make COM API support timeouts


artemsheinacn:
### What

@bharatGoswami8 @darkwisebear 
It seems COM API at the moment is not compatible with timeouts, for example using a subscription like this:
```
futures::executor::block_on(tokio::time::timeout(timeout.into(), scheduler_input.receive()))
```
with receive:
```
pub async fn receive(&mut self) -> MwComSample<'a, T> {
        let mut sample_container = self
            .subscription
            .receive(self.sample_container.take().expect("concurrent receive calls detected on the same subscription"), 1, 1)
            .await
            .unwrap();
        let result = sample_container.pop_front().unwrap();
        self.sample_container = Some(sample_container);
        result
    }
```
In this case subscription.receive call consumes the sample container and in case if timeout is triggered and the receive future is dropped there is no way to get the sample container back. I think it should provide some kind of into_inner method, so I can reuse the sample container, or better provide some kind of a helper type abstracting out a need to be dealing with sample container manually and providing a clean way to convert a subscription and a sample container into a stream. 

### Estimates for realization

8h

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

bharatGoswami8:
Hi @artemsheinacn , i am looking on this how we can support timeout and sampleContainer ownership.

bharatGoswami8:
@artemsheinacn , Added the support for cancel (timeout) `future` for async receive.

`cancellable_receive` is an extension of `receive` with an additional `cancellation` parameter, which allows the caller to specify a future that can be used to cancel the receive operation. It returns a future that resolves as soon as at least `new_samples` samples have been transferred from the communication buffer to the sample container or the `cancellation` future resolves.

please check `score/mw/com/example/com-api-example/basic-consumer-producer.rs` for more detail.

```
  fn cancellable_receive<'a>(
        &'a self,
        scratch: SampleContainer<Self::Sample<'a>>,
        new_samples: usize,
        max_samples: usize,
        cancellation: impl Future<Output = ()> + Send + 'static,
    ) -> impl Future<Output = (SampleContainer<Self::Sample<'a>>, Result<usize>)> + 'a;
```

PR #470, opened 2026-05-22
Rust::com async cancellable receive API and SampleContainer return handling

bharatGoswami8:
- Receive API now return the tuple of SampleContainer and Result.
- Receive API extended with timeout support for user to resolve / cancel the future.

https://github.com/eclipse-score/communication/issues/242

### DIFF
--- a/score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
@@ -865,19 +865,25 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
         )
     }
 
-    /// This method is extension of `receive` with an additional `timeout` parameter
+    /// This method is an extension of `receive` with an additional `cancellation` parameter
     /// which allows the caller to specify a future that can be used to cancel the receive operation.
     /// It returns a future that resolves as soon as at least `new_samples` samples have been transferred
-    /// from the communication buffer to the sample container or the `timeout` future resolves.
+    /// from the communication buffer to the sample container or the `cancellation` future resolves.
     ///
     /// # Parameters
     /// * `scratch` - Container for events from this subscription
     /// * `new_samples` - Minimum number of new events before resolution, 0 value is treated as error
     /// * `max_samples` - Maximum number of events that shall be received from the communication
     ///   buffer and transferred to the container, 0 value is treated as error
-    /// * `timeout` - Future that resolves when the receive operation should time out. If the timeout
-    ///   future resolves before the required number of samples are received, the receive operation is
-    ///   considered to have timed out.
+    /// * `cancellation` - Future that resolves when the receive operation should be cancelled. If the
+    ///   cancellation future resolves before the required number of samples are received, the receive
+    ///   operation is considered to have been cancelled.
+    ///   This can be used to implement various cancellation strategies, such as:
+    ///   - **Timeouts**: `tokio::time::sleep(Duration::from_secs(5))`
+    ///   - **Channel-based cancellation**: Waiting for a signal on a channel to indicate cancellation
+    ///   - **User-triggered cancellation**: Waiting for a keyboard interrupt or UI signal or any other user input to trigger cancellation
+    ///   - **External event cancellation**: Any other future that signals when cancellation is needed like shutdown signal, etc.
+    ///
     ///
     /// # Returns
     /// Future that resolves to `(SampleContainer<Self::Sample<'a>>, Result<usize>)`.
@@ -890,12 +896,12 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
     /// # Important Notes
     /// User can not concurrenly call `cancellable_receive` on the same subscription instance from
     /// multiple threads or tasks.
-    /// `timeout` must be `'static` because timeout futures (e.g. `tokio::time::sleep`) own all
+    /// `cancellation` must be `'static` because cancellation futures (e.g. `tokio::time::sleep`) own all
     /// their state and do not borrow anything from the caller's scope.
     /// And if you change to `'a` lifetime then it create lifetime bound not satisfied
     /// error from rust (see issue <https://github.com/rust-lang/rust/issues/100013> for more information)
     ///
-    /// The `timeout` future currently has `Output = ()` and returns nothing. However, it could be
+    /// The `cancellation` future currently has `Output = ()` and returns nothing. However, it could be
     /// enhanced in the future to return a value that would be propagated to the caller, as per the use-case.
     /// No use case for this enhancement has been identified yet, so it is deferred for later implementation.
     fn cancellable_receive<'a>(
@@ -903,7 +909,7 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
         scratch: SampleContainer<Self::Sample<'a>>,
         new_samples: usize,
         max_samples: usize,
-        timeout: impl Future<Output = ()> + Send + 'static,
+        cancellation: impl Future<Output = ()> + Send + 'static,
     ) -> impl Future<Output = (SampleContainer<Self::Sample<'a>>, Result<usize>)> + 'a;
 }
 
--- a/score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
@@ -858,7 +858,7 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
         scratch: SampleContainer<Self::Sample<'a>>,
         new_samples: usize,
         max_samples: usize,
-        timeout_future: impl Future<Output = ()> + Unpin + 'a,
+        timeout_future: impl Future<Output = ()> + Send + 'static,
     ) -> impl Future<Output = (SampleContainer<Self::Sample<'a>>, Result<()>)> + 'a;
 }
 
--- a/score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
@@ -232,7 +232,9 @@ impl InstanceSpecifier {
         }
 
         // Remove the single leading slash
-        let service_name = service_name.strip_prefix('/').unwrap();
+        let Some(service_name) = service_name.strip_prefix('/') else {
+            return false;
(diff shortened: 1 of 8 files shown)


tejaswinijoshi89 (approved):

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
```diff
@@ -844,12 +850,56 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
...
+    /// an 'Error' indicating the reason for failure of the receive operation.
+    ///
+    /// # Important Notes
+    /// User can not concurrenly call `receive_timeout` on the same subscription instance from
+    /// multiple threads or tasks.
+    /// `timeout` must be `'static` because timeout futures (e.g. `tokio::time::sleep`) own all
+    /// their state and do not borrow anything from the caller's scope.
+    /// And if you change to `'a` lifetime then it create lifetime bound not satisfied
+    /// error from rust (see issue <https://github.com/rust-lang/rust/issues/100013> for more information)
+    fn receive_timeout<'a>(
```
darkwisebear: As already discussed, naming of `receive_timeout` is a bit narrow, as this call can be used for any form of premature cancellation. Names I can think of:
* `receive_cancel`
* `receive_select`
* `cancellable_receive`

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
```diff
@@ -844,12 +850,56 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
...
+        &'a self,
+        scratch: SampleContainer<Self::Sample<'a>>,
+        new_samples: usize,
+        max_samples: usize,
+        timeout: impl Future<Output = ()> + Send + 'static,
```
darkwisebear: This future could actually return something, which is then also returned by this call. I cannot think of any use case, though, so that could also be saved for later, as this would be a non-breaking change.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -537,21 +538,84 @@ where
...
+                        scratch,
+                        Err(Error::ReceiveError(
+                            ReceiveFailedReason::InputValueOutOfBounds {
+                                max: self.max_num_samples,
+                                requested: max_samples.max(new_samples),
+                            },
+                        )),
+                    );
+                }
+                _ => {}
```
darkwisebear: Which cases does this cover? If this is a match arm that shall not be reached, I would recommend an `unreachable!` statement that helps to detect erroneous assumptions during testing.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -581,18 +646,21 @@ where
 // a waker storage for async notifications, and parameters for managing the receive operation.
 // The Future implementation for ReceiveFuture defines the polling logic,
 // which attempts to receive samples and manages the state of the receive operation.
-struct ReceiveFuture<'a, T: CommData + Debug> {
+// Unpin bound is required for the timeout future to allow it to be safely pinned on the stack without
+// heap allocation.
+struct ReceiveFuture<'a, T: CommData + Debug, F: Future<Output = ()> + Unpin> {
```
darkwisebear: Strictly speaking, the `Unpin` bound isn't needed here, as the type definition itself does not need it. However, the `impl` block needs it since you do not specify that the `timeout` member is, in fact, a pinned reference, which always is `Unpin`.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -581,18 +646,21 @@ where
...
     event_guard: Option<ProxyEventManagerGuard<'a>>,
     waker_storage: Arc<AtomicWaker>,
     max_num_samples: usize,
     scratch: Option<SampleContainer<Sample<T>>>,
     new_samples: usize,
     max_samples: usize,
     total_received: usize,
+    timeout: F,
```
darkwisebear: As stated above, I think it would be clearer if you use `Pin<&'a mut F>` and remove `Unpin` as a bound everywhere, as this is how the `ReceiveFuture` really works.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -601,6 +669,17 @@ impl<'a, T: CommData + Debug> Future for ReceiveFuture<'a, T> {
         let max_num_samples = self.max_num_samples;
         let total_received = self.total_received;
 
+        // Poll cancel/timeout future first — F: Unpin so Pin::new is safe, no allocation
```
darkwisebear: Misleading comment - `Pin::new` is always safe. However, it only _exists_ because `F: Unpin`. Otherwise, you'd need to use `Pin::new_unchecked`.

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
bharatGoswami8: Changed to `cancellable_receive`

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
bharatGoswami8: Agreed, added comment for same on API documentation.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: This was not unreachable branch, it was for positive case if all init validation pass, but this changed now from match arm to function call, moved init validation on function.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Removed and changed `timeout` type to `Pin<&'a mut F>`

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Agreed, updated same.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Updated comment and now here we do not need `Pin::new` as timeout already type of `Pin<&'a mut F>`

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -581,18 +647,21 @@ where
 // a waker storage for async notifications, and parameters for managing the receive operation.
 // The Future implementation for ReceiveFuture defines the polling logic,
 // which attempts to receive samples and manages the state of the receive operation.
-struct ReceiveFuture<'a, T: CommData + Debug> {
+// Unpin bound is required for the timeout future to allow it to be safely pinned on the stack without
```
darkwisebear: There is no `Unpin` bound anymore.

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
```diff
@@ -844,12 +850,61 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
...
+    ///
+    /// The `timeout` future currently has `Output = ()` and returns nothing. However, it could be
+    /// enhanced in the future to return a value that would be propagated to the caller, as per the use-case.
+    /// No use case for this enhancement has been identified yet, so it is deferred for later implementation.
+    fn cancellable_receive<'a>(
+        &'a self,
+        scratch: SampleContainer<Self::Sample<'a>>,
+        new_samples: usize,
+        max_samples: usize,
+        timeout: impl Future<Output = ()> + Send + 'static,
```
darkwisebear: the naming `timeout` isn't ideal either, as it can be anything actually. Maybe `cancellation`? In the description, you could then give examples what this future could be, like a timeout, or a channel to cancel the receive.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: I missed removing the comment earlier, it has now been removed, also removed the doc test which was added due to `unpin` bound.

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
bharatGoswami8: Renamed to `cancellation` and updated the documentation to reflect possible usage scenarios.

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
```diff
@@ -844,12 +850,67 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
...
+    /// User can not concurrenly call `cancellable_receive` on the same subscription instance from
+    /// multiple threads or tasks.
+    /// `cancellation` must be `'static` because cancellation futures (e.g. `tokio::time::sleep`) own all
+    /// their state and do not borrow anything from the caller's scope.
+    /// And if you change to `'a` lifetime then it create lifetime bound not satisfied
+    /// error from rust (see issue <https://github.com/rust-lang/rust/issues/100013> for more information)
+    ///
+    /// The `cancellation` future currently has `Output = ()` and returns nothing. However, it could be
+    /// enhanced in the future to return a value that would be propagated to the caller, as per the use-case.
+    /// No use case for this enhancement has been identified yet, so it is deferred for later implementation.
```
darkwisebear, line 906: Just an idea and not a review comment: One use case would be to implement a `select!` like mechanism that allows for combining multiple receive calls without losing either `SampleContainer`s.

darkwisebear (approved):

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
bharatGoswami8: multiple receive call for different subscription ?

bemerybmw (approved):

Issue #467, opened 2026-05-21
Bug: GenericSkeleton::GetEvents() retuns a const EventMap but Allocate()/Send() need mutability


NEOatNHNG:
### Description

The `GenericSkeleton::GetEvents()` returns a const EventMap which means all the elements are also const. But in order to Allocate() or Send() an event, the user needs a mutable GenericEvent which means we have to resort to ugly things like const_cast.

### Error Occurrence  Rate

None

### Classification

minor

### Affected Version

0.5

### Expected Closure Version

0.5

### Category

- [ ] Safety Related
- [ ] Security Related

NEOatNHNG:
One alternative instead of making the whole map mutable would be to wrap the events in a Mutable wrapper class or make methods like Allocate() and Send() `const`.

crimson11:
@NEOatNHNG : Good point! Even the `GenericProxy` does return a "non const" map! Although it is on the consumer side, which can't change the event content! But even there it needs to be non-const as "using the ProxyEvents" also/always means to alter state within the ProxyEvent!

crimson11:
@bemerybmw and me had a short discussion about "how to fix the issue".

### Our proposal

1. Provide a symmetric solution on both sides ( `GenericSkeleton` and `GenericProxy` )
2. Also the current solution on `GenericProxy` side is not "nice". It returns a non-const ref to some map-type, which solves the issue, that you then can call non-const methods on the elements (in this case `GenericProxyEvent`s), but it currently also allows to alter (add/remove) elements from the map, which is not intended.
3. The "symmetric" approach shall be as follows:
   1. `GetEvents()` of `GenericProxy` and `GenericSkeleton` shall return a
    ```
        template <typename Value>
        class ServiceElementMapView
    ```
     similar to what `GenericProxy::GetEvents()` is currently providing. It is a view/wrapper to the internal map.

   2. But this `ServiceElementMapView` shall not provide any methods, which mutate the underlying map!
   3. The `GenericSkeleton` and `GenericProxy` store internally (private member) a `std::unique_ptr<std::unordered_map<std::string_view, GenericProxyEvent>> owned_events_` resp. `std::unique_ptr<std::unordered_map<std::string_view, GenericSkeletonEvent>> owned_events_`
   4. `GetEvents` then creates a `ServiceElementMapView`, which holds a private member: `std::reference_wrapper<std::unordered_map<std::string_view, Value|>> owned_elements_` 
[...]

PR #475, opened 2026-05-27
Introduce ServiceElementMapView

crimson11:
Up to now GenericProxy and GenericSkeleton provided inconsistent views for their contained events.
This commit now introduces a common solution with
ServiceElementMapView, which solves both issues:
- map itself is immutable (no addition/removal of events)
- events within the maps are mutable

Issue: #467


score/mw/com/impl/service_element_map_view.h
```diff
@@ -38,61 +35,42 @@ class ServiceElementMap
...
 
     bool empty() const noexcept
     {
-        return map_.empty();
+        return owned_elements_.get().empty();
     }
 
   private:
-    map_type map_;
+    std::reference_wrapper<std::map<Key, Value>> owned_elements_;
```
bemerybmw: Why is it called owned_elements? That makes it sound like they're owned by this class.

score/mw/com/impl/service_element_map_view.h
```diff
@@ -38,61 +35,42 @@ class ServiceElementMap
...
-    iterator erase(const_iterator pos)
-    {
-        return map_.erase(pos);
-    }
-
-    size_type erase(const Key& key)
-    {
-        return map_.erase(key);
-    }
+    ServiceElementMapView(std::map<Key, Value>& service_element_map) noexcept : owned_elements_{service_element_map} {}
```
bemerybmw: The map should be an alias defined with the other aliases above

bemerybmw: Should document here or in the class doc that the provided map must never be moved / copied

score/mw/com/impl/service_element_map_view.h
```diff
@@ -10,25 +10,22 @@
...
 #include <map>
 #include <string_view>
 #include <utility>
 
 namespace score::mw::com::impl
 {
 
-/// \brief Map that will be used in GenericProxies to store GenericProxyEvents and possibly GenericProxyFields and
-/// GenericProxyMethods once they're supported by LoLa.
+/// \brief Map that will be provided to the user in case of GenericProxy and GenericSkeleton to allow access to the
```
bemerybmw: Comment should explain that it's a non owning view into the map.

score/mw/com/impl/service_element_map_view_test.cpp
```diff
@@ -88,38 +91,41 @@ TEST(ServiceElementMapTest, CanFindElementsInMap)
...
     EXPECT_FALSE(third_it->first.compare("2"));
     EXPECT_EQ(third_it->second, 2);
 
-    auto invalid_it = map.find("3");
-    ASSERT_EQ(invalid_it, map.cend());
+    auto invalid_it = map_view.find("3");
+    ASSERT_EQ(invalid_it, map_view.cend());
 }
 
 TEST(ServiceElementMapTest, MapCanBeCopied)
```
bemerybmw:
```suggestion
TEST(ServiceElementMapViewTest, MapCanBeCopied)
```
Same for all tests.

score/mw/com/impl/generic_skeleton.h
```diff
@@ -86,7 +84,7 @@ class GenericSkeleton : public SkeletonBase
     GenericSkeleton(const InstanceIdentifier& identifier, std::unique_ptr<SkeletonBinding> binding);
 
     /// @brief This map owns all GenericSkeletonEvent instances.
-    EventMap events_;
+    std::unique_ptr<std::map<std::string_view, GenericSkeletonEvent>> events_;
```
bemerybmw: mention that it has to be a unique_ptr so that the map never gets moved.

score/mw/com/impl/generic_proxy.h
```diff
@@ -71,17 +71,17 @@ class GenericProxy : public ProxyBase
...
-    EventMap& GetEvents() noexcept;
+    EventMap GetEvents() noexcept;
 
   private:
     GenericProxy(std::unique_ptr<ProxyBinding> proxy_binding, HandleType instance_handle);
 
     void FillEventMap(const std::vector<std::string_view>& event_names) noexcept;
 
-    EventMap events_;
+    std::unique_ptr<std::map<std::string_view, GenericProxyEvent>> events_;
```
bemerybmw: mention that it has to be a unique_ptr so that the map never gets moved.

score/mw/com/impl/generic_proxy.h
```diff
@@ -59,7 +59,7 @@ class GenericProxy : public ProxyBase
     friend class test::GenericProxyAttorney;
 
   public:
-    using EventMap = ServiceElementMap<GenericProxyEvent>;
+    using EventMap = ServiceElementMapView<GenericProxyEvent>;
```
bemerybmw: Should be symmetrical on proxy and skeleton side. If you want the alias, should be EventMapView.

score/mw/com/impl/generic_proxy.cpp
```diff
@@ -86,7 +86,8 @@ Result<GenericProxy> GenericProxy::Create(HandleType instance_handle) noexcept
 }
 
 GenericProxy::GenericProxy(std::unique_ptr<ProxyBinding> proxy_binding, HandleType instance_handle)
-    : ProxyBase{std::move(proxy_binding), std::move(instance_handle)}
+    : ProxyBase{std::move(proxy_binding), std::move(instance_handle)},
+      events_(std::make_unique<std::map<std::string_view, GenericProxyEvent>>())
 {
```
bemerybmw, line 89: assert events_ is not nullptr. We can use that so we don't have to add assertions elsewhere in the class.

score/mw/com/impl/generic_skeleton.cpp
```diff
@@ -138,7 +138,8 @@ void GenericSkeleton::StopOfferService() noexcept
 }
 
 GenericSkeleton::GenericSkeleton(const InstanceIdentifier& identifier, std::unique_ptr<SkeletonBinding> binding)
-    : SkeletonBase(std::move(binding), identifier)
+    : SkeletonBase(std::move(binding), identifier),
+      events_(std::make_unique<std::map<std::string_view, GenericSkeletonEvent>>())
 {
```
bemerybmw, line 144: assert events_ is not nullptr. We can use that so we don't have to add assertions elsewhere in the class.

score/mw/com/impl/generic_proxy.cpp
```diff
@@ -110,12 +111,9 @@ void GenericProxy::FillEventMap(const std::vector<std::string_view>& event_names
     }
 }
 
-GenericProxy::EventMap& GenericProxy::GetEvents() noexcept
+GenericProxy::EventMap GenericProxy::GetEvents() noexcept
```
bemerybmw: Can now be const?

score/mw/com/impl/generic_proxy_test.cpp
```diff
@@ -434,6 +434,32 @@ TEST_F(GenericProxyFixture, GenericProxyWillLogErrorMessageForEventsProvidedInCo
...
+    std::optional<GenericProxy> moved_generic_proxy;
+    std::optional<GenericProxy::EventMap> events;
+
+    {
+        // Given a valid GenericProxy with some GenericProxyEvents
+        CreateAHandle({kEventName1, kEventName2, kEventName3});
+        auto generic_proxy_result = GenericProxy::Create(*handle_);
+        EXPECT_TRUE(generic_proxy_result.has_value());
+
+        // when getting the event-map from the GenericProxy
```
bemerybmw: and given a ServiceElementMapView of the event map

score/mw/com/impl/generic_proxy_test.cpp
```diff
@@ -434,6 +434,32 @@ TEST_F(GenericProxyFixture, GenericProxyWillLogErrorMessageForEventsProvidedInCo
...
+        auto& generic_proxy = generic_proxy_result.value();
+        events = generic_proxy.GetEvents();
+        // and then moving the GenericProxy
+        moved_generic_proxy = std::move(generic_proxy);
+        // and destructing the moved-from GenericProxy
```
bemerybmw:
```suggestion
        // wan destroying the moved-from GenericProxy
```

score/mw/com/impl/generic_proxy_test.cpp
```diff
@@ -434,6 +434,32 @@ TEST_F(GenericProxyFixture, GenericProxyWillLogErrorMessageForEventsProvidedInCo
...
+    }
+
+    // expect, that EventMap acquired previously from the moved-from GenericProxy still contains the events
```
bemerybmw:
```suggestion
    // Then the ServiceElementMapView acquired previously from the moved-from GenericProxy still points to the events now owned by the moved-to GenericProxy.
```

bemerybmw (changes requested):

score/mw/com/impl/service_element_map_view.h
crimson11: You are right! How could I do this - me being the naming freak!?

score/mw/com/impl/generic_proxy.cpp
crimson11: `std::make_unique` never will return a `nullptr`, but will throw instead (which in our case would either lead to termination), which will therefore never create a `GenericProxy` instance ...

score/mw/com/impl/generic_skeleton.cpp
crimson11: see above ...

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+        return elements_.get().size();
+    }
+
+    bool empty() const noexcept
+    {
+        return elements_.get().empty();
+    }
+
+  private:
+    std::reference_wrapper<std::map<Key, Value>> elements_;
```
bemerybmw: This should use the map_type alias

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+{
+
+/// \brief A map view that will be provided to the user in case of GenericProxy and GenericSkeleton to allow access to
+///        the GenericProxyEvents/GenericSkeletonEvents and possibly fields/methods in a later stage. The view is NOT
+///        owning. I.e. the view is set up by the owner of the underlying map and provides a view to this map.
+template <typename Value>
+class ServiceElementMapView
+{
+  public:
+    using Key = std::string_view;
```
bemerybmw: The alias naming is not consistent. I guess we have some camel_case names to match the standard library, but then map_type should be MapType since it's not following the standard library? Or maybe even just change them all to CamelCase?

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+    using value_type = std::pair<const Key, Value>;
+    using iterator = typename std::map<Key, Value>::iterator;
+    using const_iterator = typename std::map<Key, Value>::const_iterator;
+    using size_type = typename std::map<Key, Value>::size_type;
+
+    /// \brief Creates a ServiceElementMapView that provides "read-only" access to the provided map. That is no elements
+    ///        can be added or removed from the map via the ServiceElementMapView, but changes to the map elements is
+    ///        possible.
+    /// @param service_element_map underlying map on which the view is created. It must not be moved or copied after
+    ///        creation.
```
bemerybmw:
```suggestion
    ///        being passed to this constructor.
```

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+    ///        Note: Copying is not possible anyhow as the elements within the map are all not copyable.
```
bemerybmw: I don't get this comment. The Value is templated so we don't know if the map is copyable from here. If you really want this limitation, then you should add a static assert.

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+#include <map>
+#include <string_view>
+#include <utility>
+
+namespace score::mw::com::impl
+{
+
+/// \brief A map view that will be provided to the user in case of GenericProxy and GenericSkeleton to allow access to
+///        the GenericProxyEvents/GenericSkeletonEvents and possibly fields/methods in a later stage. The view is NOT
+///        owning. I.e. the view is set up by the owner of the underlying map and provides a view to this map.
```
bemerybmw:
```suggestion
///        owning. I.e. the view is set up by the owner of the underlying map and provides a read-only view to this map.
```

score/mw/com/impl/generic_skeleton.h
```diff
@@ -85,8 +84,11 @@ class GenericSkeleton : public SkeletonBase
     // Private constructor, only callable by static Create methods.
     GenericSkeleton(const InstanceIdentifier& identifier, std::unique_ptr<SkeletonBinding> binding);
 
-    /// @brief This map owns all GenericSkeletonEvent instances.
-    EventMap events_;
+    /// \brief This map owns all GenericSkeletonEvent instances.
+    /// \details This map needs to be covered in a unique_ptr as it shall not be relocated by a move of the
+    /// GenericSkeleton. This is required as we hand out views to this map (see GetEvents()), which need to be valid
+    /// even after the GenericSkeleton instance has been moved.
+    std::unique_ptr<std::map<std::string_view, GenericSkeletonEvent>> events_;
```
bemerybmw: This should use an alias. Maybe we can have a quick call about how we should handle this.

score/mw/com/impl/generic_skeleton.h
```diff
@@ -70,9 +69,9 @@ class GenericSkeleton : public SkeletonBase
     [[nodiscard]] static Result<GenericSkeleton> Create(const InstanceSpecifier& specifier,
                                                         const GenericSkeletonServiceElementInfo& in) noexcept;
 
-    /// @brief Returns a const reference to the name-keyed map of events.
-    /// @note The returned reference is valid as long as the GenericSkeleton lives.
-    [[nodiscard]] const EventMap& GetEvents() const noexcept;
+    /// @brief Returns a view to the name-keyed map of events.
```
bemerybmw:
```suggestion
    /// @brief Returns a read-only view to the name-keyed map of events.
```

score/mw/com/impl/generic_proxy.h
bemerybmw: I think you forgot this

bemerybmw (changes requested):

score/mw/com/impl/generic_proxy.h
crimson11: ups. yeah! Just added the comment on the skeleton side

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+template <typename Value>
+class ServiceElementMapView
```
bemerybmw, line 32: Needs to be in types.h

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,83 @@
...
+    using const_iterator = typename std::map<Key, Value>::const_iterator;
+    using size_type = typename std::map<Key, Value>::size_type;
+
+    /// \brief Creates a ServiceElementMapView that provides "read-only" access to the provided map. That is no elements
+    ///        can be added or removed from the map via the ServiceElementMapView, but changes to the map elements is
+    ///        possible.
+    /// @param service_element_map underlying map on which the view is created. It must not be moved or copied after
+    ///        creation.
+    ///        Note: Copying is not possible anyhow as the elements within the map are all not copyable.
+    ServiceElementMapView(map_type& service_element_map) noexcept : elements_{service_element_map} {}
```
bemerybmw: this should be private and we should have a factory which stays in the impl namespace

score/mw/com/impl/service_element_map_view.h
crimson11: Changed it to lower!

score/mw/com/impl/service_element_map_view.h
crimson11: Removed the comment ... as it is misleading ... I wanted to express, that in our usage of this template class, we are using it with template args, which aren't copyable ...

score/mw/com/impl/service_element_map_view.h
crimson11: Nope. We were wrong I guess! We don't want the general class template to be accessible/visible! What the user wants/needs is already public visible via GenericProxy::EventMap resp. GenericSkeleton::EventMap!
These are exactly the template instantiations the user needs/shall use.

score/mw/com/impl/generic_skeleton.h
crimson11: After the refactoring I could use `ServiceElementMapViewFactory<GenericSkeletonEvent>::map_type` ... but does it really add value?

score/mw/com/impl/service_element_map_view_factory.h
```diff
@@ -0,0 +1,36 @@
...
+namespace score::mw::com::impl
+{
+
+template <typename ElementType>
+class ServiceElementMapViewFactory
+{
+  public:
+    using map_type = typename ServiceElementMapView<ElementType>::map_type;
+
+    static ServiceElementMapView<ElementType> CreateMapView(map_type& element_map) noexcept
```
bemerybmw: We usually just use Create since it's already clear what we're creating based on the class name. i.e. ServiceElementMapViewFactory::CreateMapView is no clearer than ServiceElementMapViewFactory::Create

score/mw/com/impl/service_element_map_view_factory.h
```diff
@@ -0,0 +1,36 @@
...
+
+#ifndef SCORE_MW_COM_IMPL_SERVICE_ELEMENT_MAP_VIEW_FACTORY_H
+#define SCORE_MW_COM_IMPL_SERVICE_ELEMENT_MAP_VIEW_FACTORY_H
+
+#include "score/mw/com/impl/service_element_map_view.h"
+
+namespace score::mw::com::impl
+{
+
+template <typename ElementType>
```
bemerybmw, line 28: Needs a docstring to explain why we need a factory

score/mw/com/impl/service_element_map_view_factory.h
bemerybmw: Should not be noexcept

score/mw/com/impl/service_element_map_view.h
```diff
@@ -0,0 +1,92 @@
...
+
+  private:
+    using map_type = std::map<std::string_view, Value>;
+
+    /// \brief Creates a ServiceElementMapView that provides "read-only" access to the provided map. That is no elements
+    ///        can be added or removed from the map via the ServiceElementMapView, but changes to the map elements is
+    ///        possible.
+    /// @param service_element_map underlying map on which the view is created. It must not be moved or copied after
+    ///        being passed to this ctor.
+    ServiceElementMapView(map_type& service_element_map) noexcept : elements_{service_element_map} {}
```
bemerybmw, line 86: mention that it can only be called via the factory.

score/mw/com/impl/generic_skeleton.h
bemerybmw: Hmmm... I think it does because it gives a bit more context to the reader but I don't have a strong opinion. You decide what you prefer

bemerybmw (approved):

PR #477, opened 2026-05-27
Fixed(docs): use doxygenclass for template class members in RST generation

AAmbuj:
Breathe cannot resolve template member functions via doxygenfunction when Doxygen does not emit standalone XML entries for inline template method definitions. Detect template class members in extract_api_items.py and emit doxygenclass with :members: instead, with deduplication to avoid documenting the same class multiple times.

### DIFF
--- a/docs/sphinx/utils/extract_api_items.py
+++ b/docs/sphinx/utils/extract_api_items.py
@@ -261,6 +261,76 @@ class RSTGenerator: # pylint: disable=too-few-public-methods
         self.project_name = project_name
         self.output_dir = Path(output_dir)
         self.output_dir.mkdir(parents=True, exist_ok=True)
+        self._documented_template_classes: set = set()
+
+    @staticmethod
+    def _extract_template_class(base_name: str) -> Optional[str]:
+        """Detect if a function is a member of a template class.
+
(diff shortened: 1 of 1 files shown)


docs/sphinx/utils/extract_api_items.py
```diff
@@ -261,6 +261,49 @@ def __init__(self, project_name: str, output_dir: str):
...
+        # outermost balanced angle brackets from each segment.
+        segments = class_part.split('::')
+        clean_segments = []
+        for seg in segments:
+            idx = seg.find('<')
+            if idx != -1:
+                seg = seg[:idx]
+            clean_segments.append(seg)
+
+        return '::'.join(clean_segments)
```
LittleHuba: What about partial template specializations where a specialization contains namespace separators? (`a::b<c::d, T>`)
They will mess this algorithm up.

gdadunashvili (approved):

docs/sphinx/utils/extract_api_items.py
AAmbuj: fix the edge cases.

castler (approved):

Issue #371, opened 2026-04-29
Improvement: Async Stream Receive API


bharatGoswami8:
### What

Rust COM API library offers the async receive APIs but for this user need to poll always and in that it may miss some samples to overcome that we are introducing a stream API which will provide a stream of data as soon as event trigger.

### How

https://github.com/eclipse-score/communication/pull/349

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [ ] Requirements / Architecture are not affected by this change?

bharatGoswami8:
Hi @artemsheinacn ,

> But why would I implement it using an additional thread and more memory consumption if I can just keep using my implementation of streaming API which provides proper error-state detection?

I did not get second statement, do you mean you have internal implementation for streaming the data using COM-API ?

artemsheinacn:
Yes, I do https://github.com/eclipse-score/feo/blob/242a28c5de30bee532e4d9094325a30ccabaca37/src/feo/src/signalling/direct/mw_com/mod.rs#L89
It was a temporary solution and may be unsound, but it asserts that there is no events overflow:
```
assert!(
                sample_container.sample_count() + self.received.borrow().len() <= self.received.borrow().capacity(),
                "must not allocate, configuration error"
            );
```

bharatGoswami8:
> Yes, I do https://github.com/eclipse-score/feo/blob/242a28c5de30bee532e4d9094325a30ccabaca37/src/feo/src/signalling/direct/mw_com/mod.rs#L89 It was a temporary solution and may be unsound, but it asserts that there is no events overflow:
> 
> ```
> assert!(
>                 sample_container.sample_count() + self.received.borrow().len() <= self.received.borrow().capacity(),
>                 "must not allocate, configuration error"
>             );
> ```

@artemsheinacn ,
Understood,  it is user specific requirement where user do not want to overwrite the value in internal container, this request we can also consider if other contributor agree, so what i feel one more bool parameter in API which ask user do you want to overwrite container when it is full or not if true then it can return a error when container is full.

```
fn to_stream<'a>(
        &'a self,
        max_samples: usize,
        **is_container_overwrite: bool,**
    ) -> impl Stream<Item = Result<Self::Sample<'a>>> + 'a;
```

artemsheinacn:
Maybe it makes sense to define it as an enum like so:
```
enum OverflowStrategy {
   Replace,
   Panic,
   Resize,
   ...
}

[...]

PR #480, opened 2026-05-28
Rust::com Rust Stream API Implementation

bharatGoswami8:
- Adds a borrow-based, non-'static async streaming API (Subscription::to_stream) with an internal fixed-capacity buffer.
- Implements the stream for the Lola runtime.

https://github.com/eclipse-score/communication/issues/371

### DIFF
--- a/score/mw/com/example/com-api-example/basic-consumer-producer.rs
+++ b/score/mw/com/example/com-api-example/basic-consumer-producer.rs
@@ -414,21 +414,26 @@ mod test {
     }
 
     async fn stream_processor_fn<R: Runtime>(subscribed: impl Subscription<Tire, R>) {
-        let mut stream = subscribed.to_stream(5);
-        let mut cnt = 5;
+        let mut stream = subscribed.to_stream();
+        let mut cnt = 5usize;
         println!("[RECEIVER] Stream processor started");
-        while let Some(sample_result) = stream.next().await {
-            match sample_result {
-                Ok(sample) => println!(
-                    "[RECEIVER] Stream received sample: {:.2} psi",
-                    sample.pressure
-                ),
-                Err(e) => eprintln!("[RECEIVER] Stream error: {:?}", e),
+        while cnt > 0 {
+            // Use timeout to avoid waiting indefinitely in case of issues with the producer or subscription
+            match tokio::time::timeout(tokio::time::Duration::from_secs(3), stream.next()).await {
+                Ok(Some(Ok(sample))) => {
+                    println!(
+                        "[RECEIVER] Stream received sample: {:.2} psi",
+                        sample.pressure
+                    )
+                }
+                Ok(Some(Err(e))) => eprintln!("[RECEIVER] Stream error: {:?}", e),
+                Ok(None) => break,
+                Err(_) => {
+                    eprintln!("[RECEIVER] Timeout while waiting for stream sample");
+                    break;
+                }
             }
             cnt -= 1;
-            if cnt == 0 {
-                break;
-            }
         }
     }
 
--- a/score/mw/com/example/com-api-example/basic-consumer-producer.rs
+++ b/score/mw/com/example/com-api-example/basic-consumer-producer.rs
@@ -413,7 +413,7 @@ mod test {
         }
     }
 
-    async fn stream_processor_fn<R: Runtime>(subscribed: impl Subscription<Tire, R>) {
+    async fn stream_processor_fn<R: Runtime>(mut subscribed: impl Subscription<Tire, R>) {
         let mut stream = subscribed.to_stream();
         let mut cnt = 5usize;
         println!("[RECEIVER] Stream processor started");
--- a/score/mw/com/example/com-api-example/basic-consumer-producer.rs
+++ b/score/mw/com/example/com-api-example/basic-consumer-producer.rs
@@ -244,6 +244,7 @@ fn main() {
 #[cfg(test)]
 mod test {
     use super::*;
+    use futures::stream::StreamExt;
     use std::sync::OnceLock;
     use std::thread;
     use std::time::Duration;
@@ -364,7 +365,7 @@ mod test {
     async fn async_data_sender_fn<R: Runtime>(
         offered_producer: VehicleOfferedProducer<R>,
     ) -> VehicleOfferedProducer<R> {
-        for i in 0..10 {
+        for i in 0..6 {
             let uninit_sample = match offered_producer.left_tire.allocate() {
                 Ok(sample) => sample,
                 Err(e) => {
@@ -393,7 +394,7 @@ mod test {
     ) {
         println!("[RECEIVER] Async data processor started");
         let mut buffer = SampleContainer::new(5);
-        for _ in 0..5 {
+        for _ in 0..3 {
             let (returned_buf, result) = if is_timeout {
                 let timeout = tokio::time::sleep(Duration::from_millis(1000));
                 subscribed.cancellable_receive(buffer, 2, 3, timeout).await
@@ -409,6 +410,25 @@ mod test {
                 println!("[RECEIVER]   Sample: {:.2} psi", sample.pressure);
             }
             buffer = buf;
+        }
+    }
+
+    async fn stream_processor_fn<R: Runtime>(subscribed: impl Subscription<Tire, R>) {
+        let mut stream = subscribed.to_stream(5);
+        let mut cnt = 5;
+        println!("[RECEIVER] Stream processor started");
+        while let Some(sample_result) = stream.next().await {
+            match sample_result {
+                Ok(sample) => println!(
+                    "[RECEIVER] Stream received sample: {:.2} psi",
+                    sample.pressure
+                ),
+                Err(e) => eprintln!("[RECEIVER] Stream error: {:?}", e),
+            }
+            cnt -= 1;
+            if cnt == 0 {
+                break;
+            }
         }
     }
 
@@ -451,6 +471,7 @@ mod test {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "This test demonstrates async receive with timeout, it can run individually if wanted to test timeout behavior"]
     async fn receive_with_timeout_and_send_using_multi_thread() {
         println!("Starting async subscription test with Lola runtime");
         //Intentionally using service instance of test1, if you face issue add new service instance in config file and use it here.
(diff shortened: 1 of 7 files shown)


darkwisebear (commented):
(AI screwed up)

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -451,6 +476,7 @@ mod test {
     }
 
     #[tokio::test(flavor = "multi_thread")]
+    #[ignore = "This test demonstrates async receive with timeout, it can run individually if wanted to test timeout behavior"]
```
bharatGoswami8, line 479: This test cases ignored in example app because of it required separate service instance.
We already have plan to optimize this example app which can run as an app and can select which API want to test using user input - https://github.com/eclipse-score/communication/issues/489

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
```diff
@@ -911,6 +912,39 @@ pub trait Subscription<T: CommData + Debug, R: Runtime + ?Sized> {
...
+    /// even though the actual stream implementation might be `Unpin`. The compiler cannot guarantee
+    /// that the returned stream is `Unpin` without the explicit bound, so we add it to ensure that
+    /// the returned stream can be safely polled without additional pinning at the user side.
+    ///
+    /// # Returns
+    /// A stream that yields individual `Sample<'a>` items one at a time.
+    ///
+    /// # Errors
+    /// Returns an error if a problem occurs during sample reception.
+    fn to_stream<'a>(&'a self) -> impl Stream<Item = Result<Self::Sample<'a>>> + Unpin + 'a;
```
darkwisebear: This should be `&mut self`, so that `receive` (or a 2nd `to_stream` call) cannot happen.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -364,7 +365,7 @@ mod test {
     async fn async_data_sender_fn<R: Runtime>(
         offered_producer: VehicleOfferedProducer<R>,
     ) -> VehicleOfferedProducer<R> {
-        for i in 0..10 {
+        for i in 0..6 {
```
darkwisebear, line 368: Why is 6 better than 10?

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -393,7 +394,7 @@ mod test {
     ) {
         println!("[RECEIVER] Async data processor started");
         let mut buffer = SampleContainer::new(5);
-        for _ in 0..5 {
+        for _ in 0..3 {
```
darkwisebear, line 397: Why is 3 better than 5? Consider turning these numbers into constants and explain why they have the limit they have.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -703,6 +726,78 @@ impl<'a, T: CommData + Debug, F: Future<Output = ()>> Future for ReceiveFuture<'
...
 
+/// A `Stream` that continuously delivers one sample at a time from the subscription.
+/// It maintains an internal `SampleContainer` to buffer samples received from the FFI callback.
+/// It also holds an exclusive `ProxyEventManagerGuard` for the lifetime of the stream to prevent
+/// concurrent receives on the same subscriber instance.
+/// On each poll, it first yields any buffered samples before attempting to receive more from the FFI callback.
+struct SampleStream<'a, T: CommData + Debug> {
+    subscriber: &'a SubscriberImpl<T>,
+    sample_container: Option<SampleContainer<Sample<T>>>,
+    event_guard: Option<ProxyEventManagerGuard<'a>>,
```
darkwisebear: Why does this need to be an `Option`? I do not see any place where this is set to `None`.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -703,6 +726,78 @@ impl<'a, T: CommData + Debug, F: Future<Output = ()>> Future for ReceiveFuture<'
...
+}
+
+impl<'a, T: CommData + Debug> Stream for SampleStream<'a, T> {
+    type Item = Result<Sample<T>>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        // Yield any buffered samples from a previous batch fetch first.
+        if let Some(mut container) = self.sample_container.take() {
```
darkwisebear: Why do you need to `take` the container? Wouldn't a simple `pop_front` be enough? You reassign it in any case.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -703,6 +726,78 @@ impl<'a, T: CommData + Debug, F: Future<Output = ()>> Future for ReceiveFuture<'
...
+
+        // Register the waker before attempting to receive so that a concurrent FFI
+        // callback cannot fire between the receive attempt and registering the waker,
+        // which would cause a missed wake-up.
+        self.subscriber.waker_storage.register(cx.waker());
+        let max_num_samples = self.subscriber.max_num_samples;
+
+        let samples_received = {
+            // Temporarily take ownership of scratch to avoid borrow issues
+            if let Some(mut scratch) = self.sample_container.take() {
```
darkwisebear: Same. Why take the container? In case of an error, the stream is then broken, since the container "gets lost".

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -703,6 +726,78 @@ impl<'a, T: CommData + Debug, F: Future<Output = ()>> Future for ReceiveFuture<'
...
+                    Err(Error::ReceiveError(ReceiveFailedReason::ReceiveError))
+                }
+            } else {
+                Err(Error::ReceiveError(ReceiveFailedReason::BufferUnavailable))
+            }
+        };
+
+        match samples_received {
+            Ok(_count) => {
+                if let Some(mut container) = self.sample_container.take() {
```
darkwisebear: Same.

score/mw/com/impl/rust/com-api/com-api-runtime-mock/runtime.rs
```diff
@@ -337,6 +338,11 @@ where
     ) -> impl Future<Output = (SampleContainer<Self::Sample<'a>>, Result<usize>)> + 'a {
         async { todo!() }
     }
+
+    #[allow(clippy::manual_async_fn)]
+    fn to_stream<'a>(&'a self) -> impl Stream<Item = Result<Self::Sample<'a>>> + Unpin + 'a {
+        stream::empty()
```
darkwisebear, line 344: Is there a ticket to extend this at some point?

darkwisebear (changes requested):

score/mw/com/impl/rust/com-api/com-api-concept/com_api_concept.rs
bharatGoswami8: yes agreed, updated the API signature.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: This issue was originally caused by the borrow checker. When calling fn `try_receive_samples`, it required mutable access to both values. To address this, we previously wrapped them in `Option`, took temporary ownership, and then reassigned them.
Now, this workaround has been removed because we are borrowing `self` mutably only once and using it to call internal fields when passing values to the function call so only one mut borrow happening with this now.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: it was because of Option and borrow checker issue, removed now.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: it was because of Option and borrow checker issue, removed now.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: it was because of Option and borrow checker issue, removed now.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: It was reduced because we have now more test and sending 10 sample is taking time because each test is invoking producer but as already i have mention in above comment , we will optimize this example app from test based to proper API function based testing using user input from CLI , ticket for same https://github.com/eclipse-score/communication/issues/489

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: When we are testing developed API using this example app test, it takes some time to run because of that iteration is reduced but we will optimize this example app from test based to proper API function-based testing using user input from CLI , ticket for same https://github.com/eclipse-score/communication/issues/489

score/mw/com/impl/rust/com-api/com-api-runtime-mock/runtime.rs
bharatGoswami8: Created ticket for same - https://github.com/eclipse-score/communication/issues/490

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -703,6 +726,55 @@ impl<'a, T: CommData + Debug, F: Future<Output = ()>> Future for ReceiveFuture<'
...
+        );
+
+        match samples_received {
+            Ok(_count) => {
+                match self.sample_container.pop_front() {
+                    Some(sample) => Poll::Ready(Some(Ok(sample))),
+                    None => Poll::Pending,
+                }
+            }
+            Err(e) => Poll::Ready(Some(Err(e))),
```
bharatGoswami8, line 771: Stream need to returns Poll::Ready(None) for dropping the stream.

Should library close the stream after error occurred or user can handle from its side to drop the stream when error occurred.

```
while let Some(item) = stream.next().await {
    match item {
        Ok(v) => {}
        Err(e) => break, // stop consuming
    }
}
drop(stream);
```
we can explicitly mention in API documentation a bout user needs to drop the stream if error occurred and implement the drop function if required in library for stream type.

library side Stream close -  
We need to introduce one more flag to track the error and if error occurred , will return the error and in iteration of next call will return the `None`.
```
// Close stream immediately in next iteration after error
        if self.errored {
            return Poll::Ready(None);
        }
```

bharatGoswami8: I think combination of both the approach will make better solution, we will ask user to explicitly call the drop on error but if it does not call then we can terminate when next iteration received to library by user.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
```diff
@@ -609,6 +613,25 @@ where
             .await
         }
     }
+
+    fn to_stream<'a>(&'a mut  self) -> impl Stream<Item = Result<Self::Sample<'a>>> + Unpin + 'a {
+        // Get the event guard to ensure no concurrent receive calls
+        // on the same subscriber instance. This guard is held for the lifetime of the stream.
+        let mut proxy_event_guard = self.event.get_proxy_event();
```
darkwisebear: That should not be necessary: Since `to_stream` can only be called when the user has unique ownership, no guard is needed as in the `receive` case.

bharatGoswami8: Agreed, but `init_async_receive` required `ProxyEventManagerGuard` because receive also use same method for initializing callback.
i will keep the scope only for init and then `SampleStream` can directly use event instance using `SubscriberImpl` instance.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/consumer.rs
bharatGoswami8: Updated the API comment, we will return the error and user take a call whether stream need to drop or continuously receive the samples.

darkwisebear (approved):

bemerybmw (approved):

Issue #387, opened 2026-05-05
Improvement: Provide basic infrastructure for a gateway approach


crimson11:
### What

Within BMW internally we showed a PoC for "gatewaying" LoLa shared-memory communication in an inter-VM setup.
The underlying architecture is depicted here: https://github.com/eclipse-score/communication/tree/main/score/mw/com/dependability/software_architectural_design/generalized_gateway

The generic code/extensions (without hypervisor specific implementations) shall be incorporated in the existing `score::mw::com` codebase.

### How

See: https://github.com/eclipse-score/communication/tree/main/score/mw/com/dependability/software_architectural_design/generalized_gateway

### Estimates for realization

2 weeks

### Category

- [x] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

PR #482, opened 2026-05-28
  Extend the generic skeleton to opening non-owned shared memory objects.

Rahul-Sutariya:
Issue: https://github.com/eclipse-score/communication/issues/387

### DIFF
--- a/score/mw/com/impl/bindings/lola/skeleton.cpp
+++ b/score/mw/com/impl/bindings/lola/skeleton.cpp
@@ -142,6 +142,49 @@ T& DereferenceWithNullCheck(std::unique_ptr<T>& ptr)
 
 }  // namespace
 
+Skeleton::ShmReuseStrategy Skeleton::DetermineShmReuseStrategy(
+    const bool previous_shm_region_unused_by_proxies) const noexcept
+{
+    // In case of inter-VM forwarded services, always open existing SHM created by the gateway on the other VM
+    if (lola_service_instance_deployment_.inter_vm_support_ && lola_service_instance_deployment_.inter_vm_forwarded_)
+    {
+        return ShmReuseStrategy::kOpenGatewayForwardedShm;
+    }
+
+    // If we can exclusively lock the usage marker file, the previous SHM region is not being used by proxies.
+    // This means either:
+    // (1) Previous skeleton stopped cleanly with no proxies, or
+    // (2) Previous skeleton crashed
+    // In both cases, we recreate the SHM to ensure a clean state.
+    if (previous_shm_region_unused_by_proxies)
+    {
+        return ShmReuseStrategy::kRecreateShm;
+    }
+
+    // If we cannot lock the usage marker file, proxies are still using the previous SHM region.
+    // This means the previous skeleton crashed while proxies were subscribed, so we reuse the existing SHM.
+    return ShmReuseStrategy::kReuseExistingShm;
+}
+
+Skeleton::ShmRemovalStrategy Skeleton::DetermineShmRemovalStrategy(
+    const bool can_exclusively_lock_usage_file) const noexcept
+{
+    // If we cannot lock the usage marker file that means that the proxies are still subscribed and using the SHM
+    if (!can_exclusively_lock_usage_file)
+    {
+        return ShmRemovalStrategy::kRemoveNeitherShmNorMarkerFile;
+    }
+
+    // Gateway-forwarded services don't own the SHM (it's on the other VM), so only remove the marker file
+    if (IsGatewayForwardedSkeleton())
+    {
+        return ShmRemovalStrategy::kRemoveMarkerFileOnly;
+    }
+
+    // Regular skeleton: we own the SHM, so remove both SHM and marker file
+    return ShmRemovalStrategy::kRemoveShmAndMarkerFile;
+}
+
 // Suppress "AUTOSAR C++14 A15-5-3" rule findings. This rule states: "The std::terminate() function shall not be
 // called implicitly". std::terminate() is implicitly called from 'service_instance_existence_marker_file.value()'
 // in case the 'service_instance_existence_marker_file' doesn't have value but as we check before with 'has_value()'
@@ -210,6 +253,7 @@ Skeleton::Skeleton(const InstanceIdentifier& identifier,
       quality_type_{InstanceIdentifierView{identifier_}.GetServiceInstanceDeployment().asilLevel_},
       lola_instance_id_{lola_service_instance_deployment.instance_id_.value().GetId()},
       lola_service_id_{lola_service_type_deployment.service_id_},
+      lola_service_instance_deployment_{lola_service_instance_deployment},
       shm_path_builder_{std::move(shm_path_builder)},
       memory_manager_{GetInstanceQualityType(),
                       DereferenceWithNullCheck(shm_path_builder_),
@@ -227,6 +271,7 @@ Skeleton::Skeleton(const InstanceIdentifier& identifier,
       method_subscription_registration_guard_qm_{},
       method_subscription_registration_guard_asil_b_{},
       was_old_shm_region_reopened_{false},
+      use_gateway_forwarded_shm_{false},
       filesystem_{std::move(filesystem)},
       prepare_offer_called_{false},
       prepare_stop_offer_called_{false},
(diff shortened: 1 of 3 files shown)


score/mw/com/impl/bindings/lola/shm_path_builder.cpp
```diff
@@ -30,6 +30,17 @@ constexpr auto kDataChannelPrefix = "lola-data-";
 constexpr auto kControlChannelPrefix = "lola-ctl-";
 constexpr auto kMethodChannelPrefix = "lola-methods-";
 constexpr auto kAsilBControlChannelSuffix = "-b";
+constexpr auto kInterVmSharedShmPrefix = "/intervm-shared-shmem/";
+
+// Inter-VM shared-memory backends can have a short NAME_MAX, use compact 5-digit service ID instead of the default
+// 16-digit padding
+void AppendServiceAndInstanceCompact(std::ostream& out,
```
crimson11: You can leave out this "compacting". If we would takes this over from the PoC ... then we would really materialize all the bugs/short-commings, which BlackBerry had in their sample-contribution :) (namely they had a weird restriction in their `qvmshm` implementation regarding path-name-sizes).

score/mw/com/impl/bindings/lola/shm_path_builder.h
```diff
@@ -29,7 +29,10 @@ namespace score::mw::com::impl::lola
 class ShmPathBuilder : public IShmPathBuilder
 {
   public:
-    explicit ShmPathBuilder(const std::uint16_t service_id) noexcept : IShmPathBuilder{}, service_id_{service_id} {}
+    explicit ShmPathBuilder(const std::uint16_t service_id, const bool inter_vm_support = false) noexcept
```
crimson11, line 41: Add a doxygen comment to this ctor.
document/comment the `inter_vm_support` parameter. Note, that it is an interim solution, to express the intent to have a shm-object, which supports sharing across VMs by encoding it into the path-name. A future solution will be based on a redesigned `lib/memory/shared` API abstraction.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -256,43 +301,70 @@ auto Skeleton::PrepareOffer(SkeletonEventBindings& events,
...
+            }
+            else
+            {
+                memory_manager_.CleanupSharedMemoryAfterCrash();
+            }
+            was_old_shm_region_reopened_ = true;
+            gateway_shm_ready_ = false;
+            break;
+
+        case ShmReuseStrategy::kUnknownStrategy:
```
crimson11, line 359: You should have a "fal through" here:

```
case ShmReuseStrategy::kUnknownStrategy:
default:
   SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(false, "Unknown SHM reuse strategy.");
   break;
```

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -256,43 +301,70 @@ auto Skeleton::PrepareOffer(SkeletonEventBindings& events,
...
+            shm_setup_result = memory_manager_.OpenExistingSharedMemory(std::move(register_shm_object_trace_callback));
+            if (!shm_setup_result.has_value())
+            {
+                score::mw::log::LogError("lola")
+                    << "Could not open existing shared memory region for gateway-forwarded service.";
+                return MakeUnexpected(ComErrc::kBindingFailure,
+                                      "Could not open existing shared memory region for gateway-forwarded service.");
+            }
+            was_old_shm_region_reopened_ = false;
+            gateway_shm_ready_ = true;
```
crimson11: I guess, we should rename `gateway_shm_ready_` -> `use_gateway_forwarded_shm_`

midterm we might need some refactoring imho ... because now we have different flags, which are mutual exclusive (was_old_shm_region_reopened_ ,  use_gateway_forwarded-shm_) which is a bit ugly...

So from a high-level perspective, we have either a "normal" skeleton ... or a "forwarding" skeleton.
So - for now we express the latter only with this flag `use_gateway_forwarded_shm_` ... which looks a bit "hacky" as a lot of behaviour is different in this case. So I'm thinking in the direction to introduce a more generic "mode" attribute in our `skeleton` class, which is either "REGULAR" or "FORWARDING" ... but this is - as I said - for future refactorings.

score/mw/com/impl/bindings/lola/skeleton.h
```diff
@@ -249,7 +302,15 @@ auto Skeleton::Register(const ElementFqId element_fq_id, SkeletonEventProperties
     // If the skeleton previously crashed and there are active proxies connected to the old shared memory, then we
     // re-open that shared memory in PrepareOffer(). In that case, we should retrieved the EventDataControl and
     // EventDataStorage from the shared memory and attempt to rollback the Skeleton tracing transaction log.
-    if (was_old_shm_region_reopened_)
+    if (gateway_shm_ready_)
```
crimson11: We should do an assert here! If we end up in this template-function call AND `gateway_shm_ready_` (now - see above `use_gateway_forwarded_shm_` ) is true, something went wrong.
Why? Because this specific interVM functionality is only used in our gateway setups, where always `GenericSkeleton` is used - no "typed skeletons". But in case of a `GenericSkeleton` this template-func will never be called! Because in this case we only deal with `GenericSkeletonEvents`, which do not call this template func, but instead the non-template func `Skeleton::RegisterGeneric()`!

So - make an assert here and document, that `use_gateway_forwarded_shm_` == true is only expected in gateway-use-cases, where `GenericSkeletons` are used ...

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -264,9 +276,19 @@ auto SkeletonMemoryManager::RetrieveEventDataFromOpenedSharedMemory(const Elemen
         return it;
     };
 
-    score::cpp::ignore = find_element(storage_->events_metainfo_, element_fq_id);
+    const auto event_metainfo_it = find_element(storage_->events_metainfo_, element_fq_id);
     const auto event_data_storage_it = find_element(storage_->events_, element_fq_id);
 
+    if (lola_service_instance_deployment_.inter_vm_support_ && lola_service_instance_deployment_.inter_vm_forwarded_)
```
crimson11: See argumentation above! In case we are in the interVM-gateway case, we expect a `GenericSkeleton` being used. But then, we would NOT end up in this template func as it is only used within "typed skeletons".

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -448,6 +535,106 @@ QualityType Skeleton::GetInstanceQualityType() const
     return quality_type_;
 }
 
+Result<void> Skeleton::ValidateDataShmObject(const SkeletonEventBindings& events,
```
crimson11: I did not see any calls to this method (and also not to `ValidateControlShmObject`)

Where did you expect the validation calls to happen?
My 1st guess would be directly after opening the shm-objects in `PrepareOffer`?

But now I had a 2nd look... and I'm not sure, whether this validation you implemented here is "added value".
My thoughts:
- in `Skeleton::RegisterGeneric()` further down you have the code every `GenericSkeletonEvent` (and later also field) will run through, to register at its parent skeleton, thereby also "checking" the shared memory (CONTROL and DATA). I.e. at least it checks, whether for its `ElementFqId` the expected control/data artefact exists in shared-memory ... so if in the interVM use case, the shm, which we opened would be a wrong one, `Skeleton::RegisterGeneric()` would fail.
- do you really check much more here, than what is implicitly checked in Skeleton::RegisterGeneric()?

You are even hyper-strict here - as you check, that the shm conains exactly the number of events/fields as you have here with `events`/ `fields` params ... if the shm-object would contain additional ones ... that would not even be a real problem/error. The `Skeleton::RegisterGeneric()` is more "flexible" ... it just checks, that the things at least exist in shm, which we expect ... if there is even more ... no problem.

My proposal:
- remove `ValidateDataShmObject` and `ValidateControlShmObject`
- if you feel that some consistency checks are missing within `Skeleton::RegisterGeneric()` ... add them there.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -500,7 +687,15 @@ auto Skeleton::RegisterGeneric(const ElementFqId element_fq_id,
                                const size_t sample_size,
                                const size_t sample_alignment) noexcept -> GenericRegistrationResult
 {
-    if (was_old_shm_region_reopened_)
+    if (gateway_shm_ready_)
```
crimson11: This can be optimized:
the if/else branch basically just differ regarding rollback! Only in one case a rollback is done - in the other not.

```
if (use_gateway_forwarded_shm_ || was_old_shm_region_reopened_)
{
   auto [event_data_control_qm, event_data_control_asil_b] =
            memory_manager_.RetrieveEventControlsFromOpenedSharedMemory(element_fq_id);

   auto& event_data_storage = memory_manager_.RetrieveEventDataFromOpenedSharedMemory<std::uint8_t>(element_fq_id);
 if (!use_gateway_forwarded_shm_)
 {
        memory_manager_.RollbackSkeletonTracingTransactions(event_data_control_qm);
        if (event_data_control_asil_b != nullptr)
        {
            memory_manager_.RollbackSkeletonTracingTransactions(*event_data_control_asil_b);
        }
   }
   return {static_cast<void*>(&event_data_storage), event_data_control_qm, event_data_control_asil_b};
}
```

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/shm_path_builder.cpp
Rahul-Sutariya: removed compacting, spilt the pr and updated to: https://github.com/eclipse-score/communication/pull/492

score/mw/com/impl/bindings/lola/shm_path_builder.h
Rahul-Sutariya: updated the doxygen comment and spilt the pr and updated to: https://github.com/eclipse-score/communication/pull/492 https://github.com/eclipse-score/communication/pull/492/changes#diff-23baa8b4571febbd6ac770b38d32efd5fdba6b4a52c5ed12e37d03531ef2ef0dR32-R44

score/mw/com/impl/bindings/lola/skeleton.cpp
Rahul-Sutariya: Adapted the changes, thanks

score/mw/com/impl/bindings/lola/skeleton.cpp
Rahul-Sutariya: remanned gateway_shm_ready_ -> use_gateway_forwarded_shm_, thanks

score/mw/com/impl/bindings/lola/skeleton.h
Rahul-Sutariya: updated the changes, Thanks

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
Rahul-Sutariya: Updated the changes. Thanks

score/mw/com/impl/bindings/lola/skeleton.cpp
Rahul-Sutariya: yes, you are right i feel Skeleton::RegisterGeneric() is enough to verify, removed both functions, Thanks

score/mw/com/impl/bindings/lola/skeleton.cpp
Rahul-Sutariya: Optimized the code, Thanks

crimson11 (approved):

PR #484, opened 2026-05-29
koma : add CodeQL GitHub Actions workflow

Komal362000:
Run CodeQL via Bazel on all relevant C++ targets (//score/...) on every pull_request and push to main, then upload the resulting SARIF file to GitHub Code Scanning via github/codeql-action/upload-sarif.

Without this upload step GitHub has no stored baseline, causing every PR to incorrectly report no new alerts introduced.

Refs: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,89 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+# Workflow: CodeQL analysis on push to main and on demand.
+#
+# Runs CodeQL via Bazel on all relevant C++ targets under //score/...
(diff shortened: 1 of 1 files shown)


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,89 @@
...
+# Runs CodeQL via Bazel on all relevant C++ targets under //score/...
+# and uploads the resulting SARIF file to GitHub Code Scanning so that:
+#   - GitHub tracks findings over time with a stable baseline on main.
+#   - New PRs are compared against that baseline to surface new findings.
+#
+# Why not run on pull_request?
+#   The full CodeQL analysis (169 MISRA rules over //score/...) exceeds
+#   GitHub's 6-hour runner limit and is cancelled before producing output.
+#   Running on push to main ensures a complete, uploadable baseline exists.
+#   GitHub Code Scanning then uses that baseline to flag new findings per PR.
```
LittleHuba: This sounds like we want a smart target-picker.
E.g. a logic that finds out what Bazel targets are affected based on the changes of the PR.

It is important that we run CodeQL on PRs, since the automated way that CodeQL is executed on PRs is not really well-suited for our highly templated code. Prs seldomly change many targets at the same time. So smart target-picking should reduce runtime significantly.

This should be introduced in a follow-up.

Komal362000: Thanks for the feedback. Agreed, we need smart target-picking for PR runs to reduce CodeQL runtime. i will make changes for this .

PR #487, opened 2026-05-29
fix: make Clang-Tidy block early on PRs, not only at gate

Komal362000:
Previously ERRORS_BLOCKING was scoped to errors in PR-changed files only. This caused the 'Fail check' step to be skipped (ERRORS_BLOCKING=0) when errors existed in unchanged files, so the PR check passed but the merge-queue (gate) run failed — feedback arrived too late after reviews/approvals.

Changes:
- Remove pull_request-specific changed-file scoping for ERRORS_BLOCKING. Any clang-tidy error now blocks both PR and gate runs consistently.
- Extract errors-only output to clang_tidy_errors.txt (grep 'error:') so the cause of failure is clearly surfaced without reading a 6MB file.
- Print first 20 errors in the job summary when the check fails.
- Print all errors in the 'Fail check' step log for inline visibility.
- Upload clang_tidy_errors.txt as part of the clang-tidy-findings artifact.
- Rename step: 'Fail check if clang-tidy errors found in changed files' -> 'Fail check if clang-tidy errors found' (reflects new behaviour).

### DIFF
--- a/.github/workflows/clang_tidy.yml
+++ b/.github/workflows/clang_tidy.yml
@@ -11,12 +11,14 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
-# Workflow: Clang-Tidy PR check (voting).
+# Workflow: Clang-Tidy PR check (blocking on changed files).
 #
-# Runs clang-tidy on every pull_request and push to main.
-# - Writes a job summary with error/warning counts.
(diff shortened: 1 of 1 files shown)


gdadunashvili (approved):

bemerybmw (approved):
Reviewed by @gdadunashvili . Approving as code owner

PR #488, opened 2026-05-29
mw/com: Fix clang tidy warnings

bemerybmw:


score/mw/com/test/partial_restart/provider_restart/consumer.cpp
```diff
@@ -128,7 +128,6 @@ void DoConsumerActionsWithProxy(CheckPointControl& check_point_control,
     //            subscription-pending. If detected notify controller, that
     //            checkpoint (2) has been reached.
     // ********************************************************************************
-    auto subscription_state = lola_proxy.simple_event_.GetSubscriptionState();
```
limdor, line 131: It feels strange to remove it. Because the description says that we wait until the state is pending. Wouldn't it make more sense to move this line after `WaitTillServiceDisappears` and then read if the state is pending?

crimson11 (approved):

score/mw/com/test/partial_restart/provider_restart/consumer.cpp
limdor: Clarified offline. We cannot really check that subscription changed because we do not have implemented the subscription state change handler.

PR #494, opened 2026-06-02
fix: bazel apt install with merged usr

lurtz:
Ubuntu 24.04 uses a merged /usr, where e.g. /bin, /lib or /sbin are symlinks into their respective /usr directories. In this case the option mergedusr has to be set to true. Otherwise bazel will produce a container image with missing applications. Which application is missing depends on the order of packages being installed.

I suppose currently this is not a problem, because very few packages are installed and none seem to trigger the errornous behavior.


LittleHuba (approved):

LittleHuba:
Thanks, nice catch!

PR #495, opened 2026-06-02
mw/com: add unit tests for MethodType

KrishaDeshkool:
Add unit tests for to_string(MethodType), covering kMethod/kGet/kSet/kUnknown and the out-of-range "Invalid" fallback. to_string is used in production when streaming the lola UniqueMethodIdentifier.


KrishaDeshkool:
This is done as part of another PR closing this one

PR #496, opened 2026-06-02
mw:com: Clean up Field impl

KrishaDeshkool:
Depends on: https://github.com/eclipse-score/communication/pull/357
Depends on: https://github.com/eclipse-score/communication/pull/534


KrishaDeshkool:
@bemerybmw can you review this?

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -244,6 +244,83 @@ score::Result<void> ExecutePartialRestartLogic(const QualityType quality_type,
...
+        // Defensive programming: This branch should never be hit as this function is only instantiated for
+        // LolaFieldInstanceDeployment and LolaMethodInstanceDeployment.
+        static_assert(always_false_v<LolaInstanceDeploymentType>,
+                      "GetEnabledMethodEntries is not implemented for this type");
+    }
+    return entries;
+}
+
+template <typename DeploymentMap>
+void AppendEnabledMethodIds(
```
bemerybmw: Instead of templating with DeploymentMap, what about templating it with ServiceElementType directly? It follows that we're clearly specifying in the function call exactly what we want i.e. EnabledMethodIds for methods of fields and then the internals of the function get the correct deployment map etc. 
otherwise, it's less clear in the calling code that we're getting the ids for methods / fields (have to read the argument lola_service_instance_deployment.methods_/fields_ to get this info. We then work backwards to get the ServiceElementType from the map provided. 

E.g.
```
template <ServiceElementType kElementType>
void AppendEnabledMethodIds(
    std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>& result,
    const LolaServiceInstanceDeployment& service_instance_deployment,
    const LolaServiceTypeDeployment& service_type_deployment)
{
    auto& deployment_map = [&service_instance_deployment]() -> const auto& {
        if constexpr (kElementType == ServiceElementType::METHOD)
        {
            return service_instance_deployment.methods_;
        }
        else if constexpr (kElementType == ServiceElementType::FIELD)
        {
            return service_instance_deployment.fields_;
        }
        else
        {
            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(false, "Unsupported service element type");
        }
    }();
    for (const auto& [name, deployment] : deployment_map)
    {
        const auto element_id = GetServiceElementId<kElementType>(service_type_deployment, name);
        for (const auto& [method_type, queue_size] : GetEnabledMethodEntries(deployment))
        {
            result.emplace_back(UniqueMethodIdentifier{element_id, method_type}, queue_size);
        }
    }
}
```

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -244,6 +244,83 @@ score::Result<void> ExecutePartialRestartLogic(const QualityType quality_type,
...
+    static constexpr ServiceElementType kType = ServiceElementType::FIELD;
+};
+
+constexpr LolaMethodInstanceDeployment::QueueSize kFieldMethodQueueSize{1U};
+
+template <typename T>
+constexpr bool always_false_v = false;
+
+template <typename LolaInstanceDeploymentType>
+std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> GetEnabledMethodEntries(
```
bemerybmw: The point of template code is to make it generic. Here, we have completely different code for different deployments so better to just use overloads. e.g. 

```
std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> GetEnabledMethodEntries(
    const LolaFieldInstanceDeployment& lola_field_instance_deployment)
{
    std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> entries{};

    for (const auto& [use_if_available, method_type] :
         {std::pair{lola_field_instance_deployment.use_get_if_available_, MethodType::kGet},
          std::pair{lola_field_instance_deployment.use_set_if_available_, MethodType::kSet}})
    {
        if (use_if_available.value_or(false))
        {
            entries.emplace_back(method_type, kFieldMethodQueueSize);
        }
    }
    return entries;
}

std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> GetEnabledMethodEntries(
    const LolaMethodInstanceDeployment& lola_method_instance_deployment)
{
    std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> entries{};
    if (lola_method_instance_deployment.enabled_.value_or(false))
    {
        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
            lola_method_instance_deployment.queue_size_.has_value(),
            "Method instance deployment must contain queue_size on proxy side!");
        entries.emplace_back(MethodType::kMethod, lola_method_instance_deployment.queue_size_.value());
    }
    return entries;
}
```

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -244,6 +244,83 @@ score::Result<void> ExecutePartialRestartLogic(const QualityType quality_type,
...
+        static_assert(always_false_v<LolaInstanceDeploymentType>,
+                      "GetEnabledMethodEntries is not implemented for this type");
+    }
+    return entries;
+}
+
+template <typename DeploymentMap>
+void AppendEnabledMethodIds(
+    const DeploymentMap& deployment_map,
+    std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>& result,
```
bemerybmw, line 286: I'm not a huge fan of output parameters, but I guess that it avoids reallocations. If you want to go this route, then the output param should be the first argument (IMO) and also it should be applied also to GetEnabledMethodEntries.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -244,6 +244,83 @@ score::Result<void> ExecutePartialRestartLogic(const QualityType quality_type,
...
+    static constexpr ServiceElementType kType = ServiceElementType::METHOD;
+};
+
+template <>
+struct ServiceElementTypeFor<LolaFieldInstanceDeployment>
+{
+    static constexpr ServiceElementType kType = ServiceElementType::FIELD;
+};
+
+constexpr LolaMethodInstanceDeployment::QueueSize kFieldMethodQueueSize{1U};
```
bemerybmw: constants should all be together at the top of the file.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -604,38 +681,9 @@ score::Result<void> Proxy::SetupMethods()
...
-            if ((kUseSetIfAvailable) && (proxy_methods_.count({field_id, MethodType::kSet}) != 0U))
-            {
-                enabled_method_data.push_back({{field_id, MethodType::kSet}, kFieldMethodQueueSize});
-            }
-        }
-    }
-    // This check has be done after looping over the fields to add the field methods to the enabled method data because,
-    // if we did this before then even when the fields are configured and when the enabled_method_names are empty we
-    // would do an early return and miss setting up the fields shm.
+    // This check has to be done after looping over the fields to add the field methods to the enabled method data
```
bemerybmw: This comment can now be removed? methods and fields are now checked together

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -764,23 +812,11 @@ Proxy::GetMethodIdAndQueueSizeForEnabledMethods() const
...
-        if (method_deployment.enabled_.has_value() && method_deployment.enabled_.value())
-        {
-            const auto method_id =
-                GetServiceElementId<ServiceElementType::METHOD>(lola_service_type_deployment, method_name);
-
-            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
-                method_deployment.queue_size_.has_value(),
-                "Method instance deployment must contain queue_size on proxy side!");
-            const auto queue_size = method_deployment.queue_size_.value();
+    AppendEnabledMethodIds(
```
bemerybmw: AppendEnabledMethodIdsAndQueueSizes?

MODULE.bazel.lock
```diff
@@ -20,7 +20,6 @@
     "https://bcr.bazel.build/modules/abseil-py/2.1.0/MODULE.bazel": "5ebe5bf853769c65707e5c28f216798f7a4b1042015e6a36e6d03094d94bec8a",
     "https://bcr.bazel.build/modules/abseil-py/2.1.0/source.json": "0e8fc4f088ce07099c1cd6594c20c7ddbb48b4b3c0849b7d94ba94be88ff042b",
     "https://bcr.bazel.build/modules/ape/1.0.1/MODULE.bazel": "37411cfd13bfc28cd264674d660a3ecb3b5b35b9dbe4c0b2be098683641b3fee",
-    "https://bcr.bazel.build/modules/ape/1.0.1/source.json": "96bc5909d1e3ccc4203272815ef874dbfd99651e240c05049f12193d16c1110b",
```
bemerybmw: I guess this was mistakenly pushed?

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -244,6 +244,83 @@ score::Result<void> ExecutePartialRestartLogic(const QualityType quality_type,
...
+        {
+            if (use_if_available.value_or(false))
+            {
+                entries.emplace_back(method_type, kFieldMethodQueueSize);
+            }
+        }
+    }
+    else if constexpr (std::is_same_v<LolaInstanceDeploymentType, LolaMethodInstanceDeployment>)
+    {
+        if (deployment.enabled_.value_or(false))
```
bemerybmw: The default value should be handled by the config parsing. On proxy side, deployment.enabled_
should always be set, so this should be an assertion, not defaulted to false.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -244,6 +244,83 @@ score::Result<void> ExecutePartialRestartLogic(const QualityType quality_type,
...
+{
+    std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> entries;
+
+    if constexpr (std::is_same_v<LolaInstanceDeploymentType, LolaFieldInstanceDeployment>)
+    {
+        for (const auto& [use_if_available, method_type] :
+             {std::pair{deployment.use_get_if_available_, MethodType::kGet},
+              std::pair{deployment.use_set_if_available_, MethodType::kSet}})
+        {
+            if (use_if_available.value_or(false))
```
bemerybmw: The default value should be handled by the config parsing. On proxy side, deployment.use_get_if_available_/use_set_if_available_ should always be set, so this should be an assertion, not defaulted to false.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -381,6 +413,180 @@ TEST_F(ProxyMethodHandlingFixture, SetsInArgsAndReturnStoragesForEachMethodInShm
     }
 }
 
+TEST_F(ProxyMethodHandlingFixture, RegisteredFieldGetAndSetAreCreatedInShmWhenEnabledInDeployment)
+{
+    // Given a field whose Get and Set are enabled in config and registered as proxy methods
+    auto use_get_if_available = true;
+    auto use_set_if_available = true;
+    GivenAConfigurationWithField(use_get_if_available, use_set_if_available)
```
bemerybmw: What about instead of having 3 duplicate tests with different values for use_get_if_available / use_set_if_available, GivenAConfigurationWithField takes a vector of fields with get , set enabled. You can then register 4 different fields with all combinations of get / set in one test and then check that the call queue only contains the fields that were enabled.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -381,6 +413,180 @@ TEST_F(ProxyMethodHandlingFixture, SetsInArgsAndReturnStoragesForEachMethodInShm
...
+              TypeErasedCallQueue::TypeErasedElementInfo{
+                  kEmptyInArgsTypeErasedDataInfo, kValidReturnTypeTypeErasedDataInfo, kDummyQueueSize0},
+              MethodType::kGet}});
+
+    // When setting up methods.
+    // Then the program terminates.
+    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(proxy_->SetupMethods());
+}
+
+TEST_F(ProxyMethodHandlingFixture, FieldGetNulloptInDeploymentConfigIsExcludedEvenWhenRegistered)
```
bemerybmw: As mentioned in the previous commit's comments, the value should always be filled in the config on proxy side, so this should check that an assertion is triggered.

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -381,6 +413,180 @@ TEST_F(ProxyMethodHandlingFixture, SetsInArgsAndReturnStoragesForEachMethodInShm
...
+    // When setting up methods
+    score::cpp::ignore = proxy_->SetupMethods();
+
+    // Then only the Set method is created; the nullopt Get is excluded even though it was registered
+    const auto& method_data = GetMethodDataFromShm();
+    EXPECT_THAT(method_data.method_call_queues_,
+                UnorderedElementsAre(Pair(UniqueMethodIdentifier{kDummyFieldId0, MethodType::kSet}, _)));
+}
+
+TEST_F(ProxyMethodHandlingFixture, FieldSetNulloptInDeploymentConfigIsExcludedEvenWhenRegistered)
```
bemerybmw: As mentioned in the previous commit's comments, the value should always be filled in the config on proxy side, so this should check that an assertion is triggered.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -731,28 +731,14 @@ auto Skeleton::SubscribeMethods(const MethodData& method_data,
...
-            // interface definition. In that case, the skeleton has no handler for it.
-            if ((unique_method_identifier.method_type == MethodType::kGet) ||
-                (unique_method_identifier.method_type == MethodType::kSet))
-            {
-                score::mw::log::LogInfo("lola")
-                    << "Proxy registered a field Get/Set method that is not available on the skeleton side. Skipping.";
-                continue;
-            }
+        // Defensive check for skeleton method population.
+        // The skeleton_methods_ map is populated at skeleton construction time
```
bemerybmw, line 741: maybe also mention that a Skeleton must always provide all methods in its interface. Only on proxy side can we be selective with which method a specific proxy method wants to use.

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: Agreed.

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: ah true. Moved it

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: You're right, it's stale.

MODULE.bazel.lock
KrishaDeshkool: Ah yeah! i will remove it

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: ah! i did not know that.
thanks, i will make this as assert here and update tests .

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: atleast here there is no default value set
`        const auto use_get_if_available = deployment_parser.RetrieveJsonElement<bool>(kFieldUseGetIfAvailableKey);
        const auto use_set_if_available = deployment_parser.RetrieveJsonElement<bool>(kFieldUseSetIfAvailableKey);
`

KrishaDeshkool: is that a mistake there ?

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: hmm., okay,
my idea was we see .field or .method in the call site and we know what we are passing,
and if we add a template like you added we would add a if branch inside.
one pro i see is that we would get rid of the trait helper above which might not be trivial for everyone to reason about.
i will update this.

score/mw/com/impl/bindings/lola/skeleton.cpp
KrishaDeshkool: added a note

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: same technique does not work for GetEnabledMethodEntries, if i make this vector of entries live outside the loop 
```
 std::vector<std::pair<MethodType, LolaMethodInstanceDeployment::QueueSize>> entries{};   
 for (const auto& [name, deployment] : deployment_map)
    {
        const auto element_id = GetServiceElementId<kElementType>(service_type_deployment, name);
        GetEnabledMethodEntries(entries, deployment);
        for (const auto& [method_type, queue_size] : entries)
        {
            result.emplace_back(UniqueMethodIdentifier{element_id, method_type}, queue_size);
        }
    }
```
    
  Then entries had duplicate entries on second run,
  
  if i put this vector inside the loop then we are anyway reallocating each loop, so just leaving this as is.

bemerybmw: I think it should work, something like this:

```
void AppendEnabledMethodIdsAndQueueSizesImpl(
    std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>& result,
    const LolaFieldInstanceDeployment& lola_field_instance_deployment,
    LolaServiceElementId field_id)
{
    for (const auto& [use_if_available, method_type] :
         {std::pair{lola_field_instance_deployment.use_get_if_available_, MethodType::kGet},
          std::pair{lola_field_instance_deployment.use_set_if_available_, MethodType::kSet}})
    {
        SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(
            use_if_available.has_value(),
            "Field instance deployment must contain use_get/set_if_available on proxy side!");
        if (use_if_available.value())
        {
            result.emplace_back(UniqueMethodIdentifier{field_id, method_type}, kFieldMethodQueueSize);
        }
    }
}

void AppendEnabledMethodIdsAndQueueSizesImpl(
    std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>& result,
    const LolaMethodInstanceDeployment& lola_method_instance_deployment,
    LolaServiceElementId method_id)
{
    // At this point the enabled_ field must be set by the config parser.
    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(lola_method_instance_deployment.enabled_.has_value(),
                                                      "Method instance deployment must contain enabled on proxy side!");

    if (lola_method_instance_deployment.enabled_.value())
    {
        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
            lola_method_instance_deployment.queue_size_.has_value(),
            "Method instance deployment must contain queue_size on proxy side!");
        result.emplace_back(UniqueMethodIdentifier{method_id, MethodType::kMethod},
                            lola_method_instance_deployment.queue_size_.value());
    }
}

template <ServiceElementType kElementType>
void AppendEnabledMethodIdsAndQueueSizes(
    std::vector<std::pair<UniqueMethodIdentifier, LolaMethodInstanceDeployment::QueueSize>>& result,
    const LolaServiceInstanceDeployment& service_instance_deployment,
    const LolaServiceTypeDeployment& service_type_deployment)
{
    auto& deployment_map = [&service_instance_deployment]() -> const auto& {
        if constexpr (kElementType == ServiceElementType::METHOD)
        {
            return service_instance_deployment.methods_;
        }
        else if constexpr (kElementType == ServiceElementType::FIELD)
        {
            return service_instance_deployment.fields_;
        }
        else
        {
            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(false, "Unsupported service element type");
        }
    }();
    for (const auto& [name, deployment] : deployment_map)
    {
        const auto element_id = GetServiceElementId<kElementType>(service_type_deployment, name);
        AppendEnabledMethodIdsAndQueueSizesImpl(result, deployment, element_id);
    }
}
```

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: reworked the configs to change the defaults here https://github.com/eclipse-score/communication/pull/668

score/mw/com/impl/bindings/lola/proxy.cpp
KrishaDeshkool: yeah, this should work, let me update it

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -374,6 +439,96 @@ TEST_F(ProxyMethodHandlingFixture, SetsInArgsAndReturnStoragesForEachMethodInShm
...
+    //  field3 (both disabled) is absent
+    const auto& method_data = GetMethodDataFromShm();
+    EXPECT_THAT(method_data.method_call_queues_,
+                UnorderedElementsAre(Pair(UniqueMethodIdentifier{kDummyFieldId0, MethodType::kGet}, _),
+                                     Pair(UniqueMethodIdentifier{kDummyFieldId0, MethodType::kSet}, _),
+                                     Pair(UniqueMethodIdentifier{kDummyFieldId1, MethodType::kGet}, _),
+                                     Pair(UniqueMethodIdentifier{kDummyFieldId2, MethodType::kSet}, _)));
+}
+
+TEST_F(ProxyMethodHandlingFixture, FieldGetNotRegisteredDoesAbort)
```
bemerybmw, line 492: Name should be clearer. E.g. FieldEnabledInConfigurationWhichIsNotCreatedInShmTerminates

score/mw/com/impl/bindings/lola/proxy_method_handling_test.cpp
```diff
@@ -374,6 +439,96 @@ TEST_F(ProxyMethodHandlingFixture, SetsInArgsAndReturnStoragesForEachMethodInShm
...
+              TypeErasedCallQueue::TypeErasedElementInfo{
+                  kValidInArgsTypeErasedDataInfo, kEmptyReturnTypeTypeErasedDataInfo, kDummyQueueSize0},
+              MethodType::kSet}});
+
+    // When setting up methods.
+    // Then the program terminates.
+    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(proxy_->SetupMethods());
+}
+
+TEST_F(ProxyMethodHandlingFixture, FieldSetNotRegisteredDoesAbort)
```
bemerybmw, line 512: name should be clearer (see above)

bemerybmw (approved):

PR #501, opened 2026-06-03
Added clang-tidy check/fix targets, autofix config, and quality docs

AAmbuj:
Added clang-tidy check/fix targets, autofix config, and quality docs

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -13,7 +13,9 @@
 
 load("@aspect_rules_lint//format:defs.bzl", "format_multirun", "format_test")
 load("@rules_python//python:pip.bzl", "compile_pip_requirements")
+load("@rules_shell//shell:sh_binary.bzl", "sh_binary")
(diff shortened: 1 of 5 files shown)


BUILD
```diff
@@ -60,3 +62,19 @@ format_test(
...
     workspace = "//:LICENSE",
 )
+
+use_clang_tidy_targets()
+
+sh_binary(
+    name = "clang-tidy.fix",
+    srcs = [":clang-tidy.fix_script"],
+    target_compatible_with = ["@platforms//os:linux"],
+    visibility = ["//visibility:public"],
```
castler: We should not make this public visible, as we do not want that any body depend on it. Please fix in follow up (same below)

castler (dismissed):

castler (dismissed):

castler (approved):

PR #508, opened 2026-06-08
Initial implementation for HV LoLa gateway sample transport layer

SebSparrowHawk:
A sample implementation of a transport layer for the Hypervisor LoLa gateway.

### DIFF
--- a/score/mw/com/gateway/transport_layer/sample/README.md
+++ b/score/mw/com/gateway/transport_layer/sample/README.md
@@ -1,0 +1,43 @@
+<!---- 
+*******************************************************************************
+Copyright (c) 2026 Contributors to the Eclipse Foundation
+
+See the NOTICE file(s) distributed with this work for additional
+information regarding copyright ownership.
+
+This program and the accompanying materials are made available under the
+terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+*******************************************************************************
+-->
+
+# Gateway Transport Layer Sample Implementation
+
+The functionality a `Gateway Transport Layer` has to provide can be roughly divided into two parts:
+
+1. Provide a message transport between the gateways, so that the gateways instances can interact with specific control 
+messages.
+2. Provide the data exchange for the service-instance related data. In case of a non-copying shared memory setup, 
+this means to make shared-memory accessible between VMs. In case of a copying `Transport Layer` this means setting up DMA 
+data transfers and the like.
+
+
+For the first part this module provides a sample implementation of a mw::com gateway's transport layer based on regular 
+POSIX network sockets for communication. Since this communication mechanism should be available in most HyperVisor 
+environments, we decided to add this as a sample implementation. 
+
+## Vendor-specific implementation details
+
+Our sample Transport Layer implementation returns `true` for an API call of `IsMemorySharingSupported`. So it is a sample for a `Gateway Transport Layer` 
+for an inter-VM setup, where memory sharing between VMs is possible.
+It can not fully implement the second part (see above) with regard to making shared-memory objects visible/accessible 
+between gateway domains (VMs) as this is vendor/hypervisor/OS specific. Thus, there are TODOs in the code.
+
+However, the basic assumption in this sample implementation is:
+
+- in the `ProvideService` request, the destination Transport Layer receives from the source side, there is an 
+`InstanceSpecifier` and the sizes of the control/data shm-objects.
+- from the `InstanceSpecifier` the Transport Layer can deduce the identification of the shared-memory objects created on the source side.
+- it can verify, whether the shm-object sizes communicated in the request match with the shm-objects it opened.
--- a/score/mw/com/gateway/transport_layer/sample/sample_hypervisor_transport.cpp
+++ b/score/mw/com/gateway/transport_layer/sample/sample_hypervisor_transport.cpp
@@ -1,0 +1,284 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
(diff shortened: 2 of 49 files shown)


score/mw/com/gateway/transport_layer/sample/sample_hypervisor_transport.h
```diff
@@ -0,0 +1,94 @@
...
+    /// @param message received message from the BidirectionalTransport (transport_ member variable)
+    void OnMessageReceived(std::unique_ptr<TransportMessage> message);
+    /// \brief Creates the shared memory region based on the given arguments, prior to the gateway application accessing
+    /// the shared memory region, i.e. before ProvideService() has been called.
+    void PreCreateInterVmSharedMemory(const impl::InstanceSpecifier& specifier,
+                                      std::uint32_t shm_control_size,
+                                      std::uint32_t shm_data_size);
+
+    GatewayCore& gateway_app_;
+    std::unique_ptr<IBidirectionalTransport> transport_;
```
crimson11: rebame to `message_transport_` -> makes reading code easier ... because now we have "transport" everywhere ... it is good to know, that this is the "endpoint" for the transport of messages.

score/mw/com/gateway/transport_layer/sample/sample_hypervisor_transport.cpp
```diff
@@ -0,0 +1,255 @@
...
+
+score::ResultBlank SampleHyperVisorTransport::Setup()
+{
+    transport_->SetMessageHandler([this](std::unique_ptr<TransportMessage> message) {
+        OnMessageReceived(std::move(message));
+    });
+    return transport_->Setup();
+}
+
+void SampleHyperVisorTransport::OnMessageReceived(std::unique_ptr<TransportMessage> message)
```
crimson11, line 96: Compact this method by calling explicit subfunctions for each message.

I.e. in case of
`if (message->GetType() == MessageType::kProvideServiceRequest)` call
`HandleProvideServiceRequest(request);`

score/mw/com/gateway/transport_layer/sample/README.md
```diff
@@ -0,0 +1,32 @@
...
+for communication. Since this communication mechanism should be available in most HyperVisor environments, we decided to
+add this as a sample implementation. 
+
+## Vendor-specific implementation details
+
+Parts that interact with the underlying shared memory have not been implemented as part of this sample implementation,
+as these parts are vendor-specific and depend on the used HyperVisor and its shared memory abilities.
+This includes the following methods: 
+
+- `ShmPaths score::mw::com::gateway::ResolveShmPaths(const impl::InstanceSpecifier&)` where a path to access the HV-shared memory region has to be created.
```
crimson11: This isn't quite correct, as you implemented `ResolveShmPaths` and `GetShmSizes` ....

score/mw/com/gateway/transport_layer/sample/README.md
```diff
@@ -0,0 +1,32 @@
...
+terms of the Apache License Version 2.0 which is available at
+https://www.apache.org/licenses/LICENSE-2.0
+
+SPDX-License-Identifier: Apache-2.0
+*******************************************************************************
+-->
+
+# Gateway Transport Layer Sample Implementation
+
+This module provides a sample implementation of a mw::com gateway's transport layer based on regular POSIX network sockets
```
crimson11: I would start maybe with the following paragraph:

The functionality a Gateway Transport Layer has to provide can be roughly divided into two parts:

1. Provide a message transport between the gateways, so that the gateways instances can interact with specific control messages.
2. Provide the data exchange for the service-instance related data. In case of a non-copying shared memory setup, this means to make shared-memory accessible between VMs. I case of copying Transport Layer this means setting up DMA data transfers and the like.

For the 1st part this module provides a sample ...

score/mw/com/gateway/transport_layer/sample/README.md
```diff
@@ -0,0 +1,32 @@
...
+for communication. Since this communication mechanism should be available in most HyperVisor environments, we decided to
+add this as a sample implementation. 
+
+## Vendor-specific implementation details
+
+Parts that interact with the underlying shared memory have not been implemented as part of this sample implementation,
```
crimson11: Maybe document:

Our sample Transport Layer implementation returns `true` in API `IsMemorySharingSupported`. So it is a sample for a Gateway Transport Layer for an inter-VM setup, where sharing of memory between VMs is possible.
It can't fully implement the part 2 (see above) with regard to making shared-memory objects visible/accessible between gateway domains (VMs) as this is vendor/hypervisor/OS specific. Thus, there are a TODOs in the code.

However, the basic assumption in this sample implementation is:
- in the `ProvideService` request, the destination Transport Layer receives from the source side, there is an `InstanceSpecifier` and the sizes for the control/data shm-objects.
- from the `InstanceSpecifier` the Transport Layer can deduce the identification of the shared-memory objects created on the source side.
- it can verify, whether the shm-object sizes communicated in the request match with the shm-objects he opened.

crimson11 (changes requested):

score/mw/com/gateway/transport_layer/sample/sample_hypervisor_transport.h
SebSparrowHawk: Applied that change.

score/mw/com/gateway/transport_layer/sample/sample_hypervisor_transport.cpp
SebSparrowHawk: Applied that change.

score/mw/com/gateway/transport_layer/sample/README.md
SebSparrowHawk: I removed that part and switched to your recommendation, mentioning the limitations and expectations for that part of the implementation.

crimson11 (dismissed):

crimson11 (dismissed):

crimson11 (approved):

PR #516, opened 2026-06-09
message_passing: create feature requirements

gdadunashvili:


score/message_passing/dependability/requirements/feature_requirements.trlc
```diff
@@ -19,9 +19,60 @@ import ScoreReq
...
+
+ScoreReq.FeatReq SynchronousAndAsynchronousCommunicationModes {
+    description = "The communication shall support synchronous and asynchronous communication."
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.SystemMessagingProtocol@1]
+    version = 1
+}
+
+ScoreReq.FeatReq FriendlyToSingletonFreeBoundedMonotonicMemoryAllocationAndResourceMockInjection {
+    description = "bazel-bin/score/message_passing/dependability/dependable_element_message_passing_index_report.html"
```
castler: Mhh?

score/message_passing/dependability/requirements/feature_requirements.trlc
```diff
@@ -19,9 +19,79 @@ import ScoreReq
...
+
+ScoreReq.FeatReq AsynchronousCommunicationMode {
+    description = "The communication shall support asynchronous communication."
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.SystemMessagingProtocol@1]
+    version = 1
+}
+
+ScoreReq.FeatReq FriendlyToSingletonFreeImplementations {
+    description = "The Message Passing library design (and the design of the interfaces in particular) shall be friendly to singleton-free implementations."
```
gdadunashvili: Not clear what friendly means. We need to state a testable expectation here

gdadunashvili: Can this be replaced with shall allow singleton-free implementation?

score/message_passing/dependability/assumed_system/assumed_system_requirements.trlc
```diff
@@ -19,7 +19,7 @@ import ScoreReq
 // System level requirements for MessagePassing SEooC
 ///////////////////////////////
 
-ScoreReq.AssumedSystemReq ASR_MP_001 {
+ScoreReq.AssumedSystemReq SystemMessagingProtocol {
     description = "The system shall provide a mechanism for inter-process communication via a client-server messaging protocol."
     safety = ScoreReq.Asil.B
     rationale = "Application processes need to exchange requests and responses reliably across process boundaries."
```
crimson11: `reliably` - is this a well-known formal definition?
Why I am asking:
With tagging this req. as `safety = Asil.B` - do we expect, that the following ISO 26262 communication fault modes shall be handled/avoided, when we say `reliably`? These are the 26262 fault modes for communication:

- Repetition: Sending the same message multiple times.
- Loss: Dropping or failing to receive a message.
- Delay: Receiving a message outside its expected time window.
- Insertion: Unwarranted addition or injection of a message.
- Incorrect Addressing: Delivering a message to the wrong destination.Incorrect Sequence: Messages arriving out of chronological order.
- Corruption: Bit-level errors or payload degradation within the message.
- Masquerading: An unauthorized sender appearing as a valid transmitter.

I do have some doubts with `Loss` and `Delay` ... I guess this we can hardly avoid ... at least our current implementation can't avoid ....

score/message_passing/dependability/requirements/feature_requirements.trlc
```diff
@@ -19,9 +19,79 @@ import ScoreReq
...
+
+ScoreReq.FeatReq SmallDataLowLatencyCommunication {
+    description = "The system shall be optimized for the usage of small data and enable low-latency inter-process communication between dedicated endpoints."
+    safety = ScoreReq.Asil.QM
+    derived_from = [MessagePassingSEooC.SystemMessagingProtocol@1]
+    version = 1
+}
+
+ScoreReq.FeatReq UnidirectionalAndTwoWayCommunicationModes {
+    description = "The communication shall support unidirectional and two-way communication."
```
crimson11: since you already introduced the notion `connection` above ... I would rephrase it to:

> The communication shall support unidirectional and bidirectional communication on a given connection.

score/message_passing/dependability/requirements/feature_requirements.trlc
crimson11: agree with you! Why not write:
> The implementation shall not use any singletons
rationale: Resource management and lifecycle aspects for singleton implementations are hard to track.

score/message_passing/dependability/requirements/feature_requirements.trlc
```diff
@@ -19,9 +19,79 @@ import ScoreReq
...
+
+ScoreReq.FeatReq FriendlyToSingletonFreeImplementations {
+    description = "The Message Passing library design (and the design of the interfaces in particular) shall be friendly to singleton-free implementations."
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.SystemMessagingProtocol@1]
+    version = 1
+}
+
+ScoreReq.FeatReq AllowsResourceMockInjection {
+    description = "The Message Passing library design shall allwo bounded monotonic memory allocation."
```
crimson11: allwo -> allow

Why we are only talking about, what the design should allow? Why not require, that the implementation shall apply "bounded monotonic memory allocation"?

score/message_passing/dependability/requirements/feature_requirements.trlc
```diff
@@ -19,9 +19,79 @@ import ScoreReq
...
+
+ScoreReq.FeatReq UnidirectionalAndTwoWayCommunicationModes {
+    description = "The communication shall support unidirectional and two-way communication."
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.SystemMessagingProtocol@1]
+    version = 1
+}
+
+ScoreReq.FeatReq SynchronousCommunicationMode {
+    description = "The communication shall support synchronous communication."
```
crimson11: Do we need to be more concise regarding synchronous and asynchronous communication? TBH: Imho there are different things to distinguish:

- what does synchronous/asynchronous mean in case of `unidirectional` communication?
- what does synchronous/asynchronous mean in case of `bidirectional` communication in the specific case of a request/response (in message passing we call it `SendWaitReply`)?

I.e. if we would require an asynchronous communication for bidirectional send/reply functionality ... we don't have it currently! So - with this requirement, someone would need to implement in message-passing 2.0. Are we aware?
Even for unidirectional communication, I guess it isn't clear, what asynchronous means? Imho, there are at least 3 different cases:

1. The async call to send a unidirectional message, directly returns, after the message_passing layer accepted the message. (the message did not even reach the other/remote receiver process and there is also no guarantee, that it ever does)
2. The async call to send a unidirectional message returns, after the message_passing layer was able to transfer the message to the other side. The message did not yet receive the handler for it (it is just queued somewhere) and there is also no guarantee, that it ever does get handled.
3. The async call to send a unidirectional message returns, after the message_passing layer was able to transfer the message to the other side and also a suitable handler to process it has been found ... it has not been processed and it might be never processed.

So - again, what is exactly our definition of async vs. sync in this case?

crimson11 (dismissed):

castler (approved):

PR #523, opened 2026-06-10
Rust::com enable the CI Job for rust lint check

bharatGoswami8:
**GitHub Actions workflow that runs Rust linter on pull requests to enforce code quality standards.**

Key Features

- Only runs when PR has `.rs` rust files or workflow of lint changed
- Only fails on warnings/errors in files you changed, not pre-existing issues
- Treats Warnings as Errors
- Job summary shows issue count + code snippets from changed files
- Full Build log available for download

When It Runs

- Pull requests with `.rs` rust files or workflow of lint changes
- Pushes to main
- Merge queue checks

Fail Job Example available here - https://github.com/bharatGoswami8/communication/actions/runs/27337081555/job/80763973850?pr=1

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -131,3 +131,6 @@ try-import %workspace%/.bazelrc.ai_checker
 
 # Enable user-defined configs
 try-import %workspace%/user.bazelrc
+
+# Import Clippy linting configuration for Rust
+import %workspace%/quality/static_analysis/clippy.bazelrc
--- a/.github/workflows/clippy_lint.yml
+++ b/.github/workflows/clippy_lint.yml
@@ -13,11 +13,14 @@
 
 # Workflow: Clippy PR check
 #
-# Runs Clippy on pull requests, pushes to main, and merge groups when Rust files or workflow changes.
-# - For PRs: Runs if PR contains changes to .rs files or this workflow file
+# Runs Clippy on every pull_request, push to main, and merge_group.
 # - Fails if Clippy errors or warnings are found in Rust files changed by the PR.
+# - The same changed-files scope is applied for both pull_request and
+#   merge_group (gate) so that a passing PR check guarantees a passing gate.
 # - Writes a job summary with error/warning counts.
 # - Uploads clippy log as a downloadable artifact.
+#
+# Cache is saved only on pushes to main.
 
 name: Clippy Check
 
@@ -60,6 +63,7 @@ jobs:
     runs-on: ubuntu-24.04
     permissions:
       contents: read
+      actions: write # Needed to delete old bazel caches
     outputs:
       artifact-name: clippy-findings
       errors: ${{ steps.findings.outputs.errors }}
@@ -106,12 +110,12 @@ jobs:
       - name: Run Clippy via Bazel
         id: run-clippy
         if: steps.check-changes.outputs.should-run == 'true'
+        # continue-on-error so we can collect findings and upload the artifact
+        # even when the Bazel invocation exits non-zero.
         continue-on-error: true
         run: |
           bazel build --config=clippy //score/... \
             2>&1 | tee clippy_raw.log
-          # Capture the exit code from bazel (PIPESTATUS[0]), not tee
-          echo "bazel_exit_code=${PIPESTATUS[0]}" >> $GITHUB_OUTPUT
       - name: Collect findings
         id: findings
         if: steps.check-changes.outputs.should-run == 'true'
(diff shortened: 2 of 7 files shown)


LittleHuba (commented):
Hi,
I'll do some general feedback first.
Please change the trigger conditions. Avoid hiding this behind the rust-api label, since this puts load on maintainers to label PRs correctly. We only use the labeling approach for QNX since it helps with the security implications of pull-request-target triggers.

Better filter PRs for changes to specific path patterns. E.g. PRs that touch files with ending `.rs`. That would achieve the same but automatically.

bharatGoswami8:
> Hi, I'll do some general feedback first. Please change the trigger conditions. Avoid hiding this behind the rust-api label, since this puts load on maintainers to label PRs correctly. We only use the labeling approach for QNX since it helps with the security implications of pull-request-target triggers.
> 
> Better filter PRs for changes to specific path patterns. E.g. PRs that touch files with ending `.rs`. That would achieve the same but automatically.

@LittleHuba , I have updated pre-check condition now Job will run only when either rust files or workflow of lint is changed or updated on PR.

.bazelrc
```diff
@@ -120,3 +120,8 @@ build --tool_java_runtime_version=remotejdk_21
 
 # Import AI checker custom configuration
 try-import %workspace%/.bazelrc.ai_checker
+
+#Clippy linting for Rust
+build --aspects=@score_rust_policies//clippy:linters.bzl%clippy_strict
+build --output_groups=+rules_lint_human
+build:lint --@aspect_rules_lint//lint:fail_on_violation=true
```
LittleHuba: Just calling the config `lint` invites future name clashes.
Why not call it `clippy`? this follows the concept we also use for clang-tidy.

Also scoping all three config lines to `test:clippy` makes more sense. We want to avoid worldwide flags as much as we can.

Also, this config ideally goes into our quality/static_analysis directory. Look for the clang-tidy config as an example.

MODULE.bazel
```diff
@@ -23,6 +23,7 @@ bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "rules_python", version = "1.8.5")
 bazel_dep(name = "rules_rust", version = "0.68.1-score")
+bazel_dep(name = "score_rust_policies", version = "0.0.5")
```
LittleHuba: Is this really a non-dev-dependency?
You only need this for clippy. And clippy is only used by developers and not our users.

.github/workflows/clippy_lint.yml
```diff
@@ -0,0 +1,302 @@
...
+          elif [[ "$EVENT_NAME" == "push" ]]; then
+            # For pushes to main, always run
+            should_run='true'
+          elif [[ "$EVENT_NAME" == "workflow_call" ]]; then
+            # For workflow_call, always run (manual or nightly)
+            should_run='true'
+          fi
+          
+          echo "should-run=${should_run}" >> $GITHUB_OUTPUT
+          echo "Clippy check: should-run=${should_run}"
```
LittleHuba: Please make it reusable and align it with the clang-format and clang-tidy workflows.

.github/workflows/clippy_lint.yml
```diff
@@ -0,0 +1,302 @@
...
+
+  clippy:
+    needs: precheck
+    if: ${{ needs.precheck.outputs.should-run == 'true' }}
```
LittleHuba: This will make this workflow impossible to make mandatory.

Once we make it mandatory, it must always run. So having an if condition here, would block any PR that does not touch a Rust file indefinitely blocked by the CI because the merge-queue would get a timeout.

Instead, run the action and return all good when no files were touched. If files are touched, do the actual check.

.github/workflows/clippy_lint.yml
```diff
@@ -0,0 +1,302 @@
...
+          disk-cache: "clippy"
+          repository-cache: true
+          cache-optimized: true
+          cache-save: ${{ github.ref == 'refs/heads/main' }}
+
+      - name: Run Clippy via Bazel
+        id: run-clippy
+        continue-on-error: true
+        run: |
+          bazel build --config=lint //score/mw/com/... \
```
LittleHuba: Even though we do not have Rust targets outside //score/mw/com currently, people will likely forget to extend the scope when they add such code.

Since we use aspec_rules_lint for this, we anyway have a fast aspect-based way of finding out what targets to check. So just extend the scope to our complete relevant scope.
```suggestion
          bazel build --config=lint //score/... \
```

LittleHuba (changes requested):
I see a lot of code duplication with the clang-tidy and the codeql workflows. We should try to reduce this.

One option is to work on https://github.com/eclipse-score/communication/issues/554

.bazelrc
bharatGoswami8: Moved this into `static_analysis` folder but i feel keeping build is fine for clippy as currently few of Rust target has `manual` bezel target tag.
And if no test target added by developer (may be initial Implementation PR kind of), then it may not give clippy warnings.

.github/workflows/clippy_lint.yml
bharatGoswami8: Removed the precheck and changed as per your suggestion.

.github/workflows/clippy_lint.yml
bharatGoswami8: Updated.

MODULE.bazel
bharatGoswami8: It is a dev-dependency, i have updated flag.

.github/workflows/clippy_lint.yml
bharatGoswami8: I have created `.github/actions/setup-bazel-lint/action.yml` but only part we kept Bazel setup block which was common but other detail both workflow have different as per the processing.

bharatGoswami8: Other duplicate kind of steps in `clang_tidy.yml `and` clippy_lint.yml`  like - `workflow_call` outputs and `env:`, I think we can not make function like replacement for those as Github action YAML does not support.

bharatGoswami8:
https://github.com/eclipse-score/communication/pull/594 Created PR with score workflow.

PR #527, opened 2026-06-11
feat: add strict/permissive compiler warning profiles

AAmbuj:
Introduce configurable warning features for GCC and Clang toolchains with two build modes controlled via --config=strict and --config=permissive.

- Add quality/compiler_warnings/{BUILD,clang/BUILD,gcc/BUILD} with cc_feature definitions for minimal, strict, and additional warnings
- Configure LLVM toolchain with extra_known_features (toolchains_llvm 1.7.0)
- Configure GCC toolchain with extra_cxxflags baseline matching minimal_warnings
- Add selective -Wno-error overrides in .bazelrc for permissive mode
- Refactor score/message_passing BUILD files to use COMPILER_WARNING_FEATURES
- Add FP/TP test targets with build_test verification
- Document usage in quality/quality.md

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -64,12 +64,20 @@ bazel_dep(name = "gcc_toolchain", version = "0.9.0", dev_dependency = True)
 gcc_toolchains = use_extension("@gcc_toolchain//toolchain:module_extensions.bzl", "gcc_toolchains", dev_dependency = True)
 gcc_toolchains.toolchain(
     name = "gcc_toolchain_x86_64",
+    # TODO: Add extra_known_features once gcc_toolchain supports it.
+    # See: https://github.com/f0rmiga/gcc-toolchain/issues/219
     extra_cxxflags = [
-        "-fdiagnostics-color=always",  # Adds copt to always use coloring in build output
-        "-Wsign-conversion",
-        "-Wswitch-enum",
-        "-Werror",
-        "-Wno-error=deprecated-declarations",
+        # Diagnostics
+        "-fdiagnostics-color=always",
+        # Minimal warning baseline (mirrors quality/compiler_warnings/gcc minimal_warnings)
+        # Ref: https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html
+        "-Wall",  # Foundation warning set (MISRA Dir 1.1)
+        "-Wswitch-enum",  # Warn on unhandled enum in switch (MISRA Rule 9.6.4)
+        "-Wno-error=deprecated-declarations",  # Allow deprecated API as warning
+        "-Wno-error=cpp",  # Allow #warning as warning
+        "-Wno-maybe-uninitialized",  # GCC 8/9 false positives (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=80635)
+        "-Wno-free-nonheap-object",  # GCC false positive (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=99578)
+        "-Wunused-but-set-parameter",  # Dead store detection (MISRA Rule 0.1.1)
     ],
     extra_ldflags = [
         "-lstdc++",
@@ -84,8 +92,31 @@ use_repo(gcc_toolchains, "gcc_toolchain_x86_64")
 bazel_dep(name = "score_bazel_cpp_toolchains", version = "0.5.4", dev_dependency = True)
 
 score_gcc = use_extension("@score_bazel_cpp_toolchains//extensions:gcc.bzl", "gcc", dev_dependency = True)
+
+# Minimal warning baseline for the S-CORE GCC/QCC toolchains.
+# Mirrors //quality/compiler_warnings/gcc:minimal_warnings. The S-CORE GCC
+# toolchain extension does not consume cc_features, so the baseline is applied
+# directly as compiler flags. Warning-as-error enforcement stays in the
+# score_communication_* cc_features (currently consumed by the LLVM toolchain).
+SCORE_GCC_MINIMAL_WARNING_COMPILE_FLAGS = [
+    "-Wall",
+    "-Wno-error=deprecated-declarations",
+    "-Wno-error=cpp",
+    "-Wno-format-y2k",
+    "-Wno-free-nonheap-object",
+    "-Wno-maybe-uninitialized",
+    "-Wunused-but-set-parameter",
+]
+
+SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS = [
+    "-Wno-literal-suffix",
+    "-Wno-noexcept-type",
+]
+
 score_gcc.toolchain(
     name = "score_gcc_x86_64_toolchain",
+    extra_compile_flags = SCORE_GCC_MINIMAL_WARNING_COMPILE_FLAGS,
+    extra_cxx_compile_flags = SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS,
     target_cpu = "x86_64",
     target_os = "linux",
     use_base_constraints_only = True,
@@ -94,6 +125,8 @@ score_gcc.toolchain(
 )
 score_gcc.toolchain(
     name = "score_gcc_aarch64_toolchain",
+    extra_compile_flags = SCORE_GCC_MINIMAL_WARNING_COMPILE_FLAGS,
+    extra_cxx_compile_flags = SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS,
     target_cpu = "aarch64",
     target_os = "linux",
     use_base_constraints_only = True,
@@ -102,6 +135,8 @@ score_gcc.toolchain(
 )
 score_gcc.toolchain(
     name = "score_qcc_x86_64_toolchain",
+    extra_compile_flags = SCORE_GCC_MINIMAL_WARNING_COMPILE_FLAGS,
+    extra_cxx_compile_flags = SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS,
     sdp_version = "8.0.4",
     target_cpu = "x86_64",
     target_os = "qnx",
@@ -110,6 +145,8 @@ score_gcc.toolchain(
 )
 score_gcc.toolchain(
     name = "score_qcc_aarch64_toolchain",
+    extra_compile_flags = SCORE_GCC_MINIMAL_WARNING_COMPILE_FLAGS,
+    extra_cxx_compile_flags = SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS,
     sdp_version = "8.0.4",
     target_cpu = "aarch64",
     target_os = "qnx",
@@ -239,26 +276,23 @@ llvm = use_extension(
     dev_dependency = True,
 )
 llvm.toolchain(
-    compile_flags = {"": [
-        "-march=nehalem",
-        "-ffp-model=strict",
-        # Security
-        "-U_FORTIFY_SOURCE",  # https://github.com/google/sanitizers/issues/247
-        "-fstack-protector",
-        "-fno-omit-frame-pointer",
-        # Diagnostics
-        "-fcolor-diagnostics",
-        "-Wno-deprecated-declarations",
-        "-Wno-error=self-assign-overloaded",
-        "-Wthread-safety",
-    ]},
     cxx_standard = {"": "c++17"},
+    extra_enabled_features = [
+        "//quality/compiler_warnings:default_flags",
+        "//quality/compiler_warnings:default_link_flags",
+        "//quality/compiler_warnings/clang:minimal_warnings",
+    ],
     extra_known_features = [
         "@score_cpp_policies//sanitizers/features:debug_symbols",
         "@score_cpp_policies//sanitizers/features:asan",
         "@score_cpp_policies//sanitizers/features:ubsan",
         "@score_cpp_policies//sanitizers/features:lsan",
         "@score_cpp_policies//sanitizers/features:tsan",
+        "//quality/compiler_warnings:treat_warnings_as_errors",
+        "//quality/compiler_warnings:additional_warnings",
+        "//quality/compiler_warnings/clang:strict_warnings",
+        "//quality/compiler_warnings/clang:strict_warnings_no_error",
+        "//quality/compiler_warnings/clang:third_party_warnings",
     ],
     link_libs = {"": [
         "-lrt",
@@ -326,6 +360,7 @@ download_file(
 
 bazel_dep(name = "aspect_rules_lint", version = "2.5.0", dev_dependency = True)
 bazel_dep(name = "rules_shell", version = "0.6.1", dev_dependency = True)
+bazel_dep(name = "rules_build_error", version = "0.11.0", dev_dependency = True)
 bazel_dep(name = "google_benchmark", version = "1.9.5", dev_dependency = True)
 bazel_dep(name = "buildifier_prebuilt", version = "8.5.1", dev_dependency = True)
 bazel_dep(name = "hedron_compile_commands", dev_dependency = True)
--- a/quality/compiler_warnings/BUILD
+++ b/quality/compiler_warnings/BUILD
@@ -1,0 +1,138 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+# Common compiler warning features shared across GCC and Clang.
+
+load("@rules_cc//cc/toolchains:args.bzl", "cc_args")
+load("@rules_cc//cc/toolchains:feature.bzl", "cc_feature")
+
+cc_args(
+    name = "default_compile_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-march=nehalem",
+        "-ffp-model=strict",
+        # Undefine _FORTIFY_SOURCE to avoid conflict with sanitizers.
+        # See https://github.com/google/sanitizers/issues/247
+        "-U_FORTIFY_SOURCE",
+        "-fstack-protector",
+        "-fno-omit-frame-pointer",
+        "-fcolor-diagnostics",
+    ],
+)
+
+cc_feature(
+    name = "default_flags",
+    args = [
+        ":default_compile_args",
+    ],
+    feature_name = "score_communication_default_flags",
+    visibility = ["//visibility:public"],
+)
+
+cc_args(
+    name = "default_link_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:link_actions",
+    ],
+    args = [
+        "-lrt",
+        # Required for targets using std::atomic.
+        "-latomic",
+    ],
+)
+
+cc_feature(
+    name = "default_link_flags",
+    args = [
+        ":default_link_args",
+    ],
+    feature_name = "score_communication_default_link_flags",
+    visibility = ["//visibility:public"],
+)
+
+cc_args(
+    name = "treat_warnings_as_errors_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-Werror",
+    ],
+)
+
+cc_feature(
+    name = "treat_warnings_as_errors",
+    args = [
+        ":treat_warnings_as_errors_args",
+    ],
+    feature_name = "score_communication_treat_warnings_as_errors",
+    visibility = ["//visibility:public"],
+)
+
+cc_args(
+    name = "minimal_warnings_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-Wall",
+        # Keep #warning visible without failing the build.
+        "-Wno-error=cpp",
+        "-Wno-error=deprecated-declarations",
+    ],
+    visibility = ["//quality/compiler_warnings:__subpackages__"],
+)
+
+cc_args(
+    name = "strict_warnings_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-Wextra",
+        "-Wpedantic",
+        "-Wconversion",
+        "-Wsign-conversion",
+    ],
+    visibility = ["//quality/compiler_warnings:__subpackages__"],
+)
+
+cc_args(
+    name = "additional_warnings_args",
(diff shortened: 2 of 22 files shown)


score/mw/common_features.bzl
LittleHuba, line 1: This should not be necessary.

score/message_passing/BUILD
```diff
@@ -15,6 +15,7 @@ load("@rules_cc//cc:defs.bzl", "cc_library")
 load("@score_baselibs//third_party/itf:py_unittest_qnx_test.bzl", "py_unittest_qnx_test")
 load("@score_tooling//bazel/rules/rules_score:rules_score.bzl", "unit")
 load("//quality/unit_testing:unit_testing.bzl", "cc_unit_test")
+load("//score/mw:common_features.bzl", "COMPILER_WARNING_FEATURES")
```
LittleHuba, line 18: Move the bzl file to `//score:common_features.bzl`, since we now share it with something outside of `//score/mw`.

.bazelrc
```diff
@@ -104,6 +104,35 @@ common --check_direct_dependencies=error # Issue an error if a direct dependency
...
+build:permissive --copt=-Wno-error=sign-conversion
+build:permissive --copt=-Wno-error=sign-compare
+build:permissive --copt=-Wno-error=unused-parameter
+build:permissive --copt=-Wno-error=delete-non-virtual-dtor
+build:permissive --copt=-Wno-error=overloaded-virtual
+# additional_warnings overrides
+build:permissive --copt=-Wno-error=cast-qual
+build:permissive --copt=-Wno-error=format-nonliteral
+build:permissive --copt=-Wno-error=undef
+
```
LittleHuba:
```suggestion
```
Do not use copts. This is the non-portable old way. You can deselect a feature using `-<feature_name>`.

In general, I would not create these configs. Users can deselect features from the CLI. That way we don't spam users with configs when they already have a straight-forward way to achieve the same.

MODULE.bazel
LittleHuba, line 1: You should also add the features to the S-CORE gcc toolchain

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,180 @@
...
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        # Downgrade all warnings back to non-fatal. Applied after -Werror
+        # to override it in permissive mode.
+        "-Wno-error",
+    ],
+    requires_any_of = [":treat_warnings_not_as_errors"],
+    visibility = ["//visibility:public"],
+)
```
LittleHuba: Drop this please. Users can just disable with `--features=-treat_warnings_as_errors`

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,180 @@
...
+        "-Wthread-safety",
+    ],
+)
+
+cc_feature(
+    name = "minimal_warnings",
+    args = [
+        ":minimal_warnings_args",
+    ],
+    feature_name = "minimal_warnings",
```
LittleHuba: Currently we see warnings also from other modules like baselibs.
This is problematic, since this is out of our hands and might block a bump.

Please prefix ALL our feature names with `score_communication_`.

This will ensure that we do not enable compiler warnings for other modules as long as they don't reuse our feature names.

When S-CORE agrees on a common set of compiler warnings that must be enabled by all modules we can then imply the S-CORE features in our features.

quality/compiler_warnings/test/BUILD
```diff
@@ -0,0 +1,63 @@
...
+        "-Wconversion",
+        "-Wshadow",
+        "-Wsign-compare",
+        "-Wunused-parameter",
+        "-Wdelete-non-virtual-dtor",
+        "-Woverloaded-virtual",
+        # additional_warnings flags
+        "-Wcast-qual",
+        "-Wformat-nonliteral",
+        "-Wundef",
```
LittleHuba: This makes no sense here. You don't test our features if you change the compiler options.

quality/compiler_warnings/test/BUILD
LittleHuba, line 1: We need three tests:

1. A bug that causes a compiler warning with disabled feature does not cause a compiler error
2. An enabled feature without a bug that causes a compiler warning does not cause a compiler error
3. A bug that causes a compiler warning with enabled feature does cause a compiler error

I suggest that for now you focus on the first two. Create a ticket for the third, since this needs some tricks to achieve.

MODULE.bazel
```diff
@@ -65,10 +65,17 @@ gcc_toolchains = use_extension("@gcc_toolchain//toolchain:module_extensions.bzl"
 gcc_toolchains.toolchain(
     name = "gcc_toolchain_x86_64",
     extra_cxxflags = [
-        "-fdiagnostics-color=always",  # Adds copt to always use coloring in build output
-        "-Wswitch-enum",
-        "-Werror",
-        "-Wno-error=deprecated-declarations",
+        # Diagnostics
```
LittleHuba, line 70: Add a comment that refers to https://github.com/f0rmiga/gcc-toolchain/issues/219
Once that issue is resolved, we should use features in this toolchain as well.

AAmbuj:
This much warring turn into error when we implement the warring as a error feature in current code base: https://github.com/eclipse-score/communication/pull/539

MODULE.bazel
AAmbuj: Done

quality/compiler_warnings/clang/BUILD
AAmbuj: Done

score/message_passing/BUILD
AAmbuj: This file move we will do later.  because if we move file position it cause too many file changes in single PR. for that i will raise another PR when current PR merge.  or I will make a separate commit for that.

score/mw/common_features.bzl
```diff
@@ -11,8 +11,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 COMPILER_WARNING_FEATURES = [
-    "treat_warnings_as_errors",
-    "additional_warnings",
-    "strict_warnings",
-    "strict_warnings_no_error",
+    "score_communication_treat_warnings_as_errors",
+    "score_communication_strict_warnings",
+    "score_communication_additional_warnings",
```
LittleHuba, line 16: What happened to strict_warnings_no_error ?

MODULE.bazel
```diff
@@ -145,29 +154,28 @@ imagefs.toolchain(
...
+    ],
+    extra_known_features = [
+        "//quality/compiler_warnings:default_flags",
+        "//quality/compiler_warnings:treat_warnings_as_errors",
+        "//quality/compiler_warnings:additional_warnings",
+        "//quality/compiler_warnings/clang:minimal_warnings",
+        "//quality/compiler_warnings/clang:strict_warnings",
+        "//quality/compiler_warnings/clang:third_party_warnings",
+    ],
+    extra_link_libs = {"": [
```
LittleHuba: These link flags should probably also go into a cc_feature that you reference in the extra_enabled_features here.

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,156 @@
...
+
+cc_args(
+    name = "third_party_warnings_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        # Suppress all warnings. Used for third-party code where we cannot
+        # fix upstream warnings and they would only add noise.
+        "--no-warnings",
```
LittleHuba, line 97:
```suggestion
        "--no-warnings",
        # NOTE: Some diagnostics are by default an error in clang and not a warning.
        #       Thus, `--no-warnings` cannot silence them. Such diagnostics have to be explicitly disabled via `-Wno-error=`, as done for the following ones:
        "-Wno-error=implicit-function-declaration",  # due to https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#potentially-breaking-changes
        # since -Wmissing-template-arg-list-after-template-kw (cf. https://github.com/llvm/llvm-project/pull/80801) is an error by default starting from clang 19.1.0 on
        # (cf. https://releases.llvm.org/19.1.0/tools/clang/docs/DiagnosticsReference.html#wmissing-template-arg-list-after-template-kw)
        # regardless of the --no-warnings flag, we demote such warning to no-error for all our third-party code
        "-Wno-error=missing-template-arg-list-after-template-kw",
```

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,156 @@
...
+
+cc_feature(
+    name = "third_party_warnings",
+    args = [
+        ":third_party_warnings_args",
+    ],
+    feature_name = "score_communication_third_party_warnings",
+    mutually_exclusive = [
+        "//quality/compiler_warnings:treat_warnings_as_errors",
+        ":minimal_warnings",
```
LittleHuba, line 113:
```suggestion
        "//quality/compiler_warnings:additional_warnings",
        ":minimal_warnings",
        ":strict_warnings",
```

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,156 @@
...
+        "-Wno-error=deprecated-declarations",
+        # Suppress false positive from Clang's self-assignment overloaded check.
+        # See https://bugs.llvm.org/show_bug.cgi?id=43124
+        "-Wno-error=self-assign-overloaded",
+        # Clang Thread Safety Analysis. Detects potential data races through
+        # annotations (GUARDED_BY, REQUIRES, etc.). Essential for concurrent code.
+        # Ref: CERT C++ CON43-CPP — do not allow data races.
+        # Clang-specific: not available in GCC.
+        # Ref: https://clang.llvm.org/docs/ThreadSafetyAnalysis.html
+        "-Wthread-safety",
```
LittleHuba: We currently do not use these annotations in our code.
```suggestion
```

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,156 @@
...
+        # Comprehensive variable shadowing detection. Includes shadowing of
+        # member variables, not just local-to-local shadowing (-Wshadow).
+        # Ref: MISRA C++:2023 Rule 6.4.1 — identifiers shall not be hidden.
+        # Clang-specific: -Wshadow-all is a superset of GCC's -Wshadow.
+        "-Wshadow-all",
+        # -Wshadow-all sub-categories are NOT subcategories of -Wshadow for
+        # -Wno-error purposes. Downgrade these individually so that the global
+        # -Werror does not reject them during migration.
+        "-Wno-error=shadow-uncaptured-local",
+        "-Wno-error=shadow-field",
```
LittleHuba: Discussed with @limdor. We decided to enable -Wshadow-all.
```suggestion
```

quality/compiler_warnings/test/BUILD
LittleHuba: For the third test, please see https://github.com/eclipse-score/communication/blob/b7c543a2ef58366564bac459be9fedcc61ae582c/quality/sanitizer/BUILD#L17-L38

Would be great if you can already start on this as well when you have time.

quality/compiler_warnings/test/clean.cpp
```diff
@@ -0,0 +1,184 @@
...
+    }
+    return static_cast<std::size_t>(index) < size;
+}
+
+// (void) cast suppresses -Wunused-parameter for callback signatures.
+// MISRA C++:2023 Rule 0.1.1 — parameter intentionally unused.
+std::int32_t callback_handler(std::int32_t event_id, std::int32_t context, std::int32_t reserved)
+{
+    (void)context;
+    (void)reserved;
```
LittleHuba: Better use the C++ way of doing this, since we also suggest that in proper production code.
```suggestion
    std::ignore = context;
    std::ignore = reserved;
```

quality/compiler_warnings/test/has_warnings.cpp
LittleHuba, line 1: You must split this to one bazel target per flag you test. Otherwise you only test the first flag that causes an error.

quality/compiler_warnings/test/has_warnings.cpp
```diff
@@ -0,0 +1,139 @@
...
+    delete b;  // warning: deleting object of polymorphic type with non-virtual dtor
+}
+
+// -Woverloaded-virtual: Derived function hides base virtual (different signature).
+// MISRA C++:2023 Rule 10.2.0 — overriding virtual functions.
+struct VBase
+{
+    virtual void process(std::int32_t x)
+    {
+        (void)x;
```
LittleHuba: Same

quality/compiler_warnings/test/has_warnings.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+    virtual ~VBase() = default;
+};
+
+struct VDerived : VBase
+{
+    void process(double x)
+    {
+        (void)x;
```
LittleHuba: Same

quality/compiler_warnings/test/README.md
```diff
@@ -0,0 +1,70 @@
...
+|----------|-------------|---------|-------------|
+| `implicit_narrowing` | `-Wfloat-conversion` | `score_communication_strict_warnings` | MISRA Rule 7.0.5, CWE-681 |
+| `variable_shadowing` | `-Wshadow` | `score_communication_strict_warnings` | MISRA Rule 6.4.1, CWE-398 |
+| `signed_unsigned_compare` | `-Wsign-compare` | `score_communication_strict_warnings` | CWE-195, CWE-697 |
+| `unused_parameter` | `-Wunused-parameter` | `score_communication_strict_warnings` | MISRA Rule 0.1.1, CWE-561 |
+| `delete_polymorphic` | `-Wdelete-non-virtual-dtor` | `score_communication_strict_warnings` (C++) | MISRA Rule 15.7.1 |
+| `VDerived::process` | `-Woverloaded-virtual` | `score_communication_strict_warnings` (C++) | MISRA Rule 10.2.0 |
+| `cast_away_const` | `-Wcast-qual` | `score_communication_additional_warnings` | MISRA Rule 8.2.3 |
+| `format_nonliteral` | `-Wformat-nonliteral` | `score_communication_additional_warnings` | CWE-134 |
+| `#if UNDEFINED_MACRO` | `-Wundef` | `score_communication_additional_warnings` | MISRA Dir 4.4 |
```
LittleHuba: Ideally we have one function per flag. That way we can make sure that actually every flag triggers as expected.

quality/compiler_warnings/test/README.md
```diff
@@ -0,0 +1,70 @@
...
+> ignores them. To exercise features with LLVM:
+
+```bash
+# Explicitly switch to LLVM
+bazel test //quality/compiler_warnings/test:fp_build_test \
+  --extra_toolchains=@llvm_toolchain//:cc-toolchain-x86_64-linux
+
+# Sanitizer configs activate LLVM automatically
+bazel test --config=asan_ubsan_lsan //quality/compiler_warnings/test:fp_build_test
+bazel test --config=tsan            //quality/compiler_warnings/test:fp_build_test
```
LittleHuba: Misleading in this docu.
```suggestion
```

quality/compiler_warnings/test/test_workspace/BUILD.tpl
```diff
@@ -0,0 +1,118 @@
...
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_cc//cc/toolchains:args.bzl", "cc_args")
+load("@rules_cc//cc/toolchains:feature.bzl", "cc_feature")
+
+# Self-contained copies of the warning features under test. The real features
+# live in //quality/compiler_warnings/{clang,gcc}:BUILD; they are duplicated
+# here (Clang flag set) because a nested workspace cannot reference the parent
+# module's targets.
```
LittleHuba: Hmm this restriction is awful. Let me think about a better way.

score/mw/common_features.bzl
```diff
@@ -11,8 +11,9 @@
...
 # *******************************************************************************
 COMPILER_WARNING_FEATURES = [
-    "treat_warnings_as_errors",
-    "additional_warnings",
-    "strict_warnings",
-    "strict_warnings_no_error",
+    # Temporarily disabled for sanitizer test runs (ASAN/TSAN).
+    # Re-enable after shadow/deprecation warnings are fixed in dependencies.
+    # "score_communication_treat_warnings_as_errors",
+    # "score_communication_strict_warnings",
```
LittleHuba: Better would be to introduce a `score_communication_strict_warnings_no_error` and degrade the flags that cause errors to warnings. That way we can enable this and slowly work towards improving the situation.

MODULE.bazel
```diff
@@ -83,8 +92,31 @@ use_repo(gcc_toolchains, "gcc_toolchain_x86_64")
...
+
+SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS = [
+    "-Wno-literal-suffix",
+    "-Wno-noexcept-type",
+]
+
 score_gcc.toolchain(
     name = "score_gcc_x86_64_toolchain",
+    extra_compile_flags = SCORE_GCC_MINIMAL_WARNING_COMPILE_FLAGS,
+    extra_cxx_compile_flags = SCORE_GCC_MINIMAL_WARNING_CXX_FLAGS,
```
LittleHuba, line 119: Is there no support for cc_features?

AAmbuj: Currently not on this path. cc_features are wired for LLVM, but GCC/QCC still require compile flag injection, so these two lines are a temporary compatibility step until upstream feature support is available.

score/mw/common_features.bzl
AAmbuj: Done

quality/compiler_warnings/test/test_workspace/BUILD.tpl
AAmbuj: Done using suggested method https://github.com/yuyawk/rules_build_error

score/mw/common_features.bzl
AAmbuj: after score_communication_strict_warnings_no_error enable this much changes only required to merge our PR.  https://github.com/eclipse-score/communication/pull/558

MODULE.bazel
limdor: I agree, but right now S-CORE gcc toolchain does not support injecting features. Let's first merge it for clang and create a ticket for adding the support in the S-CORE gcc toolchain and once that is there we can add it.

MODULE.bazel
limdor: Ambuj is right, the support is only for llvm, I did not add the support for upstream gcc or S-CORE gcc / qcc

quality/compiler_warnings/clang/BUILD
```diff
@@ -0,0 +1,130 @@
...
+)
+
+cc_args(
+    name = "strict_warnings_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-Wfloat-equal",
+        "-Wformat-security",
```
LittleHuba, line 49: Missing argumentation for presence (follow-up is okay)

quality/compiler_warnings/gcc/BUILD
```diff
@@ -0,0 +1,187 @@
...
+        "-Wpacked",
+        "-Wscalar-storage-order",
+        "-Wsuggest-attribute=format",
+        "-Wundef",
+        "-Wunused-macros",
+        "-Wvector-operation-performance",
+        "-Wwrite-strings",
+        "-Wlogical-op",
+        "-Wredundant-decls",
+        "-Wshadow",
```
LittleHuba, line 80: Same

quality/compiler_warnings/gcc/BUILD
```diff
@@ -0,0 +1,187 @@
...
+cc_args(
+    name = "strict_warnings_args_for_cpp_compile_actions",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:cpp_compile_actions",
+    ],
+    args = [
+        "-Wdelete-non-virtual-dtor",
+        "-Woverloaded-virtual",
+        "-Wregister",
+        "-Wstrict-null-sentinel",
```
LittleHuba, line 93: Same

quality/compiler_warnings/gcc/BUILD
```diff
@@ -0,0 +1,187 @@
...
+)
+
+cc_args(
+    name = "strict_warnings_args_for_c_compile_actions",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:c_compile_actions",
+    ],
+    args = [
+        "-Wold-style-definition",
+        "-Wstrict-prototypes",
```
LittleHuba, line 104: Same

quality/compiler_warnings/test/BUILD
```diff
@@ -0,0 +1,95 @@
...
+)
+
+# Test 3: bug + enabled feature must fail the build.
+
+[
+    cc_build_error_test(
+        name = "fail_{}_build_error_test".format(case),
+        src = "has_warnings/{}.cpp".format(case),
+        compile_stderr = matcher.has_substr("error:"),
+        copts = ["-Werror"] + warning,
```
LittleHuba, line 72: This does not exercise our features.
Do not supply the compiler flag here but instead enable the relevant feature.

quality/compiler_warnings/BUILD
```diff
@@ -0,0 +1,173 @@
...
+cc_args(
+    name = "strict_warnings_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-Wextra",
+        "-Wpedantic",
+        "-Wconversion",
+        "-Wsign-conversion",
```
LittleHuba, line 108: Missing argumentation

quality/compiler_warnings/test/BUILD
LittleHuba: Talked with @limdor and decided to also do this in a follow-up. The issue is that once we use features here, we would have to restrict the tests to LLVM executions only.

Additionally, the compile_stderr must actually verify that the correct error was issued and not just SOME error.

AAmbuj: Modified the algo.

LittleHuba: Nice, modification. This resolves the part about the warning.
Using the cc_feature is still open. As mentioned above, this must be done in a follow-up.

quality/compiler_warnings/BUILD
```diff
@@ -0,0 +1,138 @@
...
+)
+
+cc_feature(
+    name = "default_flags",
+    args = [
+        ":default_compile_args",
+    ],
+    feature_name = "score_communication_default_flags",
+    visibility = ["//visibility:public"],
+)
```
LittleHuba, line 43: Since we are directly depending on the args in the compiler specific variants, we should remove the cc_feature targets here to avoid confusion.

Can be done in a follow-up.

LittleHuba (approved):

PR #529, opened 2026-06-11
Avoid time penalty when unregistering receive handler after the producer is unreachable

LittleHuba:
Assume you have two applications.
One application offers an event. The second receives notifications for this event.

Now the sender shuts down. Afterwards, the receiver also tries to shut down.

When the receiver has a receive handler registered, he sends a message to the sender to unregister.

But the sender is not there and the communication channel is not ready.

The receiver will therefor try to recreate the channel. This will block him until the attempt times out.

Since the API is synchronous, this will block the shutdown of the application.

From resource POV, loosing this message has no drawbacks (see comment in code). Hence, we try to send it based on best-effort.

### DIFF
--- a/score/mw/com/impl/bindings/lola/messaging/message_passing_client_cache.cpp
+++ b/score/mw/com/impl/bindings/lola/messaging/message_passing_client_cache.cpp
@@ -50,18 +50,37 @@ MessagePassingClientCache::MessagePassingClientCache(const ClientQualityType asi
 {
 }
 
+std::shared_ptr<score::message_passing::IClientConnection>
+MessagePassingClientCache::UnlockedGetCachedMessagePassingClient(const pid_t target_node_id) noexcept
+{
+    auto search = clients_.find(target_node_id);
+    if (search != clients_.end())
+    {
+        return search->second;
+    }
+
+    return nullptr;
+}
+
+std::shared_ptr<score::message_passing::IClientConnection> MessagePassingClientCache::GetCachedMessagePassingClient(
+    const pid_t target_node_id) noexcept
+{
+    std::lock_guard<std::mutex> lck(mutex_);
+    return UnlockedGetCachedMessagePassingClient(target_node_id);
+}
+
 std::shared_ptr<score::message_passing::IClientConnection> MessagePassingClientCache::GetMessagePassingClient(
     const pid_t target_node_id) noexcept
 {
     std::lock_guard<std::mutex> lck(mutex_);
 
-    auto search = clients_.find(target_node_id);
-    if (search != clients_.end())
+    const auto cached_client = UnlockedGetCachedMessagePassingClient(target_node_id);
+    if (cached_client != nullptr)
     {
-        const auto state = search->second->GetState();
+        const auto state = cached_client->GetState();
         if (state == score::message_passing::IClientConnection::State::kReady)
         {
-            return search->second;
+            return cached_client;
         }
         // Evict a cached client that is not in kReady state. This covers:
         // - kStopped/kStopping: connection was lost (e.g. peer was SIGKILL'd)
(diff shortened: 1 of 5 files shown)


score/mw/com/impl/bindings/lola/messaging/message_passing_client_cache.cpp
```diff
@@ -50,31 +50,48 @@ MessagePassingClientCache::MessagePassingClientCache(const ClientQualityType asi
...
+}
+
 std::shared_ptr<score::message_passing::IClientConnection> MessagePassingClientCache::GetMessagePassingClient(
     const pid_t target_node_id) noexcept
 {
     std::lock_guard<std::mutex> lck(mutex_);
 
-    auto search = clients_.find(target_node_id);
-    if (search != clients_.end())
+    const auto cached_sender = UnlockedGetCachedMessagePassingClient(target_node_id);
```
crimson11: `cached_client` ... the notion `sender`is so message-passing 1.0 :)

score/mw/com/impl/bindings/lola/messaging/message_passing_client_cache.cpp
```diff
@@ -50,31 +50,48 @@ MessagePassingClientCache::MessagePassingClientCache(const ClientQualityType asi
...
-                                        << ")";
-        search->second->Stop();
-        score::cpp::ignore = clients_.erase(search);
+        score::mw::log::LogWarn("lola")
+            << "MessagePassingClientCache: Evicting non-ready client for node " << target_node_id
+            << " (state=" << static_cast<std::uint32_t>(score::cpp::to_underlying(state))
+            << ", reason=" << static_cast<std::uint32_t>(score::cpp::to_underlying(cached_sender->GetStopReason()))
+            << ")";
+        cached_sender->Stop();
+        score::cpp::ignore = clients_.erase(target_node_id);
```
crimson11, line 96: Why doing the erase/eviction? Shouldn't the caller of `GetMessagePassingClient` decide (based on the state), whether it wants to use the connection with the risk to block? But an "automatic" eviction is too harsh. You would now also influence other use-cases than UnsetReceiveHandler!

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp
```diff
@@ -1404,7 +1405,23 @@ void MessagePassingServiceInstance::UnregisterEventNotificationRemote(
         // Suppress "AUTOSAR C++14 A18-5-8" rule finding. This rule states: "Objects that do not outlive a function
         // shall have automatic storage duration". The object is a shared_ptr which is allocated in the heap.
         // coverity[autosar_cpp14_a18_5_8_violation]
-        auto sender = client_cache_.GetMessagePassingClient(target_node_id);
+        auto sender = client_cache_.GetCachedMessagePassingClient(target_node_id);
```
crimson11, line 1408: sender could be a null-ptr ... after your changes `GetMessagePassingClient` -> `GetCachedMessagePassingClient` ... But in line 1417 you aren't checking this?

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/messaging/message_passing_client_cache.cpp
LittleHuba: This is the existing behavior. We evict and then directly create a new client.

score/mw/com/impl/bindings/lola/messaging/message_passing_client_cache.cpp
LittleHuba: Oops... Fixed.

score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp
LittleHuba: Oops... Fixed.

crimson11 (approved):

PR #532, opened 2026-06-12
Rust::com Added SCTF/Integration test for Stream and Receive APIs and Documentation update 

bharatGoswami8:
- Added SCTF Test cases for `cancellable_receive` and `to_stream` APIs
- Added sequence diagram and documentation update.


score/mw/com/impl/rust/doc/user_facing_api_examples.md
```diff
@@ -256,6 +280,158 @@ fn read_tire_data<R: Runtime>(
...
+---
+
+### Stream-based Receive
+
+**Key Points**:
+- `to_stream()` converts subscription to async stream for futures-based iteration
+- Stream returns samples with `Ready` status
+- Errors don't end the stream automatically - user decides based on error severity
+- Requires mutable reference to subscription
+- `SampleContainer` lifecycle managed internally
```
darkwisebear: This is misleading. I think you should just state that this method will allocate an internal sample pointer buffer.

score/mw/com/test/basic_rust_api/consumer_async_apis/consumer_app.rs
```diff
@@ -17,28 +17,159 @@
...
+
+    println!("[bigdata-consumer] Using async cancellable receive with timeout");
+
+    while total_cycle < num_cycles {
+        // Create a timeout future using oneshot channel and thread sleep
+        let (tx, rx) = oneshot::channel();
+        thread::spawn(move || {
+            thread::sleep(Duration::from_millis(50));
+            let _ = tx.send(()); // Send signal after timeout
+        });
```
darkwisebear: That's complicated and resource wasting: You create a new thread once per cycle whose single purpose is to sleept-then-ping? Why not just using https://docs.rs/tokio/latest/tokio/time/fn.sleep.html?

score/mw/com/impl/rust/doc/sequence_diagram/com_api_async_find_service_and_subscribe.puml
```diff
@@ -0,0 +1,50 @@
...
+User -> Runtime: find_service::<VehicleInterface>(\nFindServiceSpecifier::Specific(\nInstanceSpecifier))
+Runtime --> User: ServiceDiscovery<I, R>
+
+User -> ServiceDiscovery: get_available_instances_async().await
+Note over ServiceDiscovery: Async waits until\nservice is offered
+ServiceDiscovery --> User: Result<ServiceEnumerator>
+
+Note over User: ServiceEnumerator contains\nConsumerBuilder instances
+
+User -> ConsumerBuilder: (from iterator.nth())
```
darkwisebear: That's a bit misleading. It's not necessarily `iterator.nth()` but it could be anything that returns a `ConsumerBuilder`. Also, the call goes to `ServiceEnumerator` and not to `ConsumerBuilder`.

score/mw/com/impl/rust/doc/user_facing_api_examples.md
bharatGoswami8: Rephrased the comment.

score/mw/com/impl/rust/doc/sequence_diagram/com_api_async_find_service_and_subscribe.puml
bharatGoswami8: Updated the sequence flow.

score/mw/com/test/basic_rust_api/consumer_async_apis/consumer_app.rs
bharatGoswami8: I have moved the thread to create only once, and lifecycle of this will be handle with in function.

I am thinking to keep the Integration test `tokio` free, this just to demonstrate how safe application can be use APIs without `tokio` dependecy. 

So if we want to avoid creating thread approach then i can bring some lightweight crate for creating future for timer like` futures-timer`, `async_io`

bharatGoswami8: Also, if we use `tokio::time::sleep(Duration::from_secs(1))`, it must run within a t`okio::runtime::Runtime`.
This is because the current application uses `futures::executor::block_on` to execute async APIs, and `tokio::time::sleep` requires the Tokio runtime's timer infrastructure.

score/mw/com/test/basic_rust_api/consumer_async_apis/consumer_app.rs
```diff
@@ -166,7 +177,7 @@ async fn receive_stream<R: Runtime>(mut subscription: impl Subscription<MapApiLa
             Some(Err(e)) => {
                 eprintln!("[bigdata-consumer] Stream error: {:?}", e);
             }
-            None => unreachable!("Stream never sends None")
+            None => unreachable!("Stream never sends None"),
```
darkwisebear, line 180: Probably not true for the current implementation but: What if the stream would know that the server terminated? Wouldn't this be a situation where the stream would send `None`?

darkwisebear (approved):

score/mw/com/test/basic_rust_api/consumer_async_apis/consumer_app.rs
bharatGoswami8: Currently stream is not sending `None` in any cases, in future it may send.
So, i should rephrase the error message.

bemerybmw (approved):

PR #549, opened 2026-06-15
 use llvm cov for coverage and justification

castler:


quality/coverage/llvm_cov/filter_regexes.txt
```diff
@@ -0,0 +1,20 @@
+# Coverage filter regexes (one per line).
+# Lines matching any of these patterns are excluded from the coverage report
+# via llvm-cov's --ignore-filename-regex option.
+#
+# NOTE: --experimental_use_llvm_covmap causes Bazel to instrument ALL targets
+# regardless of --instrumentation_filter. Therefore, filtering MUST happen here
+# at the report level.
+
+# Exclude mock files.
+.*mock.*\.(h|hpp|cpp)$
```
LittleHuba: Better make it `.*_mock.*\.(h|hpp|cpp)$`. I'd prefer to start very strict and relax it when we find files that should be excluded.

quality/coverage/llvm_cov/filter_regexes.txt
```diff
@@ -0,0 +1,20 @@
...
+
+# Exclude external dependencies (anything under external/).
+.*/external/.*
```
LittleHuba: Not a big fan of the starting `.*`. What is the base path? Could we live with `external/.*`?

quality/coverage/llvm_cov/merger.py
```diff
@@ -0,0 +1,313 @@
...
+                    candidate = runfiles_dir / obj_path
+                    if candidate.exists():
+                        object_files.add(str(candidate))
+                    else:
+                        object_files.add(str(exec_root / obj_path))
+
+    return object_files
+
+
+def find_llvm_bin() -> Path:
```
LittleHuba: Uff... not really sure whether this is a good idea.
This could silently break without us being aware.

Two problems I see:
- This might become unhermetic if we ever end up with a llvm-toolchain that does not provide llvm-cov for some reason
- globbing through all of "external" invites usage of an llvm-cov that ends up there for a different reason (e.g. sysroot for a different tool)

quality/coverage/llvm_cov/merger.py
```diff
@@ -0,0 +1,313 @@
...
+            return candidate
+
+    # Search in external/ for llvm-cov.
+    search_root = Path.cwd() / "external"
+    if search_root.exists():
+        for llvm_cov in search_root.glob("*/**/bin/llvm-cov"):
+            return llvm_cov.resolve().parent
+
+    # Search from ROOT.
+    root = Path(os.environ.get("ROOT", "."))
```
LittleHuba: Again, not sure whether `.` is a good fallback here. The fallback also makes no sense since it results in the same logic as ll254.

quality/coverage/llvm_cov/merger.py
```diff
@@ -0,0 +1,313 @@
...
+    gcov_link = directory / "gcov"
+    if gcov_link.is_symlink():
+        gcov_link.unlink()
+
+    # Also remove any other symlinks pointing into sandbox paths.
+    for entry in directory.iterdir():
+        if entry.is_symlink():
+            target = os.readlink(entry)
+            if "sandbox" in target:
+                entry.unlink()
```
LittleHuba, line 109: Should we at least issue a warning for such symlinks?
Better to know which ones point in and then add them here explicitly than just removing everything that is there blindly?

quality/coverage/llvm_cov/merger.py
```diff
@@ -0,0 +1,313 @@
...
+    root = os.environ.get("ROOT", str(Path.cwd()))
+    if not root.endswith("/"):
+        root += "/"
+    return root
+
+
+def get_object_files_from_manifest(source_file_manifest: Path) -> Set[str]:
+    """Parse the coverage manifest to find instrumented object files."""
+    runfiles_dir = Path(os.environ.get("RUNFILES_DIR", "")) / os.environ.get("TEST_WORKSPACE", "_main")
+    exec_root = Path(os.environ.get("ROOT", "."))
```
LittleHuba: Not sure that `.` is a great fallback for `ROOT`. Better to fail here?

quality/coverage/generate_coverage_html.sh
```diff
@@ -58,87 +59,82 @@ unset _SELF_DIR _SELF_NAME
...
+  echo "Running coverage justification processing..."
+
+  JUSTIFICATION_DIR="${TMPDIR_EXTRACT}/justification_report"
+  mkdir -p "${JUSTIFICATION_DIR}"
+
+  # Run justify.py via Bazel to produce the resolved manifest.
+  if bazel run //quality/coverage/llvm_cov:justify -- \
+      --yaml "${JUSTIFICATION_YAML}" \
+      --source-root "${BUILD_WORKSPACE_DIRECTORY}" \
+      --output "${JUSTIFICATION_DIR}/manifest.json"; then
```
LittleHuba, line 109: Why do we run this separately and not as part of the merger? If needed we could still make it possible to disable via env-var/flag/...

quality/coverage/generate_coverage_html.sh
```diff
@@ -58,87 +59,82 @@ unset _SELF_DIR _SELF_NAME
...
+
+    # Run effective_coverage.py via Bazel to post-process HTML and calculate effective coverage.
+    bazel run //quality/coverage/llvm_cov:effective_coverage -- \
```
LittleHuba, line 112: Same

quality/coverage/llvm_cov/effective_coverage.py
```diff
@@ -0,0 +1,737 @@
+#!/usr/bin/env python3
```
LittleHuba, line 1: The previous commit already refers to this file while it is not there. Better squash before merge.

quality/coverage/llvm_cov/justify.py
```diff
@@ -0,0 +1,327 @@
...
+        errors.append("Missing 'version' field")
+
+    if "justifications" not in data:
+        errors.append("Missing 'justifications' field")
+        if errors:
+            for e in errors:
+                print(f"ERROR: {e}", file=sys.stderr)
+            sys.exit(1)
+        return
+
```
LittleHuba, line 265: Check that justifications is an array.

LittleHuba: In general you are missing type checks here.

quality/coverage/llvm_cov/justify.py
```diff
@@ -0,0 +1,327 @@
...
+        print(f"ERROR: Justification YAML not found: {yaml_path}", file=sys.stderr)
+        sys.exit(1)
+
+    with open(yaml_path, "r", encoding="utf-8") as f:
+        content = f.read()
+
+    return yaml.safe_load(content)
+
+
+def validate_yaml(data: Dict[str, Any]) -> None:
```
LittleHuba, line 246: Might make sense to put this whole validation in a try-catch block. That way you will not die weirdly if you missed a check.

quality/coverage/llvm_cov/README.md
```diff
@@ -0,0 +1,249 @@
...
+└────┬────────┘
+     │  manifest.json: {file → {line → justification}}
+     ▼
+┌───────────────────────┐
+│ effective_coverage.py  │ ◄── manifest.json + html_report/
+└────┬──────────────────┘
+     │  • Modifies HTML in-place (restyled justified lines/branches)
+     │  • report.json + summary.txt
+     ▼
+  Console output: effective coverage summary
```
LittleHuba, line 49: I think it would help to argue why we do not call this from reporter.py.
If it is because we want to retain the original data without the justifications applied, we could just output the modified report next to the original report.

quality/coverage/README.md
```diff
@@ -0,0 +1,204 @@
...
+    ├── BUILD                       ← Bazel targets for Python tools
+    ├── merger.py                   ← Per-test coverage output generator
+    ├── reporter.py                 ← Final combined report generator
+    ├── justify.py                  ← Justification resolver
+    └── effective_coverage.py       ← HTML post-processor & effective coverage calculator
+```
+
+## Requirements
+
+The coverage pipeline was built to satisfy the following requirements:
```
LittleHuba, line 25: Should we put those directly into TRLC? Nit-pick can be done in a follow-up.

quality/coverage/README.md
```diff
@@ -0,0 +1,204 @@
...
+```cpp
+unreachable_code();  // COV_JUSTIFIED my-unique-id
+
+// COV_JUSTIFIED_START my-unique-id
+defensive_block();
+more_defensive_code();
+// COV_JUSTIFIED_STOP
+```
+
+Both methods can be combined. A justification covers both the line and any branches on that line.
```
LittleHuba, line 185: Add some lines that strongly suggest that using code markes is the preferred way.

Putting locations in the YAML always runs the risk of justifications being outdated without us being aware, because code is shifted.

quality/quality.md
```diff
@@ -110,13 +110,13 @@ The query configuration is defined in [`quality/static_analysis/config.yaml`](st
 
 ## Coverage
 
-Code coverage is generated using LLVM's source-based coverage instrumentation. The instrumentation filter is configured in [`quality/coverage.bazelrc`](coverage.bazelrc) to cover `//score/message_passing` and `//score/mw/com` while excluding test and benchmark code.
+Code coverage is generated using LLVM's source-based coverage instrumentation. The instrumentation filter is configured in [`quality/coverage/coverage.bazelrc`](coverage/coverage.bazelrc) to cover `//score/message_passing` and `//score/mw/com` while excluding test and benchmark code.
 
-### Running Coverage
+HTML reports are generated directly by `llvm-cov show` — no intermediate LCOV conversion or `genhtml` is needed.
```
LittleHuba: Please just drop everything after the dash. It just confuses.

quality/quality.md
```diff
@@ -143,6 +143,59 @@ The report is written to `cpp_coverage/index.html`. Open it with:
...
+- **Raw line coverage**: actual lines hit / total instrumented lines
+- **Effective line coverage**: (lines hit + justified lines) / total instrumented lines
+- **Raw branch coverage**: actual branches hit / total branches
+- **Effective branch coverage**: (branches hit + justified branches) / total branches
+
+The index page shows a banner with overall effective coverage and updates per-file percentages. The `generate_coverage_html` script prints the full summary. Set `COVERAGE_THRESHOLD` to enforce a minimum (default: 100%):
+
+```bash
+COVERAGE_THRESHOLD=95 bazel run //quality/coverage:generate_coverage_html
+```
```
LittleHuba: Not really a huge fan of this duplication. Can we not just link to the actual readme for coverage?

LittleHuba (commented):
Really nice work!

quality/coverage/llvm_cov/merger.py
castler: maybe try to hard code it - ,aybe look at it next.

quality/coverage/llvm_cov/merger.py
castler: maybe log into undeclared test outputs (maybe do later)

quality/coverage/generate_coverage_html.sh
castler: maybe later to try to bring this to the compile step

quality/coverage/generate_coverage_html.sh
castler: maybe later to try to bring this to the compile step

quality/coverage/README.md
castler: We can do in an other iteration

quality/coverage/generate_coverage_html.sh
```diff
@@ -58,87 +59,82 @@ unset _SELF_DIR _SELF_NAME
...
 cd "${BUILD_WORKSPACE_DIRECTORY}"
 
-# bazel-out/ is a symlink Bazel always creates in the workspace root, it
-# points to the real output directory.  _coverage/ sits at its root (not
-# inside a config-specific sub-directory), so we can locate the merged
-# report without calling 'bazel info' — which would require 'bazel' to be
-# on PATH inside the run environment.
-LCOV_DAT="${BUILD_WORKSPACE_DIRECTORY}/bazel-out/_coverage/_coverage_report.dat"
+# Resolve OUTPUT_DIR to absolute path (relative to workspace root).
+OUTPUT_DIR="${BUILD_WORKSPACE_DIRECTORY}/${OUTPUT_DIR}"
```
LittleHuba, line 63: What if `OUTPUT_DIR` is an absolute path?

quality/coverage/llvm_cov/reporter.py
```diff
@@ -0,0 +1,320 @@
...
+        str(llvm_bin_path),
+        "show",
+        f"--format={output_format}",
+        f"--path-equivalence=/proc/self/cwd/,{workspace_root}",
+        f"--compilation-dir={workspace_root}",
+        "--show-branches=count",
+        "--show-region-summary=0",
+    ]
+
+    cxxfilt = llvm_bin_path.parent / "llvm-cxxfilt"
```
LittleHuba, line 145:
```suggestion
    cxxfilt = r.Rlocation("llvm_toolchain/llvm-cxxfilt")
```

quality/coverage/llvm_cov/reporter.py
```diff
@@ -0,0 +1,320 @@
...
+    valid_profdata_files, valid_object_files = extract_reports(reports)
+
+    if not valid_profdata_files or not valid_object_files:
+        print("INFO: No valid profdata or object files found.", file=sys.stderr)
+        write_empty_output(args.output_file)
+        sys.exit(0)
+
+    # Get llvm tools via runfiles.
+    r = Runfiles.Create()
+    llvm_bin_path = Path(r.Rlocation("llvm_toolchain/llvm-cov"))
```
LittleHuba, line 57: It feels hacky that in some places you use r.Rlocation to determine the path. And in others you do path traversal based on this path.

quality/coverage/llvm_cov/BUILD
```diff
@@ -0,0 +1,77 @@
...
+    ],
+    outs = ["reporter_wrapper.sh"],
+    cmd = """
+cat > $@ << EOF
+#!/usr/bin/env bash
+set -euo pipefail
+if [[ -z "\\$${RUNFILES_DIR:-}" ]]; then
+  if [[ -d "\\$$0.runfiles" ]]; then
+    export RUNFILES_DIR="\\$$0.runfiles"
+  fi
```
LittleHuba, line 48: Should we have an else condition as a safeguard here?

LittleHuba (approved):
Approving as remaining issues are nitpicks that can be resolved in a follow-up.

olivembo:
@castler Moved and adapted your implementation to the cpp_policies repo. 

PR is: https://github.com/eclipse-score/score_cpp_policies/pull/9

Can you check?

PR #555, opened 2026-06-17
message_passing: derive component_requirements

gdadunashvili:


score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
 // Server Unit and Client Unit requirements
 ///////////////////////////////
 
-ScoreReq.CompReq REQ_MP_COMP_001 {
-    description = "The message passing server shall accept connections from clients on a configurable resource path."
+section "System Requirements" {
+
+ScoreReq.CompReq SafetyCertifiedTransportMechanismUnderQNX {
+    description = "Under QNX, the message passing component shall use a safety-certified transport mechanism as the underlying IPC."
```
castler: That's basically the same statement as in the upper requirement. We should define here that we are using message passing.

score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
...
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.SafetyCertifiedTransportMechanism@1]
+    version = 1
+}
+
+ScoreReq.CompReq OSProvidedSenderIdentity {
+    description = "The message passing server shall be able to identify the sender of each received message by the sender's OS-provided UID."
```
castler: UID = user-id?
or unique id, if unique id, then is the question what unique id?

If user-id, then we need an aou that every process should have a different user-id, and we should link this here somehow.

score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
...
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.ServerInterface@1]
+    version = 1
+}
+
+ScoreReq.CompReq UnforgableSenderIdentity {
+    description = "The transport mechanism shall ensure that the UID used to identify a message, cannot be forged by the client."
```
castler, line 48: This would then be a requirement towards the OS, no?

castler: According to my picture this would be an ExternalComponentRequirement

score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
...
+    derived_from = [MessagePassingSEooC.ServerInterface@1]
+    version = 1
+}
+
+}
+
+section "Behaviour Requirements" {
+
+ScoreReq.CompReq ServerCallbacksAreSequential {
+    description = "The server shall serialize all callbacks belonging to the same IServer instance, such that no two such callbacks execute concurrently."
```
castler: what is an IServer instance?

score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
...
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.ServerInterface@1]
+    version = 1
+}
+
+ScoreReq.CompReq ServerProcessesSinglePendingRequest {
+    description = "The server shall not dispatch a new sent-with-reply message callback on a given IServerConnection until the preceding request has been replied to via `IServerConnection::Reply()`."
```
castler: Again IServerConnection, this sounds like implementation detail - I think what you did below with ´ClientConnection´ is maybe a little bit more abstract?

score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
...
+    safety = ScoreReq.Asil.B
+    derived_from = [MessagePassingSEooC.SynchronousBidirectionalCommunication@1]
+    version = 1
+}
+
+ScoreReq.CompReq ClientConnectionMaintainsStateMachine {
+    description = "The `ClientConnection` shall maintain a state machine with the states `Starting`, `Ready`, `Stopping`, and `Stopped`."
```
castler, line 73: What is the allowed transitions between these states? Should we draw this in a state diagram?

score/message_passing/dependability/requirements/component_requirements.trlc
```diff
@@ -20,9 +20,249 @@ import MessagePassingSEooC
...
+
+ScoreReq.CompReq IServerStartListeningAPI {
+    description = "`IServer` shall provide a `StartListening` method that registers connection, disconnection, fire-and-forget message, and sent-with-reply message callbacks."
     safety = ScoreReq.Asil.B
     derived_from = [MessagePassingSEooC.ServerInterface@1]
     version = 1
 }
+
+ScoreReq.CompReq IServerStopListeningAPI {
+    description = "`IServer` shall provide a `StopListening` method that releases all registered callbacks and closes all active server connections."
```
castler, line 126: I really like this way of writing the API requirements, without exactly specifying the C++ API!

score/message_passing/dependability/requirements/component_requirements.trlc
castler: We also should have a requirement what we use for linux to satisfy that this is OS abstracted

score/message_passing/dependability/requirements/component_requirements.trlc
gdadunashvili: Agree. That's why it is clustered in System Requirements section, i.e. Component level Requirements towards the system. If it needs to go in a different file I can do that.

gdadunashvili: Or would it be better if this becomes an AOU?

score/message_passing/dependability/requirements/component_requirements.trlc
gdadunashvili: I already drew a state diagram and it is merged
https://github.com/eclipse-score/communication/blob/main/score/message_passing/dependability/software_architectural_design/client_connection_activity_diagram.puml



Should I link it here? If yes what is our way of doing that? Just a path to the file?

score/message_passing/dependability/requirements/component_requirements.trlc
gdadunashvili: I created a PR with tentative AoUs:
https://github.com/eclipse-score/communication/pull/597

score/message_passing/dependability/requirements/component_requirements.trlc
gdadunashvili: I understood this and the next comment as a suggestion to not use API names but more generic terms like Server & Client, and made changes accordingly

castler (approved):

Issue #489, opened 2026-05-30
Improvement: Optimize the Rust API sample application example app


bharatGoswami8:
### What

Example app has test for each API, this can be improved with each API based function and execute based on user selection.

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [ ] Requirements / Architecture are not affected by this change?

bharatGoswami8:
https://github.com/eclipse-score/communication/pull/556

PR #556, opened 2026-06-18
Rust::com Rust APIs Example app Optimize

bharatGoswami8:
- Created Example app which can run based on CLI argument like only Producer / Consumer or Both and different Async APIs of Consumer.
- Independent a test unit kept which use tokio runtime for async testing.

https://github.com/eclipse-score/communication/issues/489


score/mw/com/example/com-api-example/USAGE.md
```diff
@@ -0,0 +1,129 @@
...
+## Main Application
+
+The main application (`main.rs`) demonstrates:
+
+### Multi-threaded Architecture
+- **Producer Thread**: Sends vehicle data at regular intervals
+- **Consumer Thread**: Receives and processes data independently
+
+### Async Operations
+- Uses `futures::executor::block_on()` for async operations
```
darkwisebear: Sloppy - block_on is synchronous. AI should elaborate more about the execution model instead of making obvious but inaccurate statements like here.

score/mw/com/example/com-api-example/USAGE.md
```diff
@@ -0,0 +1,129 @@
...
+  -s ./path/to/custom/config.json
+```
+
+## Integration Tests
+
+### Tokio crate based Integration Tests
+
+A separate test target (`tests_using_tokio_runtime.rs`) provides comprehensive integration tests using the **Tokio async runtime**. These tests are independent of the main application and use the `com-api-gen` generated types together with the `com_api` runtime, service discovery, and subscription APIs directly.
+
+**Why Separate?**
```
darkwisebear: I don't get how the points below answer this question.

score/mw/com/example/com-api-example/src/lib.rs
```diff
@@ -0,0 +1,35 @@
...
+pub mod producer;
+pub mod vehicle_monitor;
+pub use vehicle_monitor::{
+    VehicleConsumer, VehicleMonitorConsumer, VehicleMonitorProducer, VehicleOfferedProducer,
+};
+
+/// All execution modes supported by the example application.
+/// Used by clap for CLI argument parsing and by the consumer/producer to select behaviour.
+#[derive(clap::ValueEnum, Clone, Debug)]
+pub enum ExampleType {
```
darkwisebear: Why is this here? Imo that's clearly something that does not belong in a library but to the main CLI crate.

score/mw/com/example/com-api-example/BUILD
```diff
@@ -11,28 +11,46 @@
 # SPDX-License-Identifier: Apache-2.0
 # *******************************************************************************
 
-load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_test")
+load("@rules_rust//rust:defs.bzl", "rust_binary", "rust_library", "rust_test")
+
+rust_library(
```
darkwisebear, line 16: Please choose the 2024 edition.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+            };
+            let sample = uninit_sample.write(Tire {
+                pressure: 1.0 + i as f32,
+            });
+            match sample.send() {
+                Ok(_) => (),
+                Err(e) => eprintln!("[SENDER] Failed to send sample: {:?}", e),
+            }
+            println!("[SENDER] Sent sample with pressure: {}", 1.0 + i as f32);
+            tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
darkwisebear: Magic number.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+        println!("Starting async subscription test with Lola runtime");
+        let service_id = InstanceSpecifier::new("/Vehicle/Service3/Instance")
+            .expect("Failed to create InstanceSpecifier");
+        let service_id_clone = service_id.clone();
+        //consumer create
+        let consumer_runtime = get_test_runtime();
+        //starting service discovery in async way, so that it can be discovered when producer offer service after some delay, and consumer is waiting for discovery result
+        let consumer = tokio::spawn(create_consumer_async(consumer_runtime, service_id));
+        //simulate some delay before producer offer service, so that consumer is waiting for discovery
+        tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
darkwisebear: Magic number.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+        //Intentionally using service instance of test1, if you face issue add new service instance in config file and use it here.
+        let service_id = InstanceSpecifier::new("/Vehicle/Service1/Instance")
+            .expect("Failed to create InstanceSpecifier");
+        let service_id_clone = service_id.clone();
+        //consumer create
+        let consumer_runtime = get_test_runtime();
+        //starting service discovery in async way, so that it can be discovered when producer offer service after some delay, and consumer is waiting for discovery result
+        let consumer = tokio::spawn(create_consumer_async(consumer_runtime, service_id));
+        //simulate some delay before producer offer service, so that consumer is waiting for discovery
+        tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
darkwisebear: Magic... you know what.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+        println!("Starting async subscription test with Lola runtime");
+        let service_id = InstanceSpecifier::new("/Vehicle/Service1/Instance")
+            .expect("Failed to create InstanceSpecifier");
+        let service_id_clone = service_id.clone();
+        //consumer create
+        let consumer_runtime = get_test_runtime();
+        //starting service discovery in async way, so that it can be discovered when producer offer service after some delay, and consumer is waiting for discovery result
+        let consumer = tokio::spawn(create_consumer_async(consumer_runtime, service_id));
+        //simulate some delay before producer offer service, so that consumer is waiting for discovery
+        tokio::time::sleep(tokio::time::Duration::from_millis(1000)).await;
```
darkwisebear: Same.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+    //it will receive data 10 times and print the received samples
+    async fn async_data_processor_fn<R: Runtime>(
+        subscribed: impl Subscription<Tire, R>,
+        is_timeout: bool,
+    ) {
+        println!("[RECEIVER] Async data processor started");
+        let mut buffer = SampleContainer::new(5);
+        for _ in 0..3 {
+            let (returned_buf, result) = if is_timeout {
+                let timeout = tokio::time::sleep(Duration::from_millis(1000));
```
darkwisebear: Magic number.

score/mw/com/example/com-api-example/BUILD
```diff
@@ -11,28 +11,46 @@
...
+    name = "com-api-example-lib",
+    srcs = glob(["src/**/*.rs"]),
```
rpreddyhv, line 18: question: Is there any better alternative possible here? One potential issue could be that even an unrelated change to the files in `src/**/*.rs`, will endup in rebuilding the `com-api-example-lib`. I am okay, if that is intentionally choosen.

score/mw/com/example/com-api-example/main.rs
```diff
@@ -0,0 +1,157 @@
...
+        RunMode::Producer => None,
+    };
+
+    // Wait for both threads to complete
+    if let Some(h) = producer_handle {
+        h.join().expect("Producer thread panicked");
+    }
+    if let Some(h) = consumer_handle {
+        h.join().expect("Consumer thread panicked");
+    }
```
rpreddyhv: nitpick: this can be generalized to avoid duplication.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+    };
+    use com_api_gen::{Tire, VehicleInterface, VehicleOfferedProducer};
+
+    use futures::stream::StreamExt;
+    use std::sync::OnceLock;
+    use std::time::Duration;
+    const TEST_CONFIG_PATH: &str = "./score/mw/com/example/com-api-example/etc/mw_com_config.json";
+
+    static LOLA_RUNTIME: OnceLock<com_api::LolaRuntimeImpl> = OnceLock::new();
+    // it will create a singleton instance of LolaRuntime for testing,
```
rpreddyhv, line 53: clarification required: I hope the Singleton LolaRuntime object being created here is stateless and hence it is safe to use across tests?

score/mw/com/example/com-api-example/src/consumer.rs
rpreddyhv, line 1: suggestion: In my personal opinion having separate `VehicleMonitorConsumer` for different types of operations (`sync` or `async` & it's variations) is better. We can still share the common functionality across such different types. Such types become concrete examples for user needs.

score/mw/com/example/com-api-example/main.rs
```diff
@@ -0,0 +1,157 @@
...
+/// Controls which roles are started in this process.
+#[derive(clap::ValueEnum, Clone, Debug, Default)]
+enum RunMode {
+    /// Start both the producer and consumer (default).
+    #[default]
+    Both,
+    /// Start only the producer.
+    Producer,
+    /// Start only the consumer.
+    Consumer,
```
rpreddyhv, line 39: question: What happens if it is run using `RunMode::Consumer`? Does it hang forever waiting for producer to offer?

Are these modes (`Producer` & `Consumer`) provided here to run this `main.rs` as producer & consumer separately and show that they communicate with each other on a platform?

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+
+    //receiver function which use async receive to get data, it waits for new data and process it once it arrives,
+    //it will receive data 10 times and print the received samples
+    async fn async_data_processor_fn<R: Runtime>(
+        subscribed: impl Subscription<Tire, R>,
+        is_timeout: bool,
+    ) {
+        println!("[RECEIVER] Async data processor started");
+        let mut buffer = SampleContainer::new(5);
+        for _ in 0..3 {
```
rpreddyhv: question: Is there any relation between 0..3 and 5?

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
```diff
@@ -0,0 +1,299 @@
...
+        consumer_builder
+            .build()
+            .expect("Failed to build consumer instance")
+    }
+
+    //sender will send data in each 1 second
+    async fn async_data_sender_fn<R: Runtime>(
+        offered_producer: VehicleOfferedProducer<R>,
+    ) -> VehicleOfferedProducer<R> {
+        for i in 0..6 {
```
rpreddyhv: question: Why only `0..6`?

rpreddyhv (commented):
few comments & queries to clarify.

score/mw/com/example/com-api-example/src/consumer.rs
```diff
@@ -0,0 +1,193 @@
...
+
+use crate::vehicle_monitor::VehicleMonitorConsumer;
+use crate::ExampleType;
+
+/// Timeout used when performing a cancellable async receive.
+const RECEIVE_TIMEOUT_MS: u64 = 500;
+/// Polling interval between iterations in synchronous receive loops.
+const SYNC_RECEIVE_POLL_INTERVAL_MS: u64 = 1000;
+
+impl<R: Runtime> VehicleMonitorConsumer<R> {
```
darkwisebear, line 36: Is there a reason why the impl block is separate from the type definition. Imo this only makes it harder to comprehend the logic.

score/mw/com/example/com-api-example/src/consumer.rs
```diff
@@ -0,0 +1,193 @@
...
+    /// asynchronous service discovery for all other modes.
+    pub async fn new(
+        runtime: &R,
+        service_id: InstanceSpecifier,
+        mode: &ExampleType,
+    ) -> Result<Self> {
+        let consumer_discovery =
+            runtime.find_service::<VehicleInterface>(FindServiceSpecifier::Specific(service_id));
+        let instances = match mode {
+            ExampleType::Sync => consumer_discovery
```
darkwisebear: Using a parameter to distinguish between sync and async doesn't make sense imo, as the call is enclosed in an async function in any case. I think you should split this into a sync and an async create function instead.

Now thinking about it, the definition of `get_available_instances_async` is odd: What's the async part of this method actually waiting for? Presence of at least one instance? If yes, then this behavior should be documented.

score/mw/com/example/com-api-example/src/consumer.rs
```diff
@@ -0,0 +1,193 @@
...
+            }
+        };
+        Self::from_service_instances(instances)
+    }
+
+    /// Read tire data sample using the strategy defined by `mode`.
+    ///
+    /// Takes `&self` to support all receive modes including streaming,
+    /// which requires exclusive access to the subscriber.
+    pub async fn read_tire_data(&self, mode: &ExampleType) -> Result<String> {
```
darkwisebear: Same comment as in the `new` case: Having a sync code path in an async function isn't wrong (as log as it's nonblocking, which is the case), but kinda pointless: Why have the burden of an async function without any benefit?

score/mw/com/example/com-api-example/src/consumer.rs
```diff
@@ -0,0 +1,193 @@
...
+                match result {
+                    Ok(0) => Ok("No tire data received".to_string()),
+                    Ok(x) => {
+                        let sample = sample_buf.pop_front().unwrap();
+                        Ok(format!("{x} samples received: sample[0] = {:?}", *sample))
+                    }
+                    Err(e) => Err(e),
+                }
+            }
+            _ => unreachable!(
```
darkwisebear: Another code smell that reveals the non-ideal design here: This method isn't defined for all modes, which leads to mixing static and dynamic properties in this regard.

darkwisebear (changes requested):

score/mw/com/example/com-api-example/BUILD
bharatGoswami8: This is for library related files to build.

score/mw/com/example/com-api-example/main.rs
bharatGoswami8: Updated with `tuple`.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: standardize the iteration number using const.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: standardize the iteration number using const.

score/mw/com/example/com-api-example/USAGE.md
bharatGoswami8: Rephrased the statement.

score/mw/com/example/com-api-example/USAGE.md
bharatGoswami8: It was added to give impression why we have iteration test separately with `tokio`, so the simple reason for that is we do not want to use tokio in application and as we have already existing test case with tokio which can be used for quick and async runtime related testing.

score/mw/com/example/com-api-example/src/lib.rs
bharatGoswami8: Yes in first iteration of implementation, it was in main file and we were having different functions for each operation, but as it was calling separate function for each APIs so i thought to generalize and keep minimum function call from main and parsing can be done on library,
But I agree in that case sync execution is done with async function which is not proper way.

I have moved to the first iteration design.

score/mw/com/example/com-api-example/BUILD
bharatGoswami8: Added.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: created `const`

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: created `const`

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: created `const`

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: created `const`

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: created `const`

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: Moved async and sync APIs in separate function and behavior is documented.

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: Agreed, it may give wrong impression to user as well, so moved async and sync function separately.

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: As `enum` moved to main so now design is purely based on each APIs function call.

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: i don't think so we need to create separate `VehicleMonitorConsumer` for async and sync APIs because calling of operation is from subscriber instance, if you mean separate module /files then i think as of now keeping in same file is fine as we don't have much processing logic and sync has only two APIs.

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: I was thinking to keep `vehicleMonitor` on separate module for better modularity but agreed to your point moved type into same file and removed vehicle_monitor file.

score/mw/com/example/com-api-example/main.rs
bharatGoswami8: As discussed, updated behavior on `usage.md` file as a note.

score/mw/com/example/com-api-example/tests_using_tokio_runtime.rs
bharatGoswami8: This is just share the Runtime instance and which is due to config loading , we can not load same config on different test it will create warning from underlying implementation.
Producer and Consumer creating for each test is seperate.

rpreddyhv (approved):
Looks okay to me!

rpreddyhv (approved):

score/mw/com/example/com-api-example/src/producer.rs
```diff
@@ -0,0 +1,71 @@
...
+use com_api::{
+    Builder, InstanceSpecifier, OfferedProducer, Producer, Publisher, Result, Runtime,
+    SampleMaybeUninit, SampleMut,
+};
+use com_api_gen::{Tire, VehicleInterface};
+use std::thread;
+use std::time::Duration;
+
+// Delay between sample data sending in the producer loop.
+const PPRODUCER_SEND_INTERVAL_MS: u64 = 1000;
```
tejaswinijoshi89: Change PPRODUCER to PRODUCER

score/mw/com/example/com-api-example/src/consumer.rs
```diff
@@ -0,0 +1,180 @@
...
+    /// Shared constructor logic: picks the first available service instance and subscribes.
+    fn from_service_instances<B>(instances: impl IntoIterator<Item = B>) -> Result<Self>
+    where
+        B: ConsumerBuilder<VehicleInterface, R>,
+    {
+        let handle_index = 0; // or any index you need from vector of instances
+        let consumer_builder = instances
+            .into_iter()
+            .nth(handle_index)
+            .expect("Failed to get consumer builder at specified handle index");
```
tejaswinijoshi89, line 46: I think instead of .expect it should be .ok_or_else as expect will generate panic. As function already returns the Result, error can be just propagated instead of introducing panic

score/mw/com/example/com-api-example/src/consumer.rs
```diff
@@ -0,0 +1,180 @@
...
+
+        let consumer = consumer_builder
+            .build()
+            .expect("Failed to build consumer instance");
```
tejaswinijoshi89: I think instead of .expect it should be .ok_or_else as expect will generate panic. As function already returns the Result, error can be just propagated instead of introducing panic

score/mw/com/example/com-api-example/src/producer.rs
```diff
@@ -0,0 +1,71 @@
...
+
+// Example struct demonstrating composition with VehicleConsumer
+pub struct VehicleMonitorProducer<R: Runtime> {
+    pub(crate) producer: VehicleOfferedProducer<R>,
```
tejaswinijoshi89: Is it necessary to make producer public?

anmittag:
@bharatGoswami8 there are some comments/findings in your PR, please adapt

score/mw/com/example/com-api-example/src/producer.rs
bharatGoswami8: Done, it was typo.

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: Done.

score/mw/com/example/com-api-example/src/producer.rs
bharatGoswami8: It was there because we were using in crate but later with updated design it is not required now, removed.

score/mw/com/example/com-api-example/src/consumer.rs
bharatGoswami8: `.nth()` returns `Option`. Propagating it as `Result` would require an application-level error enum, which this example does not define.

tejaswinijoshi89 (approved):

LittleHuba (approved):

PR #558, opened 2026-06-18
fix: resolve strict warning build breaks in consumer test sources

AAmbuj:
remove unused lambda capture and mark reserved UIDs as maybe_unused to pass strict warning-as-error builds


castler:
@AAmbuj please converto to non-draft again once CI is passing.

PR #559, opened 2026-06-18
mw/com: Add integration tests for moving SkeletonEvent

bemerybmw:
In the near future, we should implement similar tests for SkeletonFields and SkeletonMethods. We should also implement similar tests on proxy side.

### DIFF
--- a/score/mw/com/test/methods/methods_test_resources/BUILD
+++ b/score/mw/com/test/methods/methods_test_resources/BUILD
@@ -55,7 +55,7 @@ cc_library(
     hdrs = ["process_synchronizer.h"],
     features = COMPILER_WARNING_FEATURES,
     visibility = [
-        "//score/mw/com/test/methods:__subpackages__",
+        "//score/mw/com/test:__subpackages__",
     ],
     deps = [
         "//score/mw/com/test/common_test_resources:fail_test",
--- a/score/mw/com/test/skeleton_move_semantics/consumer.h
+++ b/score/mw/com/test/skeleton_move_semantics/consumer.h
@@ -1,0 +1,31 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_TEST_SKELETON_MOVE_SEMANTICS_CONSUMER_H
+#define SCORE_MW_COM_TEST_SKELETON_MOVE_SEMANTICS_CONSUMER_H
+
+#include "score/mw/com/test/skeleton_move_semantics/test_parameters.h"
+#include "score/mw/com/types.h"
+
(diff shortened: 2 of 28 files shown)


score/mw/com/test/methods/methods_test_resources/BUILD
```diff
@@ -21,7 +21,7 @@ cc_library(
     hdrs = ["proxy_container.h"],
     features = COMPILER_WARNING_FEATURES,
     visibility = [
-        "//score/mw/com/test/methods:__subpackages__",
```
bemerybmw: These changes are temporary, Sahithi has a pr moving them into shared resources

score/mw/com/test/skeleton_move_semantics/main_consumer.cpp
```diff
@@ -0,0 +1,138 @@
...
+    score::mw::com::runtime::InitializeRuntime(argc, argv);
+
+    score::cpp::stop_source stop_source{};
+    const bool sig_term_handler_setup_success = score::mw::com::SetupStopTokenSigTermHandler(stop_source);
+    if (!sig_term_handler_setup_success)
+    {
+        std::cerr << "Unable to set signal handler for SIGINT and/or SIGTERM, cautiously continuing" << std::endl;
+    }
+
+    const auto num_send_iterations = [&test_configuration]() {
```
bemerybmw: Extract this since it's duplicated in all 3 mains

score/mw/com/test/skeleton_move_semantics/integration_test/move_assign_not_offered_skeleton_different_process_test.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+from test_fixture import consumer, provider, SkeletonMoveScenario
+
+NUMBER_OF_SAMPLES_TO_SEND_PER_OFFER = 10
```
bemerybmw: This value is not really important for the test and so could be a constant defined in c++ rather than making it configurable here.

score/mw/com/test/skeleton_move_semantics/consumer.h
```diff
@@ -0,0 +1,31 @@
...
+#include "score/mw/com/types.h"
+
+#include <score/stop_token.hpp>
+
+namespace score::mw::com::test
+{
+
+void RunConsumer(const InstanceSpecifier& instance_specifier,
+                 const std::size_t num_samples_to_receive,
+                 const std::size_t num_send_iterations,
```
castler, line 26: you are missing an include for `[<cstddef>](https://en.cppreference.com/cpp/header/cstddef)`

score/mw/com/test/skeleton_move_semantics/main_consumer_and_provider.cpp
```diff
@@ -0,0 +1,146 @@
...
+            //   - Create a skeleton
+            //   - offer the service
+            //   - send samples
+            //   - move construct the skeleton while the service is offered
+            //   - send samples
+            //   - stop offering the service
+            //   - offer the service again
+            //   - send samples again
+            // The receiver therefore expects three send iterations.
+            return 3U;
```
castler, line 81: This needs to be one

castler (approved):
I tried my best to find something - but just found nits. One for you ;)

PR #566, opened 2026-06-18
mw/com: Add ToString for logging purposes

bemerybmw:


castler (changes requested):
As discussed, please use respective overloads for `operator<<` for logging. Esp. to avoid dynamic memory allocation creating ´std::string´

castler (approved):

Issue #431, opened 2026-05-14
Improvement: Optimization of unsafe call in rust FFI


bharatGoswami8:
I think this unsafety is misplaced. Afaiu, the main point is that `FatPtr` is essentially a type-erased Rust fat ptr. However, we could turn this method into something safe if we wrap the `FatPtr` into a `FindServiceCallable` and embed the `FatPtr` inside this type. The invariant of this type would then be the guarantee that it points to the appropriate callback. If the Rust compiler cannot verify that, then the unsafety is pushed to the point where this new type is instantiated. And this is imo the better place, since it is directly at the point where the original type gets "encoded into the Rust type system".

_Originally posted by @darkwisebear in https://github.com/eclipse-score/communication/pull/388#discussion_r3234005667_

bharatGoswami8:
`get_allocatee_ptr` FFI API should take type rather than c_void pointer
https://github.com/eclipse-score/communication/pull/388#discussion_r3288607246

PR #577, opened 2026-06-22
Rust::com FFI Unsafe Optimization

bharatGoswami8:
* FindServiceCallback optimized
* Updated callback with catch_unwind

https://github.com/eclipse-score/communication/issues/431

### DIFF
--- a/score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_lola.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_lola.rs
@@ -42,8 +42,13 @@ unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_sample(
     // The caller must ensure the lifetime of the closure outlives this call.
     let callable: &mut dyn FnMut(*mut std::ffi::c_void) = unsafe { std::mem::transmute(*ptr) };
 
-    // Invoke the closure with the void* sample pointer
-    callable(sample_ptr);
+    // Invoke the closure with the void* sample pointer.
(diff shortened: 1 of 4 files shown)


score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi_lola.rs
```diff
@@ -42,8 +42,13 @@ unsafe extern "C" fn mw_com_impl_call_dyn_ref_fnmut_sample(
...
     let callable: &mut dyn FnMut(*mut std::ffi::c_void) = unsafe { std::mem::transmute(*ptr) };
 
-    // Invoke the closure with the void* sample pointer
-    callable(sample_ptr);
+    // Invoke the closure with the void* sample pointer.
+    // catch_unwind prevents a Rust panic from unwinding across the C++ stack boundary.
+    if std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| callable(sample_ptr))).is_err() {
+        // LOG the error here, Once logging mechanism is available.
+        // Abort to prevent unwinding across FFI boundary
+        std::process::abort();
```
bharatGoswami8, line 50: Should we `abort` the process, or is there a better way to address this situation?

tejaswinijoshi89: As per my understanding, this situation might arrive only in case of undefined or partially destructed state due to memory corruption and if this understanding is correct then aborting the service is preferred solution.

tejaswinijoshi89 (approved):

LittleHuba (approved):

PR #587, opened 2026-06-24
Showcase QNX SDP bug

LittleHuba:
This PR is not meant to be merged. It only exists to showcase a bug to QNX.


LittleHuba:
Showcase no longer required. Bug in QNX SDP 8.0.4 acknowledged.

PR #589, opened 2026-06-24
Avoid Console logging as validation mechanism

ahmed0mousa:


LittleHuba (commented):
In many tests you drop the check that the logs contain a specific substring. Was this an active decision by you? Could you highlight the process how you came to this decision please?

ahmed0mousa:
I add the missed expectations for checking that the logs contain a specific substring. now there are only two tests in `runtime_configuration_test.cpp` that don't have the expectation of the logged message, but I think the current expectation which `EXPECT_DEATH` is enough

LittleHuba (approved):

PR #596, opened 2026-06-25
mw/com: Undo change which breaks binding independence of factories 

bemerybmw:
This commit undoes most of the changes to the binding factories added in
https://github.com/eclipse-score/communication/commit/5804ab699490b062877296599826211a30075f0e. This change removed the
variant handling in the factory create functions which are needed to
support different bindings in mw/com.

It also adds ServiceElementType to ProxyEventBindingFactory::Create so
that it can be reused by ProxyFieldBindingFactory.

### DIFF
--- a/score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
+++ b/score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
@@ -97,35 +97,47 @@ std::unique_ptr<ProxyMethodBinding> ProxyMethodBindingFactoryImpl<ReturnType(Arg
     const std::string_view method_name,
     MethodType method_type) noexcept
 {
-    // When method_type is Get or Set, method_name holds a field name, not a method name. A field
-    // only has one id at the lola level, so the Get and the Set of the same field both resolve to
-    // the same element id here. The Get-vs-Set split is added further down when we pair the id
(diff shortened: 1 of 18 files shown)


score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
```diff
@@ -97,35 +97,47 @@ std::unique_ptr<ProxyMethodBinding> ProxyMethodBindingFactoryImpl<ReturnType(Arg
...
-
-    // Pairing the id with method_type is what keeps Get and Set separate from here on. They share
-    // the same id, but the pair (id, method_type) is different, so the two end up in separate
-    // entries of the proxy_methods_ / skeleton_methods_ maps in the binding layer.
-    const lola::ProxyMethodInstanceIdentifier proxy_method_instance_identifier{
-        lookup->parent.GetProxyInstanceIdentifier(), {lookup->element_fq_id.element_id_, method_type}};
-
-    return std::make_unique<lola::ProxyMethod>(
-        lookup->parent, proxy_method_instance_identifier, type_erased_element_info);
+    auto method_name_str = std::string{method_name};
```
castler, line 100: I don't think it makes sense to generate here a std::string out of the std::string_view

castler (approved):
Agreed to tackle in a follow up.

PR #598, opened 2026-06-25
Replace all instances of score::cpp::optional for std::optional in LoLa

ahmed0mousa:


ahmed0mousa:
now in `mw/com` we still have to use `score::cpp::optional` in two examples:
- In few of our tests as we use `CompositeRecorder::StartRecord` which returns `score::cpp::optional`.
- Calling `analysis::tracing::AraComProperties` which takes `score::cpp::optional` as one of the arguments.

switching to `std::optional` in these two examples should happened after the called functions switch to `std::optional`

castler (approved):

Issue #432, opened 2026-05-14
Improvement: Cleanup deprecated Rust APIs


bharatGoswami8:
### What

As we have new rust API ready and released - https://github.com/eclipse-score/communication/tree/main/score/mw/com/impl/rust/com-api
we need to clean-up the deprecated rust -apis .

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

bharatGoswami8:
We are planning to remove old APIs on 06 July 2026.

bharatGoswami8:
PR is merged.

PR #602, opened 2026-06-25
Rust::com Cleanup deprecated Rust APIs

bharatGoswami8:
- Removed old APIs example.
- Removed legacy integration tests and related test documentation.
- The LoLa runtime builder, consumer, and producer switch to the bridge-local instance and handle types and call the bridge initialization path.
- Cleaned up obsolete configuration, generated artifacts, and compatibility layers no longer used by the product.

https://github.com/eclipse-score/communication/issues/432


score/mw/com/impl/rust/com-api/com-api-ffi-lola/bridge_ffi.rs
```diff
@@ -56,10 +56,9 @@
 // - This enables type-safe communication without C++ needing to know Rust types at compile time
 //
 // Dependencies:
-// - It relies on proxy_bridge_rs crate for FatPtr and ProxyWrapperClass definitions
+// - C FFI implementations (mw_com_impl_*) are all in registry_bridge_macro.cpp
```
tejaswinijoshi89: Suggestion:
Better to add bridge_ffi.rs in separate folder as its generic FFI interface abstraction and currently implemented by lola but eventually it can be implemented by any other communication mechanism

bharatGoswami8: Yes, we can plan to move with common file but may be with separate PR or when Method/Field FFI implementation we are doing that time it may help to create separate interface file.
As this PR for cleanup old APIs so  i think we can keep it in same folder as of now.

tejaswinijoshi89 (approved):

bharatGoswami8:
PR required review from @darkwisebear, so please do not merge it.

LittleHuba (approved):

PR #603, opened 2026-06-26
refactor: clean up shadowed and unused identifiers across com impl/tests

AAmbuj:
-rename shadowing lambda captures and local variables for clarity -replace unused result bindings with std::ignore in tests -remove unused constants and lambda captures
-rename GenericProxy events map member to generic_events -silence unused parameter warnings in test overrides -keep behavior unchanged while improving static-analysis compliance


AAmbuj:
This task is complied by other PR

PR #605, opened 2026-06-26
Race condition on SubscriptionStateMachine::current_state_idx

ahmed0mousa:
mw/com: eliminate data race on current_state_idx_ in GetSlotCollectorLockFree
GetSlotCollectorLockFree() read current_state_idx_ through GetCurrentEventState()
without holding state_mutex_. Concurrently, StopOfferEvent() and ReOfferEvent()
write current_state_idx_ under state_mutex_ from the middleware notification
thread. This constitutes a data race (undefined behaviour per C++11) and is
caught by ThreadSanitizer.

All three state objects (SubscribedState, SubscriptionPendingState,
NotSubscribedState) return the same subscription_data_.slot_collector_ field
from their GetSlotCollector() implementations, so the state dispatch was
redundant. Accessing the field directly eliminates the race entirely:
StopOfferEvent() and ReOfferEvent() never modify subscription_data_.slot_collector_,
so the direct access remains safe under the existing single-threaded AoU for
user-facing calls (SubscribeEvent / UnsubscribeEvent / GetSlotCollectorLockFree).

Issue: SWP-209400

### DIFF
--- a/score/mw/com/impl/bindings/lola/subscription_state_machine.cpp
+++ b/score/mw/com/impl/bindings/lola/subscription_state_machine.cpp
@@ -59,19 +59,19 @@ SubscriptionStateMachineState SubscriptionStateMachine::GetCurrentState() const 
 
 SubscriptionStateMachineState SubscriptionStateMachine::GetCurrentStateNoLock() const noexcept
 {
-    return current_state_idx_;
+    return current_state_idx_.load();
 }
 
 SubscriptionStateBase& SubscriptionStateMachine::GetCurrentEventState() noexcept
 {
     // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index): current_state_idx_ can't be const expression
-    return *states_[static_cast<std::uint8_t>(current_state_idx_)];
+    return *states_[static_cast<std::uint8_t>(current_state_idx_.load())];
 }
 
 const SubscriptionStateBase& SubscriptionStateMachine::GetCurrentEventState() const noexcept
 {
     // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-constant-array-index): current_state_idx_ can't be const expression
-    return *states_[static_cast<std::uint8_t>(current_state_idx_)];
+    return *states_[static_cast<std::uint8_t>(current_state_idx_.load())];
 }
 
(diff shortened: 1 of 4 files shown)


score/mw/com/impl/bindings/lola/subscription_state_machine.cpp
```diff
@@ -130,12 +130,12 @@ std::optional<std::uint16_t> SubscriptionStateMachine::GetMaxSampleCount() const
 
 score::cpp::optional<SlotCollector>& SubscriptionStateMachine::GetSlotCollectorLockFree() noexcept
 {
-    return GetCurrentEventState().GetSlotCollector();
+    return subscription_data_.slot_collector_;
```
bemerybmw: This is now bypassing the state machine completely which IMO should be a last resort. Because we now for example have no reaction if this function is called before subscribe is called (which was a previous invariant of the state machine).

My understanding is that we're getting the race condition on current_state_idx_. Couldn't we just make that atomic? If we're in a race with changing from subscribed / pending then the behaviour of GetSlotCollector is the same in both so it doesn't matter whether we read or write  current_state_idx_ first. We'd still need the justification for GetSlotCollectorLockFree why this is safe to basically access the slot collector while we're doing a state transition.

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/subscription_state_machine.cpp
ahmed0mousa: I had three options in my mind:
1. Bypass state dispatch in `GetSlotCollectorLockFree`
2. Make `current_state_idx_` atomic
3. Acquire the mutex in `GetSlotCollectorLockFree` (which will not be lock free anymore)

I did chose option 1 because the root cause is unnecessary indirection through `current_state_idx_` when all states delegate to the same field. Removing that indirection eliminates the race without any overhead and without weakening any invariants that actually matter at runtime. but now I see that The `NotSubscribedState::GetSlotCollector()` assert is a debug-time invariant that catches misuse (calling `GetSlotCollectorLockFree` before subscribing) that actually matter at runtime. So option 2 (atomic) is objectively the better fix.

bemerybmw (approved):

Issue #483, opened 2026-05-28
mw::com: Add a integration test that proves that moved from object is still valid and can call all public methods.


KrishaDeshkool:
### What

after finding a bug in move semantics and fixing it we now need a integration test that proves that moved from object is still valid and can call all public methods.

### Estimates for realization

It can be done by using existing integration tests as examples.
same test but before calling public methods move the object explicitly and then call public methods on moved to object.

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

PR #607, opened 2026-06-29
mw/com: Add the integration tests for moving proxyEvent

sahithi-nukala:
Adding the integration tests that verify a Proxy (and its ProxyEvent) can be moved without disrupting service discovery, subscription, or sample reception. 

Issue: https://github.com/eclipse-score/communication/issues/483

### DIFF
--- a/score/mw/com/test/common_test_resources/proxy_container.h
+++ b/score/mw/com/test/common_test_resources/proxy_container.h
@@ -22,6 +22,7 @@
 #include <memory>
 #include <mutex>
 #include <string>
+#include <utility>
 
 namespace score::mw::com::test
 {
@@ -39,11 +40,13 @@ class ProxyContainer
         return *proxy_;
     }
 
-    Proxy&& Extract()
+    Proxy Extract()
     {
         SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(proxy_ != nullptr,
                                                     "Proxy was not successfully created! Cannot extract it!");
-        return std::move(*proxy_);
+        auto proxy = std::move(*proxy_);
+        proxy_.reset();
+        return proxy;
     }
 
   private:
--- a/score/mw/com/test/common_test_resources/proxy_container.h
+++ b/score/mw/com/test/common_test_resources/proxy_container.h
@@ -40,13 +40,15 @@ class ProxyContainer
         return *proxy_;
     }
 
-    Proxy Extract()
+    /// \brief Extracts the proxy, leaving the ProxyContainer in a valid but unspecified state.
+    ///
+    /// This function can only be called on an rvalue ProxyContainer, to make it clear that the ProxyContainer should
+    /// not be used after calling Extract()
+    Proxy&& Extract() &&
     {
         SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(proxy_ != nullptr,
                                                     "Proxy was not successfully created! Cannot extract it!");
-        auto proxy = std::move(*proxy_);
-        proxy_.reset();
-        return proxy;
+        return std::move(*proxy_);
     }
 
   private:
--- a/score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
+++ b/score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
@@ -1,0 +1,48 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_TEST_COMMON_TEST_RESOURCES_SEND_INCREMENTING_SEQUENCE_OF_SAMPLES_H
+#define SCORE_MW_COM_TEST_COMMON_TEST_RESOURCES_SEND_INCREMENTING_SEQUENCE_OF_SAMPLES_H
+
+#include "score/mw/com/test/common_test_resources/fail_test.h"
+
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <thread>
+
+namespace score::mw::com::test
+{
+
+/// Sends an incrementing sequence of number_of_samples_to_send_per_offer samples via
+/// skeleton.moved_event_, starting at initial_value.
+template <typename Skeleton>
+void SendIncrementingSequenceOfSamples(Skeleton& skeleton,
+                                       const std::size_t number_of_samples_to_send_per_offer,
+                                       const std::uint32_t initial_value)
+{
+    std::cout << "\nProvider: Sending " << number_of_samples_to_send_per_offer << " samples" << std::endl;
+    for (std::uint32_t i = 0; i < number_of_samples_to_send_per_offer; ++i)
+    {
+        auto send_result = skeleton.moved_event_.Send(i + initial_value);
+        if (!send_result.has_value())
+        {
+            FailTest("Provider: Send failed: ", send_result.error());
+        }
+        std::this_thread::sleep_for(std::chrono::milliseconds(20));
+    }
+}
+
+}  // namespace score::mw::com::test
+
+#endif  // SCORE_MW_COM_TEST_COMMON_TEST_RESOURCES_SEND_INCREMENTING_SEQUENCE_OF_SAMPLES_H
--- a/score/mw/com/test/move_semantics/proxy_event/consumer.cpp
+++ b/score/mw/com/test/move_semantics/proxy_event/consumer.cpp
@@ -1,0 +1,303 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/test/move_semantics/proxy_event/consumer.h"
+
+#include "score/mw/com/test/common_test_resources/fail_test.h"
+#include "score/mw/com/test/common_test_resources/process_synchronizer.h"
+#include "score/mw/com/test/common_test_resources/proxy_container.h"
+#include "score/mw/com/test/common_test_resources/proxy_event_receiver.h"
+#include "score/mw/com/test/common_test_resources/proxy_event_state_change_notifier.h"
+#include "score/mw/com/test/move_semantics/proxy_event/proxy_event_move_semantics_interface.h"
+#include "score/mw/com/test/move_semantics/proxy_event/test_parameters.h"
+#include "score/mw/com/types.h"
+
+#include <cstdint>
+#include <string>
+#include <utility>
+
+namespace score::mw::com::test
+{
+namespace
+{
+
+const std::string kConsumerBatchReceivedNotificationShmPath{
+    "/proxy_event_move_semantics_consumer_batch_received_notification"};
+const std::string kReofferTriggerNotificationShmPath{"/proxy_event_move_semantics_provider_withdraw_notification"};
+
+void CheckIfProxyIsFullyFunctional(ProxyMoveSemanticsProxy& proxy,
+                                   ProcessSynchronizer& process_synchronizer,
+                                   ProcessSynchronizer& reoffer_trigger_synchronizer,
+                                   const std::size_t num_samples_to_receive,
+                                   const score::cpp::stop_token& stop_token)
+{
+    // Step 3. Register receive handler
+    std::cout << "\nConsumer: Step 3 - Register receive handler" << std::endl;
+    ProxyEventReceiver proxy_event_receiver{proxy.moved_event_};
+
+    // Step 4. Register state change handler
+    std::cout << "\nConsumer: Step 4 - Register state change handler" << std::endl;
+    ProxyEventStateChangeNotifier proxy_event_state_change_notifier{proxy.moved_event_};
+
+    // Step 6. Receive the first batch and notify the provider when done.
+    std::cout << "\nConsumer: Step 6 - Receive first batch of samples" << std::endl;
+    if (!proxy_event_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    }
+    if (!proxy_event_receiver.WaitForSamples(stop_token, kInitialSamplesToSend))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForSamples was interrupted by stop_token");
+    }
+    process_synchronizer.Notify();
+
+    // Step 7. Unsubscribe and subscribe again
+    // std::cout << "\nConsumer: Step 7 - Unsubscribe and subscribe again" << std::endl;
+    // proxy.moved_event_.Unsubscribe();
+    // std::cout << "\nConsumer:Current Subscription state after calling kNotSubscribed:"
+    //           << static_cast<std::uint32_t>(proxy.moved_event_.GetSubscriptionState()) << std::endl;
+    // if (!proxy_event_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kNotSubscribed))
+    // {
+    //     FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    // }
+    // proxy.moved_event_.Subscribe(num_samples_to_receive);
+    // if (!proxy_event_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    // {
+    //     FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    // }
+
+    // Step 8.  Notify provider to re-offer the service
+    std::cout << "\nConsumer: Step 8 - Notify provider to re-offer the service." << std::endl;
+    reoffer_trigger_synchronizer.Notify();
+
+    // Step 9. wait for the state to change to kSubscriptionPending
+    std::cout << "\nConsumer: Step 9 - Wait for the state to change to kSubscriptionPending" << std::endl;
+    std::cout
+        << "\nConsumer: Current Subscription state during waiting for the state to change to kSubscriptionPending:"
+        << static_cast<std::uint32_t>(proxy.moved_event_.GetSubscriptionState()) << std::endl;
+    if (!proxy_event_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscriptionPending))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    }
+
+    // Step 10. Wait for the state to change to kSubscribed
+    std::cout << "\nConsumer: Step 10 - Wait for the state to change to kSubscribed" << std::endl;
+    std::cout << "\nConsumer: Current Subscription state before wait:"
+              << static_cast<std::uint32_t>(proxy.moved_event_.GetSubscriptionState()) << std::endl;
+    if (!proxy_event_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    }
+    std::cout << "Consumer: Step 10 - current subscription state after wait = "
+              << static_cast<std::uint32_t>(proxy.moved_event_.GetSubscriptionState()) << std::endl;
+
+    // step 11. Wait for the newSamples
+    std::cout << "\nConsumer: Step 11 - Wait for the new samples" << std::endl;
+    std::cout << "\nConsumer: expected second batch of samples: " << kSamplesToSendAfterReOffer.size() << std::endl;
+    if (!proxy_event_receiver.WaitForSamples(stop_token, kSamplesToSendAfterReOffer))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForSamples was interrupted by stop_token");
+    }
+    std::cout << "\nConsumer: Received the second batch of samples successfully" << std::endl;
+}
+
+void RunConsumerMoveConstructProxyBeforeSubscribe(ProcessSynchronizer& process_synchronizer,
+                                                  ProcessSynchronizer& reoffer_trigger_synchronizer,
+                                                  const std::size_t num_samples_to_receive,
+                                                  const score::cpp::stop_token& stop_token)
+{
+    ProxyContainer<ProxyMoveSemanticsProxy> proxy_container{};
+
+    // Step 1. Find service and create original proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+
+    // Step 2. Move construct proxy before subscribe
+    std::cout << "\nConsumer: Step 2 - Move construct proxy before subscribe" << std::endl;
+    auto original_proxy = std::move(proxy_container).Extract();
+
+    // Step 5. Subscribe
+    std::cout << "\nConsumer: Step 5 - Subscribe" << std::endl;
+    auto subscribe_result = original_proxy.moved_event_.Subscribe(num_samples_to_receive);
+    if (!subscribe_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: Subscribe failed: ", subscribe_result.error());
+    }
+
+    CheckIfProxyIsFullyFunctional(
+        original_proxy, process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+
+    // Step 12. Notify to provider we're done.
+    std::cout << "\nConsumer: Step 12 - Notify provider we're done." << std::endl;
+    process_synchronizer.Notify();
+    std::cout << "Consumer: Done with all iterations, exiting" << std::endl;
+}
+
+void RunConsumerMoveConstructProxyWhileSubscribed(ProcessSynchronizer& process_synchronizer,
+                                                  ProcessSynchronizer& reoffer_trigger_synchronizer,
+                                                  const std::size_t num_samples_to_receive,
+                                                  const score::cpp::stop_token& stop_token)
+{
+    ProxyContainer<ProxyMoveSemanticsProxy> proxy_container{};
+
+    // Step 1. Find service and create original proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create original proxy" << std::endl;
+    proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+    auto& original_proxy = proxy_container.GetProxy();
+
+    // Step 2. Subscribe on original proxy
+    std::cout << "\nConsumer: Step 2 - Subscribe on original proxy" << std::endl;
+    auto subscribe_result = original_proxy.moved_event_.Subscribe(num_samples_to_receive);
+    if (!subscribe_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: Subscribe failed: ", subscribe_result.error());
+    }
+
+    // Step 5. Move construct while subscribed. The receive and state-change handlers are preserved on the moved-to
+    // proxy, so no re-registration is needed.
+    std::cout << "\nConsumer: Step 5 - Move construct while subscribed" << std::endl;
+    auto moved_proxy = std::move(proxy_container).Extract();
+
+    CheckIfProxyIsFullyFunctional(
+        moved_proxy, process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+}
+
+void RunConsumerMoveAssignProxyBeforeSubscribe(ProcessSynchronizer& process_synchronizer,
+                                               ProcessSynchronizer& reoffer_trigger_synchronizer,
+                                               const std::size_t num_samples_to_receive,
+                                               const score::cpp::stop_token& stop_token)
+{
+    ProxyContainer<ProxyMoveSemanticsProxy> moved_from_proxy_container{};
+    ProxyContainer<ProxyMoveSemanticsProxy> moved_to_proxy_container{};
+
+    // Step 1. Find service and create original proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create original proxy" << std::endl;
+    moved_from_proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+    std::cout << "\nConsumer: Step 2 - Find service and create proxy" << std::endl;
+    moved_to_proxy_container.CreateProxy(kInstanceSpecifierMovedFrom, "proxy_event_move_semantics");
+
+    auto& moved_from_proxy = moved_from_proxy_container.GetProxy();
+    auto& moved_to_proxy = moved_to_proxy_container.GetProxy();
+
+    // Step 3. Move assign proxy = move(original proxy) before subscribe
+    std::cout << "\nConsumer: Step 3 - Move assign proxy = move(original proxy) before subscribe" << std::endl;
+    moved_to_proxy = std::move(moved_from_proxy_container).Extract();
+
+    // Step 5. Subscribe
+    std::cout << "\nConsumer: Step 5 - Subscribe on proxy" << std::endl;
+    auto subscribe_result = moved_to_proxy.moved_event_.Subscribe(num_samples_to_receive);
+    if (!subscribe_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: Subscribe failed: ", subscribe_result.error());
+    }
+
+    CheckIfProxyIsFullyFunctional(
+        moved_to_proxy, process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+
+    // Step 12. Notify to provider we're done.
+    std::cout << "\nConsumer: Step 12 - Notify provider we're done." << std::endl;
+    process_synchronizer.Notify();
+    std::cout << "Consumer: Done with all iterations, exiting" << std::endl;
+}
+
+void RunConsumerMoveAssignProxyWhileSubscribed(ProcessSynchronizer& process_synchronizer,
+                                               ProcessSynchronizer& reoffer_trigger_synchronizer,
+                                               const std::size_t num_samples_to_receive,
+                                               const score::cpp::stop_token& stop_token)
+{
+    ProxyContainer<ProxyMoveSemanticsProxy> moved_from_proxy_container{};
+    ProxyContainer<ProxyMoveSemanticsProxy> moved_to_proxy_container{};
+
+    // Step 1. Create two proxies
+    std::cout << "\nConsumer: Step 1 - Find service and create active proxy" << std::endl;
+    moved_from_proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+    std::cout << "\nConsumer: Step 2 - Find service and create passive proxy" << std::endl;
+    moved_to_proxy_container.CreateProxy(kInstanceSpecifierMovedFrom, "proxy_event_move_semantics");
+
+    auto& moved_from_proxy = moved_from_proxy_container.GetProxy();
+    auto& moved_to_proxy = moved_to_proxy_container.GetProxy();
+
+    // Step 3. Subscribe
+    std::cout << "\nConsumer: Step 3 - Subscribe active proxy" << std::endl;
+    auto subscribe_result = moved_from_proxy.moved_event_.Subscribe(num_samples_to_receive);
+    if (!subscribe_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: Subscribe failed: ", subscribe_result.error());
+    }
+
+    // Step 6. Move assign while active. The receive and state-change handlers are preserved on the moved-to proxy, so
+    // no re-registration is needed.
+    std::cout << "\nConsumer: Step 6 - Move assign while active" << std::endl;
+    moved_to_proxy = std::move(moved_from_proxy_container).Extract();
+
+    CheckIfProxyIsFullyFunctional(
+        moved_to_proxy, process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+}
+}  // namespace
+
+void RunConsumer(const ProxyMoveScenario& scenario,
+                 const std::size_t num_samples_to_receive,
+                 const score::cpp::stop_token& stop_token)
+{
+    auto process_synchronizer_result = ProcessSynchronizer::Create(kConsumerBatchReceivedNotificationShmPath);
+    if (!process_synchronizer_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: could not create ready synchronizer");
+    }
+
+    ExitFunctionGuard done_guard{[&process_synchronizer_result]() {
+        process_synchronizer_result->Notify();
+    }};
+
+    auto reoffer_trigger_synchronizer_result = ProcessSynchronizer::Create(kReofferTriggerNotificationShmPath);
+    if (!reoffer_trigger_synchronizer_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: could not create reoffer trigger synchronizer");
+    }
+
+    auto& process_synchronizer = *process_synchronizer_result;
+    auto& reoffer_trigger_synchronizer = *reoffer_trigger_synchronizer_result;
+
+    switch (scenario)
+    {
+        case ProxyMoveScenario::kMoveConstructBeforeSubscribe:
+        {
+            RunConsumerMoveConstructProxyBeforeSubscribe(
+                process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+            break;
+        }
+        case ProxyMoveScenario::kMoveConstructWhileSubscribed:
+        {
+            RunConsumerMoveConstructProxyWhileSubscribed(
+                process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+            break;
+        }
+        case ProxyMoveScenario::kMoveAssignBeforeSubscribe:
+        {
+            RunConsumerMoveAssignProxyBeforeSubscribe(
+                process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+            break;
+        }
+        case ProxyMoveScenario::kMoveAssignWhileSubscribed:
+        {
+            RunConsumerMoveAssignProxyWhileSubscribed(
+                process_synchronizer, reoffer_trigger_synchronizer, num_samples_to_receive, stop_token);
+            break;
+        }
+        case ProxyMoveScenario::kNumberOfScenarios:
+            [[fallthrough]];
+        default:
+            FailTest("Unknown proxy move scenario in consumer");
+    }
+}
+
+}  // namespace score::mw::com::test
--- a/score/mw/com/test/move_semantics/proxy_event/main_consumer.cpp
+++ b/score/mw/com/test/move_semantics/proxy_event/main_consumer.cpp
@@ -1,0 +1,45 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/runtime.h"
+
+#include "score/mw/com/test/common_test_resources/assert_handler.h"
+#include "score/mw/com/test/common_test_resources/stop_token_sig_term_handler.h"
+#include "score/mw/com/test/move_semantics/proxy_event/consumer.h"
+#include "score/mw/com/test/move_semantics/proxy_event/test_parameters.h"
+#include "score/string_manipulation/arguments/arguments.h"
+
+int main(int argc, const char** argv)
+{
+    auto test_configuration{score::mw::com::test::ReadCommandLineArguments(argc, argv)};
+
+    score::mw::com::test::SetupAssertHandler();
+    score::mw::com::runtime::InitializeRuntime(score::string_manipulation::GetArguments(argc, argv));
+
+    score::cpp::stop_source stop_source{};
+    const bool sig_term_handler_setup_success = score::mw::com::SetupStopTokenSigTermHandler(stop_source);
+    if (!sig_term_handler_setup_success)
+    {
+        std::cerr << "Unable to set signal handler for SIGINT and/or SIGTERM, cautiously continuing" << std::endl;
+    }
+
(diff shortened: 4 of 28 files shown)


score/mw/com/test/common_test_resources/proxy_container.h
```diff
@@ -39,6 +40,13 @@ class ProxyContainer
         return *proxy_;
     }
 
+    Proxy&& Extract()
```
bemerybmw: directory structure: 
move_semantics/
....proxy_event/
....skeleton_event/

score/mw/com/test/proxy_event_move_semantics/consumer.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        proxy_event_.UnsetSubscriptionStateChangeHandler();
+    }
+
+    ProxyStateChangeNotifier(const ProxyStateChangeNotifier&) = delete;
+    ProxyStateChangeNotifier& operator=(const ProxyStateChangeNotifier&) = delete;
+    ProxyStateChangeNotifier(const ProxyStateChangeNotifier&&) = delete;
+    ProxyStateChangeNotifier& operator=(const ProxyStateChangeNotifier&&) = delete;
+
+    bool WaitForStateChange(const score::cpp::stop_token& stop_token, SubscriptionState desired_state)
```
bemerybmw:
```suggestion
    [[nodiscard]] bool WaitForStateChange(const score::cpp::stop_token& stop_token, SubscriptionState desired_state)
```
If false is returned, we never got the expected state change so should terminate.

score/mw/com/test/proxy_event_move_semantics/consumer.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        proxy_event_.UnsetReceiveHandler();
+    }
+
+    ProxyEventReceiver(const ProxyEventReceiver&) = delete;
+    ProxyEventReceiver& operator=(const ProxyEventReceiver&) = delete;
+    ProxyEventReceiver(const ProxyEventReceiver&&) = delete;
+    ProxyEventReceiver& operator=(const ProxyEventReceiver&&) = delete;
+
+    void WaitForSamples(const score::cpp::stop_token& stop_token, const std::size_t num_samples_to_receive)
```
bemerybmw:
```suggestion
    [[nodiscard]] bool WaitForSamples(const score::cpp::stop_token& stop_token, const std::size_t num_samples_to_receive)
```
if false, didn't receive expected samples so should fail test.

score/mw/com/test/common_test_resources/proxy_container.h
sahithi-nukala: done in the PR - https://github.com/eclipse-score/communication/pull/610

score/mw/com/test/move_semantics/proxy_event/provider.cpp
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+}  // namespace
+
+void RunProvider(const std::size_t num_samples_to_send,
+                 const std::size_t num_send_iterations,
+                 const score::cpp::stop_token& stop_token)
+{
+    const auto moved_to_name = filesystem::Path{kInstanceSpecifierMovedTo.ToString()}.Filename().Native();
```
bemerybmw: Do we need to add moved_to_name to kInterprocessNotificationShmPath?

score/mw/com/test/move_semantics/proxy_event/provider.cpp
```diff
@@ -0,0 +1,101 @@
...
+            process_synchronizer_result->Reset();
+
+            std::cout << "\nProvider: Re-offer skeleton" << std::endl;
+            skeleton_container.OfferService("proxy_event_move_semantics");
+        }
+
+        const auto initial_value = static_cast<std::uint32_t>(iteration * num_samples_to_send) + 1U;
+        std::cout << "\nProvider: Iteration " << (iteration + 1U) << " of " << num_send_iterations << " - Send "
+                  << num_samples_to_send << " samples" << std::endl;
+        SendSamples(skeleton_container.GetSkeleton(), num_samples_to_send, initial_value);
```
bemerybmw: If you want to extract this into a function, then the name should be more descriptive IMO. Maybe something like SendIncrementingSequenceOfSamples. Otherwise passing in "initial_value" is a bit strange to me

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,384 @@
...
+{
+    ProxyContainer<ProxyMoveSemanticsProxy> proxy_container{};
+
+    // Step 1. Find service and create proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+
+    // Step 2. Move construct proxy before subscribe
+    std::cout << "\nConsumer: Step 2 - Move construct proxy before subscribe" << std::endl;
+    auto moved_proxy = proxy_container.Extract();
```
bemerybmw: IMO, it's better to make the move construction explicit like in the skeleton move test:
```
    skeleton_container.CreateSkeleton(kInstanceSpecifierMovedTo, "skeleton_event_move_semantics");
    auto original_skeleton = skeleton_container.Extract();

    // Step 2. Move construct skeleton
    std::cout << "\nProvider: Step 2 - Move construct skeleton" << std::endl;
    auto moved_to_skeleton = std::move(original_skeleton);
```

score/mw/com/test/common_test_resources/proxy_event_consumer.h
```diff
@@ -0,0 +1,109 @@
...
+
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <optional>
+
+namespace score::mw::com::test
+{
+
+inline auto MakeSampleSequenceCallback(std::optional<std::uint32_t>& latest_value, const char* failure_message_prefix)
```
bemerybmw: IMO, these functions are quite specific to the move semantic tests, so I wouldn't put them in the test resources.

score/mw/com/test/common_test_resources/proxy_event_consumer.h
```diff
@@ -0,0 +1,109 @@
...
+                      ProxyStateChangeNotifierType& proxy_state_change_notifier,
+                      ProcessSynchronizerType& process_synchronizer,
+                      const std::size_t num_samples_to_receive,
+                      const std::size_t num_iterations,
+                      const score::cpp::stop_token& stop_token)
+{
+    for (std::size_t iteration = 0U; iteration < num_iterations; ++iteration)
+    {
+        std::cout << "\nConsumer: Iteration " << (iteration + 1U) << " of " << num_iterations << std::endl;
+        const bool subscribed =
```
bemerybmw: Why are you waiting for a state change on every iteration?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,384 @@
...
+    std::cout << "\nConsumer: Step 5 - Subscribe" << std::endl;
+    auto subscribe_result = moved_proxy.moved_event_.Subscribe(num_samples_to_receive);
+    if (!subscribe_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: Subscribe failed: ", subscribe_result.error());
+    }
+
+    // Step 6. Wait for provider to send the first batch of values and notify
+    std::cout << "\nConsumer: Step 6 - Receive first batch of samples" << std::endl;
+    ReceiveAndNotify(proxy_event_receiver,
```
bemerybmw: In general, I don't think it's a good idea to extract too much actual test logic. We want the test sequence to be as clear as possible. Something like proxy_state_change_notifier.WaitForStateChange or proxy_event_receiver.WaitForSamples are basically doing one thing and it's completely clear from the name what that is. Here, looking at the name and the args, it's not really clear what it does. 
e.g. the name doesn't hint that it's also checking the subscription state (which I'm not sure it should be there at all), it's not really clear that it will receive samples, notify and then reset the process_synchronizer and then repeat that multiple times. IMO, that logic should be here in the main test sequence or it for example you could make it simpler like ReceiveSamplesAndNotify but do the loop here and reset the process_synchronizer here. But tbh at that point, I don't see much point in the function. I also think the fact that the function takes 6 params and has 3 template params is a bit of a code smell.

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,384 @@
...
+                     proxy_state_change_notifier,
+                     process_synchronizer,
+                     num_samples_to_receive,
+                     num_send_iterations - 1U,
+                     stop_token);
+
+    // Step 7. Unsubscribe and subscribe again across the provider's re-offer
+    std::cout << "\nConsumer: Step 7 - Unsubscribe and subscribe again" << std::endl;
+    ResubscribeAcrossReoffer(moved_proxy.moved_event_,
```
bemerybmw: similar to my comments above. This should either be made clearer or put here in the main.

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
bemerybmw: This is also not an exact science but is a bit of developer preference. We can also get an outside opinion. @limdor what do you think?

score/mw/com/test/common_test_resources/proxy_event_consumer.h
```diff
@@ -0,0 +1,109 @@
...
+template <typename ProxyEventType, typename ProxyStateChangeNotifierType, typename ProcessSynchronizerType>
+void ResubscribeAcrossReoffer(ProxyEventType& proxy_event,
+                              ProxyStateChangeNotifierType& proxy_state_change_notifier,
+                              ProcessSynchronizerType& process_synchronizer,
+                              const std::size_t num_samples_to_receive,
+                              const score::cpp::stop_token& stop_token)
+{
+    std::cout << "\nConsumer: Waiting for provider to withdraw its offer" << std::endl;
+    const bool withdrawn =
+        proxy_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscriptionPending);
```
bemerybmw: kSubscriptionPending shouldn't really be used to determine whether a provider is alive or not. I think you should probably use a ProcessSynchronizer instead. Although to be honest, I think it would also be fine to unsubscribe and subscribe without the provider reoffering. Is there a particular reason why you want to wait for the skeleton to reoffer?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    // Step 3. Register handlers on original proxy and receive first batch
+    std::cout << "\nConsumer: Step 3 - Register handlers on original proxy" << std::endl;
+    {
+        std::optional<std::uint32_t> latest_value{0U};
+        ProxyEventReceiver proxy_event_receiver{
+            original_proxy.moved_event_,
+            MakeSampleSequenceCallback(latest_value, "proxy_event_move_semantics consumer failed:")};
+        ProxyEventStateChangeNotifier proxy_state_change_notifier{original_proxy.moved_event_};
```
bemerybmw: We want to make sure that the set receive handler is not affected when moving. so this should not be destroyed and we shouldn't create a new one.

score/mw/com/test/move_semantics/proxy_event/provider.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+void RunProvider(const std::size_t num_samples_to_send,
+                 const std::size_t num_send_iterations,
+                 const score::cpp::stop_token& stop_token)
+{
+    const auto moved_to_name = filesystem::Path{kInstanceSpecifierMovedTo.ToString()}.Filename().Native();
+    auto process_synchronizer_result =
+        ProcessSynchronizer::CreateUniquePtr(kInterprocessNotificationShmPath + std::string{moved_to_name});
+
+    // Step 1. Create and offer the skeleton the consumer receives samples from. A second instance is also offered so
```
bemerybmw: You could have both proxy instances connect to the same skeleton?

bemerybmw (changes requested):

score/mw/com/test/common_test_resources/proxy_event_consumer.h
sahithi-nukala: correct, I'm moving this MakeSampleSequenceCallback to consumer.cpp under proxy_event

score/mw/com/test/common_test_resources/proxy_event_consumer.h
sahithi-nukala: in all 4 scenarios ReceiveAndNotify is always called with num_iterations=1 so the loop only ever runs once per call  WaitForStateChange is not actually being called repeatedly across iterations. the call is needed because ReceiveAndNotify is also invoked after ResubscribeAcrossReoffer, where the consumer has re-subscribed but the provider hasn't re-offered yet, so the state is still kSubscriptionPending and we need to wait for kSubscribed before receiving samples. For the earlier calls where the service is already offered, WaitForStateChange returns immediately since it checks the current state first

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: done

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: yes updated to SendIncrementingSequenceOfSamples

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: The second skeleton (MovedFrom) is necessary it gives the placeholder proxy its own shm so the MovedTo lifecycle is unaffected

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: SendIncrementingSequenceOfSamples is used in both proxy_event/provider.cpp and skeleton_event/provider.cpp with identical logic, only the skeleton type differs. To avoid the duplication, I've extracted it as a header-only template function into common_test_resources/send_incrementing_sequence_of_samples.h

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: good point. no, we don't need to append moved_to_name here.
i'll update it accordingly.

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: The proxy_state_change_notifier object is neither destroyed nor recreated after the move. Reattach() only rebinds the internal event reference so that subsequent WaitForStateChange() calls target the correct event; it does not call SetSubscriptionStateChangeHandler() again.

The original handler, registered in the constructor on original_proxy.moved_event_, continues to fire for the remainder of the test. This verifies that the handler survives the proxy move intact.

score/mw/com/test/move_semantics/proxy_event/provider.cpp
```diff
@@ -0,0 +1,106 @@
...
+const std::string kInterprocessNotificationShmPath{"/proxy_event_move_semantics_interprocess_notification"};
+const std::string kProviderWithdrawNotificationShmPath{"/proxy_event_move_semantics_provider_withdraw_notification"};
+
+}  // namespace
+
+void RunProvider(const std::size_t num_samples_to_send,
+                 const std::size_t num_send_iterations,
+                 const score::cpp::stop_token& stop_token)
+{
+    auto process_synchronizer_result = ProcessSynchronizer::CreateUniquePtr(kInterprocessNotificationShmPath);
```
bemerybmw: Why are you using CreateUniquePtr instead of Create? CreateUniquePtr was mainly added for when we need to put the ProcessSynchronizer in a vector.

score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
```diff
@@ -0,0 +1,48 @@
...
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+#include <iostream>
+#include <thread>
+
+namespace score::mw::com::test
+{
+
+/// Sends an incrementing sequence of \p number_of_samples_to_send_per_offer samples via
```
bemerybmw: Why have you added \p?

score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
```diff
@@ -0,0 +1,48 @@
...
+/// \p skeleton.moved_event_, starting at \p initial_value.
+template <typename SkeletonT>
```
bemerybmw:
```suggestion
template <typename Skeleton>
```

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,539 @@
...
+    // not be re-created afterwards, so that we verify they are preserved on the moved-to proxy.
+    std::cout << "\nConsumer: Step 3 - Register handlers on original proxy" << std::endl;
+    std::optional<std::uint32_t> latest_value{0U};
+    ProxyEventReceiver proxy_event_receiver{
+        original_proxy.moved_event_,
+        MakeSampleSequenceCallback(latest_value, "proxy_event_move_semantics consumer failed:")};
+    ProxyEventStateChangeNotifier proxy_state_change_notifier{original_proxy.moved_event_};
+    // Notify provider that subscription and handler registration are complete so it can start sending.
+    // Without this the provider might send the entire first batch before Subscribe() is called; LoLa SHM
+    // does not deliver pre-subscription samples, so GetNewSamples would return 0 forever.
```
bemerybmw: I don't think this is correct. The ProxyEvent should be able to access event samples sent before it subscribes. You just have to make sure that the number of samples sent is less than or equal to the number of configured sample slots in the configuration. So you shouldn't need this notify call here.

score/mw/com/test/move_semantics/proxy_event/provider.cpp
bemerybmw: How would the lifecycle of the MovedTo proxy be affected?

score/mw/com/test/move_semantics/proxy_event/provider.cpp
```diff
@@ -0,0 +1,106 @@
...
+    // final batch the offer is withdrawn and re-offered so the consumer can unsubscribe and re-subscribe on a
+    // withdrawn service (the consumer notifies once it has re-subscribed).
+    for (std::size_t iteration = 0U; iteration < num_send_iterations; ++iteration)
+    {
+        const bool is_final_iteration = (iteration + 1U == num_send_iterations);
+        if (is_final_iteration)
+        {
+            // Stop offering and immediately re-offer before notifying the consumer. This ensures the consumer
+            // always subscribes to an already-offered service and never enters kSubscriptionPending. Entering
+            // kSubscriptionPending and relying on LoLa's auto-reconnect is unreliable under concurrent load: the
```
bemerybmw: If you think that auto reconnect is "unreliable", then please be specific about the issue. I guess what you're referring to is that if the service sends data before the proxy autoreconnects, then it would miss the event notification. But aren't you waiting for subscribed state in the consumer before calling WaitForSamples?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,539 @@
...
+        FailTest("proxy_event_move_semantics consumer failed: Subscribe failed: ", subscribe_result.error());
+    }
+    // Notify provider that subscription and handler registration are complete so it can start sending.
+    // Without this the provider might send the entire first batch before Subscribe() is called; LoLa SHM
+    // does not deliver pre-subscription samples, so GetNewSamples would return 0 forever.
+    process_synchronizer.Notify();
+
+    // Step 6. Receive batches of samples and notify the provider after each one
+    std::cout << "\nConsumer: Step 6 - Receive first batches of samples" << std::endl;
+    for (std::size_t iteration = 0U; iteration < num_send_iterations - 1U; ++iteration)
```
bemerybmw: I don't think that we need to iterate here, it just adds complexity and additional sync points. We should just wait for subscribed state, wait for the expected samples and then notify the provider that we're done. The provider can then stop offer and reoffer the service and then send the next batch of samples. We should check in our sub state change handler that we transitioned to subscription pending and back to subscribed (you will probably need a custom state change handler to check this). We can then unsubscribe and resubscribe and then wait for the final samples.

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,539 @@
...
+    // Step 2. Move construct proxy before subscribe
+    std::cout << "\nConsumer: Step 2 - Move construct proxy before subscribe" << std::endl;
+    auto moved_proxy = std::move(original_proxy);
+
+    // Step 3. Register receive handler
+    std::cout << "\nConsumer: Step 3 - Register receive handler" << std::endl;
+    std::optional<std::uint32_t> latest_value{0U};
+    ProxyEventReceiver proxy_event_receiver{
+        moved_proxy.moved_event_,
+        MakeSampleSequenceCallback(latest_value, "proxy_event_move_semantics consumer failed:")};
```
bemerybmw: The whole point of adding the callback to the call to WaitForSamples is that you don't need to reuse the same callback and call  latest_value.reset(); (which introduces a race condition). You should use the overload which allows you to provide the exact samples that should be received (https://github.com/eclipse-score/communication/blob/2df18a1b312d02cbba177b601f4f04a9c4e95fac/score/mw/com/test/fields/set_and_get_and_notifier/consumer.cpp#L81)

score/mw/com/test/common_test_resources/proxy_event_receiver.h
```diff
@@ -21,41 +21,81 @@
...
         };
-        const auto result = proxy_event_or_field_.SetReceiveHandler(receive_handler);
+        const auto result = proxy_event_or_field_.get().SetReceiveHandler(receive_handler);
         if (!result.has_value())
         {
             std::cerr << "ProxyEventReceiver: Failed to set up receive handler!" << std::endl;
         }
     }
 
+    explicit ProxyEventReceiver(ProxyOrFieldType& proxy_event_or_field, StoredCallback callback)
```
bemerybmw: I don't think we should make thes changes to try to support the use case of these new tests in ProxyEventReceiver / ProxyEventStateChangeNotifier. The ProxyEventReceiver / ProxyEventStateChangeNotifier are intended as helpers for the simple use case and what we want to do in this test is very specific and falls outside the scope of that class. This class takes a reference to a proxe event or field. If this reference is invalidated by the caller e.g. because the proxy was moved, this is outside the scope of this class. We should instead write the handlers manually in the tests.

score/mw/com/test/common_test_resources/proxy_event_receiver.h
```diff
@@ -146,9 +186,16 @@ class ProxyEventReceiver
 
   private:
     score::concurrency::Notification received_sample_notification_;
-    ProxyOrFieldType& proxy_event_or_field_;
+    std::reference_wrapper<ProxyOrFieldType> proxy_event_or_field_;
+    StoredCallback stored_callback_;
 };
 
+template <typename ProxyOrFieldType>
```
bemerybmw: Why is this added?

score/mw/com/test/common_test_resources/proxy_event_receiver.h
bemerybmw: The goal of this class is to register a receive handler on construction and unregister it on destruction because we want to handle the notification of events in this class. The GetNewSamples callback is provided every time you call WaitForSamples.

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: done

score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
sahithi-nukala: removed

score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
sahithi-nukala: done

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: done

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: removed the  process_synchronizer.Notify();

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: agreed. I removed the broad auto-reconnect reliability claim

score/mw/com/test/common_test_resources/proxy_event_receiver.h
sahithi-nukala: Agreed. I simplified ProxyEventReceiver back to its intended purpose. It now only handles registering and unregistering receive handlers, while the sample-processing callbacks are passed directly to each WaitForSamples call.

score/mw/com/test/common_test_resources/proxy_event_receiver.h
sahithi-nukala: it’s there so this helper can be reused for both proxy events and proxy fields without duplicating code

score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
```diff
@@ -0,0 +1,48 @@
...
+#include <iostream>
+#include <thread>
+
+namespace score::mw::com::test
+{
+
+/// Sends an incrementing sequence of number_of_samples_to_send_per_offer samples via
+/// skeleton.moved_event_, starting at initial_value.
+template <typename Skeleton>
+void SendIncrementingSequenceOfSamples(Skeleton& skeleton,
```
bemerybmw: This should take the event rather than the skeleton. As it is now, this will only work for an event called moved_event_

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,526 @@
...
+        if (*sample != expected_value)
+        {
+            FailTest("proxy_event_move_semantics consumer failed: received value ",
+                     *sample,
+                     " does not match expected value ",
+                     expected_value);
+        }
+        last_received_value = *sample;
+    };
+    if (!proxy_event_receiver.WaitForSamples(stop_token, num_samples_to_receive, get_new_samples_callback))
```
bemerybmw: As mentioned here https://github.com/eclipse-score/communication/pull/607#discussion_r3861263537, wouldn't it be much simpler to use the other overload of WaitForSamples:
if (!proxy_event_receiver.WaitForSamples(stop_token, {1, 2, 3...}))

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,526 @@
...
+    // Step 8. Receive the remaining batch after re-subscribing. LoLa replays the last buffered
+    // batch (sent before stop-offer) upon re-subscription, so we expect those samples.
+    std::cout << "\nConsumer: Step 8 - Receive samples again after re-subscribing" << std::endl;
+    if (!proxy_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    }
+    // Reset last_received_value to the base of the expected batch so the callback checks the right range.
+    last_received_value = static_cast<std::uint32_t>((num_send_iterations - 2U) * num_samples_to_receive);
+    if (!proxy_event_receiver.WaitForSamples(stop_token, num_samples_to_receive, get_new_samples_callback))
```
bemerybmw: Same as above, use the other overload of WaitForSamples

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,526 @@
...
+        }
+        last_received_value = *sample;
+    };
+    if (!proxy_event_receiver.WaitForSamples(stop_token, num_samples_to_receive, get_new_samples_callback))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForSamples was interrupted by stop_token");
+    }
+    process_synchronizer.Notify();
+
+    // Step 7. Unsubscribe and subscribe again across the provider's re-offer
```
bemerybmw: What does "across the provider's re-offer" mean?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,526 @@
...
+        FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    }
+    provider_withdraw_synchronizer.Reset();
+    moved_proxy.moved_event_.Unsubscribe();
+    const auto resubscribe_result = moved_proxy.moved_event_.Subscribe(num_samples_to_receive);
+    if (!resubscribe_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: Re-subscribe failed: ", resubscribe_result.error());
+    }
+    // Tell the provider we have re-subscribed so that it can re-offer the service.
```
bemerybmw: Isn't the service already offered?

score/mw/com/test/move_semantics/proxy_event/provider.cpp
```diff
@@ -0,0 +1,92 @@
...
+    moved_from_skeleton_container.CreateSkeleton(kInstanceSpecifierMovedFrom, "proxy_event_move_semantics");
+    moved_from_skeleton_container.OfferService("proxy_event_move_semantics");
+
+    // Step 2. Send one batch of samples per iteration and wait for the consumer to acknowledge each one. Before the
+    // final batch the offer is withdrawn and re-offered so the consumer can unsubscribe and re-subscribe on a
+    // withdrawn service (the consumer notifies once it has re-subscribed).
+    for (std::size_t iteration = 0U; iteration < num_send_iterations; ++iteration)
+    {
+        const bool is_final_iteration = (iteration + 1U == num_send_iterations);
+        if (is_final_iteration)
```
bemerybmw: Let's separate the num_send_iterations from this offering / reoffering. num_send_iterations should be the number of times that the provider sends samples and waits on process_synchronizer_result. Then after that is all done, we can reoffer the service and send samples which has a different process synchronizer.

bemerybmw (changes requested):

score/mw/com/test/common_test_resources/send_incrementing_sequence_of_samples.h
sahithi-nukala: done

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: used the shared MakeExpectedSampleBatch helper to minimise the overload

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: no its already offered updated the comment

score/mw/com/test/move_semantics/proxy_event/provider.cpp
sahithi-nukala: Added a new, separate synchronizer for the reoffering step

score/mw/com/test/move_semantics/proxy_event/main_consumer.cpp
```diff
@@ -0,0 +1,46 @@
...
+#include "score/mw/com/test/common_test_resources/stop_token_sig_term_handler.h"
+#include "score/mw/com/test/move_semantics/proxy_event/consumer.h"
+#include "score/mw/com/test/move_semantics/proxy_event/test_parameters.h"
+
+int main(int argc, const char** argv)
+{
+    auto test_configuration{score::mw::com::test::ReadCommandLineArguments(argc, argv)};
+
+    score::mw::com::test::SetupAssertHandler();
+    score::mw::com::runtime::InitializeRuntime(argc, argv);
```
crimson11: This is deprecated! Please use the new API instaed.

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+
+namespace score::mw::com::test
+{
+namespace
+{
+
+const std::string kInterprocessNotificationShmPath{"/proxy_event_move_semantics_interprocess_notification"};
```
crimson11: This name is completely generic/unspecific, while the other two below explicitly specify for what this Notification is used for ... thus, you should rename it...

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+    if (!proxy_event_receiver.WaitForSamples(stop_token, resubscribe_batch_expected_samples))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForSamples was interrupted by stop_token");
+    }
+    process_synchronizer.Notify();
+
+    std::cout << "Consumer: Done with all iterations, exiting" << std::endl;
+}
+
+#if 0
```
crimson11: Should this stay? Why this "exclusion"? Since it is "commented out" I did not check/review this code.

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+    {
+        FailTest("proxy_event_move_semantics consumer failed: could not create provider withdrawal synchronizer");
+    }
+    auto resubscribe_synchronizer_result = ProcessSynchronizer::Create(kConsumerResubscribeNotificationShmPath);
+    if (!resubscribe_synchronizer_result.has_value())
+    {
+        FailTest("proxy_event_move_semantics consumer failed: could not create resubscribe synchronizer");
+    }
+
+    ExitFunctionGuard done_guard{[&process_synchronizer_result]() {
```
crimson11, line 257: So this is an RAII/clean-up? Doesn't this then need to be shifted to line 469?
I.e. - if you exit/fail in line 472 or line 477 your "clean-up" won't be done? It is also much more clear imho if you shift this RAII-guard directyl after process_synchronizer_result creation/verification.

score/mw/com/test/move_semantics/proxy_event/proxy_event_move_semantics_interface.h
```diff
@@ -0,0 +1,37 @@
+/********************************************************************************
```
crimson11, line 1: why is this file named `test_event_data_type.*` ... and not `proxy_move_semantics_interface` ?

score/mw/com/test/common_test_resources/proxy_container.h
```diff
@@ -39,11 +40,13 @@ class ProxyContainer
         return *proxy_;
     }
 
-    Proxy&& Extract()
+    Proxy Extract()
```
crimson11: Hm. OK. You want to re-use our glorious :D `ProxyContainer` in your tests and aren't using the existing `GetProxy` in your tests, because it just returns a reference and the issue is, that in your "move-tests" you would then do a move on this reference which basically means, that you leave `ProxyContainer` (the contained unique_ptr in an invalid state...?
It is just a bit "confusing", that `Extract()` returns a `Proxy` byValue ... although a Proxy isn't copyable ....
So - you are completely relying on `RVO`? I'm not the 100% C++ expert ;) - so question:
Is the sequence, where the move is done in line 47 and then you return the moved `Proxy` (which is an `Proxy&&`) as a `Proxy` byVal assured to be always done as RVO by the compiler?

If this is indeed the case: Please document this `Extract` method accordingly! Explain, what it does, that afterwards the `ProxyContainer` is in an invalid state (i.e. you can't call `GetProxy`anymore) and that the implementation relies on RVO happening (this is just for me dumb guy ;))

And a hint: Since you leave the ProxyContainer in an invalid state after `Extract` ... would it be then be a good idea to declare the `Extract`method as:
`Proxy Extract() &&` ?
I.e. stating, that `Extract` would be just viable on a "temporary". So in code you would then access it as:
`auto proxy = std::move(proxy_container).Extract()`

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+                                                  const std::size_t num_samples_to_receive,
+                                                  const std::size_t num_send_iterations,
+                                                  const score::cpp::stop_token& stop_token)
+{
+    ProxyContainer<ProxyMoveSemanticsProxy> proxy_container{};
+
+    // Step 1. Find service and create proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+    auto original_proxy = proxy_container.Extract();
```
crimson11: See all my blabla above!
Within your `Extract` method implementation you will always have alread a move-construction!
It happens with this line:
`auto proxy = std::move(*proxy_);` within the `Extract` implementation.
If the compiler does `(N)RVO` optimization, there is just no 2nd move-construction triggered by the `return` ... as the proxy is already move-constructed at the caller side!
If the compiler does NO `(N)RVO` optimization, you will have TWO move-constructions for proxy! One within `Extract` and the other from the return!

Why am I coming up with all this? Because your lines/comments line 63 and 67 are imho misleading! In line 63 you already get a MOVED proxy! This will never be the original proxy from within the `ProxyContainer` ... you only get it out from the container with a move-construction via the `Extract` call!

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+    // Step 2. Move construct proxy before subscribe
+    std::cout << "\nConsumer: Step 2 - Move construct proxy before subscribe" << std::endl;
+    auto moved_proxy = std::move(original_proxy);
```
crimson11: i.e. following my comments above ... this most-likely will become:
`auto moved_proxy = proxy_container.Extract()`  ... or (if you follow my other hint):
`auto moved_proxy = std::move(proxy_container).Extract()`

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+
+    // Step 3. Register receive handler
+    std::cout << "\nConsumer: Step 3 - Register receive handler" << std::endl;
+    ProxyEventReceiver proxy_event_receiver{moved_proxy.moved_event_};
+
+    // Step 4. Register state change handler
+    std::cout << "\nConsumer: Step 4 - Register state change handler" << std::endl;
+    ProxyEventStateChangeNotifier proxy_state_change_notifier{moved_proxy.moved_event_};
```
crimson11: `proxy_state_change_notifier` -> `proxy_event_state_change_notifier`
or
`subscription_state_change_notifier` ...

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
crimson11: @sahithi-nukala  - you didn't answer Bendans question?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+    }
+    const auto first_batch_expected_samples = MakeExpectedSampleBatch(0U, num_samples_to_receive);
+    if (!proxy_event_receiver.WaitForSamples(stop_token, first_batch_expected_samples))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForSamples was interrupted by stop_token");
+    }
+    process_synchronizer.Notify();
+
+    // Step 7. Unsubscribe and subscribe again across the provider's re-offer
+    std::cout << "\nConsumer: Step 7 - Unsubscribe and subscribe again" << std::endl;
```
crimson11: Where does the `Unsubscribe`happen you mention here?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+    std::cout << "\nConsumer: Waiting for provider to withdraw its offer" << std::endl;
+    if (!provider_withdraw_synchronizer.WaitWithAbort(stop_token))
+    {
+        FailTest(
+            "proxy_event_move_semantics consumer failed: Waiting for provider withdrawal was interrupted by "
+            "stop_token");
+    }
+    if (!proxy_state_change_notifier.WaitForStateChange(stop_token, SubscriptionState::kSubscribed))
```
crimson11: I didn't see any `Unsubscribe` up to this point. For me it seems like you are basically expecting/relying on "auto-reconnect"!?
If I'm correct, why then the activity in line 100?
You could have waited for a subscription state change to SubscriptionState::kPending ... and then wait for kSubscribed, what you are already doing in line 106?

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,540 @@
...
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForStateChange was interrupted by stop_token");
+    }
+    provider_withdraw_synchronizer.Reset();
+    moved_proxy.moved_event_.Unsubscribe();
+    const auto resubscribe_result = moved_proxy.moved_event_.Subscribe(num_samples_to_receive);
```
crimson11: TBH: Eventually I do not completely get, what is the testing approach.
Afais - the idea was:
There are different points in times, where you move-construct a proxy (in this case) and afterwards you check, whether ALL functionalities a proxy (and its events/fields) provides still work?
From the existing code I understand, that the different points in time are, what the enum `ProxyMoveScenario` expresses. Fine!
But then I would expect the test structured as follows:
You have one (reusable through ALL scenarios) method/function, which checks all functionalities of a given proxy, like you seem to do also here in this specific scenarion. So please refactor the code and implement ONE function `CheckIfProxyIsFullyFunctional` ... this tests/checks alle the APIs/functionalities! There are some easy/obvious functionalities (subscribe/unsubscribe/GetNewSamples/...) and some functionalities, which are NOT so obvious (e.g. "proxy-auto-reconnect")

Then your methods which test the various scenarios become "simple":
You do the "move" from in different situations (this code is small/simple for each different scenarion) and then you hand over byRef the moved-proxy to the `CheckIfProxyIsFullyFunctional` func to check it.

-

crimson11 (changes requested):

score/mw/com/test/move_semantics/proxy_event/main_consumer.cpp
sahithi-nukala: done

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: As per the discussion with @bemerybmw i just made the updates/changes only to RunConsumerMoveConstructProxyBeforeSubscribe scenario. once it looks good then will update the remaining scenarios accordingly.

score/mw/com/test/move_semantics/proxy_event/main_consumer.cpp
sahithi-nukala: done

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: Updated to kConsumerBatchReceivedNotificationShmPath

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: As per the discussion with @bemerybmw currently i have updated the chnages only for RunConsumerMoveConstructProxyBeforeSubscribe scenario. If its fine then will update the remaining scenarios accordingly.

score/mw/com/test/move_semantics/proxy_event/proxy_event_move_semantics_interface.h
sahithi-nukala: updated to proxy_event_move_semantics_interface

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
sahithi-nukala: updated to proxy_event_state_change_notifier

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,303 @@
...
+    std::cout << "\nConsumer: Step 11 - Wait for the new samples" << std::endl;
+    std::cout << "\nConsumer: expected second batch of samples: " << kSamplesToSendAfterReOffer.size() << std::endl;
+    if (!proxy_event_receiver.WaitForSamples(stop_token, kSamplesToSendAfterReOffer))
+    {
+        FailTest("proxy_event_move_semantics consumer failed: WaitForSamples was interrupted by stop_token");
+    }
+    std::cout << "\nConsumer: Received the second batch of samples successfully" << std::endl;
+}
+
+void RunConsumerMoveConstructProxyBeforeSubscribe(ProcessSynchronizer& process_synchronizer,
```
crimson11, line 113: OK. I didn't find the single-location, where we list all the test-scenarios we need? Currently I can try to find out from the `consumer.cpp` file ... by looking at the function names and the code ... but as I said: we need a documentation 1st, so that we know, what to do!

Currently you only have some function names in this .cpp and some *.py file-names, which give some "hint", what might be tested ... but imho this is just weak!
I don't care WHERE you document! Interim you can write it down in some .md file ... and once you have implemented everything, it is also fine for me to remove the .md file and instead write a BIG doxygen comment above each test-case/method in the .cpp ...


So - now - what do we think should be tested in the case of Proxy move-functionality?
What comes to my mind:

1. Check, that moved proxy instance is fully functional after an initial move directly after proxy construction. (The most basic test)
2. Check, that the moved proxy instance is fully functional when the move happened AFTER the original proxy had already done certain activities:
   - subscribed to an event
   - unsubscribed from an event
   - received samples
   - registered an event receive-handler
   - called a method
   - registered a subscription-state-change handler
3. Check, that proxy-auto-reconnect also works on a moved-proxy (move BEFORE auto-reconnect):
   - you create the proxy
   - you move the proxy
   - you stop offer the server
   - you re-offer the server
   - you wait till the moved proxy gets auto-reconnected.
   - you verify, that the proxy is still fully functional.
4. Check, that the moved proxy instance is fully functional when the move happened AFTER a proxy-auto-reconnect had happened. I.e. you create a proxy, then you make the skeleton StopOffer, then you wait for the proxy to be disconnected, then you make the skeleton Offer again, then you wait till the proxy is reconnected and then you do the move and then verify, that the proxy is still fully functional.

Good, that you already introduced `CheckIfProxyIsFullyFunctional` as your re-usable function block!
Now, if you look at (2) above, you could say - wait - these are basically 6 or more tests, because I could do the "move" at all the differnet steps (subscribe/unsubscribe/receive sample/call a method) ... but it is overdone to do all combinations! So - for (2) I feel, it is enough, that after creation of the initial proxy (before the move) you do EVERYTHING/each API call you can think of with this proxy! So - similar to `CheckIfProxyIsFullyFunctional`! I.e. you can maybe re-use it. And then after you have called everything possible, you do the move and then call `CheckIfProxyIsFullyFunctional` on the move-constructed instance.

If you are done with this, you could do the same pattern for `MoveAssign` - maybe you can handle this with some parametrized test pattern ... at least try to minimize code duplication as much as possible!

score/mw/com/test/move_semantics/proxy_event/consumer.cpp
```diff
@@ -0,0 +1,303 @@
...
+    // Step 1. Find service and create original proxy
+    std::cout << "\nConsumer: Step 1 - Find service and create proxy" << std::endl;
+    proxy_container.CreateProxy(kInstanceSpecifierMovedTo, "proxy_event_move_semantics");
+
+    // Step 2. Move construct proxy before subscribe
+    std::cout << "\nConsumer: Step 2 - Move construct proxy before subscribe" << std::endl;
+    auto original_proxy = std::move(proxy_container).Extract();
+
+    // Step 5. Subscribe
+    std::cout << "\nConsumer: Step 5 - Subscribe" << std::endl;
```
crimson11, line 129: I guess you have an issue with counting ;) - I see always 5. coming after 2. in your tests ...

crimson11 (changes requested):

PR #614, opened 2026-06-29
Mark as nodiscard the public API from LoLa that can return an error

ahmed0mousa:


castler (changes requested):
This will break our API. We need to discuss if we do it and how we handle it.

castler:
As discussed, we have no intention right now to go this way, as it will be quite some breaking change.

PR #615, opened 2026-06-30
Config validation extraction

Thomas-Mikhael:
+ Extract the common config validation checks into a common file/lib config_validate (preparing for use in both parsing strategies, JSON & Flatbuffer).
+ Unify the checking of duplicates while emplacing, EmplaceOrFatal()
+ Remove redundant checks of duplicate Events and Fields (CheckContainsEvent, CheckContainsField)
+ Change ParseLolaxxx functions to return void instead of ignoring the return


score/mw/com/impl/configuration/config_validate.h
```diff
@@ -0,0 +1,39 @@
...
+#include <score/assert.hpp>
+
+#include <string_view>
+#include <utility>
+
+namespace score::mw::com::impl::configuration
+{
+
+template <typename Map, typename Key, typename Value>
+void EmplaceOrFatal(Map& map, Key&& key, Value&& value, std::string_view element_description)
```
castler, line 36: This function is local to `config_parser.cpp` thus, it should be within an anonymous namespace within the config parser.

This statement is also true for all other extractions that are done in later commits.

score/mw/com/impl/configuration/config_parser.cpp
```diff
@@ -478,8 +452,6 @@ auto ParseLolaFieldInstanceDeployment(const score::json::Object& json_map, LolaS
         ServiceElementInstanceDeploymentParser deployment_parser{field_object};
 
         const auto& field_name_it = field_object.find(kFieldNameKey);
-        deployment_parser.CheckContainsField(field_name_it, service);
```
castler, line 483: I do not fully understand why we remove this check

score/mw/com/impl/configuration/config_parser.cpp
```diff
@@ -422,8 +401,6 @@ auto ParseLolaEventInstanceDeployment(const score::json::Object& json_map, LolaS
         ServiceElementInstanceDeploymentParser deployment_parser{event_object};
 
         const auto& event_name_it = event_object.find(kEventNameKey.data());
-        deployment_parser.CheckContainsEvent(event_name_it, service);
```
castler, line 427: I also do not understand fully, why we remove this check?

score/mw/com/impl/configuration/config_parser_test.cpp
```diff
@@ -402,6 +402,38 @@ TEST_F(ConfigParserFixture, NoEventsOrFieldsWillCauseTermination)
...
+      ]
+    }
+  ],
+  "serviceInstances": []
+}
+)"_json;
+
+    // When parsing the JSON
+    // That the application will terminate
+    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(score::mw::com::impl::configuration::Parse(std::move(j2)));
```
castler: Its not a bug - its a feature :)

There have been use cases in the past, where an empty service was used. We see no reason why an empty service (e.g. as marker via the service discovery) shall not be allowed.

Can you maybe elaborate a bit, why you think this is a bug?

castler (changes requested):
Thank you for the contribution!

I think there are two main points that contradict our current style:

1st)
We try to keep functions as local as possible - as the config checks are just used in the config parser, we should have them there as anonymous functions.

2nd)
We strictly want to allow empty services.

On a side-note it would be nice trying to avoid merge commits in Pull-Requests.

Thank you!

score/mw/com/impl/configuration/config_validate.h
Thomas-Mikhael: This function will mostly be also used for the flatbuffer parser, that's why I put it in the common config_validate.h

If you suggest adding these files directly to the 'configuration_json_parsing_strategy' lib directly instead of making it as a separate common lib 'config_validate' that to be used later with 'configuration_flatbuffer_parsing_strategy' I would see it okay.

Or did I missunderstood something?

score/mw/com/impl/configuration/config_parser.cpp
Thomas-Mikhael: Because it's redundant check.
CheckContainsField() and CheckContainsEvent() are using misleading naming, it is just ensuring that there's no duplicates, which would never happen using .emplace()

score/mw/com/impl/configuration/config_parser.cpp
Thomas-Mikhael: Because it's redundant check.
CheckContainsField() and CheckContainsEvent() are using misleading naming, it is just ensuring that there's no duplicates, which would never happen using .emplace()

score/mw/com/impl/configuration/config_parser_test.cpp
Thomas-Mikhael: In the [original code](https://github.com/etas-contrib/score_communication/blob/397a89307caf47e9cdc79097d6befcb8355b3b24/score/mw/com/impl/configuration/config_parser.cpp#L1011) you can see the in config_parser.cpp the error message description as following:
`score::mw::log::LogFatal("lola") << "Configuration should contain at least one event, field, or method.";`
That's why I considered it as an unintended bug.

Thomas-Mikhael: If we consider it as a feature then we shall remove the check at all. What do you suggest?

score/mw/com/impl/configuration/config_validate.h
castler: No you can keep it then in a separate library. Maybe a explanation comment would be good, that explains why this is in a custom file and what the purpose if it would be.

score/mw/com/impl/configuration/config_parser_test.cpp
castler: > If we consider it as a feature then we shall remove the check at all. What do you suggest?

That would be great! Then things should be more consistent!

score/mw/com/impl/configuration/config_validate.h
Thomas-Mikhael: I have that already in the description of the PR:
_Extract the common config validation checks into a common file/lib config_validate (preparing for use in both parsing strategies, JSON & Flatbuffer)._
Or you mean to comment it in the Bazel BUILD file?

score/mw/com/impl/configuration/config_parser_test.cpp
Thomas-Mikhael: Removed and reverted the Tests changes

score/mw/com/impl/configuration/config_validate.h
Thomas-Mikhael: I added some comments

castler (commented):
Thanks the changes look good.

The only thing I would ask you to clean up the git history, right now its quite some mess. I am fine with having few proper commits (with proper messages).

In addition - but this could be done in a follow up if you want - we have the best practice that for each library we have our custom tests.

Since you now extracted these functions into `config_validate` they are most certainly covered via the existing test. But we would now normally rather create a custom `config_validate_test.cpp` to ensure that only this unit is correctly covered.

Thomas-Mikhael:
@castler I agree.
Added config_validate_test.cpp
Fixed up the commits into 2 commits

castler (approved):

PR #617, opened 2026-06-30
feat(itf): add dual-QEMU ivshmem integration-test environment

Rahul-Sutariya:
Add a `dual_qemu` ITF/pytest plugin that boots two QNX VMs sharing one QEMU ivshmem-plain region, extending the upstream single-VM `qemu` plugin.

- plugin: config.py, ivshmem_qemu.py, dual_qemu_process.py, __init__.py (target_a/target_b fixtures; sequential boot to avoid KVM race), BUILD, example config, README, example/run_example_manually.sh.
- integration_testing.bzl: dual_qemu_integration_test macro (QNX-only, flaky).

### DIFF
--- a/quality/integration_testing/environments/dual_qemu/BUILD
+++ b/quality/integration_testing/environments/dual_qemu/BUILD
@@ -1,0 +1,42 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+load("@rules_python//python:defs.bzl", "py_library")
+load("@score_itf//bazel:py_itf_plugin.bzl", "py_itf_plugin")
+
+exports_files(["dual_qemu_config.example.json"])
+
+py_library(
+    name = "dual_qemu_lib",
+    srcs = [
+        "__init__.py",
+        "config.py",
+        "dual_qemu_process.py",
+        "ivshmem_qemu.py",
+    ],
+    imports = [".."],
+    deps = [
+        "@score_itf//score/itf/core/process",
+        "@score_itf//score/itf/core/utils",
+        "@score_itf//score/itf/plugins/qemu",
+    ],
+)
+
+# ---- ITF plugin target (used by the py_itf_test symbolic macro) ----
+py_itf_plugin(
+    name = "dual_qemu_plugin",
+    enabled_plugins = [
+        "dual_qemu",
+    ],
+    py_library = ":dual_qemu_lib",
+    visibility = ["//:__subpackages__"],
+)
--- a/quality/integration_testing/environments/dual_qemu/README.md
+++ b/quality/integration_testing/environments/dual_qemu/README.md
@@ -1,0 +1,87 @@
+# `dual_qemu` ITF plugin
+
+A pytest/ITF plugin that boots **two** QNX QEMU VMs which share a single QEMU
+[`ivshmem`](https://www.qemu.org/docs/master/system/devices/ivshmem.html) region. It is a
+thin extension of the upstream single-VM `qemu` plugin
+(`@score_itf//score/itf/plugins/qemu`).
+
+## What it adds over the upstream `qemu` plugin
+
+| Aspect            | upstream `qemu`        | `dual_qemu`                                            |
+| ----------------- | ---------------------- | ----------------------------------------------------- |
+| Number of VMs     | 1 (`target_init`)      | 2 (`target_a`, `target_b`)                            |
+| Shared memory     | none                   | one `ivshmem-plain` region backed by one host file    |
+| SSH ports         | one                    | distinct host port per VM (e.g. `2222` / `2223`)      |
+| Inter-VM NIC      | n/a                    | optional point-to-point socket NIC (off by default)   |
+
+The shared region uses **`ivshmem-plain`** — a passive shared-memory PCI device with
+**no MSI-X and no doorbell**. Cross-VM notifications travel over a separate
+socket-based control plane, so the data plane only needs plain shared memory.
+
+## Fixtures
+
+| Fixture            | Scope   | Description                                              |
+| ------------------ | ------- | ------------------------------------------------------- |
+| `target_a`         | session | First VM (`QemuTarget`).                                 |
+| `target_b`         | session | Second VM (`QemuTarget`).                                |
+| `ivshmem_backend`  | session | Path of the shared host backing file.                   |
+| `config`           | session | Loaded `DualQemuConfigModel` + `qemu_image`.            |
+
+Both VMs run the upstream `pre_tests_phase` checks (ping / SSH / SFTP) before the tests
+start.
+
+### Boot reliability
+
+Booting two QNX guests **concurrently** under KVM occasionally wedges the second guest
+during device bring-up (it can hang at SMP AP / secondary-CPU startup under the `-kernel`
+boot path, so its `sshd` never comes up and QEMU's SLIRP resets the harness connections).
+To keep the test reliable the plugin:
+
+- boots the VMs **sequentially** — it starts a VM, waits until SSH is *stably* reachable
+  and runs `pre_tests_phase`, and only then starts the next one, so the two guests never
+  initialise their devices at the same time;
+- gives each VM a single core and a **distinct NIC MAC**;
+- the `dual_qemu_integration_test` macro additionally marks the test `flaky = True` so
+  bazel transparently retries the rare residual KVM boot hiccup.
+
+## CLI options
+
+| Option               | Required | Description                                        |
+| -------------------- | -------- | -------------------------------------------------- |
+| `--dual-qemu-config` | yes      | JSON config (see `dual_qemu_config.example.json`). |
+| `--qemu-image`       | yes      | Shared QEMU IFS image booted by both VMs.          |
+
+## Configuration
+
+See [`dual_qemu_config.example.json`](dual_qemu_config.example.json). A config is just
+**two ordinary QEMU configs** (each an upstream `QemuConfigModel`) plus an `ivshmem`
+block:
+
+```json
+{
+    "ivshmem": { "size": "4M" },
+    "intervm_network": { "enabled": false, "host_port": 12345 },
+    "vms": [ { "...": "QemuConfigModel for VM-A" },
+             { "...": "QemuConfigModel for VM-B" } ]
+}
+```
+
+- `ivshmem.size` — size of the shared region (e.g. `4M`).
+- `ivshmem.mem_path` (optional) — explicit host backing file; when empty a temporary file
+  is created for the session and removed on teardown.
+- `intervm_network.enabled` — when `true`, adds a socket NIC where VM-A listens and VM-B
+  connects on `host_port` (for the future socket control plane).
+- Each VM's `ssh_port` and `port_forwarding` must use **distinct host ports**.
+
+## Debugging guest boot
+
+Set `DUAL_QEMU_SERIAL_DIR` to capture each VM's serial console to
+`<dir>/vm<index>_serial.log` (instead of the normal `mon:stdio`). For example:
+
+```bash
+bazel test //path/to/your:dual_qemu_test \
+    --config=qnx_x86_64 --strategy=TestRunner=local \
+    --test_env=DUAL_QEMU_SERIAL_DIR=/tmp/dqdbg
+# then inspect /tmp/dqdbg/vm0_serial.log and /tmp/dqdbg/vm1_serial.log
+```
+
--- a/quality/integration_testing/environments/dual_qemu/README.md
+++ b/quality/integration_testing/environments/dual_qemu/README.md
@@ -73,6 +73,16 @@ block:
   connects on `host_port` (for the future socket control plane).
 - Each VM's `ssh_port` and `port_forwarding` must use **distinct host ports**.
 
+## Cross-VM lib-memory example
+
+[`example_xvm/`](example_xvm/README.md) drives the `score::memory::shared` allocator (the
+one LoLa uses) across both VMs over the ivshmem BAR. The one BAR is split into two
+page-aligned lib-memory regions, so the exchange is **bidirectional**: the producer creates
+a forward region holding a `Vector` (the consumer opens it), and the consumer creates a
+separate reverse region holding a `Map` (the producer opens it) — each container's
+`OffsetPtr`s resolve despite each VM mapping the BAR at a different base. See
+[`example_xvm/README.md`](example_xvm/README.md) for the details.
+
 ## Debugging guest boot
 
 Set `DUAL_QEMU_SERIAL_DIR` to capture each VM's serial console to
@@ -84,4 +94,3 @@ bazel test //path/to/your:dual_qemu_test \
     --test_env=DUAL_QEMU_SERIAL_DIR=/tmp/dqdbg
 # then inspect /tmp/dqdbg/vm0_serial.log and /tmp/dqdbg/vm1_serial.log
 ```
-
--- a/quality/integration_testing/environments/dual_qemu/__init__.py
+++ b/quality/integration_testing/environments/dual_qemu/__init__.py
@@ -1,0 +1,139 @@
+# *******************************************************************************
+# Copyright (c) 2026 Contributors to the Eclipse Foundation
+#
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Pytest plugin that boots **two** QNX QEMU VMs sharing a QEMU ``ivshmem`` region.
+
+It is a thin extension of the single-VM ``qemu`` plugin
+(``@score_itf//score/itf/plugins/qemu``): it reuses ``QemuTarget`` / ``pre_tests_phase``
+and only adds (a) a second VM, (b) an ``ivshmem-plain`` device backed by one shared host
+file, and (c) distinct host SSH ports per VM.
+
+Exposed session fixtures:
+    - ``target_a`` / ``target_b`` -- the two booted VMs (``QemuTarget``).
+    - ``ivshmem_backend``         -- path of the shared host backing file.
+"""
+import logging
+import socket
+
+import pytest
+
+from score.itf.core.utils.bunch import Bunch
+
+from .config import load_configuration, parse_size
+from .dual_qemu_process import DualQemuProcess
+
+logger = logging.getLogger(__name__)
+
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--dual-qemu-config",
+        action="store",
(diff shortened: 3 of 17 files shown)


quality/visibility_guard/public_targets.golden
```diff
@@ -19,6 +19,9 @@
 //quality/compiler_warnings:default_link_flags
 //quality/compiler_warnings:treat_warnings_as_errors
 //quality/coverage:enable_llvm_coverage_for_death_tests
+//quality/integration_testing/environments/dual_qemu/example_xvm:ivshmem_lib_memory_support
+//quality/integration_testing/environments/dual_qemu:dual_qemu_lib
+//quality/integration_testing/environments/dual_qemu:dual_qemu_plugin
```
castler: These should not be public :) This is only something we want to use within the communication repository for now.

quality/integration_testing/integration_testing.bzl
```diff
@@ -142,3 +142,78 @@ def integration_test(name, srcs, filesystem, **kwargs):
         env = {"DOCKER_HOST": ""},
         **kwargs
     )
+
+def dual_qemu_integration_test(
+        name,
+        srcs,
+        filesystem,
```
castler: Do we not need two filesystems here?

quality/integration_testing/integration_testing.bzl
```diff
@@ -142,3 +142,78 @@ def integration_test(name, srcs, filesystem, **kwargs):
...
+    # Two VMs require even more resources than a single one.
+    if "size" not in kwargs:
+        kwargs["size"] = "enormous"
+    if "timeout" not in kwargs:
+        kwargs["timeout"] = "moderate"
+
+    # Driving two real QNX guests under KVM has rare, environment-induced boot
+    # nondeterminism (e.g. a guest occasionally wedging during device bring-up).
+    # The fixtures already stagger boots and wait for stable SSH; mark the test
+    # flaky so bazel transparently retries such infrastructure hiccups.
```
castler, line 218: Ouch - is this possible to be fixed? Should we add a ticket with a todo?

quality/integration_testing/environments/dual_qemu/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+| File                              | Purpose                                                        |
+| --------------------------------- | -------------------------------------------------------------- |
+| `__init__.py`                     | `pytest_addoption` + fixtures (`target_a`/`target_b`/...).      |
+| `config.py`                       | `DualQemuConfigModel` (two `QemuConfigModel` + `ivshmem`).      |
+| `ivshmem_qemu.py`                 | QEMU launcher that appends the `ivshmem-plain` device.         |
+| `dual_qemu_process.py`            | Process wrapper around `IvshmemQemu`.                           |
+| `BUILD`                           | `py_library` + `py_itf_plugin` (`:dual_qemu_plugin`).          |
+| `dual_qemu_config.example.json`   | Sample two-VM configuration.                                   |
+| `example_xvm/`                    | Cross-VM `score::memory::shared` demo (see below).             |
```
castler: Let's remove this section. AI loves to add this - but it will get out of sync quickly.

quality/integration_testing/environments/dual_qemu/README.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Using it from a test
+
+Wire `@//quality/integration_testing/environments/dual_qemu:dual_qemu_plugin` as the
+plugin of a `py_itf_test`-based target (instead of the single-VM `qemu_plugin`) and pass
+`--dual-qemu-config` / `--qemu-image`. The easiest way is the
+`dual_qemu_integration_test` macro in
+`//quality/integration_testing:integration_testing.bzl`, which builds the shared IFS image
+and wires everything for you.
```
castler: This is what you have done already, we do not need this section.

quality/integration_testing/environments/dual_qemu/BUILD
```diff
@@ -0,0 +1,43 @@
...
+py_library(
+    name = "dual_qemu_lib",
+    srcs = [
+        "__init__.py",
+        "config.py",
+        "dual_qemu_process.py",
+        "ivshmem_qemu.py",
+    ],
+    imports = [".."],
+    visibility = ["//visibility:public"],
```
castler: I do not think it is required, that this is of public visibility.

quality/integration_testing/environments/dual_qemu/BUILD
```diff
@@ -0,0 +1,43 @@
...
+)
+
+# ---- ITF plugin target (used by the py_itf_test symbolic macro) ----
+py_itf_plugin(
+    name = "dual_qemu_plugin",
+    enabled_plugins = [
+        "dual_qemu",
+    ],
+    py_library = ":dual_qemu_lib",
+    visibility = ["//visibility:public"],
```
castler: This should also not be public visibility.

quality/integration_testing/environments/dual_qemu/__init__.py
```diff
@@ -0,0 +1,215 @@
...
+
+    logger.info(f"Creating ivshmem backing file {path} ({ivshmem.size})")
+    with open(path, "wb") as backing_file:
+        backing_file.truncate(size_bytes)
+
+    yield path
+
+
+@pytest.fixture(scope="session")
+def targets(config, ivshmem_backend):
```
castler: Can we mark this somehow private, that its clear that users should not call it?

quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py
```diff
@@ -0,0 +1,220 @@
...
+"""
+import logging
+import os
+import subprocess
+import sys
+
+logger = logging.getLogger(__name__)
+
+
+class IvshmemQemu:
```
castler: We now copied here a lot from: https://github.com/eclipse-score/itf/blob/main/score/itf/plugins/qemu/qemu.py can we not use some software mechanisms to reduce code duplication? Inheritance? Decorators?

quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py
```diff
@@ -0,0 +1,220 @@
...
+
+        self._subprocess = None
+
+    def __enter__(self):
+        return self.start()
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.stop()
+
+    def start(self, subprocess_params=None):
```
castler: should this not only start one instance? and then when we have another instance of this class, we will start the otherone?

quality/integration_testing/environments/dual_qemu/dual_qemu_process.py
```diff
@@ -0,0 +1,80 @@
...
+import subprocess
+
+from score.itf.core.process.console import PipeConsole
+
+from .ivshmem_qemu import IvshmemQemu
+
+logger = logging.getLogger(__name__)
+
+
+class DualQemuProcess:
```
castler: Should we not inherit from here: https://github.com/eclipse-score/itf/blob/main/score/itf/plugins/qemu/qemu_process.py#L22?

Where is our equivalent to https://github.com/eclipse-score/itf/blob/main/score/itf/plugins/qemu/qemu_target.py ?

quality/integration_testing/environments/dual_qemu/__init__.py
```diff
@@ -0,0 +1,215 @@
...
+from score.itf.plugins.qemu.checks import pre_tests_phase
+from score.itf.plugins.qemu.qemu_target import QemuTarget
+
+from .config import load_configuration, parse_size
+from .dual_qemu_process import DualQemuProcess
+
+logger = logging.getLogger(__name__)
+
+
+def _wait_for_ssh(target, total_timeout: int = 180, interval: int = 3, stable_successes: int = 3):
```
castler: These helper functions must be used already somewhere else in the other Qemu plugin, no?

quality/integration_testing/environments/dual_qemu/example_xvm/run_example_manually.sh
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/env bash
```
castler: This script should go away.

castler (commented):
Let's first align on the test infrastructure - then we can check out the example :)

quality/integration_testing/integration_testing.bzl
Rahul-Sutariya: We can have multiple binaries in one filesystem Ex.
```
pkg_application(
    name = "gateway_test_apps",
    app_name = "gateway_test",
    bin = [":app1", ":app2"],  # both binaries in one package
)
```
If we want separate, then i can extend the macro to accept two filesystem targets and build two IFS images

quality/integration_testing/integration_testing.bzl
Rahul-Sutariya: It's a KVM + QNX -kernel boot path timing issue — the second guest occasionally wedges during SMP AP startup when both initialize under the same KVM instance. The plugin already mitigates by booting VMs sequentially and using single-core guests, which makes it very rare but not fully eliminable from our side.

A proper fix would require either:

1. QNX fixing the boot-path race in their -kernel loader (upstream QNX issue), or
2. Using the -bios/UEFI boot path instead of -kernel (different QNX IFS packaging, not trivial).

quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py
Rahul-Sutariya: The Qemu class uses double-underscore name mangling (__build_qemu_command, __find_available_kvm_support, etc.), which Python mangles to _Qemu__build_qemu_command — making inheritance/override impossible from a subclass. There's no extra_args hook or any other extension point to inject additional -object/-device flags.

Rahul-Sutariya: we can rename __build_qemu_command to _build_qemu_command (drop the name mangling) and add an overridable _extra_qemu_args() hook that the base command assembly calls. (itf repo changes) then we can avoid duplication, want me creat pr for ITF repo?

quality/integration_testing/integration_testing.bzl
castler: Yes - I think we should have two filesystems to be more variable (e.g. imagine having different IP configurations one day or different boot ups or what ever)

quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py
Rahul-Sutariya: Yes, that's exactly how it works. Each IvshmemQemu instance starts one QEMU process. In the plugin's _targets fixture, the loop creates a separate instance per VM

quality/integration_testing/environments/dual_qemu/dual_qemu_process.py
Rahul-Sutariya: QemuProcess: Can't inherit — it hardcodes Qemu(...) in __init__ with no way to inject a different launcher. If upstream accepted the Qemu instance as a constructor parameter (DI), DualQemuProcess could be eliminated entirely.

QemuTarget: We already reuse it directly (from score.itf.plugins.qemu.qemu_target import QemuTarget). No custom equivalent needed — DualQemuProcess exposes the same .stop()/.restart()/.console interface that QemuTarget expects.

quality/integration_testing/environments/dual_qemu/__init__.py
Rahul-Sutariya: No — the single-VM plugin only does pre_tests_phase (5 SSH retries with fixed timeouts). It doesn't need stable-success counting or boot-restart logic because a single QNX guest under KVM boots reliably. These helpers address problems specific to the dual-VM scenario (sshd instability during concurrent boots, occasional total boot wedges, and the first VM going quiet while the second initializes). They're new logic rather than duplication from upstream.

quality/integration_testing/environments/dual_qemu/example_xvm/run_example_manually.sh
Rahul-Sutariya: Done, Thanks

quality/visibility_guard/public_targets.golden
Rahul-Sutariya: Chnaged visibility as per requirement, Thanks

quality/integration_testing/environments/dual_qemu/README.md
Rahul-Sutariya: Done, removed

quality/integration_testing/environments/dual_qemu/README.md
Rahul-Sutariya: Done, Removed

quality/integration_testing/environments/dual_qemu/BUILD
Rahul-Sutariya: yes, Changed visibility private, Thanks

quality/integration_testing/environments/dual_qemu/BUILD
Rahul-Sutariya: yes, Changed visibility to com, Thanks!

quality/integration_testing/environments/dual_qemu/__init__.py
Rahul-Sutariya: Done — made it prefixed with underscore (_targets). That's the standard Python convention and the only mechanism pytest supports for marking a fixture as internal. The public API for tests is target_a / target_b.

quality/integration_testing/environments/dual_qemu/__init__.py
```diff
@@ -0,0 +1,215 @@
...
+        # Boot the VMs one at a time: start a VM, wait until it is fully reachable, and
+        # only then start the next one. Booting two QNX guests *concurrently* under KVM
+        # occasionally leaves the second guest's device bring-up (SMP AP / virtio-net)
+        # wedged; staggering the boots makes each guest initialize in isolation, exactly
+        # like the single-VM tests, which is reliable. _boot_vm additionally restarts an
+        # individual guest in place on the rare boot where its sshd never serves.
+        makers = []
+        for index, vm in enumerate(dual_config.vms):
+            def make_process(index=index, vm=vm):
+                return DualQemuProcess(
```
castler: Why are we not using

 ```
with DualQemuProcess(...) as target_a: 
   with DualQemuProcess(...) as target_b:
      yield target_a, tarbet_b
```

This would make things IMHO way more clear, would avoid the complicated loop and would enable to give each QemuProcess also its custom configuration.

Comment (Rahul-Sutariya) [reply quality/integration_testing/environments/dual_qemu/__init__.py]: Adapted the changes, Thanks

File: quality/integration_testing/integration_testing.bzl (hunk 2 above)
Comment (Rahul-Sutariya) [reply quality/integration_testing/integration_testing.bzl]: Adapted the changes for second filesystem, THanks

File: quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py (hunk 9 above)
Comment (Rahul-Sutariya) [reply quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py]: S-Core/ITF pr: https://github.com/eclipse-score/itf/pull/122

Comment (Rahul-Sutariya) [reply quality/integration_testing/environments/dual_qemu/ivshmem_qemu.py]: Adapted the pr changes to this pr

File: quality/integration_testing/environments/dual_qemu/example_xvm/README.md [hunk 15]
```diff
@@ -0,0 +1,78 @@
...
+> Two cross-VM hurdles the example has to clear:
+> 1. **Separate shm namespaces** — the same name is a *different* object in each guest, so the
+>    reader must first bind its name to the region's BAR physical sub-range (`shm_ctl(SHMCTL_PHYS)`
+>    via a custom `TypedMemory` provider) *before* calling `Open()`.
+> 2. **Init ordering** — ivshmem-plain has no doorbell, so the two roles rendezvous by polling
+>    a small handshake header placed in the **last page** of the BAR, outside both lib-memory
+>    arenas. Start both roles roughly together; each waits for the other.
+>
+> Even with single-writer regions this **still** relies on the lock-free lib-memory patch
+> (`//third_party/score_baselibs:lib_memory_lockfree_alloc.patch`): the stock
```
castler: This does not exist?

quality/integration_testing/environments/dual_qemu/example_xvm/bar_typed_memory.cpp
```diff
@@ -0,0 +1,120 @@
...
+        (void)::shm_unlink(shm_name.c_str());
+        return score::cpp::make_unexpected(score::os::Error::createFromErrno(saved));
+    }
+
+    // The object persists (name + physical binding) until Unlink(); our fd is no longer
+    // needed because the factory re-opens by name.
+    (void)::close(fd);
+    return {};
+#else
+    (void)shm_name;
```
castler: also in demo code, we should try to be more clean. Use e.g. `std::ignore = ` in these cases.

quality/integration_testing/environments/dual_qemu/example_xvm/README.md
castler: But let's not add the patch - we own baselibs, so lets fix it directly.

quality/integration_testing/environments/dual_qemu/example_xvm/BUILD
```diff
@@ -0,0 +1,85 @@
...
+#     app_name = "lib_memory_ivshmem_xvm",
+#     bin = [":lib_memory_ivshmem_xvm"],
+# )
+#
+# dual_qemu_integration_test(
+#     name = "test_lib_memory_ivshmem_xvm",
+#     srcs = ["lib_memory_ivshmem_xvm_test.py"],
+#     filesystem_a = ":lib_memory_ivshmem_xvm-pkg",
+#     filesystem_b = ":lib_memory_ivshmem_xvm-pkg",
+# )
```
castler: Let's merge this only once this works...otherwise we have no clue if what we have here actually works...and the adde value is then not to big.

quality/integration_testing/environments/dual_qemu/example_xvm/README.md
Rahul-Sutariya: Hi Jan, sorry it was added earlier but not created fix: https://github.com/eclipse-score/baselibs/pull/365,
Removed from the patch details, Thanks

quality/integration_testing/environments/dual_qemu/example_xvm/BUILD
Rahul-Sutariya: Ok, as of now the lib/memory pr: https://github.com/eclipse-score/baselibs/pull/365 merge is block due to lib/mem migration to S-Core/Com. once it merge, we can enable it.

quality/integration_testing/environments/dual_qemu/example_xvm/bar_typed_memory.cpp
Rahul-Sutariya: Adapted the same, Thanks

anmittag:
@Rahul-Sutariya some checks are failing in your PR. Please adapt.

castler:
> @Rahul-Sutariya some checks are failing in your PR. Please adapt.

@anmittag as already stated in https://github.com/eclipse-score/communication/pull/617#discussion_r3619737482, this is right now blocked.

quality/integration_testing/environments/dual_qemu/example_xvm/BUILD
Rahul-Sutariya: Baselib pr#https://github.com/eclipse-score/baselibs/pull/365 got merged, and last bump(23/07/2026: https://github.com/eclipse-score/communication/pull/754/changes/d6890d1dbad3b957e224d4e8bf041195603a0541) to new baselibs version containing this chagnes, Thanks

castler (approved):

PR #621, opened 2026-06-30
Refactor CI Infrastructure

castler:


.github/workflows/nightly_cache_recreation.yml
```diff
@@ -0,0 +1,189 @@
...
+      SCORE_QNX_PASSWORD: ${{ secrets.SCORE_QNX_PASSWORD }}
+    permissions:
+      contents: read
+      actions: write
+      pull-requests: read
+
+  # ======================================================================
+  # Phase 2: Build (parallel)
+  # All jobs restore the repo cache from Phase 1 and build their disk cache.
+  # ======================================================================
```
hoe-jo, line 121: potentially cache race:
Every job has cache mode "recreate-update" and on recreate update **deleteOldCaches** in post.js executes:

```
for (const entry of data.actions_caches || []) {
      if (entry.key !== currentKey) {
        core.info(`Deleting old cache: ${entry.key} (id=${entry.id})`);
        await octokit.rest.actions.deleteActionsCacheById({
          owner,
          repo,
          cache_id: entry.id,
        }).catch(() => {});
      }
    }
```

.github/actions/00_infrastructure/setup_bazel_cache/src/main.js
```diff
@@ -0,0 +1,90 @@
...
+    const lines = [`common --repository_cache=${repoCacheDir}`];
+    if (diskCacheName) {
+      lines.push(`common --disk_cache=${diskCacheDir}`);
+    }
+    fs.appendFileSync(bazelrc, lines.join("\n") + "\n");
+    core.info(`Bazel cache paths written to ${bazelrc}`);
+
+    core.info(`Cache setup complete (mode: ${mode}, disk-cache: ${diskCacheName || "(none)"}). Post hook registered for save.`);
+  } catch (error) {
+    core.setFailed(`Cache setup failed: ${error.message}`);
```
hoe-jo, line 86: Do you want to make a failed cache restore break the CI job or run it without cache instead?

.github/workflows/_build_and_test_gcc15.yml
```diff
@@ -0,0 +1,59 @@
...
+        if: inputs.fetch-only != 'true'
+        run: bazel build //...
+
+      - name: Run all tests
+        if: inputs.fetch-only != 'true'
+        run: bazel test //... --build_tests_only
+
+      - name: Build examples
+        if: inputs.fetch-only != 'true'
+        run: cd examples && bazel build //... && bazel mod deps --lockfile_mode=update
```
hoe-jo, line 53: does update fail? I though only error fails

.github/workflows/_codeql.yml
```diff
@@ -11,67 +11,58 @@
...
-        description: Job conclusion (success/failure)
-        value: ${{ jobs.run-codeql.result }}
-
-permissions:
-  contents: read
-  security-events: write  # Required to upload SARIF results to GitHub Code Scanning
-
-concurrency:
-  group: codeql-${{ github.ref }}
-  cancel-in-progress: false  # Never cancel an in-progress CodeQL run; it takes hours
```
hoe-jo, line 35: Intended to be deleted? Change in strategy

.github/actions/00_infrastructure/setup_bazel_cache/dist/main/index.js
hoe-jo, line 1: Dist not validated agains srcs, in GCC15 a lockfile check is run, should we do the same here?

.github/actions/00_infrastructure/setup_bazel_cache/src/main.js
hoe-jo, line 1: not trivial, should we add unit tests for js files somehow?

.github/workflows/nightly_quality.yml
```diff
@@ -89,13 +90,10 @@ jobs:
...
       - name: Setup Bazel with shared caching
-        uses: castler/setup-bazel@cache-optimized
+        uses: ./.github/actions/00_infrastructure/bazel_cache_restore
         with:
-          bazelisk-cache: true
-          disk-cache: "deploy_quality_reports"
-          repository-cache: true
-          cache-optimized: true
-          cache-save: ${{ github.ref == 'refs/heads/main' }}
+          mode: disabled
```
hoe-jo: why don´t we reuse the js cache action here?

.github/actions/00_infrastructure/setup_bazel_cache/src/post.js
```diff
@@ -0,0 +1,135 @@
...
+          owner,
+          repo,
+          cache_id: entry.id,
+        }).catch(() => {});
+      }
+    }
+  } catch (error) {
+    core.warning(`Cache cleanup failed: ${error.message}`);
+  }
+}
```
hoe-jo, line 71: only retrieves the first 30 entries, if we have somehow a mess in the caches from the 31st onwards they will never get cleaned

.github/workflows/_clang_tidy.yml
```diff
@@ -0,0 +1,136 @@
...
+          ERRORS=$(find -L bazel-bin -name "*.AspectRulesLintClangTidy.out" 2>/dev/null \
+            | xargs cat 2>/dev/null \
+            | grep -c "error:" || true)
+          WARNINGS=$(find -L bazel-bin -name "*.AspectRulesLintClangTidy.out" 2>/dev/null \
+            | xargs cat 2>/dev/null \
+            | grep -c "warning:" || true)
+          TOTAL=$((ERRORS + WARNINGS))
+
+          CHANGED_CPP=$(git diff --name-only "origin/${GITHUB_BASE_REF}...HEAD" 2>/dev/null \
+            | grep -E '\.(cpp|h|cc|cxx|hpp)$' || true)
```
hoe-jo, line 66: Actions checkout only does a shallow checkout, so most probably main is never checked out?

.github/actions/00_infrastructure/bazel_cache_restore/action.yml
hoe-jo, line 1: duplicated code only used for docs and nightly, shall we here also use the node action?

.github/actions/00_infrastructure/bazel_cache_save/action.yml
hoe-jo, line 1: same as for cache restore action

.github/workflows/nightly_cache_recreation.yml
castler: Yes, but only for a specific prefix, and this is different for all disk caches.

.github/actions/00_infrastructure/setup_bazel_cache/src/main.js
castler: Fail, that we do not wonder why our builds are so slow

.github/workflows/_build_and_test_gcc15.yml
castler: yes, it can fail

.github/actions/00_infrastructure/bazel_cache_restore/action.yml
castler: yes

.github/actions/00_infrastructure/bazel_cache_save/action.yml
castler: yes

PR #627, opened 2026-07-01
Small fixes

LittleHuba:


score/mw/com/impl/proxy_event_base.cpp
```diff
@@ -40,7 +40,7 @@ thread_local bool ProxyEventBase::is_in_receive_handler_context = false;
 // coverity[autosar_cpp14_a3_1_1_violation]
 
 ProxyEventBase::ProxyEventBase(std::string_view event_name,
-                               ProxyBinding* proxy_binding_ptr,
+                               ProxyBinding* /*proxy_binding_ptr*/,
```
bemerybmw: This should be removed if it's not used.

bemerybmw (changes requested):

bemerybmw (dismissed):

bemerybmw (approved):

PR #628, opened 2026-07-01
Use StringComparisonAdaptor as key to allow lookup with string_view

ahmed0mousa:


score/mw/com/impl/configuration/binding_service_type_deployment.h
```diff
@@ -28,9 +29,9 @@ template <typename EventIdType, typename FieldIdType, typename MethodIdType, typ
 class BindingServiceTypeDeployment
 {
   public:
-    using EventIdMapping = std::unordered_map<std::string, EventIdType>;
-    using FieldIdMapping = std::unordered_map<std::string, FieldIdType>;
-    using MethodIdMapping = std::unordered_map<std::string, MethodIdType>;
+    using EventIdMapping = std::unordered_map<score::memory::StringComparisonAdaptor, EventIdType>;
```
limdor, line 32: Using StringComparisonAdapter here make very hard to reason about lifetimes. It could be that you own the string and it could be that you do not know the string. If this is only for lookup, is it not possible with custom comparison?

limdor (changes requested):

PR #629, opened 2026-07-01
Force arithmetic_utils_test evaluated at runtime

castler:
This way, we can genearte the necessary coverage information.


LittleHuba (commented):
I consider this a compiler bug, so I would at least like to have a ticket linked that refers to the bug.
From what I can see, this seems to be a good fit: https://github.com/llvm/llvm-project/issues/35434

LittleHuba (approved):

PR #632, opened 2026-07-02
mw/com: Add PRIVATE support for semi dynamic data types

bemerybmw:
Adds support for dynamic types which allocate once at runtime. An
additional size in bytes can be passed to a private proxy Create
function which increases the amount of shared memory allocated for
methods. An application can then create a data type which guarantees to
not allocate more than the provided value.

An example test application is created
mw/com/test/methods/semi_dynamic_methods.

This should only be used in the context of SWP-269486. Any other usage
of this is a violation of our AOU: NoApisFromImplementationNamespace

### DIFF
--- a/score/mw/com/impl/methods/proxy_method.h
+++ b/score/mw/com/impl/methods/proxy_method.h
@@ -13,9 +13,11 @@
 #ifndef SCORE_MW_COM_IMPL_METHODS_PROXY_METHOD_H
 #define SCORE_MW_COM_IMPL_METHODS_PROXY_METHOD_H
 
+#include "score/mw/com/impl/bindings/lola/proxy.h"
 #include "score/mw/com/impl/com_error.h"
 #include "score/mw/com/impl/methods/method_signature_element_ptr.h"
 #include "score/mw/com/impl/methods/proxy_method_binding.h"
+#include "score/mw/com/impl/proxy_binding.h"
 #include "score/mw/com/impl/util/type_erased_storage.h"
 
 #include "score/containers/dynamic_array.h"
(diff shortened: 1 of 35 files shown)


score/mw/com/impl/methods/proxy_method.h
```diff
@@ -146,12 +148,35 @@ score::Result<std::tuple<impl::MethodInArgPtr<ArgTypes>...>> AllocateImpl(
...
+    }
+}
+
 /// \brief Initializes all InArgs by calling the default constructor for each argument.
 ///
 /// This step is important to avoid undefined behaviour (interpreting uninitialized memory) and also to ensure that any
 /// non-trivially constructible types are properly initialized.
 template <typename... ArgTypes>
-Result<void> InitializeInArgs(ProxyMethodBinding& binding, const std::size_t queue_size)
+Result<void> InitializeInArgs(ProxyBinding& proxy_binding, ProxyMethodBinding& binding, const std::size_t queue_size)
```
crimson11, line 183: maybe a small hint (doxygen \param), why we hand down `ProxyBinding` here ...?

\param proxy_binding proxy binding is handed over for an interim solution for support of "semi dynamic" in-arg data-types. In case the data-type is using a PolymorphicOffsetPtrAllocator, we extract the corresponding memory_resource from the proxy_binding

score/mw/com/test/methods/semi_dynamic_methods/consumer.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+    Proxy& GetProxy()
+    {
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(proxy_ != nullptr,
+                                                    "Proxy was not successfully created! Cannot get it!");
+        return *proxy_;
+    }
+
+  private:
+    std::unique_ptr<typename Proxy::HandleType> handle_{nullptr};
```
crimson11: this should be rather an std::optional ... and not a heap allocated unique_ptr ...

score/mw/com/test/methods/semi_dynamic_methods/consumer.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+constexpr std::size_t kNumberOfRuntimeSizedArrayElements = 10U;
+constexpr std::size_t kNumberOfMethodCalls = 10U;
+
+// Additional required shared memory is the number of elements in the runtime sized array multiplied by the size of each
+// element (ArrayValueType). The factor of 2 is because we allocate once for the in arg and once for the return value.
+constexpr std::size_t kAdditionalShmSizeBytes = 2 * kNumberOfRuntimeSizedArrayElements * sizeof(ArrayValueType);
+
+template <typename Proxy>
+class SemiDynamicProxyContainer
```
crimson11, line 47: Confusing name? So whether it is a "semidynamic" proxy is simply a property of template arg Proxy?
The (template) class is rather `ProxyFinder`or so ... as it just encapsulates finding/instantiating a proxy?

score/mw/com/test/methods/semi_dynamic_methods/consumer.cpp
```diff
@@ -0,0 +1,240 @@
...
+{
+    auto allocated_args_result = proxy.with_in_args_and_return.Allocate();
+    if (!allocated_args_result.has_value())
+    {
+        FailTest("Consumer: Could not allocate method args: ", allocated_args_result.error());
+    }
+    auto allocated_args = std::move(allocated_args_result).value();
+    auto& [runtime_sized_array_in_arg_ptr] = allocated_args;
+
+    runtime_sized_array_in_arg_ptr->ReserveOnce(kNumberOfRuntimeSizedArrayElements);
```
crimson11, line 133: Why is there no need to check, whether `runtime_sized_array_in_arg_ptr` is already sized? You are doing this on the skeleton side (which I expect). But why not here?
I.e. in the 2nd call to `proxy.with_in_args_and_return.Allocate();` wouldn't you get typically (queue size 1) te same instance of `RuntimeSizedArray` back? Which has already been sized? Now you are resizing?

crimson11: Ahh - now I see. This func gets only called once ... after this call you call `CallMethodZeroCopyWithoutAllocation` ...

The "Allocate" confused me, because you are ALWAYS calling `proxy.with_in_args_and_return.Allocate()`!

Maybe rename:
CallMethodZeroCopyWithAllocation -> CallMethodZeroCopyWithDynamicDataTypeResize
CallMethodZeroCopyWithoutAllocation -> CallMethodZeroCopyWithoutDynamicDataTypeResize

IDK - you decide - maybe I was just dumb.

crimson11 (changes requested):

score/mw/com/test/methods/semi_dynamic_methods/consumer.cpp
bemerybmw: No, it's basically a copy/paste of ProxyContainer (a helper class which I created https://github.com/eclipse-score/communication/blob/main/score/mw/com/test/common_test_resources/proxy_container.h) except that it calls the private proxy create function which takes the additional_shm_size_bytes as an argument. So it's specific to SemiDynamic types. I updated the template arg name to be clearer.

score/mw/com/test/methods/semi_dynamic_methods/consumer.cpp
bemerybmw: Yeah, I meant dynamic memory allocation, but I see the confusion. I'll change name.

crimson11 (approved):

PR #641, opened 2026-07-03
Reduce visibility & Introduce visibility check

castler:
The first step for a clear public API is, what Bazel targets user can depend on.
Thus we clean this in the first step, and ensure that this will not be changed by accident anymore


score/mw/com/BUILD
```diff
@@ -107,7 +107,10 @@ cc_library(
     srcs = ["test_types.cpp"],
     hdrs = ["test_types.h"],
     features = COMPILER_WARNING_FEATURES,
-    visibility = ["//visibility:public"],
+    visibility = [
```
castler: This change needs to be reverted

quality/compiler_warnings/gcc/BUILD
```diff
@@ -115,7 +115,7 @@ cc_feature(
     ],
     feature_name = "score_communication_strict_warnings",
     implies = ["minimal_warnings"],
-    visibility = ["//visibility:public"],
+    visibility = ["@toolchains_llvm//:__subpackages__"],
```
castler: gcc flags with visibility for llvm? This does not make sense.

quality/visibility_guard/parser.py
```diff
@@ -0,0 +1,110 @@
+# *******************************************************************************
```
castler, line 1: Why are we not using bazel query for this?

score/mw/com/impl/plumbing/BUILD
```diff
@@ -20,7 +20,7 @@ cc_library(
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
     visibility = [
-        "//score/mw/com/impl:__subpackages__",
+        "//score/mw/com:__subpackages__",
```
castler: why do we need this change? Why do we increase the visibility in this instance and in others?

castler (commented):
Submitting pending review before posting thread updates.

quality/compiler_warnings/gcc/BUILD
castler: Fixed. Updated GCC feature visibilities in quality/compiler_warnings/gcc/BUILD to //visibility:public for strict_warnings, strict_warnings_no_error, and third_party_warnings. The @toolchains_llvm visibility there was incorrect for the GCC file.

score/mw/com/impl/plumbing/BUILD
castler: Fixed. Reverted the visibility broadening in score/mw/com/impl/plumbing/BUILD back to //score/mw/com/impl:__subpackages__, and restored the platform_only mock targets to //visibility:public.

score/mw/com/BUILD
castler: Kept intentionally. //score/mw/com:test_types is required to stay publicly visible and listed in the visibility golden set; reverting it would reduce the public visibility surface.

quality/visibility_guard/parser.py
castler: Current implementation intentionally parses source BUILD files to track public visibility declarations in-repo. A raw bazel query returns additional generated/macro targets, which would expand this guard beyond direct source declarations. If you want the guard to include generated targets too, I can switch it to query-based collection.

score/mw/com/BUILD
castler: Fixed. Simplified //score/mw/com:test_types visibility to a single entry: //visibility:public. The extra //score/mw/com:__subpackages__ entry was redundant.

quality/visibility_guard/parser.py
castler: Done. Switched parser.py to use `bazel query "attr(visibility, //visibility:public, //...)"` — this captures both source-declared and macro-generated public targets. The golden file now has 60 entries (was 22). Two challenges solved for the test context: (1) nested bazel would deadlock on the server lock → use `--output_base` pointing to `TEST_TMPDIR`; (2) `bazel test` strips HOME and the binary is bazelisk which needs HOME to find its cached download → restore HOME from the password database.

LittleHuba (approved):

PR #657, opened 2026-07-06
Support QNX coverage measurement

castler:


quality/coverage/generate_coverage_html.sh
```diff
@@ -11,36 +11,56 @@
...
       OUTPUT_DIR="$1"
       shift
       ;;
   esac
 done
 
+# Set default output directory based on platform if not explicitly provided.
+if [[ -z "${OUTPUT_DIR}" ]]; then
+  if [[ "${PLATFORM}" == "linux" ]]; then
+    OUTPUT_DIR="cpp_coverage"
```
LittleHuba: Ideally we get rid of this extra case (e.g. make this `OUTPUT_DIR="cpp_coverage_linux"`). This can be done in a follow-up.

quality/coverage/lcov_to_html.py
```diff
@@ -0,0 +1,270 @@
...
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Generate HTML coverage reports from LCOV data using gcovr.
+
+Converts LCOV trace data to gcovr's JSON format (v0.14), then invokes gcovr
```
LittleHuba, line 16: gcovr can directly read GCOV itself. Why do we not collect those files via bazel but instead translate the information from the LCOV file?

quality/coverage/llvm_cov/effective_coverage.py
```diff
@@ -730,8 +746,444 @@ def parse_args() -> argparse.Namespace:
...
+    unjustified_uncovered = raw_uncovered - total_justified
+    effective_branch_covered = raw_branch_covered + total_justified_branches
+
+    stats = {
+        "total_instrumented_lines": raw_total,
+        "covered_lines": raw_covered,
+        "justified_lines": total_justified,
+        "unjustified_uncovered_lines": max(0, unjustified_uncovered),
+        "stale_justifications": total_stale,
+        "raw_line_coverage_pct": round(100.0 * raw_covered / raw_total, 2) if raw_total > 0 else 0.0,
```
LittleHuba: This should be a `floor` instead of a `round` to not get confusing results.

quality/coverage/llvm_cov/effective_coverage.py
```diff
@@ -730,8 +746,444 @@ def parse_args() -> argparse.Namespace:
...
+        "effective_line_coverage_pct": round(
```
LittleHuba: Same here

quality/coverage/llvm_cov/effective_coverage.py
```diff
@@ -730,8 +746,444 @@ def parse_args() -> argparse.Namespace:
...
+            100.0 * (raw_covered + total_justified) / raw_total, 2
+        ) if raw_total > 0 else 0.0,
+        "total_branches": raw_branch_total,
+        "covered_branches": raw_branch_covered,
+        "justified_branches": total_justified_branches,
+        "raw_branch_coverage_pct": round(100.0 * raw_branch_covered / raw_branch_total, 2) if raw_branch_total > 0 else 0.0,
+        "effective_branch_coverage_pct": round(
```
LittleHuba: Same here

quality/coverage/effective_coverage.py
```diff
@@ -730,8 +746,444 @@ def parse_args() -> argparse.Namespace:
...
+
+    if args.output:
+        with open(args.output, "w", encoding="utf-8") as f:
+            json.dump(report, f, indent=2)
+        print(f"Effective coverage report written to: {args.output}")
+    else:
+        print(json.dumps(report, indent=2))
+
+
+def _parse_gcovr_index_totals(html_dir: Path) -> Dict[str, Tuple[int, int]]:
```
LittleHuba, line 898: Why not generate a json report next to the html repot with gcovr that contains all this data? That would safe you this parsing.

quality/coverage/README.md
```diff
@@ -67,40 +68,67 @@ The pipeline must support a configurable effective coverage threshold (default:
...
-# Full project
+# Full project (Linux, LLVM source-based coverage)
 bazel coverage //...
 
-# Specific target
+# Full project (QNX x86_64, gcov-based coverage)
+bazel coverage //score/... --config=qnx
+
+# Specific target (Linux)
 bazel coverage //score/message_passing:client_connection_test_linux
```
LittleHuba: Does specific target coverage not work for QNX?

quality/coverage/README.md
```diff
@@ -67,40 +68,67 @@ The pipeline must support a configurable effective coverage threshold (default:
...
+
+The report is written to `cpp_coverage/` (Linux) or `cpp_coverage_qnx/` (QNX). Open it:
 
 ```bash
+# Linux
 xdg-open cpp_coverage/index.html
+
+# QNX
+xdg-open cpp_coverage_qnx/index.html
 ```
```
LittleHuba, line 108: I feel that these sections would be better separated by platform. That gives one clean section of commands to follow.

quality/coverage/README.md
```diff
@@ -196,6 +257,19 @@ better that justifications get outdated.
 | `platform_specific` | Code path only reachable on platforms not under test |
 | `other` | Any other valid reason |
 
+### Platform Scoping
+
+Each justification can optionally specify which platforms it applies to via the `platforms` field:
+
+| Value | Meaning |
+|-------|---------|
+| *(omitted)* | Justification applies to **all** platforms (Linux, QNX, ...) |
```
LittleHuba: I see this as problematic, since it makes adding future platforms error-prone.
I would prefer to make platform scoping mandatory.

quality/coverage/lcov_to_html.py
castler: We do so in order to keep the full bazel collection mechanism. We either: have to rewrite that as we did for llvm-cov, or have this translation layer or use genhtml.

I thought for the beginning, this way is the best. We have to see if true over time, or we need to adjust it.

quality/coverage/effective_coverage.py
castler: Because to avoid that we have again different information in the data. I thought it would be just easier to stay with one data source, instead of being afraid that two data sources are again consistent.

quality/coverage/README.md
castler: for each of our `cc_unit_test` we generate two targets. One compatible with Linux and one with QNX.
These underlying targets need different names and will always have the suffix `_linux` or ´_qnx´.

Thus, this specific target is the instance of `client_connection_test` for _Linux_.

As per my understanding.

quality/coverage/README.md
castler: Can we please leave it this way? I want to highlight as close as possible where the differences are.

quality/coverage/README.md
LittleHuba: I was more going in the direction of suggesting to add a line for the QNX equivalent.

quality/coverage/effective_coverage.py
```diff
@@ -534,7 +551,7 @@ def update_index_page(html_dir: Path, stats: Dict[str, Any], per_file_stats: Dic
             covered = int(branch_cell.group(3))
             total = int(branch_cell.group(4))
             eff_covered = covered + justified_branches
-            eff_pct = round(100.0 * eff_covered / total, 2) if total > 0 else 0.0
+            eff_pct = math.floor(100.0 * eff_covered / total) if total > 0 else 0.0
```
LittleHuba: This rounds to full integers which is quite coarse. Better would be to floor but retain two digits.

quality/coverage/effective_coverage.py
LittleHuba: My expectation would have been that both reports coming from gcovr that we don't have problems. But fair point with no proof for that assumption.

LittleHuba (approved):

PR #659, opened 2026-07-07
mw/com : Add the integration tests for moving skeletonMethod.

Tejveerpratap2803:
Adding integration tests that verify a Skeleton (and its SkeletonMethod) can be move-constructed or move-assigned without disrupting method handler registration or invocation.

Issue : #483

### DIFF
--- a/score/mw/com/test/move_semantics/skeleton_method/config/mw_com_config.json
+++ b/score/mw/com/test/move_semantics/skeleton_method/config/mw_com_config.json
@@ -1,0 +1,102 @@
+{
+  "serviceTypes": [
+    {
+      "serviceTypeName": "/test/move_semantics/skeleton_method/SkeletonMethodMoveInterface",
+      "version": {
+        "major": 1,
+        "minor": 0
+      },
+      "bindings": [
+        {
+          "binding": "SHM",
+          "serviceId": 2200,
+          "methods": [
+            {
+              "methodName": "moved_method",
+              "methodId": 1
+            }
+          ]
+        }
+      ]
+    }
+  ],
+  "serviceInstances": [
+    {
+      "instanceSpecifier": "test/move_semantics/skeleton_method/SkeletonMethodMoveInterfaceMovedTo",
+      "serviceTypeName": "/test/move_semantics/skeleton_method/SkeletonMethodMoveInterface",
+      "version": {
+        "major": 1,
+        "minor": 0
+      },
+      "instances": [
+        {
+          "instanceId": 1,
+          "asil-level": "B",
+          "binding": "SHM",
+          "methods": [
+            {
+              "methodName": "moved_method",
+              "queueSize": 1
+            }
+          ],
+          "allowedConsumer": {
+            "QM": [
+              0
+            ],
+            "B": [
+              0
+            ]
+          },
+          "allowedProvider": {
+            "QM": [
+              0
+            ],
+            "B": [
+              0
+            ]
+          }
+        }
+      ]
+    },
+    {
+      "instanceSpecifier": "test/move_semantics/skeleton_method/SkeletonMethodMoveInterfaceMovedFrom",
+      "serviceTypeName": "/test/move_semantics/skeleton_method/SkeletonMethodMoveInterface",
+      "version": {
+        "major": 1,
+        "minor": 0
+      },
+      "instances": [
+        {
+          "instanceId": 2,
+          "asil-level": "B",
+          "binding": "SHM",
+          "methods": [
+            {
+              "methodName": "moved_method",
+              "queueSize": 1
+            }
+          ],
+          "allowedConsumer": {
+            "QM": [
+              0
+            ],
+            "B": [
+              0
+            ]
+          },
+          "allowedProvider": {
+            "QM": [
+              0
+            ],
+            "B": [
+              0
+            ]
+          }
+        }
+      ]
+    }
+  ],
+  "global": {
+    "asil-level": "B"
+  }
+}
--- a/score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
+++ b/score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
@@ -21,6 +21,7 @@
 
 #include <chrono>
 #include <iostream>
+#include <optional>
 #include <thread>
 
 namespace score::mw::com::test
@@ -28,9 +29,7 @@ namespace score::mw::com::test
 namespace
 {
 
-constexpr auto kDetectionWindow = std::chrono::milliseconds{5000};
-
-constexpr auto kDetectionCallGap = std::chrono::milliseconds{5};
+constexpr auto kDetectionCallGap = std::chrono::milliseconds{2};
 
 std::int32_t CallMethodOrFail(SkeletonMethodMoveProxy& proxy)
 {
@@ -42,21 +41,9 @@ std::int32_t CallMethodOrFail(SkeletonMethodMoveProxy& proxy)
     return *(result.value());
 }
 
-void VerifyResultIsOneOf(std::int32_t actual,
-                         std::int32_t first_expected,
-                         std::int32_t second_expected,
-                         std::size_t call_index)
-{
-    if (actual != first_expected && actual != second_expected)
-    {
-        FailTest(
-            "Consumer: call ", call_index, " expected ", first_expected, " or ", second_expected, " but got ", actual);
-    }
-}
-
 }  // namespace
 
-void RunConsumer(const SkeletonMoveScenario& scenario, const score::cpp::stop_token& stop_token)
+void RunConsumer(const score::cpp::stop_token& stop_token)
 {
     static_cast<void>(stop_token);
 
@@ -71,24 +58,37 @@ void RunConsumer(const SkeletonMoveScenario& scenario, const score::cpp::stop_to
     //         (fuzzy) scenarios the move may already have happened, so any call may return either
     //         handler's result — each call is just checked against the two known-valid results.
     std::cout << "\nConsumer: Step 2 - Loop calling until second handler is detected" << std::endl;
-    const std::int32_t first_expected = GetFirstHandlerExpectedResult(scenario);
-    const std::int32_t second_expected = GetSecondHandlerExpectedResult(scenario);
-    const auto deadline = std::chrono::steady_clock::now() + kDetectionWindow;
     std::size_t call_count = 0U;
-    std::int32_t actual = 0;
-    do
+    std::optional<std::int32_t> actual{};
+    while (!actual.has_value() || actual.value() != kSecondHandlerExpectedResult)
     {
-        if (std::chrono::steady_clock::now() >= deadline)
-        {
-            FailTest("Consumer: timed out waiting to observe the second handler's result");
-        }
         actual = CallMethodOrFail(proxy_moved_to);
-        VerifyResultIsOneOf(actual, first_expected, second_expected, call_count);
         ++call_count;
-        std::this_thread::sleep_for(kDetectionCallGap);
-    } while (actual != second_expected);
 
-    std::cout << "\nConsumer: Step 2 done (" << call_count << " calls, second handler result=" << actual << ")"
+        // If the first handler is still registered, then we continue calling the method.
+        if (actual == kFirstHandlerExpectedResult)
+        {
+            std::this_thread::sleep_for(kDetectionCallGap);
+            continue;
+        }
+
+        // If the second handler has been registered, then we can finish.
+        if (actual == kSecondHandlerExpectedResult)
+        {
+            break;
+        }
+
+        FailTest("Consumer: call ",
+                 call_count,
+                 " expected ",
+                 kFirstHandlerExpectedResult,
+                 " or ",
+                 kSecondHandlerExpectedResult,
+                 " but got ",
+                 actual.value());
+    }
+
+    std::cout << "\nConsumer: Step 2 done (" << call_count << " calls, second handler result=" << actual.value() << ")"
               << std::endl;
 }
 
--- a/score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
+++ b/score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
@@ -17,8 +17,6 @@
 #include "score/mw/com/test/move_semantics/skeleton_method/test_method_datatype.h"
 #include "score/mw/com/test/move_semantics/skeleton_method/test_parameters.h"
 
-#include <score/stop_token.hpp>
-
 #include <chrono>
 #include <iostream>
 #include <optional>
@@ -29,7 +27,7 @@ namespace score::mw::com::test
 namespace
 {
 
-constexpr auto kDetectionCallGap = std::chrono::milliseconds{2};
+constexpr auto kMethodCallInterval = std::chrono::milliseconds{2};
 
 std::int32_t CallMethodOrFail(SkeletonMethodMoveProxy& proxy)
 {
@@ -43,10 +41,8 @@ std::int32_t CallMethodOrFail(SkeletonMethodMoveProxy& proxy)
 
 }  // namespace
 
-void RunConsumer(const score::cpp::stop_token& stop_token)
+void RunConsumer()
 {
-    static_cast<void>(stop_token);
-
     // Step 1. Create proxy for kInstanceSpecifierMovedTo.
     std::cout << "\nConsumer: Step 1 - Create proxy (kMovedTo)" << std::endl;
     ProxyContainer<SkeletonMethodMoveProxy> proxy_moved_to_container{};
@@ -59,21 +55,22 @@ void RunConsumer(const score::cpp::stop_token& stop_token)
     //         handler's result — each call is just checked against the two known-valid results.
     std::cout << "\nConsumer: Step 2 - Loop calling until second handler is detected" << std::endl;
     std::size_t call_count = 0U;
-    std::optional<std::int32_t> actual{};
-    while (!actual.has_value() || actual.value() != kSecondHandlerExpectedResult)
+    std::optional<std::int32_t> actual_method_return_value{};
+    while (!actual_method_return_value.has_value() ||
+           actual_method_return_value.value() != kSecondHandlerExpectedResult)
     {
-        actual = CallMethodOrFail(proxy_moved_to);
+        actual_method_return_value = CallMethodOrFail(proxy_moved_to);
         ++call_count;
 
         // If the first handler is still registered, then we continue calling the method.
-        if (actual == kFirstHandlerExpectedResult)
+        if (actual_method_return_value == kFirstHandlerExpectedResult)
         {
-            std::this_thread::sleep_for(kDetectionCallGap);
+            std::this_thread::sleep_for(kMethodCallInterval);
             continue;
         }
 
         // If the second handler has been registered, then we can finish.
-        if (actual == kSecondHandlerExpectedResult)
+        if (actual_method_return_value == kSecondHandlerExpectedResult)
         {
             break;
         }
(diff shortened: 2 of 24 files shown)


score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,120 @@
...
+{
+
+const std::string kProxyDoneShmPath{"/skeleton_method_move_semantics_proxy_done_sync"};
+const std::string kSkeletonReadyShmPath{"/skeleton_method_move_semantics_skeleton_ready_sync"};
+
+}  // namespace
+
+void RunConsumer(const SkeletonMoveScenario& scenario, const score::cpp::stop_token& stop_token)
+{
+    auto proxy_done_sync = ProcessSynchronizer::Create(kProxyDoneShmPath);
```
bemerybmw: We generally avoid abbreviations: proxy_done_process_synchronizer

score/mw/com/test/move_semantics/skeleton_method/config/mw_com_config.json
```diff
@@ -0,0 +1,99 @@
...
+      "instanceSpecifier": "test/move_semantics/skeleton_method/SkeletonMethodMoveInterfaceMovedTo",
+      "serviceTypeName": "/test/move_semantics/skeleton_method/SkeletonMethodMoveInterface",
+      "version": {
+        "major": 1,
+        "minor": 0
+      },
+      "instances": [
+        {
+          "instanceId": 1,
+          "asil-level": "QM",
```
bemerybmw: I think in general, we should use asil b by default unless there's a good reason not to. The most important thing is that our code works for asil b applications. Also, an asil b provider sets up infrastructure for both asil b and qm, so we're technically testing more code paths.

score/mw/com/test/move_semantics/skeleton_method/test_parameters.h
```diff
@@ -0,0 +1,74 @@
...
+const InstanceSpecifier kInstanceSpecifierMovedTo =
+    InstanceSpecifier::Create(std::string{"test/move_semantics/skeleton_method/SkeletonMethodMoveInterfaceMovedTo"})
+        .value();
+const InstanceSpecifier kInstanceSpecifierMovedFrom =
+    InstanceSpecifier::Create(std::string{"test/move_semantics/skeleton_method/SkeletonMethodMoveInterfaceMovedFrom"})
+        .value();
+
+enum class SkeletonMoveScenario : std::uint8_t
+{
+    kMoveConstructBeforeOffered = 0,
```
bemerybmw: No need to manually specify the numbers unless they're actually important e.g. if the counting starts from a non zero value.

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
```diff
@@ -0,0 +1,444 @@
...
+    RegisterHandlerOrFail(
+        skeleton_a.moved_method_,
+        [](std::int32_t& result, const std::int32_t& a, const std::int32_t& b) {
+            result = a + b;
+        },
+        "Provider: Failed to register Handler A: ");
+
+    // Step 3. Move construct: Skeleton B = std::move(Skeleton A)  [before OfferService]
+    std::cout << "\nProvider: Step 3 - Move construct Skeleton B = std::move(Skeleton A)" << std::endl;
+    auto skeleton_b = std::move(skeleton_a);
```
bemerybmw: What about moving once before RegisterHandlerOrFail and again here?

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,161 @@
...
+    std::optional<ProxyContainer<SkeletonMethodMoveProxy>> proxy_moved_from_container{};
+    if (scenario == SkeletonMoveScenario::kMoveAssignAfterOffered)
+    {
+        std::cout << "\nConsumer: Step 2 - Create proxy (kMovedFrom)" << std::endl;
+        proxy_moved_from_container.emplace();
+        proxy_moved_from_container->CreateProxy(kInstanceSpecifierMovedFrom, "skeleton_method_move_semantics");
+    }
+
+    // Step 3. For after-offered scenarios: signal provider that proxy/proxies are created
+    if (IsAfterOffered(scenario))
```
bemerybmw: Having a single function and using these flags makes the test harder to read IMO. I think it's clearer to have separate functions for the different tests to make the sequences clear.

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
```diff
@@ -0,0 +1,444 @@
...
+{
+
+const std::string kProxyDoneShmPath{"/skeleton_method_move_semantics_proxy_done_sync"};
+const std::string kSkeletonReadyShmPath{"/skeleton_method_move_semantics_skeleton_ready_sync"};
+
+void SleepRandomDuration()
+{
+    std::random_device rd{};
+    std::mt19937 gen{rd()};
+    std::uniform_int_distribution<int> dist{0, 500};
```
bemerybmw: We shouldn't have a magic number here of 500. This value is very important for the test so should be a constant (probably defined in test_parameters).

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,161 @@
...
+    {
+        std::cout << "\nConsumer: Step 3 - Notify provider: proxy/proxies created" << std::endl;
+        proxy_done_sync->Notify();
```
bemerybmw: After discussing with @limdor, we agreed that we shouldn't go with the approach of calling move in a loop which is more of a stress test. This minimises load in the ci and more importantly keeps the test sequence clearer i.e. that it is just testing one thing, i.e. what happens when we move the handler once. Since we have the nightly job which runs the test many times, we should get a lot of runs with different timings. Based on this and our discussion on Friday, I think that we should have 4 tests: 

- move construct before offer
- move construct after offer (fuzzy)
- move assign before offer
- move assign after offer (fuzzy) 

**Before offer**
The before offer tests should move the skeleton twice: before and after registering the handler (it then offers the service). After it detects that the first method handler was called (using a concurrency::Notification: it should notify a concurrency::Notification to indicate that the handler was called. the main thread of the test should wait on this notification), it should then register a new handler an once that handler is called, it can finish the test which destroys the skeleton / stop offers the service.

In the consumer, it should calling the first method. It should then keep calling the method until it gets a return value uniquely identifying the second handler and then can finish. If I haven't missed something, I think this sequence would not even require a ProcessSynchronizer for indicating to the skeleton when the consumer is done (since it gets this when the last registered handler is called).

**After offer**

The provider should offer the service and then sleep a random time and then move the skeleton. After this, it should register a new handler and repeat (similar to the before offer test). The consumer should behave the same as in the before offer test.

The "fuzzy" behaviour is basically part of the after offer tests.

bemerybmw (changes requested):

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Removed now we are using concurrency::Notification.

score/mw/com/test/move_semantics/skeleton_method/test_parameters.h
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
Tejveerpratap2803: Refined the code

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/test_parameters.h
```diff
@@ -0,0 +1,69 @@
...
+{
+    SkeletonMoveScenario scenario;
+    std::string service_instance_manifest;
+};
+
+CombinedTestConfiguration ReadCommandLineArguments(int argc, const char** argv);
+
+/// \brief Returns the expected return value of moved_method_(kTestArgA, kTestArgB) for the first,
+///        deterministic call (always resolves to Handler A: a + b = 15).
+std::int32_t GetFirstHandlerExpectedResult(SkeletonMoveScenario scenario);
```
bemerybmw: Why is this a function which takes SkeletonMoveScenario? It doesn't depend on the scenario so why not just make it a constant?

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    //         handler's result — each call is just checked against the two known-valid results.
+    std::cout << "\nConsumer: Step 2 - Loop calling until second handler is detected" << std::endl;
+    const std::int32_t first_expected = GetFirstHandlerExpectedResult(scenario);
+    const std::int32_t second_expected = GetSecondHandlerExpectedResult(scenario);
+    const auto deadline = std::chrono::steady_clock::now() + kDetectionWindow;
+    std::size_t call_count = 0U;
+    std::int32_t actual = 0;
+    do
+    {
+        if (std::chrono::steady_clock::now() >= deadline)
```
bemerybmw: This timeout is handled by the test infrastructure. I don't think we should also do a timeout check within the test itself (since it's redundant and also complicates the test sequence).

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    // Step 2. Keep calling the method until the return value identifies the second handler. For
+    //         "before offered" scenarios the first call is always Handler A. For "after offered"
+    //         (fuzzy) scenarios the move may already have happened, so any call may return either
+    //         handler's result — each call is just checked against the two known-valid results.
+    std::cout << "\nConsumer: Step 2 - Loop calling until second handler is detected" << std::endl;
+    const std::int32_t first_expected = GetFirstHandlerExpectedResult(scenario);
+    const std::int32_t second_expected = GetSecondHandlerExpectedResult(scenario);
+    const auto deadline = std::chrono::steady_clock::now() + kDetectionWindow;
+    std::size_t call_count = 0U;
+    std::int32_t actual = 0;
```
bemerybmw: We should not use 0 to represent an uninitialized value. We should instead use std::optional<std::int32_t>. You can then also change the do while loop to a while loop with the condition being `while (!actual.has_value() || actual.value() != second_expected)`

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    do
+    {
+        if (std::chrono::steady_clock::now() >= deadline)
+        {
+            FailTest("Consumer: timed out waiting to observe the second handler's result");
+        }
+        actual = CallMethodOrFail(proxy_moved_to);
+        VerifyResultIsOneOf(actual, first_expected, second_expected, call_count);
```
bemerybmw: I think that this function makes the test logic harder to understand. I would rather something like this to make the sequence clearer:

```
// If the first handler is still registered, then we continue calling the method
if (actual == first_expected)
{
    // sleep
    continue;
}

// If the second handler has been registered, then we can finish
if (actual == second_expected)
{
    break;
}

FailTest(true, ...);
```

What do you think?

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
bemerybmw: See my comment below, it would also simplify the condition.

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+    // Step 4. Sleep a random duration (0-500ms). Fuzzy Scenario: Move operation is done at random point while
+    //         the consumer is independently calling the method in a loop.
+    SleepRandomDuration();
+
+    // Step 5. Move construct: Skeleton B = std::move(Skeleton A)  [already offered, random timing]
+    std::cout << "\nProvider: Step 5 - Move construct Skeleton B = std::move(Skeleton A)" << std::endl;
+    auto skeleton_b = std::move(skeleton_a);
+
+    // Step 6. Register Handler B on Skeleton B. Its body notifies notification_b when called.
```
bemerybmw: We should wait for the first handler to be called at least once before registering the new one.

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
```diff
@@ -0,0 +1,336 @@
...
+    OfferSkeletonOrFail(skeleton_a, "Provider: OfferService (Skeleton A) failed: ");
+    OfferSkeletonOrFail(skeleton_b, "Provider: OfferService (Skeleton B) failed: ");
+
+    // Step 4. Sleep a random duration (0-500ms).
+    SleepRandomDuration();
+
+    // Step 5. Move-assign: Skeleton B = std::move(Skeleton A)  [already offered, random timing]
+    std::cout << "\nProvider: Step 5 - Move-assign Skeleton B = std::move(Skeleton A)" << std::endl;
+    skeleton_b = std::move(skeleton_a);
+
```
bemerybmw, line 295: Wait for first handler to be called.

bemerybmw (changes requested):

score/mw/com/test/move_semantics/skeleton_method/test_parameters.h
Tejveerpratap2803: Removed

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Make sense! changed it

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,95 @@
...
+        FailTest("Consumer: moved_method_ call failed: ", result.error());
+    }
+    return *(result.value());
+}
+
+}  // namespace
+
+void RunConsumer(const score::cpp::stop_token& stop_token)
+{
+    static_cast<void>(stop_token);
```
bemerybmw: If the stop token is not used, then I'd remove it from the function signature

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    proxy_moved_to_container.CreateProxy(kInstanceSpecifierMovedTo, "skeleton_method_move_semantics");
+    auto& proxy_moved_to = proxy_moved_to_container.GetProxy();
+
+    // Step 2. Keep calling the method until the return value identifies the second handler. For
+    //         "before offered" scenarios the first call is always Handler A. For "after offered"
+    //         (fuzzy) scenarios the move may already have happened, so any call may return either
+    //         handler's result — each call is just checked against the two known-valid results.
+    std::cout << "\nConsumer: Step 2 - Loop calling until second handler is detected" << std::endl;
+    std::size_t call_count = 0U;
+    std::optional<std::int32_t> actual{};
```
bemerybmw:
```suggestion
    std::optional<std::int32_t> actual_method_return_value{};
```

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
```diff
@@ -0,0 +1,95 @@
...
+    while (!actual.has_value() || actual.value() != kSecondHandlerExpectedResult)
+    {
+        actual = CallMethodOrFail(proxy_moved_to);
+        ++call_count;
+
+        // If the first handler is still registered, then we continue calling the method.
+        if (actual == kFirstHandlerExpectedResult)
+        {
+            std::this_thread::sleep_for(kDetectionCallGap);
```
bemerybmw: kMethodCallInterval instead of kDetectionCallGap?

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
bemerybmw: Still not done.

bemerybmw (changes requested):

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
```diff
@@ -0,0 +1,349 @@
...
+    // Step 4. Wait until Handler A has been called at least once before proceeding.
+    std::cout << "\nProvider: Step 4 - Wait for Handler A to be called" << std::endl;
+    WaitOrFail(handler_a_called, stop_token, "Provider: Notification wait (Handler A) timed out");
+
+    // Step 5. Sleep a random duration (0-500ms).
+    SleepRandomDuration();
+
+    // Step 6. Move-assign: Skeleton B = std::move(Skeleton A)  [already offered, random timing]
+    std::cout << "\nProvider: Step 6 - Move-assign Skeleton B = std::move(Skeleton A)" << std::endl;
+    skeleton_b = std::move(skeleton_a);
```
bemerybmw, line 294: Wait for first handler to be called.

Tejveerpratap2803: Already made the changes.  waitorFail(handler_a_called)   ----> sleep -----> move

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Removed

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/consumer.cpp
Tejveerpratap2803: Done

score/mw/com/test/move_semantics/skeleton_method/provider.cpp
bemerybmw: Yes but shouldn't we check that it's called again after moving?

Tejveerpratap2803: I’m not sure it would make sense to check this again after moving, since the shared memory has already been created and we’ve already verified that scenario.

bemerybmw (approved):

PR #662, opened 2026-07-07
Add bazel test rule for cpp compile time errors

yolani:
Goal of this PR is to make the static_assert / compile time checks in score/mw/com/impl/util/arithmetic_utils.h testable.

Since there is no built-in support in bazel to do this at least for C++, i propose to implement a new bazel rule that utilizes a small shell script to invoke the compilation while re-using the default cpp toolchain from the repository.

The tests have to be extracted into individual compilation units which are tested by the new rule.

Alternative: Extract the logic form the call to static_assert into a standalone testable function. That would work for this file but not for the remaining ~80 compile time checks throughout the repository.


LittleHuba (commented):
There is https://registry.bazel.build/modules/rules_build_error. Could try this and propose changes upstream if we miss something?

LittleHuba (dismissed):

LittleHuba (approved):
Move of all tests is really relevant because rules_build_error is a dev_dependency and therefore in using modules not present. If the `load()` statement in one of the packages that contains productiont targets references rules_build_error we must either make it a public dependency or move the load statement out of the package.

This PR correctly chooses the latter, since our tests are not part of our public API and therefore making a test-only dependency a public dependency would be incorrect.

PR #664, opened 2026-07-07
Add a workflow to test dependency versions

LittleHuba:
Tests different combinations of dependency versions. This enables users to have some certainty on which version windows our module is buildable.

For the moment we only run this on a nightly basis. Once we know what version windows we actually want to support, we may enforce this on a stricter basis.

Versions are selected using single_version_overrides. Conflicting git_overrides are overwritten.


castler:
Succeeded by https://github.com/eclipse-score/communication/pull/773

PR #667, opened 2026-07-07
Brem methods sctfs multiple proxies

bemerybmw:


LittleHuba:
Likely an issue with inotify in QNX. Events seem to be reordered.
We will wait for https://github.com/eclipse-score/bazel_cpp_toolchains/issues/68 to be resolved and retest.
If the problem persists, we will raise a new bug ticket to investigate further.

PR #668, opened 2026-07-07
mw::com: update the default config value to true

KrishaDeshkool:
previously there was no default value for use_get/set_if_available paramter,
now we explicitly set it to true.

### DIFF
--- a/score/mw/com/impl/configuration/config_parser_test.cpp
+++ b/score/mw/com/impl/configuration/config_parser_test.cpp
@@ -127,10 +127,8 @@ TEST_F(ConfigParserFixture, ParseExampleJson)
     EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft")
                   .lola_event_instance_deployment_.max_concurrent_allocations_.value(),
               1);
-    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_get_if_available_,
-              std::optional<bool>{true});
-    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_set_if_available_,
-              std::optional<bool>{true});
+    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_get_if_available_, true);
+    EXPECT_EQ(secondDeploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_set_if_available_, true);
     ASSERT_TRUE(secondDeploymentInfo.methods_.at("SetPressure").enabled_.has_value());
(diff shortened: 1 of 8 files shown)


score/mw/com/impl/configuration/config_parser_test.cpp
```diff
@@ -2186,10 +2186,11 @@ TEST(ConfigParser, LolaFieldUseGetIfAvailableSetToTrue)
 
     const auto deploymentInfo = std::get<LolaServiceInstanceDeployment>(deployment.bindingInfo_);
 
-    // Then use_get_if_available_ is true and use_set_if_available_ is not set
+    // Then use_get_if_available_ is true and use_set_if_available_ defaults to true
```
bemerybmw, line 2187: We also need to change the json schema so that the default is true instead of false.

score/mw/com/impl/configuration/config_parser_test.cpp
```diff
@@ -2255,13 +2256,14 @@ TEST(ConfigParser, LolaFieldUseSetIfAvailableSetToTrue)
 
     const auto deploymentInfo = std::get<LolaServiceInstanceDeployment>(deployment.bindingInfo_);
 
-    // Then use_set_if_available_ is true and use_get_if_available_ is not set
-    EXPECT_EQ(deploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_get_if_available_, std::nullopt);
+    // Then use_set_if_available_ is true and use_get_if_available_ defaults to true
+    EXPECT_EQ(deploymentInfo.fields_.at("CurrentTemperatureFrontLeft").use_get_if_available_,
+              std::optional<bool>{true});
```
bemerybmw: This should no longer be an optional in the config struct, since it will have a default value if not listed.

score/mw/com/impl/configuration/config_parser_test.cpp
bemerybmw, line 2169: Since the default is now true, we should use false here. If the parsing was broken and this value wasn't parsed, then the test would still pass since this is the default value.

bemerybmw (changes requested):

score/mw/com/impl/configuration/config_parser_test.cpp
KrishaDeshkool: wait, does this also apply to the enabled_ flag for methods?
it is also a optional now

score/mw/com/impl/configuration/config_parser_test.cpp
KrishaDeshkool: i added a new test case for that here LolaFieldBothFlagsExplicitlySetToFalse

score/mw/com/impl/configuration/config_parser_test.cpp
KrishaDeshkool: or we plan on reworking this when we do the split config for skeleton and proxy

bemerybmw (approved):

PR #670, opened 2026-07-08
Add alias self-assignment CopyableAtomic test

AAmbuj:
its showing warrning as error. cause the build fail. so, i fixed it.


castler (changes requested):
We should rather ignore the warning Here, as the Test was s self assignment specificaly

AAmbuj:
Hi @castler going this way is ok? 

#pragma clang diagnostic push
#pragma clang diagnostic ignored "-Wself-assign-overloaded"
    unit = unit;
#pragma clang diagnostic pop

or 

i remove target it self  
  features = COMPILER_WARNING_FEATURES + ["-score_communication_treat_warnings_as_errors"],

AAmbuj:
I modified target.

PR #673, opened 2026-07-08
Refactor method enabled non optional

KrishaDeshkool:


score/mw/com/impl/configuration/lola_method_instance_deployment.cpp
```diff
@@ -23,17 +23,17 @@ namespace
...
-                                                           std::optional<bool> enabled)
+LolaMethodInstanceDeployment::LolaMethodInstanceDeployment(std::optional<QueueSize> queue_size, bool enabled)
     : queue_size_{queue_size}, enabled_{enabled}
 {
 }
 
 LolaMethodInstanceDeployment::LolaMethodInstanceDeployment(
     const score::json::Object& serialized_lola_method_instance_deployment)
-    : queue_size_{std::nullopt}, enabled_{std::nullopt}
+    : queue_size_{std::nullopt}, enabled_{kIsMethodEnabledByDefault}
```
bemerybmw: We shouldn't set a default here. The default should be handled by the config parser. The args should be default initialised and we should use GetOptionalValueFromJson / GetValueFromJson (see score/mw/com/impl/configuration/lola_event_instance_deployment.cpp for example)

score/mw/com/impl/configuration/lola_method_instance_deployment.h
```diff
@@ -39,11 +39,10 @@ class LolaMethodInstanceDeployment
      * @brief Construct LolaMethodInstanceDeployment with optional queue size, because LolaMethodInstanceDeployment for
      * a consumer will have a value while one for a provider will not.
      * @param queue_size The maximum number of pending method requests that can be queued.
-     * @param enabled Optional flag to disable/enable the method. It is always filled on proxy side and always empty
+     * @param enabled  Flag to disable/enable the method. It is always filled on proxy side and defaulted to true
      *                on skeleton side.
```
bemerybmw: I wouldn't even mention the default value on skeleton side because that makes it seems like it's used. I would rather say something like "It is always filled on proxy side and not used on skeleton side".

score/mw/com/impl/configuration/lola_method_instance_deployment_test.cpp
```diff
@@ -91,7 +91,7 @@ TEST(LolaMethodInstanceDeploymentSerializationTest, CreateFromJsonWithoutQueueSi
     EXPECT_EQ(unit.queue_size_, std::nullopt);
 }
 
-TEST(LolaMethodInstanceDeploymentSerializationTest, CreateFromJsonWithoutEnabledLeavesEnabledUnset)
+TEST(LolaMethodInstanceDeploymentSerializationTest, CreateFromJsonWithoutEnabledDefaultsToTrue)
```
bemerybmw: It should actually terminate if the json doesn't contain the enabled flag.

bemerybmw (changes requested):

score/mw/com/impl/configuration/lola_method_instance_deployment.cpp
KrishaDeshkool: updated.

score/mw/com/impl/configuration/lola_method_instance_deployment_test.cpp
KrishaDeshkool: now it will terminate

bemerybmw (approved):

PR #676, opened 2026-07-08
Rework message passing threading model for QNX

kitsnet:
Move the background activity for client and server functionalities into separate threads to avoid possible deadlocks on cyclic client-server graphs.

### DIFF
--- a/score/message_passing/qnx_dispatch/dispatch_thread_runner.cpp
+++ b/score/message_passing/qnx_dispatch/dispatch_thread_runner.cpp
@@ -1,0 +1,181 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/message_passing/qnx_dispatch/dispatch_thread_runner.h"
+
+#include <iostream>
+
+namespace score::message_passing::detail
+{
+
+namespace
+{
+
+template <typename T>
+// Suppress "AUTOSAR C++14 A9-5-1" rule finding: "Unions shall not be used.".
+// coverity[autosar_cpp14_a9_5_1_violation: FALSE]. False positive: no any union in the following line. (Ticket-219101)
+// coverity[autosar_cpp14_m7_3_1_violation] false-positive: namespace-scoped function, not a global (Ticket-234468)
+auto ValueOrTerminate(const score::cpp::expected<T, score::os::Error> expected, const std::string_view error_text) -> T
+{
+    if (!expected.has_value())
+    {
+        std::cerr << "DispatchThreadRunner: " << error_text << ": "
+                  << expected.error().ToString()
+                  // Suppress AUTOSAR C++14 M8-4-4, rule finding: "A function identifier shall either be used to call
+                  // the function or it shall be preceded by &". Passing std::endl to std::cout object with the stream
+                  // operator follows the idiomatic way that both features in conjunction were designed in the C++
+                  // standard.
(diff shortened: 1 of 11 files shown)


score/message_passing/qnx_dispatch/qnx_dispatch_engine.h
```diff
@@ -244,23 +245,28 @@ class QnxDispatchEngine final : public ISharedResourceEngine
 
     bool IsOnCallbackThread() const noexcept override
     {
-        return std::this_thread::get_id() == thread_.get_id();
+        const std::thread::id id = std::this_thread::get_id();
+        return client_runner_.IsMyThread(id) || server_runner_.IsMyThread(id);
```
crimson11, line 249: IDK, when this func will be called. Isn't the "or" here watering down, what you want to exactly know?
I get, that `QnxDispatchEngine` instance might be used across ServerConnections and ClientConnections.
I.e. you don't need a separation into `IsOnClientCallbackThread` and `IsOnServerCallbackThread`?

score/message_passing/qnx_dispatch/dispatch_thread_runner.h
```diff
@@ -0,0 +1,116 @@
...
+    DispatchThreadRunner& operator=(DispatchThreadRunner&&) = delete;
+
+    bool IsMyThread(const std::thread::id id) const noexcept
+    {
+        return id == thread_.get_id();
+    }
+
+    static constexpr std::int32_t kQuitPulseCode = _PULSE_CODE_MINAVAIL;
+
+    template <typename Init>
```
crimson11, line 57: Init -> InitCallable?

score/message_passing/qnx_dispatch/dispatch_thread_runner.cpp
```diff
@@ -0,0 +1,181 @@
...
+                          "Unable to attach quit pulse code");
+}
+
+void DispatchThreadRunner::StartPostInit() noexcept
+{
+    // NOLINTNEXTLINE(score-banned-function) implementing FFI wrapper
+    context_pointer_ =
+        ValueOrTerminate(dispatch_.dispatch_context_alloc(dispatch_pointer_), "Unable to allocate context pointer");
+
+    // Normally, during the application lifecycle initialization, LifeCycleManager blocks the SIGTERM on the main
```
crimson11, line 95: So now we live in S-CORE ... does S-CORE have a `LifeCycleManager` infrastructure? IDK - did it copy the same approach from Adaptive AUTOSAR?

crimson11 (approved):

score/message_passing/qnx_dispatch/dispatch_thread_runner.cpp
kitsnet: It's something that I just moved from another place in the code, and something that SSP, and Logging in particular, relies on. The idea is that there is a dedicated thread waiting for the SIGTERM, and we don't want it to be delivered to another thread. We may want to discuss how we handle lifecycle in S-CORE, or we may provide API for the user to create and pre-init our internal threads.

score/message_passing/qnx_dispatch/dispatch_thread_runner.h
kitsnet: Will change.

score/message_passing/qnx_dispatch/qnx_dispatch_engine.h
kitsnet: The idea is to queue any blocking client operation onto a client thread when we are either in a client or in a server callback, to avoid deadlocks. But given that we may have multiple engines, probably the better solution would be to have a TLS variable that is true for any message passing callback thread. Do we have any restrictions on using thread local storage?

TimofteBogdan1900:
I independently hit this deadlock and can confirm this PR fixes it
Two QNX processes mutually proxy each other's services — each offers a skeleton and creates a proxy to the other, i.e. exactly the cyclic client-server graph this PR targets

score/message_passing/qnx_dispatch/qnx_dispatch_engine.h
crimson11: No. We are using TLS in mw::com::impl. Just grep for it.

Issue #681, opened 2026-07-09
Fix the Interface macro syntax


bharatGoswami8:
### What

Interface macro has comma (,) for one pattern and another pattern does not. remove the comma with backend capabilities

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [ ] Requirements / Architecture are not affected by this change?

bharatGoswami8:
Fix for this issue is merged.
For existing user, we have added backward capability, but they should try to adapt new syntax for interface macro.

Before:
```
interface!(
            interface Vehicle, {
                Id = "CustomVehicleInterface",
                left_tire: Event<Tire>,
               exhaust: Event<Exhaust>,
            }
    );
```

After:
```
interface!(
            interface Vehicle {
                Id = "CustomVehicleInterface",
                left_tire: Event<Tire>,
               exhaust: Event<Exhaust>,
            }
    );
```

PR #682, opened 2026-07-09
Rust::com Rust interface macro syntax fix 

bharatGoswami8:
https://github.com/eclipse-score/communication/issues/681

### DIFF
--- a/score/mw/com/impl/rust/com-api/com-api-concept/interface_macros.rs
+++ b/score/mw/com/impl/rust/com-api/com-api-concept/interface_macros.rs
@@ -59,7 +59,7 @@
 /// mod abc {
 ///     use com_api_concept::interface;
 ///     interface!(
-///         interface Vehicle, {
+///         interface Vehicle {
 ///             Id = "AbcInterface",
 ///             left_tire: Event<Tire>,
 ///             exhaust: Event<Exhaust>,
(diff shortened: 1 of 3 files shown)


score/mw/com/impl/rust/com-api/com-api-concept/interface_macros.rs
```diff
@@ -95,6 +95,18 @@ macro_rules! interface {
         $crate::interface_producer!($id, $($event_name, Event<$event_type>),+);
     };
 
+    // This is for backward compatibility for existing users with comma (,)
```
rpreddyhv, line 98: question: do we know if there existing users who are using this? When do we plan to eliminate this?

rpreddyhv (commented):
question to clarify

rpreddyhv (approved):
looks good

score/mw/com/impl/rust/com-api/com-api-concept/interface_macros.rs
bharatGoswami8: Yes, as i know FEO is using, any other is using that i am not sure.
We will announce in COM CFT about this changes.

darkwisebear (approved):
Thanks so much, it was hurting my eyes :)

castler (approved):

Issue #680, opened 2026-07-09
Bug: impl::InstanceIdentifier works on dangling configuration pointer


crimson11:
### Description

For the impl::InstanceIdentifier::Create() functionality, impl::InstanceIdentifier needs access to the configuration, which is also being used/setup by the impl::Runtime.
Currently impl::InstanceIdentifier::SetConfiguration() is being called with a pointer to a configuration, which gets "dangling" after the impl::Runtime::getInstanceInternal() get called for the 1st time. It moves the configuration to a an Runtime internal member, thus the access later by impl::InstanceIdentifier becomes a "use after move".

The solution:
impl::InstanceIdentifier shall instead get a pointer to the configuration AFTER the move happened. This is a stable location for the whole lifetime of the score::mw::com app/Runtime.

### Error Occurrence  Rate

Reproducible

### Classification

minor

### Affected Version

0.5

### Expected Closure Version

0.5

### Category

- [ ] Safety Related
- [ ] Security Related

crimson11:
Fixed with #683

PR #683, opened 2026-07-09
Fix dangling config pointer

crimson11:
Fixed a dangling pointer issue in
impl::InstanceIdentifier::SetConfiguration():
It was called with a pointer to a configuration, which got dangling after the Runtime Meyer-singleton had been initialized.

impl::InstanceIdentifier::SetConfiguration() gets now called with a pointer to the config held internally by impl::Runtime, which remains stable throughout the lifetime.

Issue: #680

### DIFF
--- a/score/mw/com/impl/instance_identifier.h
+++ b/score/mw/com/impl/instance_identifier.h
@@ -45,6 +45,15 @@ class InstanceIdentifier final
      * InstanceIdentifier::ToString()
      * \param serialized_format The serialized format to create the InstanceIdentifier from
      * \return The created InstanceIdentifier or an error
+     *
+     * \pre The score::mw::com::impl::Runtime singleton must have been created (then the configuration is "locked")
+     * before calling this API, i.e. Only then has a valid Configuration pointer been registered (by the Runtime
+     * constructor) into which the reconstructed deployments are added.
+     * Currently, there is no public API whose sole purpose is to initialize/create the Runtime singleton; in practice
+     * any prior public score::mw::com call that reaches Runtime::getInstance()
+     * (e.g. creating a Skeleton/Proxy, FindService, or - as a last resort - score::mw::com::runtime::ResolveInstanceIDs
+     * called with a dummy InstanceSpecifier) triggers this.
+     * If the configuration has not been set, kInvalidConfiguration is returned.
      */
     static score::Result<InstanceIdentifier> Create(std::string&& serialized_format) noexcept;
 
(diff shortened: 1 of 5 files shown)


score/mw/com/impl/instance_identifier.h
```diff
@@ -45,6 +45,15 @@ class InstanceIdentifier final
      * InstanceIdentifier::ToString()
      * \param serialized_format The serialized format to create the InstanceIdentifier from
      * \return The created InstanceIdentifier or an error
+     *
+     * \pre The score::mw::com Runtime configuration must have been locked before calling this API, i.e. the Runtime
```
bemerybmw: What do you mean by "locked" here?

score/mw/com/impl/runtime.h
```diff
@@ -166,6 +168,12 @@ class Runtime final : public IRuntime
     ///        their context
     ///        Should stay last attribute so that it is destructed first
     concurrency::LongRunningThreadsContainer long_running_threads_;
+
+    // Suppress "AUTOSAR C++14 A11-3-1", The rule states: "Friend declarations shall not be used".
+    // Design decision: The "*Attorney" class is a test-only helper, which accesses private members of this class to
+    // verify internal invariants (e.g. that the ctor wires InstanceIdentifier to the configuration_ member).
+    // coverity[autosar_cpp14_a11_3_1_violation]
+    friend class RuntimeAttorney;
```
bemerybmw: Usually we put friend classes at the start of the class declaration. 

```
class Runtime 
{
    friend class RuntimeAttorney;

public:
...
```

score/mw/com/impl/instance_identifier_test.cpp
```diff
@@ -306,6 +310,54 @@ TEST_F(InstanceIdentifierFixture, CanCreateFromSerializedObject)
                                             InstanceIdentifierView{identifier}.GetServiceTypeDeployment());
 }
 
+TEST_F(InstanceIdentifierFixture, CreateStoresDeploymentsIntoConfiguredConfiguration)
```
bemerybmw, line 313: Should have a test that checks: If the configuration has not been set, kInvalidConfiguration is returned.

bemerybmw (changes requested):

score/mw/com/impl/instance_identifier.h
crimson11: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/runtime.h#L140

This is a "notion" from our runtime impl - and means, that the singleton has been created - I will add this info.

score/mw/com/impl/instance_identifier_test.cpp
crimson11: Its already there -> CreatingWithoutInitializingConfigurationReturnsError in line [378](https://github.com/eclipse-score/communication/pull/683/changes#diff-4e319ad78b491459acd5afc518c4845888c6ce21286f369372465944535dd299R378)

bemerybmw (approved):

PR #692, opened 2026-07-13
Add heap-free mw::com skeleton and proxy examples

jorgecasal:
## Summary
This pull request adds examples that show how to use the `mw::com` provider (skeleton) and consumer (proxy) sides without allocating on the heap during the operational phase.

The examples are integrated into the `mw::com` tutorial as **chapter 13**, under `score/mw/com/doc/tutorial/chapter_13/`. They are registered in the tutorial build and documentation (`score/mw/com/doc/tutorial/BUILD` and `README.rst`), so they render as part of the existing tutorial.

## Structure
Chapter 13 contains four self-contained scenarios. Each scenario has its own sources, config, logging config, and build file so it can be built and run on its own:

- **event_send_receive**: provider sends events, consumer receives them.
- **event_field_update**: provider updates a field, consumer reads it.
- **start_find_service**: consumer finds a service, then uses it.
- **bidirectional_discovery**: two applications (`application_a` / `application_b`) that both provide and consume.

Each scenario splits the two sides into `provider` and `consumer` (the bidirectional scenario uses `application_a` / `application_b`). The shared service interface reuses chapter 7's tire-pressure interface, in `common/tire_pressure_service.h`.

A small helper in `heap_check/heap_check.h` verifies that no heap allocation happens once the operational phase begins. It overrides `operator new`/`delete` and aborts on any allocation after the init/operational boundary.

## Known limitation
The async receive handler cannot yet run without the heap, because the provider side allocates when a consumer registers a receive handler. This is a known framework limit (see the TODO in `scoped_event_receive_handler.h`). The examples use polling instead.


score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,385 @@
...
+
+
+In all previous chapters we used ``score::mw::com`` without placing any constraints on heap allocation. The
+middleware creates internal data structures, shared-memory handles, and subscription records during startup, and
+we never had to think about when those allocations happen.
+
+In safety-qualified (ASIL-B) and Freedom-From-Interference (FFI) applications, this changes. Once the
+application enters its **operational phase**, every uncontrolled heap allocation is a potential violation: it
+can interfere with pre-allocated resources, introduce unpredictable latency, and is prohibited by the relevant
+safety standards. The application must therefore separate its lifecycle into two distinct phases:
```
castler: > and is prohibited by the relevant safety standards.

I am not to sure about this, this strongly depends on the system build and safety requirements. Thus, I would not formulate that rough, but state that this can be prohibited.

NEOatNHNG:
Is it also possible to create a proxy after forbid_heap() if I pass an allocator? That way we could handle the case where the subscriber is started before the publisher, right?
Then the subscriber would create the memory pool, StartFindService() and then forbid_heap(). Afterwards when the publisher becomes available when the find service callback is called it would create the proxy passing an allocator using the memory pool created earlier.

crimson11:
> Is it also possible to create a proxy after forbid_heap() if I pass an allocator? That way we could handle the case where the subscriber is started before the publisher, right? Then the subscriber would create the memory pool, StartFindService() and then forbid_heap(). Afterwards when the publisher becomes available when the find service callback is called it would create the proxy passing an allocator using the memory pool created earlier.

I guess we need a deeper look! In the past we had some memory allocation taking place in the message-passing layer of LoLa during "runtime". I.e. our layer, which USES message-passing. NOT in message-passing itself. 
During "runtime" means even after proxy/skeleton instances had been all set-up.
This got much better in the meantime! But there are most likely still cases, where memory allocation takes place outside of skeleton/proxy initialization!
Imho this needs to be analyzed/identified.
Then (this is, what you are hinting at):

- either we can get rid of this allocation by pre-allocating a sensible/needed amountz of memory (deducing from configs or maybe extend config when needed to give the right hints)
- or by overloading potentially allocating public APIs with the possibility to hand over custom-allocators (working on pre-allocated pools.

score/mw/com/doc/tutorial/chapter_12/common/service_interface.h
```diff
@@ -0,0 +1,52 @@
...
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_TUTORIAL_CHAPTER_12_COMMON_SERVICE_INTERFACE_H
```
crimson11: Do we really need a new service-interface? We should restrict ourself in the tutorials.
If you can show stuff with an existing service interface, then "re-use" it. (It is a pattern in the whole tutorial, that we COPY elements from previous chapters opposed to cross-reference or introduce a "common_chapter" location! This shall be kept! Jumping/referencing around makes it hard for the user, who maybe picks just one chapter!)

This is the reason, that we 1st try to use "hello_world_service interface" in each chapter 1st. If we really need to show a new service-interface specific concept, then we extend hello_service or introduce a new interface.

Bottomline:
Is it possible for you to use the tire_pressure service.
Makes it more easy for the user.

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
+The table below shows which ``score::mw::com`` calls touch the heap and which do not. All
+allocating calls must be completed **before** ``heap_check::forbid_heap()`` is called. Calls
+marked "heap-free" are safe to use in the operational loop.
+
++-------------------------------------------+-------------------------+-----------------------------------+
+| API call                                  | Heap behaviour          | Basis                             |
++===========================================+=========================+===================================+
+| ``SensorSkeleton::Create()``              | Allocates               | Binding setup, shared_ptr chains  |
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``sk.calibration_status.Update()``        | Allocates (init only)   | ``make_unique<FieldType>`` before |
```
crimson11: see my annotation in code sk -> skeleton

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
+| (before ``OfferService()``)               |                         | ``OfferService()``                |
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``sk.OfferService()``                     | Allocates               | SHM setup, PrepareOffer checks    |
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``sk.reading.Allocate()``                 | **Heap-free**           | Returns ``SampleAllocateePtr<T>`` |
+| (after ``OfferService()``)                |                         | from LoLa SHM ring buffer         |
```
crimson11: why this `after OfferService()` notion here? A call to allocate before OfferService would fail. I guess you should simply leave out this notion in brackets?

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``sk.reading.Send()``                     | **Heap-free**           | Writes directly to SHM            |
```
crimson11: FYI: In case tracing is enabled, unfortunately there currently happens heap-allocation :(
(it happens in TracingRuntime::ConvertToTracingServiceInstanceElement(), where a std::string gets temporary created for a std::unordered_map lookup ... we have an internal ticket for this issue ...)

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
+| ``proxy.reading.GetSubscriptionState()``  | **Heap-free**           | Atomic read                       |
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``Notification::waitForWithAbort()``      | **Heap-free**           | ``std::unique_lock`` on           |
+|                                           |                         | member mutex                      |
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``Notification::notify()`` / ``reset()``  | **Heap-free**           | ``std::lock_guard`` on            |
+|                                           |                         | member mutex                      |
++-------------------------------------------+-------------------------+-----------------------------------+
+| ``stop_source::get_token()``              | **Heap-free**           | Atomic shared_ptr ref-count copy  |
++-------------------------------------------+-------------------------+-----------------------------------+
```
crimson11: I do like the 3rd column in the sense, that I can double-check with my memory ;) ... but do you think this is "helpful" for a user/tutorial reader? I see some high likelihood, that this 3rd column info can quickly be outdated (as its an impl. detail)

score/mw/com/doc/tutorial/chapter_12/event_send_receive/provider.cpp
```diff
@@ -0,0 +1,88 @@
...
+
+    // ---- INIT PHASE ----
+
+    score::Result<score::mw::com::InstanceSpecifier> specifier_result =
+        score::mw::com::InstanceSpecifier::Create(std::string{"/sensor/event_send_receive/SensorInterface"});
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(specifier_result.has_value(), "Failed to create InstanceSpecifier!");
+
+    score::Result<sensor::SensorSkeleton> skeleton_result = sensor::SensorSkeleton::Create(specifier_result.value());
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(skeleton_result.has_value(), "Failed to create SensorSkeleton!");
+    sensor::SensorSkeleton& sk = skeleton_result.value();
```
crimson11: we "hate" these abbreviations :) - what does the score coding guide say? Please rename sk -> skeleton

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
+
+Which API calls allocate
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+The table below shows which ``score::mw::com`` calls touch the heap and which do not. All
+allocating calls must be completed **before** ``heap_check::forbid_heap()`` is called. Calls
+marked "heap-free" are safe to use in the operational loop.
+
++-------------------------------------------+-------------------------+-----------------------------------+
```
crimson11: I guess this table covers the API usage aggregated over all .cpp files in this chapter? Maybe note this.

score/mw/com/doc/tutorial/chapter_12/start_find_service/consumer.cpp
```diff
@@ -0,0 +1,123 @@
...
+
+    score::Result<score::mw::com::InstanceSpecifier> specifier_result =
+        score::mw::com::InstanceSpecifier::Create(std::string{"/sensor/start_find_service/SensorInterface"});
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(specifier_result.has_value(), "Failed to create InstanceSpecifier!");
+
+    std::optional<sensor::SensorProxy> proxy_opt{};
+    score::concurrency::Notification proxy_ready{};
+    score::cpp::stop_source stop_source{};
+
+    score::Result<score::mw::com::FindServiceHandle> fshandle_result = sensor::SensorProxy::StartFindService(
```
crimson11: `fshandle_result`  -> `find_service_handle_result`

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
+.. literalinclude:: event_field_update/provider.cpp
+   :language: cpp
+   :lines: 52-95
+   :caption: event_field_update/provider.cpp — operational phase with event and field update
+
+Async service discovery with StartFindService
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+
+The ``FindService()`` polling pattern used in ``event_send_receive/consumer.cpp`` has a problem
```
crimson11: Part of this sections describe content, which is presented in chapter 3.
If you think, that chapter 3 misses important stuff, shouldn't we then rather estend chapter 3 and here only discuss allocation related questions!?

score/mw/com/doc/tutorial/chapter_12/README.rst
```diff
@@ -0,0 +1,458 @@
...
+
+Two constraints together determine the required init-phase ordering:
+
+- ``Proxy::Create()`` allocates heap memory and must therefore complete **before**
+  ``forbid_heap()`` is called.
+- ``Proxy::Create()`` requires the provider to have already called ``OfferService()``, because
+  ``OfferService()`` is the call that creates the shared-memory region the proxy attaches to.
+  A proxy cannot be created against a service that has not yet been offered.
+
+Each application therefore calls ``OfferService()`` first, making its shared-memory region
```
crimson11: You are focusing here on the shm-objects (only) ... the step before that is visibility in the service-discovery ... so this is the even more natural view on the topic :) - if one side hasn't called Offer() nothing is registered in the SD, thus a FindService can't find it ...

score/mw/com/doc/tutorial/chapter_12/common/service_interface.h
```diff
@@ -0,0 +1,52 @@
...
+#define SCORE_MW_COM_TUTORIAL_CHAPTER_12_COMMON_SERVICE_INTERFACE_H
+
+// Service interface for the heap-free examples. Instantiate with AsSkeleton<SensorInterface> or
+// AsProxy<SensorInterface>.
+
+#include "score/mw/com/types.h"
+
+#include <cstdint>
+
+namespace sensor
```
crimson11: I guess we should be consistent with the tutorial namespacing .... `namespace score::mw::com::tutorial`

score/mw/com/doc/tutorial/chapter_13/event_field_update/consumer.cpp
```diff
@@ -0,0 +1,124 @@
...
+constexpr std::chrono::milliseconds kFindRetryDelay{50};
+constexpr std::uint32_t kNumIterations{20U};
+constexpr std::chrono::milliseconds kCycleTime{50};
+constexpr std::size_t kMaxSamples{16U};
+
+}  // namespace
+
+int main()
+{
+    score::mw::log::LogInfo("PxEf") << "Starting event_field_update/consumer";
```
crimson11, line 40: good! At least you are consistent with the logging context!! I was lazy and used default in all my chapters :( need to fix it!

score/mw/com/doc/tutorial/chapter_12/event_field_update/consumer.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+    // ---- OPERATIONAL PHASE (see README.rst for heap behavior of each API) ----
+    score::mw::log::LogInfo("PxEf") << "Entering operational phase (heap forbidden)";
+    heap_check::forbid_heap();
+
+    for (std::uint32_t i = 0U; i < kNumIterations; ++i)
+    {
+        score::Result<std::size_t> reading_result = proxy.reading.GetNewSamples(
+            [](score::mw::com::SamplePtr<sensor::SensorReading> sample) noexcept {
+                static_cast<void>(sample->sequence);
```
crimson11: these cast shall simulate a sample access without triggering compiler warnings?

score/mw/com/doc/tutorial/chapter_12/event_field_update/provider.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+    // ---- INIT PHASE ----
+
+    score::Result<score::mw::com::InstanceSpecifier> specifier_result =
+        score::mw::com::InstanceSpecifier::Create(std::string{"/sensor/event_field_update/SensorInterface"});
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(specifier_result.has_value(), "Failed to create InstanceSpecifier!");
+
+    score::Result<sensor::SensorSkeleton> skeleton_result = sensor::SensorSkeleton::Create(specifier_result.value());
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(skeleton_result.has_value(), "Failed to create SensorSkeleton!");
+    sensor::SensorSkeleton& sk = skeleton_result.value();
```
crimson11: sk -> skeleton

score/mw/com/doc/tutorial/chapter_12/event_send_receive/consumer.cpp
```diff
@@ -0,0 +1,153 @@
...
+    // ---- CLEANUP ----
+    heap_check::allow_heap();
+    score::mw::log::LogInfo("PxEs") << "Operational phase complete (" << kNumIterations << " iterations)";
+
+    proxy.reading.Unsubscribe();
+
+    score::mw::log::LogInfo("PxEs") << "Completed successfully";
+    return 0;
+
+    // ---- ASYNC RECEIVE HANDLER (NOT HEAP-FREE — kept for reference) ----
```
crimson11: Why this as a big comment and not as a stand-alone `consumer_receive_handler.cpp`?

crimson11 (commented):
KUDOs to you! I really really appreciate your effort! Nice reading, nice examples.
This is an advanced chapter - as the topic of heap-alloc and when it takes place and what patterns to use to "manage" it are most likely things people learn/discover after the basic API usage.

Because of that: May I ask you, to renumber your chapter from 12 -> 13.
Because of this: https://github.com/eclipse-score/communication/pull/724

I.e. I added a tutorial chapter about how-to-unit-test. I feel, that this is more "basic" as your topic and should therefore come before ...

Final question:
When we dig deeper into these heap allocations (try to get rid of/allow users to provide custom allocators - see also my previous comment: https://github.com/eclipse-score/communication/pull/692#issuecomment-5004713638) ... can we count on you? Are you willing to support/take over parts?

score/mw/com/doc/tutorial/chapter_13/heap_check/heap_check.h
```diff
@@ -0,0 +1,93 @@
...
+// Include this file in exactly ONE translation unit per binary (main.cpp only).
+//
+// Replaces the global operator new with a version that aborts if the calling
+// thread allocates after forbid_heap() has been called on that thread.
+// Check is per-thread: library background threads may still allocate freely.
+
+#include <cstdlib>
+#include <new>
+
+namespace heap_check
```
castler, line 25: I am not to fluent in this topic - but my personal understanding was that there are way more operators that would need to be overwritten.

See:

Image

Most of the information came from a talk one of our team members gave @limdor (https://limdor.github.io/talks/ "Dynamic memory allocation challenges in safety critical systems
")

Can you maybe elaborate a little bit more on the topic, why you think that these overwrites are "enough"?

castler (commented):
Thanks again - first for rebasing to quickly on the tutorial structure and second for the topic!

score/mw/com/doc/tutorial/chapter_12/common/service_interface.h
jorgecasal: I dropped the custom `SensorInterface` and now reuse the existing tire pressure service interface from chapter 7, copied into this chapter following the same self contained pattern as the other chapters. The chapter needs both an event and a subscribable field to show event send and receive together with field update, so I extended the interface with a `tire_pressure_update` event. The namespace is now `score::mw::com::tutorial` to match the rest of the tutorial.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: Removed.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: Renamed sk to skeleton here and in the other examples and in the table.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: Thanks for the heads up. I added a note that Send allocates on the heap when tracing is enabled, and pointed at the temporary string used for the map lookup in the tracing runtime as the cause.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: I removed the 3rd column and kept the table to the API call and its heap behaviour.

score/mw/com/doc/tutorial/chapter_12/common/service_interface.h
jorgecasal: Done. The interface now lives under namespace `score::mw::com::tutorial`, consistent with the rest of the tutorial.

score/mw/com/doc/tutorial/chapter_13/event_field_update/consumer.cpp
jorgecasal: Thanks!

score/mw/com/doc/tutorial/chapter_12/event_field_update/consumer.cpp
jorgecasal: Yes, the static cast lines stand in for reading the sample fields without triggering unused variable warnings. They represent whatever real processing an application would do at that point. Not sure, if that would be OK from your point of view, though.

score/mw/com/doc/tutorial/chapter_12/event_field_update/provider.cpp
jorgecasal: Done.

score/mw/com/doc/tutorial/chapter_12/event_send_receive/consumer.cpp
jorgecasal: I moved it out of the comment into a standalone consumer_receive_handler binary with its own BUILD target, and reference it from the chapter.

score/mw/com/doc/tutorial/chapter_12/start_find_service/consumer.cpp
jorgecasal: Renamed to find_service_handle_result.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: Added a note that the table aggregates the API usage across all example files in the chapter.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: I moved the general FindService polling and deadlock explanation into chapter 3 and kept only the allocation specific discussion here, with a pointer to chapter 3 for the general concept.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: you're right, I reframed this to lead with service discovery visibility. If a provider has not called OfferService it is not registered in service discovery, so the other side cannot find it. The shared memory region follows from that. Can you please double check?

jorgecasal:
> KUDOs to you! I really really appreciate your effort! Nice reading, nice examples. This is an advanced chapter - as the topic of heap-alloc and when it takes place and what patterns to use to "manage" it are most likely things people learn/discover after the basic API usage.
> 
> Because of that: May I ask you, to renumber your chapter from 12 -> 13. Because of this: #724
> 
> I.e. I added a tutorial chapter about how-to-unit-test. I feel, that this is more "basic" as your topic and should therefore come before ...
> 
> Final question: When we dig deeper into these heap allocations (try to get rid of/allow users to provide custom allocators - see also my previous comment: [#692 (comment)](https://github.com/eclipse-score/communication/pull/692#issuecomment-5004713638)) ... can we count on you? Are you willing to support/take over parts?

Thanks, yes you can count on me. I will take on the heap allocation follow-up.

score/mw/com/doc/tutorial/chapter_12/README.rst
jorgecasal: reworded, it now says heap allocation can be prohibited depending on the system's safety requirements. Can you please double check?

score/mw/com/doc/tutorial/chapter_13/heap_check/heap_check.h
jorgecasal: the basic new/new[] pair wasn't enough, was just a starting point to prove/test the heap check mechanism. I've added the remaining replaceable global allocation functions.

score/mw/com/doc/tutorial/chapter_13/README.rst
```diff
@@ -0,0 +1,566 @@
...
+   ``TracingRuntime::ConvertToTracingServiceInstanceElement()`` for an ``unordered_map`` lookup.
+   This is a known limitation tracked in an internal ticket. If tracing must be active during
+   the heap-free operational phase, treat ``Send()`` as allocating.
+
+.. note::
+
+   ``SetReceiveHandler`` cannot be used in a fully heap-free provider-consumer pair today. When
+   the proxy registers a receive handler, the LoLa runtime's skeleton-side background thread
+   creates a ``std::make_shared<MoveOnlyScopedFunction<...>>`` to record the registration. If
+   the provider has already called ``forbid_heap()``, that allocation triggers the abort. A TODO
```
crimson11: I think this is incorrect. The `MoveOnlyScopedFunction` get created on the consumer/proxy side. Yes, it requires dynamic mem-alloc, but in this case, it is not an issue, because `SetReceiveHandler` will be called once in the init-phase at the consumer side ....

But we DO have an issue at the provider side also! The provider side (skeleton) has no clue, when a consumer calls `SetReceiveHandler`. It might do, when the provider is already in its run-phase. There we do also/indeed have an issue at the provider side. But not with creation of an `MoveOnlyScopedFunction` (this isn't done on the provider side in  this context), but because of this: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/messaging/message_passing_service_instance.cpp#L484

I.e. in the messaging-layer, we are using STL containers with default-allocators, where we are inserting elements during runtime. In this case: Consumer calls `SetReceiveHandler`, we might end up inserting the consumers node-id on the provider-side eventually leading to allocation.

score/mw/com/doc/tutorial/chapter_13/README.rst
```diff
@@ -0,0 +1,566 @@
...
+   The examples run at ASIL-B configuration (``asil-level: "B"`` in every config file) but are
+   not themselves ASIL-qualified artifacts. They are blueprints. The section
+   :ref:`chapter_13_asil_guidance` at the end of this chapter lists what you need to address
+   before using these patterns in production ASIL-B code.
+
+Files/artifacts used
+~~~~~~~~~~~~~~~~~~~~~
+
+
+The ``bazel`` project for this chapter is located in ``score/mw/com/doc/tutorial/chapter_13``
```
crimson11: Maybe use a real link?
`score/mw/com/doc/tutorial/chapter_13 <https://github.com/eclipse-score/communication/blob/main/score/mw/com/doc/tutorial/chapter_13/>`__

crimson11 (changes requested):
I'm basically happy! Just look at my comment regarding `MoveOnlyScopedFunction` at eventually correct/fix.

score/mw/com/doc/tutorial/chapter_13/README.rst
jorgecasal: thanks. I have corrected the note so it names the actual reason as you described in your comment above. The runtime allocation happens on the provider side, where the messaging layer inserts the consumer node identifier into STL containers that use the default allocator. The conclusion remains unchanged, `SetReceiveHandler` cannot be part of a complete heap-free provider/consumer pair as of now, so the examples use `GetNewSamples()` polling instead. Can you please review the updated note and let me know if it's clear now?

score/mw/com/doc/tutorial/chapter_13/README.rst
jorgecasal: done, the chapter_13 path is now an external link

score/mw/com/doc/tutorial/chapter_13/heap_check/heap_check.h
```diff
@@ -0,0 +1,203 @@
...
+    {
+        std::abort();
+    }
+    void* ptr = std::malloc(sz);  // NOLINT(cppcoreguidelines-no-malloc): malloc is correct inside operator new
+    if (ptr == nullptr)
+    {
+        throw std::bad_alloc{};
+    }
+    return ptr;
+}
```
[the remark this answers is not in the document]
jorgecasal: `sz==0` is structurally unreachable in these examples. This is a tutorial heap detector, not a production operator new replacement, so the risk flagged can't be fired here.

score/mw/com/doc/tutorial/chapter_13/heap_check/heap_check.h
```diff
@@ -0,0 +1,203 @@
...
+    // aligned_alloc requires size to be a multiple of alignment; round up.
+    std::size_t align = static_cast<std::size_t>(al);
+    std::size_t rounded = (sz + align - 1U) & ~(align - 1U);
+    void* ptr = std::aligned_alloc(align, rounded);  // NOLINT(cppcoreguidelines-no-malloc)
+    if (ptr == nullptr)
+    {
+        throw std::bad_alloc{};
+    }
+    return ptr;
+}
```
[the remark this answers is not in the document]
jorgecasal: `sz==0` is structurally unreachable in these examples. This is a tutorial heap detector, not a production operator new replacement, so the risk flagged can't be fired here.

score/mw/com/doc/tutorial/chapter_13/heap_check/heap_check.h
```diff
@@ -0,0 +1,203 @@
...
+void* operator new(std::size_t sz, std::align_val_t al, const std::nothrow_t&) noexcept
+{
+    if (heap_check::tl_heap_forbidden)
+    {
+        std::abort();
+    }
+    std::size_t align = static_cast<std::size_t>(al);
+    std::size_t rounded = (sz + align - 1U) & ~(align - 1U);
+    return std::aligned_alloc(align, rounded);  // NOLINT(cppcoreguidelines-no-malloc)
+}
```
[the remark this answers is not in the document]
jorgecasal: `sz==0` is structurally unreachable in these examples. This is a tutorial heap detector, not a production operator new replacement, so the risk flagged can't be fired here.

score/mw/com/doc/tutorial/chapter_13/heap_check/heap_check.h
```diff
@@ -0,0 +1,203 @@
...
+
+// NOLINTNEXTLINE(misc-new-delete-overloads): nothrow operator new, replaceable global
+void* operator new(std::size_t sz, const std::nothrow_t&) noexcept
+{
+    if (heap_check::tl_heap_forbidden)
+    {
+        std::abort();
+    }
+    return std::malloc(sz);  // NOLINT(cppcoreguidelines-no-malloc)
+}
```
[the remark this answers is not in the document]
jorgecasal: `sz==0` is structurally unreachable in these examples. This is a tutorial heap detector, not a production operator new replacement, so the risk flagged can't be fired here.

score/mw/com/doc/tutorial/chapter_13/start_find_service/logging.json
```diff
@@ -0,0 +1,7 @@
+{
+    "appId": "PXSF",
+    "appDesc": "start_find_service_proxy_example",
```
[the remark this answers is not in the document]
jorgecasal: established project convention, `event_send_receive` and `event_field_update` share one logging.json across provider+consumer identically.

score/mw/com/doc/tutorial/chapter_13/event_send_receive/consumer_receive_handler.cpp
```diff
@@ -0,0 +1,120 @@
...
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(subscribe_result.has_value(),
+                                                "Failed to subscribe to tire_pressure_update!");
+
+    score::mw::log::LogInfo("PxEsH") << "Entering receive loop";
+
+    for (std::uint32_t i = 0U; i < kNumIterations; ++i)
+    {
+        event_received.waitWithAbort(stop_source.get_token());
+        event_received.reset();
+
```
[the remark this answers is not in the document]
jorgecasal: indeed the risk exists, but this binary is explicitly documented as the not-heap-free async demo. It's an API demonstration, not robustness guidance. Keeping as it is.

score/mw/com/doc/tutorial/chapter_13/start_find_service/consumer.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    std::optional<TirePressureProxy> proxy_opt{};
+    score::concurrency::Notification proxy_ready{};
+    score::cpp::stop_source stop_source{};
+
+    score::Result<score::mw::com::FindServiceHandle> find_service_handle_result = TirePressureProxy::StartFindService(
+        [&proxy_opt, &proxy_ready](score::mw::com::ServiceHandleContainer<TirePressureProxy::HandleType> handles,
+                                   score::mw::com::FindServiceHandle /*fsh*/) {
+            if (handles.empty())
+            {
```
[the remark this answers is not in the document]
crimson11: Imho this is a wrong finding!
The documentation of `StopFindService()` is clear:
https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/proxy_base.h#L106

I.e. after `StopFindService()` returned, there is a "guarantee", that the user provided handler won't be called anymore.
Could it be the case, that Copilot encountered a bug?
I don't think so. I checked our implementation. The core mechanism, where we avoid, that user-provided handler get called AFTER `StopFindService` returned is here:
https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/service_discovery/client/service_discovery_client.cpp#L274https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/bindings/lola/service_discovery/client/service_discovery_client.cpp#L274

I.e. calling user-provided handlers and stopping-search, both sync on the `worker_mutex_`! I.e. if a a call to the user-provided handler is already "underway", the `StopFindService` call would block until the user-provided handler call has finished ....

To me - this is a false-positive reporting by copilot.

score/mw/com/doc/tutorial/chapter_13/bidirectional_discovery/application_a.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    std::optional<TirePressureProxy> remote_proxy{};
+    score::concurrency::Notification proxy_ready{};
+    score::cpp::stop_source stop_source{};
+
+    score::Result<score::mw::com::FindServiceHandle> find_service_handle_result = TirePressureProxy::StartFindService(
+        [&remote_proxy, &proxy_ready](score::mw::com::ServiceHandleContainer<TirePressureProxy::HandleType> handles,
+                                      score::mw::com::FindServiceHandle /*fsh*/) {
+            if (handles.empty())
+            {
```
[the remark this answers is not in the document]
crimson11: see my comment above. Imho false-positive.

score/mw/com/doc/tutorial/chapter_13/bidirectional_discovery/application_b.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    std::optional<TirePressureProxy> remote_proxy{};
+    score::concurrency::Notification proxy_ready{};
+    score::cpp::stop_source stop_source{};
+
+    score::Result<score::mw::com::FindServiceHandle> find_service_handle_result = TirePressureProxy::StartFindService(
+        [&remote_proxy, &proxy_ready](score::mw::com::ServiceHandleContainer<TirePressureProxy::HandleType> handles,
+                                      score::mw::com::FindServiceHandle /*fsh*/) {
+            if (handles.empty())
+            {
```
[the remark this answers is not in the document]
crimson11: see my comment above. Imho false-positive.

crimson11:
@jorgecasal 
I'm happy with the changes you did, regarding my last review!
I now checked the copilot "findings". I guess you coud fix the missing <string> includes .... the other "findings" of copilot I commented as false-positive. See my explanation! TBH: I do NOT expect, that copilot did a deeper analysis and found a bug within our SD synchronization, which assures no user-handler being called AFTER `StopFindService` returned! We did implement/test this thoroughly back then ... I'm too lazy to throw AI on it to do a deep analysis, whether there is an unknown bug. Feel free to do so, if you want ;)

Bottom line: After you fixed the string includes, I would queue the PR for getting merged.

crimson11 (dismissed):

score/mw/com/doc/tutorial/chapter_13/start_find_service/consumer.cpp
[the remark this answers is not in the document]
jorgecasal: @crimson11 
thanks, for signalling this Copilot finding as false-positive, it saved me quite some time.

crimson11 (approved):

crimson11:
@jorgecasal 
After your PR has been merged, we have to (currently) manually trigger the HTML-generation out of the RST/sphinx-needs docs. I will take care. Once it has run, you should see your chapter_13 here: https://eclipse-score.github.io/communication/latest/tutorial/README.html

PR #698, opened 2026-07-13
coverage: Provide baseline support for Linux and QNX

castler:
With this commit we provide baseline support for Linux and QNX coverage.
We can remove manual filtering based on patterns and ensure that only the files (but also all files) from the respective public libraries are covered.


quality/coverage/coverage_scope.bzl
```diff
@@ -0,0 +1,186 @@
...
+        "actual": attr.label(
+            mandatory = True,
+            cfg = coverage_transition,  # Applying the transition here
+        ),
+        # Mandatory attribute needed when a rule uses a transition
+        "_allowlist_function_transition": attr.label(
+            default = "@bazel_tools//tools/allowlists/function_transition_allowlist",
+        ),
+    },
+)
```
LittleHuba, line 164: Why do we need the transition?

quality/coverage/generate_coverage_html.sh
```diff
@@ -117,12 +117,68 @@ else
   # Copy the raw LCOV data for later archiving.
   cp "${COVERAGE_REPORT}" "${TMPDIR_EXTRACT}/lcov.dat"
 
+  echo "Building collected coverage allowlist for QNX..."
+  # coverage_scope depends on Linux-only dependable element indices.
+  # Build it in the host/default config and reuse the workspace-relative allowlist for QNX filtering.
+  bazel build --config=qnx //quality/coverage:coverage_scope --output_groups=allowlist
+  ALLOWLIST_FILE="$(bazel info bazel-bin)/quality/coverage/coverage_scope_allowlist.txt"
```
LittleHuba, line 124: This is not very portable. Better would be if this target is provided as input artifact via bazel.

quality/coverage/generate_coverage_html.sh
```diff
@@ -117,12 +117,68 @@ else
...
+        baseline_input="${lcov_input%coverage.dat}baseline_coverage.dat"
+        if [[ -f "${baseline_input}" ]]; then
+          printf "\n" >> "${BASELINE_LCOV}"
+          cat "${baseline_input}" >> "${BASELINE_LCOV}"
+          printf "\n" >> "${BASELINE_LCOV}"
+          BASELINE_COUNT=$((BASELINE_COUNT + 1))
+        fi
+      fi
+    done < <(sort -u "${LCOV_INPUT_LIST}")
+  fi
```
LittleHuba, line 162: This section would benefit from a comment that explains what its goal is.

quality/coverage/lcov_to_html.py
```diff
@@ -123,6 +123,153 @@ def parse_lcov(lcov_path: Path) -> List[Dict[str, Any]]:
...
+    return entries
+
+
+def _normalize_entry_path(filename: str, source_root: str) -> str:
+    """Normalize LCOV paths to workspace-relative paths for comparison/output."""
+    normalized_source_root = source_root.rstrip("/")
+    if normalized_source_root and filename.startswith(normalized_source_root + "/"):
+        return filename[len(normalized_source_root) + 1:]
+    if filename.startswith("/proc/self/cwd/"):
+        return filename[len("/proc/self/cwd/"):]
```
LittleHuba: Better not duplicate this string.

quality/coverage/lcov_to_html.py
```diff
@@ -123,6 +123,153 @@ def parse_lcov(lcov_path: Path) -> List[Dict[str, Any]]:
...
+
+
+def _filter_file_entries_by_allowlist(
+    file_entries: Sequence[Dict[str, Any]],
+    allowlist: Set[str],
+    source_root: str,
+) -> List[Dict[str, Any]]:
+    """Keep only LCOV entries whose normalized path is present in the allowlist."""
+    if not allowlist:
+        return list(file_entries)
```
LittleHuba, line 174: This does not fulfill the contract of the function. Either extend the description of the function or change this to return an empty list.

quality/coverage/lcov_to_html.py
```diff
@@ -123,6 +123,153 @@ def parse_lcov(lcov_path: Path) -> List[Dict[str, Any]]:
...
+    return [merged_by_file[filename] for filename in order]
+
+
+def _merge_single_file_entry(target: Dict[str, Any], incoming: Dict[str, Any]) -> None:
+    """Merge one file entry into another in-place."""
+    target_lines = {line["line_number"]: dict(line) for line in target["lines"]}
+    for incoming_line in incoming.get("lines", []):
+        line_number = incoming_line["line_number"]
+        incoming_copy = dict(incoming_line)
+        incoming_copy["branches"] = [dict(branch) for branch in incoming_line.get("branches", [])]
```
LittleHuba: Code smell: Based on access-pattern in line 214 `incoming_line` is a dict. Based on this, the assignment in line 215 results in a shallow copy. This means, in line 216 you do not only modify `incoming_copy` but also `incoming_line`.

Possible options:
1. If the modification of `incoming_line` is non-destructive, just drop the shallow copy
2. If the modification of `incoming_line` is destructive use `copy.deepcopy()`.

quality/coverage/lcov_to_html.py
```diff
@@ -123,6 +123,153 @@ def parse_lcov(lcov_path: Path) -> List[Dict[str, Any]]:
...
+
+        existing_func = target_functions[name]
+        if existing_func.get("lineno", 0) == 0 and incoming_lineno != 0:
+            existing_func["lineno"] = incoming_lineno
+        existing_func["execution_count"] = max(
+            existing_func.get("execution_count", 0),
+            incoming_execution_count,
+        )
+
+    target["functions"] = sorted(target_functions.values(), key=lambda func: func["name"])
```
LittleHuba: This function would benefit from splitting into dedicated subfunctions for line and function coverage as well.

quality/coverage/lcov_to_html.py
```diff
@@ -123,6 +123,153 @@ def parse_lcov(lcov_path: Path) -> List[Dict[str, Any]]:
...
+                "functions": [],
+            }
+            order.append(filename)
+        _merge_single_file_entry(merged_by_file[filename], entry)
+
+    return [merged_by_file[filename] for filename in order]
+
+
+def _merge_single_file_entry(target: Dict[str, Any], incoming: Dict[str, Any]) -> None:
+    """Merge one file entry into another in-place."""
```
LittleHuba, line 213: I think it would help to describe the merge strategy here, since it is not straightforward.
Instead of actually merging the counts, you perform a selection of the maximum value between the two files.
This results in completely different counts than a user would expect.

Should this be a concious choice, I would prefer an explaination here.
If this was not concious, I would prefer to actually add the counts (considering overflows) so that the numbers are as a user would expect them.

MODULE.bazel
```diff
@@ -301,7 +301,7 @@ llvm.toolchain(
         # This is the agreed way to ensure linking for targets using std::atomic operations.
         "-latomic",
     ]},
-    llvm_version = "19.1.7",
+    llvm_version = "22.1.7",
```
LittleHuba, line 304: Why do we need this bump?

quality/coverage/coverage_scope.bzl
castler: Because by default this will otherwise not create coverage information. See line 136

quality/coverage/generate_coverage_html.sh
castler: It is actually as portable as possible - only better approach is to not do it in this preprocessing step. I have local changes for this refactoring, but this is right now not performant enough.

MODULE.bazel
castler: Because the previous llvm-cov versions were not able to create empty baselines.

quality/coverage/lcov_to_html.py
```diff
@@ -123,6 +125,187 @@ def parse_lcov(lcov_path: Path) -> List[Dict[str, Any]]:
...
+
+def _merge_line_entries(
+    existing_lines: Sequence[Dict[str, Any]],
+    incoming_lines: Sequence[Dict[str, Any]],
+) -> List[Dict[str, Any]]:
+    """Merge line coverage and branch data for a single source file.
+
+    We keep the maximum execution count per line/branch identity instead of summing:
+    the same line can appear in multiple LCOV inputs for identical instrumentation,
+    and summing those overlapping records inflates execution counts.
```
LittleHuba, line 239: Talked offline: Summing and capping at maximum

LittleHuba: Found out that this number will never be fully accurate in LCOV reports when templates are involved.
This means that there is no benefit in doing the more complex logic of summing and capping.
Thus, we leave it as it currently is.

LittleHuba (approved):

Issue #579, opened 2026-06-22
Improvement: Field and Method APIs design in Rust::com


bharatGoswami8:
### What

As Field and Method feature is almost complete from LoLa(C++) side, so to adapt this API in Rust side we need to prepare design with current Event support rust library.

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

PR #700, opened 2026-07-14
Rust::com Field APIs design and example of usage

bharatGoswami8:
edit -

Since the Field and Method changes are closely related, and the Field implementation also uses the Method interface while both share the same macro for the interface and type-state pattern, I believe it would be better to review the overall design in a single PR. This will help ensure that any suggestions or feedback can be addressed more effectively and consistently across both implementations.
Therefore, I am closing this PR. Please review the PR - https://github.com/eclipse-score/communication/pull/818

---------------------------------

- Added field-based publish/subscribe with field get/set and asynchronous field subscriptions.
- Extended interface macros to support field members and generated producer/consumer and compile-time offer API enable.
- Added a new example demonstrating offered field producers and consumer field updates.

https://github.com/eclipse-score/communication/issues/579


score/mw/com/rust/score_com_concept/interface_macros.rs
```diff
@@ -76,6 +92,12 @@
 ///   "left_tire" and "exhaust" events.
 /// - `VehicleOfferedProducer<R>` struct that implements `OfferedProducer` trait for offering
 ///   "left_tire" and "exhaust" events.
+// TODO: We need to enable the support for mixed types (Event, Method, Field) in the same interface definition.
```
bharatGoswami8: ticket for same - https://github.com/eclipse-score/communication/issues/701

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -179,6 +188,163 @@ fn create_producer<R: Runtime>(
     producer.offer().expect("Failed to offer producer instance")
 }
 
+// Below function just demonstrate the field APIs usage
+// This build fine but it can not run because we have not implemented the field APIs in Lola runtime yet.
+
+// Producer creation and intialization of fields with initial values and set handlers for the fields
+// It will return the offered producer instance which can be used to update the fields.
```
rpreddyhv:
```suggestion
// Below function demonstrate the field APIs usage
// This builds fine, but it cannot run as the field APIs in Lola runtime are not implemented yet.

// Producer creation and intialization of fields with initial values and set handlers for the fields
// It will return the offered producer instance which can be used to update the fields.
```

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -179,6 +188,163 @@ fn create_producer<R: Runtime>(
...
+    let offered = producer
+        .init_field()
+        .register_set_handler_left_tire(move |val: &Tire| {
+            println!("Received tire pressure update: {:?}", val);
+            // Additional logic to handle the tire pressure update can be added here
+            // For example, we can increment value or conver unit and update the field again.
+            // TODO: in working example add that logic to demonstrate the set handler usage.
+            // Note: I think producer may be need clone ?
+        })
+        .expect("Failed to register set handlers")
```
rpreddyhv: Why would register_* ever fail? As discussed for Method, it shouldn't return Result.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -179,6 +188,163 @@ fn create_producer<R: Runtime>(
...
+        .expect("Failed to build producer instance");
+
+    // Use validator pattern with compile-time type-state validation
+    // Must register handlers and initialize all fields before offer() is available
+    let offered = producer
+        .init_field()
+        .register_set_handler_left_tire(move |val: &Tire| {
+            println!("Received tire pressure update: {:?}", val);
+            // Additional logic to handle the tire pressure update can be added here
+            // For example, we can increment value or conver unit and update the field again.
```
rpreddyhv:
```suggestion
            // For example, we can increment value or convert unit and update the field again.
```

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -11,23 +11,25 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 
+#![allow(unused)]
```
darkwisebear: Please put this to the specific places and do not globally turn it on for the module.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -179,6 +188,163 @@ fn create_producer<R: Runtime>(
...
+    let producer_builder = runtime.producer_builder::<VehicleFieldInterface>(service_id);
+    let producer = producer_builder
+        .build()
+        .expect("Failed to build producer instance");
+
+    // Use validator pattern with compile-time type-state validation
+    // Must register handlers and initialize all fields before offer() is available
+    let offered = producer
+        .init_field()
+        .register_set_handler_left_tire(move |val: &Tire| {
```
darkwisebear: I wonder whether `&Tire` is correct. Why not hand over ownership to the callable? This way, the implementer wouldn't have to clone the value in the likely case the value is written somewhere.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
```diff
@@ -179,6 +188,163 @@ fn create_producer<R: Runtime>(
...
+            // For example, we can increment value or conver unit and update the field again.
+            // TODO: in working example add that logic to demonstrate the set handler usage.
+            // Note: I think producer may be need clone ?
+        })
+        .expect("Failed to register set handlers")
+        .register_set_handler_exhaust(|_val: &Exhaust| {
+            println!("Received exhaust update");
+        })
+        .expect("Failed to register set handlers")
+        .update_left_tire(&initial_tire_value)
```
darkwisebear: Why by-reference? This forces cloning as the implementation most likely has to clone the value to send it across the network. It might be a corner case, though.

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+// Note: We are using the event related trait as a base trait for where ever we have same common
+// APIs or functionality, as of now there are derived from concept crate but
+// we will create a module which will have common trait for event and field which will be used by both event and field as a super trait and
+// for this we need to create marker trait for event.
+
+use crate::*;
+use std::fmt::Debug;
+use std::future::Future;
+
+#[allow(dead_code)]
```
darkwisebear, line 31: Why is this dead code?

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+// Temp for build test
+// We will remove this once memory layout of same created in rust side like SamplePtr.
+#[repr(C)]
```
darkwisebear, line 34: Doesn't this trigger an "improper_ctypes" lint? `Result` isn't `#[repr(C)]`, so it cannot be used here properly.

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+/// FieldSubscriber trait is provides the receiving APIs for the field subscription and
+/// it is derived from `com_api_concept::Subscription` trait which provides the receiving APIs for the field subscription.
+/// Additional methods which the field subscription provides are added in this trait.
+pub trait FieldSubscription<T: CommData + Debug, R: Runtime + ?Sized>:
+    com_api_concept::Subscription<T, R>
+{
+    /// Returns the number of new samples a call to try_receive (given parameter max_num_samples
+    /// doesn't restrict it) would currently provide.
+    /// How many new sample available for the user of this field subscription to receive.
+    fn get_num_new_samples_available(&self) -> Result<usize>;
```
darkwisebear, line 77: Last I checked this was not part of `Subscription` as it exists at the moment. Why is it there?

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Get the number of samples that can still be received by the user of this field.
+    /// This is for checking the capacity of the field subscription and to avoid overflow of the field subscription limit.
+    fn get_free_sample_count(&self) -> Result<usize>;
```
darkwisebear, line 81: Same.

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    /// Set the value of the field.
+    ///
+    /// # Parameters
+    /// * `value` - The value to set for the field.
+    ///
+    /// # Returns
+    /// Return the result of `MethodReturnTypePtr<T>` which contains the status of the set operation.
+    /// with the current value of the field.
+    fn set(&self, value: &T) -> Result<MethodReturnTypePtr<T>>;
```
darkwisebear, line 65: I do not understand why `set` returns `T`. It would be enough to communicate errors.

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    ///Set the value of the field.
+    ///
+    /// # Parameters
+    /// * `value` - The value to set for the field.
+    ///
+    /// # Returns
+    /// Return the result of `MethodReturnTypePtr<T>` which contains the status of the set operation.
+    /// with the current value of the field.
+    fn set(&self, value: &T) -> Result<MethodReturnTypePtr<T>>;
```
darkwisebear, line 97: See comment on `FieldSubscriber`.

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+
+// TODOs:
+// 1.Get and Set methods for field it is enabled based on tag, do we want to keep same kind of mechanism
+// or by default we will enable for user,
+// -> we can keep it default enable as of now,
+// and later we can add tag based mechanism if required because Interface side we need to check how we can do this
+// 2. We are offering get method after subscrption async but before subscription it is sync,
```
darkwisebear, line 19: Just turn the one before subscription into async. Or is there anything that prevents you from doing so?

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+// It is because subscribe API take the consumer instance by value and if we offer async get method then it will create issue with subscription.
+
+// Note: We are using the event related trait as a base trait for where ever we have same common
+// APIs or functionality, as of now there are derived from concept crate but
+// we will create a module which will have common trait for event and field which will be used by both event and field as a super trait and
+// for this we need to create marker trait for event.
+
+use crate::*;
```
darkwisebear, line 27: Please import specific types. Yes, it's tedious, but being explicit here also saves some headaches.

score/mw/com/rust/score_com_concept/field_concept.rs
```diff
@@ -0,0 +1,145 @@
...
+    /// It will create new task or thread to handle the set operation callback function,
+    /// which will be mostly done using thread pool or async task pool, will be decided at the time of implementation.
+    ///
+    /// # Parameters
+    /// * `callback` - The callback function to handle the set operation for the field.
+    ///
+    /// # Returns
+    /// Return the result of `Result<()>` which contains the status of the register operation.
+    // TODO: Do we need to make callback lifetime 'static or we keep same as field publisher lifetime.
+    fn register_set_handler<'a>(&self, callback: impl Fn(&T) + Send + 'a) -> Result<()>;
```
darkwisebear, line 135: There is no equivalent method for registering the get handler. It may sound weird, but the C++ counterpart has that, since mw::com treats set and get as ordinary methods.

darkwisebear, line 135: TO me this feels as if we're treating ordinary methods and field methods in a distinct way. Maybe we should think about getting methods ready first and then incorporate them as the foundation of set and get here.

score/mw/com/rust/score_com_concept/concept.rs
```diff
@@ -329,18 +342,10 @@ where
 ///
 /// # Type Parameters
 /// * `T` - The relocatable event data type
-pub trait SampleMut<T>: DerefMut<Target = T> + Debug
+pub trait EventSampleMut<T>: SampleMut<T>
```
darkwisebear, line 345: Is this just because of `update` vs `send` (aka naming)?

darkwisebear (changes requested):
I mainly reviewed the concept part; I think we need to nail that part before it makes sense to look at the rest.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: This is temporary placeholder of  `MethodReturnTypePtr` of C++ like `SamplePtr` for Event.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: Yes, this PR required to rebase with that individually module will be created as optimization example app is merged.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: Method Design is ready - https://github.com/eclipse-score/communication/pull/777
Working on Combined Field and Method changes.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: Updated callable with `Fn(T)`

bharatGoswami8:
@darkwisebear , I will create a separate PR based on the Method Design PR, and all review feedback will be addressed in that PR.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: Result return removed.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: Done , PR will be update on comment.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: Done , PR will be update on comment.

score/mw/com/example/com-api-example/basic-consumer-producer.rs
bharatGoswami8: Agreed, i was trying to match with C++ semantic but this will add additional clone, will go with same approach of `send` for Event.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: Updated example app.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: Now we are using method interface to create Field's method (set and get), so this is removed.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: Now we are using method interface to create Field's method (set and get), so this is removed but
We have another issue consumer partially moved when we call `subscribe` on field because it take self by value.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: Done

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: Added.

score/mw/com/rust/score_com_concept/concept.rs
bharatGoswami8: Yes, because of that SampleMut is just marker trait and then we have feature specific trait which provide the API like send and update.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/proxy_field_base.h#L108

This API available for Field Event and in rust case, this should be available after subscription because of that these are part of` FieldSubscription`
yes we can add on Event `concept::Subscription<T, R>` but we do not have back-end implementation so keeping only for Field.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: This is to align with `C++ `side Set Method - https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/proxy_field.h#L322

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/proxy_field_base.h#L108

This API available for Field Event and in rust case, this should be available after subscription because of that these are part of FieldSubscription
yes we can add on Event concept::Subscription<T, R> but we do not have back-end implementation so keeping only for Field.

score/mw/com/rust/score_com_concept/field_concept.rs
bharatGoswami8: updated.

bharatGoswami8:
Hi @darkwisebear , @rpreddyhv ,

I have created PR https://github.com/eclipse-score/communication/pull/818 for field design based on Method design interface.

I already addressed most of the review comment on this PR.

Request for review mentioned PR.

Closing this PR.

PR #703, opened 2026-07-14
Provide tutorial in examples

castler:


MODULE.bazel
```diff
@@ -23,6 +23,7 @@ bazel_dep(name = "platforms", version = "1.0.0")
 bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "rules_python", version = "1.8.5")
 bazel_dep(name = "rules_rust", version = "0.68.1-score")
+bazel_dep(name = "rules_pkg", version = "1.2.0")
```
LittleHuba, line 26: In score_communication it should remain a dev-dependency. In the examples it would IMHO be a normal dependency.
Our tutorial is not meant to be pulled in by downstream users as a dependency. The tutorial is meant to work when this module is the root module.

castler:
We decided that we will not do that - because of the big drawbacks and rather do this:
https://github.com/eclipse-score/communication/pull/706

PR #706, opened 2026-07-14
refactor: rename examples to module_integration_test

castler:
Rename the top-level examples directory to module_integration_test to better reflect its role as a module-level integration test package.

Update references to the old path in repository docs and CI workflow steps so links and build commands remain correct after the move.


.github/workflows/_build_and_test_gcc15.yml
```diff
@@ -54,10 +54,10 @@ jobs:
 
       - name: Build examples
```
LittleHuba:
```suggestion
      - name: Build module_integration_test
```

README.md
```diff
@@ -122,7 +107,7 @@ communication/
 ### For Users
 - [User Guide](score/mw/com/README.md) - Getting started with the API
 - [API Reference](score/mw/com/design/README.md) - Detailed API documentation
-- [Examples](examples/) - Code examples and tutorials
+- [Examples](module_integration_test/) - Code examples and tutorials
```
LittleHuba: Would be better if we link the tutorial here now?

The module integration tests are not really relevant for users.

LittleHuba (approved):

PR #707, opened 2026-07-14
mw::com: missing test cases

KrishaDeshkool:


score/mw/com/impl/method_type_test.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include <gtest/gtest.h>
+
+namespace score::mw::com::impl
+{
+namespace
+{
+
+TEST(MethodTypeToStringTest, ReturnsMethodForKMethod)
+{
+    // Given a MethodType of kMethod
```
bemerybmw: Nitpick which doesn't have to be fixed but just for future reference, the given part is about setting up the environment / context for the test. In this case, the MethodType isn't part of the setup but is what's actually being tested. So it would be more correct to say: "When converting a MethodType of kMethod to string" without a given section.

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -330,5 +338,118 @@ TEST(ProxyFieldNotifierGatingTest, OnlyNotifierApiExistsWhenNoTagsArePresent)
...
+{
+    // Given a Get-enabled ProxyField and a mock method binding
+    auto field = CreateFieldWithGetOnly();
+    EXPECT_CALL(get_method_binding_mock_, GetReturnValueBuffer(0U));
+    EXPECT_CALL(get_method_binding_mock_, DoCall(0U));
+
+    // When calling Get()
+    auto result = field.Get();
+
+    // Then the call is delegated to the underlying method binding and succeeds
```
bemerybmw: The expecting parts should go with the expectations above:

```
// Expecting that the binding returns a buffer for a return value and then calls the method
EXPECT_CALL(get_method_binding_mock_, GetReturnValueBuffer(0U));    EXPECT_CALL(get_method_binding_mock_, DoCall(0U));

    // When calling Get()
    auto result = field.Get();

    // Then the call succeeds
```

bemerybmw: Same for all the other tests

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -330,5 +338,118 @@ TEST(ProxyFieldNotifierGatingTest, OnlyNotifierApiExistsWhenNoTagsArePresent)
...
+
+TEST_F(ProxyFieldGetSetFixture, GetDelegatesToProxyMethodBinding)
+{
+    // Given a Get-enabled ProxyField and a mock method binding
+    auto field = CreateFieldWithGetOnly();
+    EXPECT_CALL(get_method_binding_mock_, GetReturnValueBuffer(0U));
+    EXPECT_CALL(get_method_binding_mock_, DoCall(0U));
+
+    // When calling Get()
+    auto result = field.Get();
```
bemerybmw, line 404: Can you also put a value in the return buffer and check that it's returned in result?

score/mw/com/impl/plumbing/proxy_method_binding_factory_test.cpp
```diff
@@ -108,6 +136,25 @@ TYPED_TEST(ProxyMethodFactoryTypedFixture, CanConstructLolaProxyMethod)
...
+    // Given a valid lola binding whose deployment carries a field named kDummyFieldName
+    const auto handle = this->GetValidLoLaHandleWithField();
+    this->InitialiseProxyWithConstructor(handle.GetInstanceIdentifier());
+    using MethodSignature = TypeParam;
+
+    // When creating a binding for both the Get and the Set of that field
+    for (const auto method_type : {MethodType::kGet, MethodType::kSet})
+    {
+        auto proxy_method =
+            ProxyMethodBindingFactory<MethodSignature>::Create(handle, *this->proxy_, kDummyFieldName, method_type);
```
bemerybmw, line 150: I guess we should also have a death test in case MethodType::kUnknown is provided (if we don't already have one)

bemerybmw (changes requested):

score/mw/com/impl/plumbing/proxy_method_binding_factory_test.cpp
KrishaDeshkool: added an assert and added a test too.

anmittag:
@KrishaDeshkool some comments are pending in your PR, please check

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -330,5 +338,129 @@ TEST(ProxyFieldNotifierGatingTest, OnlyNotifierApiExistsWhenNoTagsArePresent)
     static_assert(!has_set<DefaultField>::value);
 }
 
+/// \brief Test fixture for the runtime behavior of ProxyField's Get / Set API.
+class ProxyFieldGetSetFixture : public ::testing::Test
```
bemerybmw, line 342: These tests will probably need to be adapted based on https://github.com/eclipse-score/communication/pull/735

bemerybmw:
Taken over in: https://github.com/eclipse-score/communication/pull/756

PR #711, opened 2026-07-14
Remove flaky gateway BidirectionalTransport unit tests

SebSparrowHawk:
Tests caused an issue in some configuration (e.g. tsan) because tests would time out. Temporarily removed those tests and it will be analyzed if and how to bring them back or how to cover those cases with different test types.


SebSparrowHawk:
Closed in favor of https://github.com/eclipse-score/communication/pull/713

PR #716, opened 2026-07-15
Service discovery vibe redesign

kitsnet:
Testing the ability to GitHub Copilot to be integrated into our infrastructure by vibe coding a replacement of filesystem-based service discovery.


score/mw/com/service_discovery/dependability/assumed_system/assumed_system_requirements.trlc
```diff
@@ -0,0 +1,36 @@
...
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+package ServiceDiscovery
+
+import ScoreReq
+
+ScoreReq.AssumedSystemReq ServiceDiscoveryTransportAvailable {
+    description = "The system shall provide process-local IPC transport suitable for communication with the service discovery daemon."
```
crimson11, line 18: What does `process-local IPC transport` mean? Sounds like an anti-thesis? IPC is NOT process local. It is INTER-Process. I don't get it!

score/mw/com/service_discovery/dependability/assumed_system/assumed_system_requirements.trlc
```diff
@@ -0,0 +1,36 @@
...
+    rationale = "Service discovery relies on message passing transport and does not access the network stack under QNX."
+    safety = ScoreReq.Asil.B
+    version = 1
+}
+
+ScoreReq.AssumedSystemReq ProcessIdentityProvidedByOS {
+    description = "The operating system shall provide process identity data (UID and PID) for each daemon session."
```
crimson11, line 25: Do we need to be more explicit? IDK. I.e. explicitly state, that both parties (daemon and its client) need the process identity data of its "partner"?

score/mw/com/service_discovery/dependability/assumed_system/assumed_system_requirements.trlc
```diff
@@ -0,0 +1,36 @@
...
+    rationale = "The daemon validates ownership and prevents spoofed unregister operations based on OS-provided identity."
```
crimson11, line 26: Also the client needs to be sure to be talking to the "right" daemon! I guess from safety perspective this is essential. Maybe we could "relax" this, if we make sure, that the daemon-provided message-passing endpoint can ONLY be occupied/used by the correct/right daemon instance. In this case the client doesn't need "process identity" from the daemon. Your thoughts?

Bottomline: For the client, we simply have the requirement, that there is a mechanism to assure, that he connects to/talks to the correct daemon instance. Whether this is achieved by:
- clients connecting to an "endpoint", which can't be spoofed/taken over by a malicious daemon instance.
- clients checking the process identity of the daemon/communication partner after message-passing connect

we don't care.

score/mw/com/service_discovery/dependability/assumed_system/aous.trlc
```diff
@@ -0,0 +1,27 @@
...
+import ScoreReq
+
+ScoreReq.AoU DaemonStartedBeforeClients {
+    description = "The service discovery daemon shall be started before service providers and consumers initiate registration or lookup calls."
+    safety = ScoreReq.Asil.B
+    version = 1
+}
+
+ScoreReq.AoU SingleDaemonAuthority {
+    description = "Exactly one service discovery daemon instance shall act as registry authority at a time."
```
crimson11, line 24: In what context? I.e. if we have an interVM setup our current design says: Each VM has its own "score::mw::com communication domain" and each such domain has its own service-discovery -> daemon instance.

Thus, makes it sense to phrase it:
>Exactly one service discovery daemon instance shall act as registry authority at a time in a given score::mw::com communication domain

or are we just opening up Pandoras box ... needing to come up with definition, what such a domain exactly is?

score/mw/com/service_discovery/dependability/requirements/feature_requirements.trlc
```diff
@@ -0,0 +1,101 @@
...
+
+ScoreReq.FeatReq OsIndependentDaemonProtocol {
+    description = "The service discovery component shall provide an OS-independent protocol over the message passing interface for daemon communication."
+    safety = ScoreReq.Asil.B
+    derived_from = [ServiceDiscovery.ServiceDiscoveryTransportAvailable@1]
+    version = 1
+}
+
+ScoreReq.FeatReq PartitionedVisibilityForIntegrityLevels {
+    description = "The service discovery component shall enforce ASIL-B and ASIL-QM partitioned visibility and prevent lower-integrity claims from escalating service quality."
```
crimson11, line 35: What does this exactly mean? I.e.: It is a valid use case, that an ASIL-B client does see QM-quality offerings.
It is also a good idea, that a QM client doesn't see an offering of a service-instance in ASIL-B quality.
(This req. we can then only support, when the SD daemon has a complete trustable list of uid->ASIL-level assignments? This we need to generate from a static deployment info ...)

Sidenote: A LoLa service-instance provider of an instance in ASIL-B quality will allways also offer the instance in QM initially, but he may withdraw the QM offering selectively, if he thinks, that QM clients did mess up.

score/mw/com/service_discovery/dependability/requirements/feature_requirements.trlc
```diff
@@ -0,0 +1,101 @@
...
+
+ScoreReq.FeatReq AnyInstanceServiceSelection {
+    description = "The service discovery component shall support both explicit instance selection and any-instance service discovery queries."
+    safety = ScoreReq.Asil.B
+    derived_from = [ServiceDiscovery.ServiceDiscoveryTransportAvailable@1]
+    version = 1
+}
+
+ScoreReq.FeatReq DaemonManagedServiceInstanceLocks {
+    description = "The service discovery component shall provide daemon-managed exclusive and shared service-instance lock semantics that replace filesystem flock markers."
```
crimson11, line 80: the last part `that replace filesystem flock markers` makes no sense ... pointing back to some previous implementation decision in a feature-req. feels wrong to me.

score/mw/com/service_discovery/dependability/requirements/feature_requirements.trlc
```diff
@@ -0,0 +1,101 @@
...
+
+ScoreReq.FeatReq BatchedDaemonOperations {
+    description = "The service discovery component shall support batching of daemon operations to reduce message-passing roundtrips during startup load."
+    safety = ScoreReq.Asil.B
+    derived_from = [ServiceDiscovery.ServiceDiscoveryTransportAvailable@1]
+    version = 1
+}
+
+ScoreReq.FeatReq StaticOfferAuthorizationPolicy {
+    description = "The service discovery component shall support a static authorization policy that constrains which UIDs may offer configured service instances."
```
crimson11, line 97: see aboive: Shall we also require within the same req. or a related one, that the daemon has a static config from UID to ASIL-level?

score/mw/com/service_discovery/service_discovery_types.h
```diff
@@ -0,0 +1,124 @@
...
+{
+    return key.instance_id == kAnyInstanceId;
+}
+
+struct Registration
+{
+    ServiceKey key{};
+    Endpoint endpoint{};
+    IntegrityLevel offered_integrity{IntegrityLevel::kAsilQm};
+    IntegrityLevel provider_integrity{IntegrityLevel::kAsilQm};
```
crimson11, line 58: Do we need this? See my comments in the TLRC. If we want to avoid, that a provider with a certain (process)identification offers/registers a service with ASIL-B, although he isn't qualified with ASIL-B ... then SD daemon needs a static config, which tells, which process-identification has which ASIL level. If SD has this info already, the `provider_integrity` member makes no sense? Daemon deduces it from the process-info bound to the session/connection.

score/mw/com/service_discovery/service_discovery_types.h
```diff
@@ -0,0 +1,124 @@
...
+    std::uint32_t provider_uid{0U};
+    std::uint32_t provider_pid{0U};
+    std::uint64_t provider_session_id{0U};
```
crimson11, line 61: Seems redundant? An instance of `Registration` is send across a message-passing session? So the "session" if required will be implicitly deduced by the receiver/daemon?

score/mw/com/service_discovery/service_discovery_types.h
```diff
@@ -0,0 +1,124 @@
...
+};
+
+struct ProviderContext
+{
+    std::uint32_t provider_uid{0U};
+    std::uint32_t provider_pid{0U};
+    std::uint64_t provider_session_id{0U};
+};
+
+constexpr std::size_t kMaxRegistrationsPerService{32U};
```
crimson11, line 71: Where does this come from?
Is `kMaxRegistrationsPerService` basically a cap how many different instances can be registered/offered for a given service-instance-type?
I this a theoretical maximum (then 32 is too small) ... or is it some value, which is deduced from a static configuration model? I.e. for a given setup designed by the integrators? Then it shouldn't be a constant, but a config value read by the daemon from its startup config.

score/mw/com/test/service_discovery_daemon/service_discovery_daemon_application.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "score/mw/com/test/common_test_resources/stop_token_sig_term_handler.h"
+
+#include <score/stop_token.hpp>
+
+#include <chrono>
+#include <cstdlib>
+#include <iostream>
+#include <thread>
+
+int main()
```
crimson11, line 27: Unexpected, that this daemonm-application is in the tests-folder? I would have expected it to be part of the `mw::com::service_discovery` implementationn folder.

score/mw/com/test/service_discovery_daemon/service_discovery_daemon_application.cpp
```diff
@@ -0,0 +1,56 @@
...
+    score::mw::com::service_discovery::ServiceDiscoveryDaemon daemon{registry};
+    score::mw::com::service_discovery::ServiceDiscoveryMessagePassingServer server{daemon};
+
+    if (!server.Start())
+    {
+        std::cerr << "Service discovery daemon app: Failed to start message passing server." << std::endl;
+        return EXIT_FAILURE;
+    }
+
+    while (!stop_source.get_token().stop_requested())
```
crimson11, line 49: Really - a busy loop waking up every 50msec to check a variable?
Should be replaced by a `cond_var.wait()` ... and the signal-handler shall notify the cond_var.

Alternative: In case we introduce a worker/background-thread for the daemon, this thread could then also "supervise" the stop_source.
But currently there is obviously no need for such a "worker-thread"? I.e. all activities the daemon does are triggered by a message-reception and are then synchronously done in the context of the message-passing reception thread. This is good. Don't want to arteficially introduce a worker/background thread, when there is no need!

score/mw/com/service_discovery/service_discovery_compat.h
```diff
@@ -0,0 +1,116 @@
...
+#include <memory>
+#include <mutex>
+#include <optional>
+#include <unordered_map>
+#include <vector>
+
+namespace score::mw::com::impl
+{
+
+class ServiceDiscoveryCompat final : public IServiceDiscovery
```
crimson11, line 42: What is `Compat` ?

... and I feel the whole class-hierarchy/interface tree is "broken"?
This class should rather implement IServiceDiscoveryClient!

crimson11 (changes requested):
Generally the "architecture" is unclear to me.
What shall

PR #720, opened 2026-07-16
feat: display CodeQL MISRA findings on nightly quality dashboard

Komal362000:
Enable the nightly quality dashboard to surface CodeQL MISRA compliance reports generated in CI, and


.github/workflows/_codeql.yml
```diff
@@ -49,17 +49,74 @@ jobs:
...
+            echo "::error::Unable to locate CodeQL pack root (cpp/common/src)"
+            exit 1
+          fi
+
+          if [[ ! -x "${CODEQL_BIN}" ]]; then
+            echo "::error::CodeQL binary not found or not executable"
+            exit 1
+          fi
+
+          "${CODEQL_BIN}" pack install "${PACK_ROOT}"
```
castler: We do not want todo these steps in the CI. Are we sure that this is required? Because this would be a lot of logic outside of Bazel - which cannot be reproduced locally.

IMHO we should avoid this whole section.

bazel/rules/generate_quality_links.bzl
```diff
@@ -36,47 +36,35 @@ def _generate_quality_links_impl(ctx):
...
-            if host.endswith(".github.io") and repo:
-                owner = host[:-len(".github.io")]
-                if owner.endswith("."):
-                    owner = owner[:-1]
-                if owner:
-                    release_coverage_asset_url = (
-                        "https://github.com/" + owner + "/" + repo +
-                        "/releases/download/" + docs_version + "/" +
-                        repo + "_coverage_report_" + docs_version + ".zip"
-                    )
```
castler, line 55: These changes were required to get correct links for release builds - why are we removing it?

quality/dashboard/generate_dashboard.py
```diff
@@ -20,6 +20,7 @@
     bazel run //quality/dashboard:generate_dashboard -- \\
         --lcov           /tmp/coverage_zip/extracted/artifacts/coverage_report.dat \\
         --clang-tidy     /tmp/clang_tidy/clang_tidy_findings.txt \\
+    --codeql-csv     /tmp/codeql-results/codeql-nightly.csv \\
```
castler: Why do we need this? If we do not use the attribute?

docs/sphinx/quality_reports.rst
```diff
@@ -35,17 +35,21 @@ nightly run of the `Nightly Quality Jobs`_ workflow.
...
    Quality reports are generated by the nightly CI job and published
    exclusively alongside the ``latest`` documentation on GitHub Pages.
-   In local Sphinx builds the link cells above show equivalent ``bazel run``
-   commands. In versioned release docs, coverage links point to the uploaded
-   release artifact while the remaining quality links point to the ``latest``
-   reports.
+   In local Sphinx builds and in versioned release archives the link cells
+   above show the equivalent ``bazel run`` command or a link to the
+   ``latest`` reports respectively. The deploy workflow controls this
+   automatically via the ``DOCS_VERSION`` build variable.
```
castler: Why do we want to change this?

quality/static_analysis/codeql_lint.py
```diff
@@ -60,7 +66,6 @@ def analyze_database(code_ql_path, database_path, source_root, analysis_report_p
     csv_path = f"{output_base}/{output_prefix}.csv"
 
     # Run CodeQL analysis (generates SARIF)
-    print("\n Running CodeQL analysis...")
```
castler, line 63: Why are we removing this?

quality/static_analysis/codeql_lint.py
```diff
@@ -74,60 +79,40 @@ def analyze_database(code_ql_path, database_path, source_root, analysis_report_p
 
     # Generate reports using CodeQL analysis_report tool
     if analysis_report_path and os.path.exists(analysis_report_path):
-        print(" Generating MISRA C++ compliance reports...")
```
castler, line 77: Can we keep this?

quality/static_analysis/codeql_lint.py
```diff
@@ -74,60 +79,40 @@ def analyze_database(code_ql_path, database_path, source_root, analysis_report_p
...
+            print(f" CodeQL bin dir exists: {os.path.isdir(codeql_bin_dir)}")
             env["PATH"] = f"{codeql_bin_dir}:{env.get('PATH', '')}"
+            print(f" PATH for analysis_report: {env['PATH']}")
 
             # analysis_report expects positional args: database-dir sarif-file output-dir
             reports_output_dir = os.path.join(output_base, "analysis_reports")
 
-            # Remove existing reports directory if it exists
-            if os.path.exists(reports_output_dir):
-                shutil.rmtree(reports_output_dir)
```
castler, line 103: Why are we removing this?

.github/workflows/_codeql.yml
Komal362000: You're right, that section was working around something Bazel already does for us. I moved the patch and try to  it runs identically whether invoked via bazel run locally or in CI.

bazel/rules/generate_quality_links.bzl
Komal362000: You're right to flag this is a genuine mistake, not an intentional removal. The **release_coverage_asset_url** logic computes a GitHub release artifact download link for the coverage reports.

quality/static_analysis/codeql_lint.py
Komal362000: I removed it to keep the CI logs clean and keep the PR small. If you think it's useful, I can add it back.

quality/static_analysis/codeql_lint.py
Komal362000: This removes the old reports before creating new ones. I first thought the new reports would replace the old ones, but they don't always do that. Some old files can still stay there and get included. Keeping this cleanup makes sure only the new reports are used.

castler (approved):

PR #727, opened 2026-07-18
Handle get/set field method enable flags in proxy method binding factory

sahithi-nukala:


score/mw/com/impl/plumbing/proxy_method_binding_factory_test.cpp
```diff
@@ -245,4 +281,92 @@ TYPED_TEST(ProxyMethodFactoryTypedFixture, GetQueueSizeTerminatesForMethodInLola
         score::cpp::ignore = detail::GetQueueSize(handle, kDummyMethodName, MethodType::kMethod));
 }
 
+TYPED_TEST(ProxyMethodFactoryTypedFixture, CanConstructLolaProxyFieldGetMethodWhenGetIsEnabled)
+{
+    // Given a valid lola binding with a field whose Get method is enabled in the configuration
+    const auto handle = this->GetFieldHandle(/*use_get_if_available=*/true, /*use_set_if_available=*/false);
```
bemerybmw: This approach is clear now, but if someone refactors GetFieldHandle in the future and for example changes the order of the args, these comments would no longer be correct. This is a general issue with comments i.e. that they're not compiled and so always have to be manually checked which is why we usually only want to add them when they really add value, because they always add a maintenance burden. In this case, I think it would be cleaner to have 2 enum classes: GetterStatus::Enabled/Disabled and same for the setter. This line will now be `this->GetFieldHandle(GetterStatus::Enabled, SetterStatus::Disabled);` This now conveys the same information but with compiled code instead of comments. If someone swaps the arguments in the function signature, we'd get a compilation error.

score/mw/com/impl/plumbing/proxy_method_binding_factory_impl.h
```diff
@@ -123,15 +123,38 @@ Result<std::unique_ptr<ProxyMethodBinding>> ProxyMethodBindingFactoryImpl<Return
             const auto& lola_service_instance_deployment =
                 GetServiceInstanceDeploymentBinding<LolaServiceInstanceDeployment>(service_instance_deployment);
 
-            const auto method_it = lola_service_instance_deployment.methods_.find(method_name_str);
-            SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(method_it != lola_service_instance_deployment.methods_.end(),
-                                                        "Could not find method deployment information for method");
-            if (!method_it->second.enabled_)
+            if ((method_type == MethodType::kGet) || (method_type == MethodType::kSet))
```
bemerybmw: I would put this into a function `IsMethodOrFieldEnabled` which returns a bool. Then here we react to this bool by logging a message and returning a nullptr if it had returned false. Th message can then say Proxy Field or method ... was disabled. The log is slightly less informative (because technically a method and field could have the same name) but that's unlikely IMO and this is anyway just a hint to the user. I would favour keeping the code clean by encapsulating this logic despite the shortcoming mentioned above.

bemerybmw (approved):

PR #735, opened 2026-07-21
mw/com: Various field fixes

bemerybmw:


score/mw/com/doc/tutorial/chapter_9/consumer/consumer.cpp
```diff
@@ -173,11 +173,22 @@ void HandleSingleWarning(TirePressureExtendedProxy& proxy, const Tire tire)
     auto set_result = proxy.tire_pressure_thresholds.Set(new_thresholds);
     if (!set_result.has_value())
     {
-        std::cerr << "Failed to Set() tire_pressure_thresholds: " << set_result.error() << std::endl;
+        std::cerr << "Failed to Set() tire_pressure_thresholds due to issue with method call: " << set_result.error()
+                  << std::endl;
         return;
     }
+
+    auto& setter_return_value_ptr = set_result.value();
```
bemerybmw: discussed with @crimson11 that the result should be serialized in shm so we don't rely on score::Result being shm compatible. Should also flatten result in ProxyField::Set so that the user just has to check one mw/com result.

score/mw/com/impl/BUILD
```diff
@@ -1149,6 +1149,21 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "shm_compatible_result",
```
castler: I would strongly vote that we do not reimplement score::Result, your statement that we cannot control it is IMHO not true. If we see it like that - then we can also not control std::vector, or map or...Thus, let's use it IMHO and IF at some point in time we would get a problem (which cannot be because it would require dynamic memory allocation on all hot paths) then we can still reimplement it.

castler (dismissed):

castler (approved):

Issue #732, opened 2026-07-21
Improvement: Move public rust API out of the impl directory


bharatGoswami8:
### What

Public APIs should be out of impl directory.

### Estimates for realization

0

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [ ] Requirements / Architecture are not affected by this change?

bharatGoswami8:
following updates required from user side as there is no backward compatibility for old targets.

- `com_api` crate name changed to `score_com` and bazel target path and name changed for same - 
>   "//score/mw/com/impl/rust/com-api/com-api" --> "//score/mw/com/rust:score_com",

- Bazel target path and name changed for CPP macro invocation from user side -
` "//score/mw/com/impl/rust/com-api/com-api-ffi-lola:registry_bridge_macro_cpp" --> "//score/mw/com/rust/score_com_cpp_bridge:register_interface",`

I hope all the users are importing all the public interface using `score_com`(old crate name `com_api`) if not then please it because concept crate visibility is changed now.
Please refer example app for change in targets and crate import - https://github.com/eclipse-score/communication/tree/main/score/mw/com/example/com-api-example

PR #730, opened 2026-07-20
Rust::com Enable the logging support in Rust Lib

bharatGoswami8:
Enable the logging support for com-api rust library.

https://github.com/eclipse-score/communication/issues/285

Depends-on: https://github.com/eclipse-score/communication/pull/556

PR #736, opened 2026-07-21
Rust::com Move Rust Public API to com/rust

bharatGoswami8:
Moves the Rust COM public API from the internal impl/rust/ folder to a proper public location at com/rust.

- New targets created at rust (`score_com`, `score_com_concept`, `score_com_macros`, `score_com_cpp_bridge`)
**- `Old targets support is not there.`**
- Internal type names prefixed with Lola/Mock to avoid clashes with public trait names (e.g. Publisher - LolaPublisher)
- Docs and examples updated to reference the new paths
- Golden file ordering for the visibility guard test
- Changes from the original text: com_api - score_com.

https://github.com/eclipse-score/communication/issues/732


score/mw/com/example/com-api-example/com-api-gen/vehicle_gen.cpp
rpreddyhv, line 1: suggestion: I prefer to see a path like `score/mw/com/example/rust/api/gen/vehicle.cpp`

score/mw/com/rust/score_com_cpp_bridge/BUILD
rpreddyhv, line 1: suggestion: I guess rust doesn't like `_` in directory names:
`score/mw/com/rust/ffi-bridge/BUILD`. or otherwise, keep up the directory naming consistency.

score/mw/com/rust/score_com_macros/lib.rs
rpreddyhv, line 1: suggestion: This could be renamed as: `score/com/rust/macro/lib.rs`

question: Do we need to expose the macro implementation to user? Similar to the rust code, is it possible to only expose the macro declaration/signature and have the implementation inside `impl/rust/macro`?

score/mw/com/example/com-api-example/BUILD
```diff
@@ -21,7 +21,7 @@ rust_binary(
     visibility = ["//visibility:public"],
     deps = [
         "//score/mw/com/example/com-api-example/com-api-gen",
-        "//score/mw/com/impl/rust/com-api/com-api",
+        "//score/mw/com/rust:api",
```
rpreddyhv: praise: Good to see a concise path with enough information.

tejaswinijoshi89 (approved):

quality/visibility_guard/public_targets.golden
```diff
@@ -111,6 +111,10 @@
 //score/mw/com/performance_benchmarks/macro_benchmark:lola_benchmarking_client
 //score/mw/com/performance_benchmarks/macro_benchmark:lola_benchmarking_service
 //score/mw/com/performance_benchmarks/macro_benchmark:mw_com_config.json
+//score/mw/com/rust/com-api-macros:com-api-macros
+//score/mw/com/rust/concept:concept
+//score/mw/com/rust/ffi_bridge:register_interface
+//score/mw/com/rust:api
```
castler: IMHO it would be good to have only _one_ crate here and we should name it: 

`//score/mw/com/rust:score_com`

score/mw/com/example/com-api-example/com-api-gen/vehicle_gen.cpp
```diff
@@ -12,7 +12,7 @@
  ********************************************************************************/
```
castler, line 12: This example folder actually does not make sense.

If this is an example, we should move this into chapter 14 of our tutorial (and also write the respective explanatory)

score/mw/com/rust/design/high_level_design_detail.md
```diff
@@ -121,7 +121,7 @@ The LoLa Runtime is responsible for translating generic trait operations into co
 
 **FFI bridge layer**
 
-**Location**: [com-api-ffi-lola/](../com-api/com-api-ffi-lola/)
```
castler, line 124: Is this a user docu, or a detailed design? I rather think its a detailed design, thus I would rename the folder form `doc` to `design`

castler (changes requested):

quality/visibility_guard/public_targets.golden
bharatGoswami8: We actually really need this two
```
//score/mw/com/rust/ffi_bridge:register_interface (c++ bridge)
//score/mw/com/rust:api (rust)
```
The other two targets are already imported by //score/mw/com/rust:api, so users can access them through that target.

Since both of these are part of the public interface, I kept their visibility as public.

Regarding the naming, we're already directly under the score/com folder, so I think using the name `score_com` is redundant.

score/mw/com/example/com-api-example/com-api-gen/vehicle_gen.cpp
bharatGoswami8: agreed, this can be moved in tutorial folder i can create ticket for this.

We have some pending PR for example app optimization once those merge, this can be planned - https://github.com/eclipse-score/communication/pull/556 and https://github.com/eclipse-score/communication/pull/730

score/mw/com/rust/design/high_level_design_detail.md
bharatGoswami8: So we have detail design, as well user related document in this folder, because of that we kept all the documentation inside `doc` folder.

score/mw/com/example/com-api-example/com-api-gen/vehicle_gen.cpp
bharatGoswami8: Ticket created for same - https://github.com/eclipse-score/communication/issues/741

score/mw/com/rust/design/high_level_design_detail.md
bharatGoswami8: Created two different folders for `design` and user related document it is `doc`.

quality/visibility_guard/public_targets.golden
bharatGoswami8: I have updated `quality/visibility_guard/public_targets.golden` and target visibility.

bharatGoswami8: Updated to `//score/mw/com/rust:com_api`

rpreddyhv (commented):
Few early review finding

score/mw/com/rust/BUILD
```diff
@@ -0,0 +1,30 @@
...
+    name = "com_api",
+    srcs = ["com_api.rs"],
+    crate_name = "com_api",
+    edition = "2024",
+    visibility = [
+        "//visibility:public",  # platform_only
+    ],
+    deps = [
+        "//score/mw/com/impl/rust/com-api/com-api-runtime-lola",
+        "//score/mw/com/impl/rust/com-api/com-api-runtime-mock",
```
LittleHuba, line 45: I would not put mocks in a production target. This will force you to either have an AoU that they shall not be used or to qualify them.
Better put them in a separate target that is marked as test_only.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/runtime.rs
LittleHuba, line 1: Just be aware that LoLa is only our IPC binding. We will have other bindings as well. If these classes are really specific to LoLa, all fine. If they are targeting our binding independent layer better drop the LoLa from the names.

LittleHuba (commented):
Your migration aliases fall short. While they point to the right targets, the types in the targets were renamed.
I see two sensible ways forward:

1. Drop the aliases completely. We do not provide a migration path. This API was never communicated as stable to our users.
2. Provide actual migration shims where you not only alias the right bazel target but also alias all types within.

I'm fine with either as long as you don't do something in the middle.

score/mw/com/example/com-api-example/com-api-gen/vehicle_gen.cpp
bharatGoswami8: I think keeping gen make right impression to use like this file code is generated.

score/mw/com/rust/score_com_macros/lib.rs
bharatGoswami8: This derive macro is directly used by user on the type/struct, we already proving common point for import using `com_api `crate.

As this is kind of public interface so keeping outside the `impl`.

score/mw/com/rust/design/sequence_diagram/async_cancellable_receive_data.puml
rpreddyhv, line 1: thought: we may not need `com_api` prefix for all files in `design` & `design/sequence_diagram` dirctories.

score/mw/com/rust/score_com_cpp_bridge/BUILD
bharatGoswami8: both work fine but i will update to maintain the same format.

score/mw/com/example/com-api-example/BUILD
bharatGoswami8: This is changed, please check above conversation.

score/mw/com/impl/rust/com-api/com-api-runtime-lola/runtime.rs
bharatGoswami8: This runtime is implemented specific for `LoLa.`

score/mw/com/rust/BUILD
bharatGoswami8: Agreed , created separate test only target

score/mw/com/rust/design/sequence_diagram/async_cancellable_receive_data.puml
bharatGoswami8: removed prefix.

bharatGoswami8:
> Your migration aliases fall short. While they point to the right targets, the types in the targets were renamed. I see two sensible ways forward:
> 
> 1. Drop the aliases completely. We do not provide a migration path. This API was never communicated as stable to our users.
> 2. Provide actual migration shims where you not only alias the right bazel target but also alias all types within.
> 
> I'm fine with either as long as you don't do something in the middle.

@LittleHuba , I think better to remove aliases, as we already notified on slack channel to adapt as per updated bazel target.

score/mw/com/rust/score_com_cpp_bridge/BUILD
bharatGoswami8: com/rust has now all the folder name with` _`

quality/visibility_guard/public_targets.golden
bharatGoswami8: Updated to //score/mw/com/rust:score_com

castler (approved):

Issue #110, opened 2025-12-01
Improvement:  Add Flatbuffers as configuration alternative


arsibo:
### What

As discussed in https://github.com/orgs/eclipse-score/discussions/472?sort=new#discussioncomment-15063052 point 3.  Migration from JSON to Flatbuffers shall be showcased in communication module.

### Estimates for realization

 -

### Category

- [x] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

arsibo:
@OliverHeilwagen

OliverHeilwagen:
This issue and respective PR were only used as a POC, the correct architectural changes and support of flatbuffers is in discussion and can be followed within #685

PR #738, opened 2026-07-21
Implement fbs schema & json to fbs converter tool

Thomas-Mikhael:
+ Added Flatbuffers Schema as the (Single Source of Truth)
+ Script to generate the JSON schema from Flatbuffers Schema (using baselibs/flatbuffers).
+ JSON schema attributes are ordered to be consistent with each schema generation
+ Script to convert config json -> fbs
+ Added Schema drift test (including config "rountrip conversion" test: json -> fbs -> json)
+ Introduce Flag to control compiling the Flatbuffers (--config=flatbuffers), default is off


.bazelrc
```diff
@@ -104,6 +104,10 @@ common:qnx_arm64 --test_env=QEMU_CPU=Cascadelake-Server-v5
 
 common:qnx --config=qnx_x86_64
 
+# Enables FlatBuffers-based mw::com configuration tooling (disabled by default,
+# see score/mw/com/impl/configuration/flatbuffers_flags.bzl).
+common:flatbuffers --//score/mw/com/impl/configuration:enable_flatbuffers=true
```
castler: Can we please name this `experimental_enable_flatbuffer_configuration`

castler:
As discussed via Slack - I love that we go for the Flatbuffers support and the contribution you perform here! But it would be very welcome to see the roadmap and map for its qualification before continuing here. Stating again: This does not require a full and ready qualification, we would just like to judge if the qualification at the end can succeed.

Thomas-Mikhael:
> As discussed via Slack - I love that we go for the Flatbuffers support and the contribution you perform here! But it would be very welcome to see the roadmap and map for its qualification before continuing here. Stating again: This does not require a full and ready qualification, we would just like to judge if the qualification at the end can succeed.

Hi Jan, I agree.
Today I have internal meeting with the responsible Arch/Manager to discuss that
Will keep you updated.

quality/visibility_guard/public_targets.golden
```diff
@@ -87,6 +87,8 @@
 //score/mw/com/dependability:mw_com_rst
 //score/mw/com/example/com-api-example:com-api-example
 //score/mw/com/impl/bindings/mock_binding:generic_skeleton_event
+//score/mw/com/impl/configuration/converter:json_to_flatbuffer
+//score/mw/com/impl/configuration/converter:mw_com_config.bin
```
castler: We should guard this via the experimental flag - and since the experimental flag is per default off, these targets should net yet be publicly visible.

score/mw/com/impl/configuration/converter/json_to_flatbuffer_rule.bzl
```diff
@@ -0,0 +1,69 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Bazel rule that converts an mw::com config JSON into a FlatBuffer binary.
+
+The rule runs the ``json_to_flatbuffer`` Python tool as a build action, passing
+the pinned ``@flatbuffers//:flatc`` compiler, so ``.bin`` artifacts become part
+of the build graph and can be consumed by other targets (e.g. as ``data``).
+"""
+
```
castler, line 18: you are missing a visibility declaration in this *.bzl file

score/mw/com/impl/configuration/mw_com_config.fbs
```diff
@@ -0,0 +1,232 @@
...
+// This program and the accompanying materials are made available under the
+// terms of the Apache License Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// SPDX-License-Identifier: Apache-2.0
+// *******************************************************************************
+//
+// FlatBuffers schema mirroring the mw::com runtime configuration.
+//
+// This schema is the binary-serialization counterpart of the JSON configuration
```
castler: I would have expected that we generate this file out of the JSON schema - otherwise, we get a problem with maintainability - as we have to take care in various places the the configuration is maintained.

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
```diff
@@ -0,0 +1,310 @@
...
+# See the NOTICE file(s) distributed with this work for additional
+# information regarding copyright ownership.
+#
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Convert an mw::com configuration JSON file into a FlatBuffer binary.
```
castler: My hope is, that if we auto generate the `*.fbs`, we also get rid of this tool (or make it way more easy - as we do not have yet another place where we translate between the two tools

castler (commented):
Thanks this is going into the right direction.

I still believe that we need a flatbuffers certifcation for the 1.0 - and not only in 2027. And we should have a parallel workstream with weekly (or bi-weekly) reporting about the tasks there. 

Anyhow - for this PR, my major concern is that we know have 4 places where we need to know the different schemas:

* In the JSON Parser (soonish also flatbuffers parser +1)
* In the JSON Schema
* In the Flatbuffers Schema
* In the JSON to Flatbuffers converter

I think this is to much duplication and will make it not maintainable. We should strive to have a generic flatbuffers schema generatioin and also a generci json to flatbuffers converter - to reduce the duplication as much as possible.

score/mw/com/impl/configuration/mw_com_config.fbs
Thomas-Mikhael: That would not be safe, JSON Schema is loosy and defines only the  structure.
1- It doesn't specify the size (Integer widths: uint8, 16, ...)
For an ASIL / safety-relevant config, silently inferring a wrong width (e.g. int64 where the C++ side expects uint16) is exactly the kind of defect that wouldn't be caught by flatc.
2- Enum sentinels e.g. INVALID = 0 is a deliberate "unset detection" (design mirroring) for QualityType
3- `= null` optional-scalar vs baked default
4- Table names for nested objects (e.g. ServiceTypeEvent vs InstanceEvent)
5- json_to_flatbuffer.py hardcodes the same key->field mapping

So yes, maintaining 2 Schemas is not great, but it's the safest way here (with existing one common testing for both - to come in next stories)
Also, JSON Schema is not changing that often, mostly stable, right?

What I suggest, is to implement a "drift-detection test" that parses the JSON schema and the .fbs and fails the build if their structure diverges.
This catches if someone edits the JSON schema and forgets the .fbs.
What do you think?

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
Thomas-Mikhael: This converter tool is only a facilitating option for whom already has a JSON config and wants to use flatbuffer config binaries for its benefits.
It's converting to flatbuffer convertable json format (snake case)

User has the freedom to use any other method to get flatbuffer binaries directly.
For example:
+ Programmatic builder (FlatBufferBuilder in C++, the flatbuffers package in Python, etc.) with any other text formats: YAML, TOML, XML, CSV, .ini.
+ A database or network service — query rows/records and build directly.
+ In-memory C++ objects

quality/visibility_guard/public_targets.golden
Thomas-Mikhael: Is it okay if I remove them from here and add them to EXCLUDED_PUBLIC_TARGETS in quality/visibility_guard/parser.py?

.bazelrc
Thomas-Mikhael: I will

score/mw/com/impl/configuration/converter/json_to_flatbuffer_rule.bzl
Thomas-Mikhael: I'll add it as
visibility(["//score/mw/com/impl/configuration/..."])

Thomas-Mikhael (commented):
I added my replies/questions for each comment

quality/visibility_guard/public_targets.golden
castler: No, as Long as it is Experimental it should Not be public

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
castler: I understand that, but Here we have then again the key mapping. Thats why auch a tool ja fine, but it should ne generic without hardcoding again the mappings

score/mw/com/impl/configuration/mw_com_config.fbs
castler: Its not only the two schemas, we will also have another instance of this mapping in the json to flatbuffer tool.
This will be (hopefully) a long-term solution, if flatbuffers is safety certified, and thus it will be a maintaince nightmare. I agree that the config does not change often, but it will change - and then you do not want to edit so many places.

I see multiple options:

1st) Have integers always as uint64. We have the very same problem with JSON and there we check the boundary values within the C++ code. If we go this way, it is the same mechanichs in JSON ant Flatbuffers. Same is true for enum sentinels or "null optional scalars.  We should just handle both the same and not have defaults in the one, where we have optional values in the other

2nd) Generate the JSON schema from flatbuffers. As per my understanding this is directly supported by flatc. <- But this will be though since this will only work with "experimental flag turned on" most certainly?

Thomas-Mikhael: I'll investigate the second option as I see it more robust.

Thomas-Mikhael: I think we can have the JSON schema committed to the repo, and when a speial flag is on then it will be regenerated and overwritten, but this would be needed only with Schema change/update.
Then we will not need to maintain both schemas manually, but just the Flatbuffers schema and regenerate the JSON one.
Would this be okay if feasible?

castler: yes that would be better IMHO

Thomas-Mikhael: Ok, here's some analysis limitations for using fbs schema as our main schema to maintain and generate the JSON schema from it using 'flatc --jsonschema':
+ Hyphen '-' is forbidden in fbs, so I need to convert it to '_' for example, and restore it when generating the JSON schema (doable)
+ flatc --jsonschema output can't reproduce the rich hand-authored JSON schema (it loses titles/descriptions/defaults/min-max/string-enums)
To not tradeoff here, I'd try adding custom attributes to fbs and try writing a script that appends these attributes to the generated JSON Schema (will be much of code but if once done, would save the maintenance nightmare you mentioned)

I'll ping you once I have more updates

Thomas-Mikhael:
@castler 
I did core changes based on our discussions, so I reverted previous commit and forced pushing my new commit.
Please re-review.

quality/visibility_guard/public_targets.golden
```diff
@@ -52,6 +52,8 @@
 //score/mw/com/dependability:mw_com_doc_needs
 //score/mw/com/dependability:mw_com_index
 //score/mw/com/dependability:mw_com_rst
+//score/mw/com/impl/configuration/converter:json_to_flatbuffer_rule.bzl
+//score/mw/com/impl/configuration:flatbuffers_flags.bzl
```
castler: Why should these two files be public? I think you are missing a visibility within them.

score/mw/com/impl/configuration/converter/BUILD
```diff
@@ -0,0 +1,33 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+load("@rules_python//python:defs.bzl", "py_binary")
+load(":json_to_flatbuffer_rule.bzl", "json_to_flatbuffer")
+
+# Exported so the configuration package's schema_drift_test can import the preprocessor.
+exports_files(["json_to_flatbuffer.py"])
```
castler: Its not a good practice to `exports_files` especially not if you want to import them. In this case, you want a `py_library`

Thomas-Mikhael:
@castler It's ready. can you please re-review?

score/mw/com/impl/configuration/converter/json_to_flatbuffer_rule.bzl
```diff
@@ -0,0 +1,49 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Build rule turning a mw::com JSON configuration into its FlatBuffer binary.
+
+Drives the thin ``json_to_flatbuffer`` preprocessor (``-`` -> ``_`` key/enum
+normalization) plus ``flatc --binary`` against ``mw_com_config.fbs``.
+"""
+
+def json_to_flatbuffer(
```
castler, line 75: Should we not have a generic rule for this in baselibs, and then have here one that just uses the `mw_com_config.fbs`?

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
```diff
@@ -0,0 +1,145 @@
...
+
+    normalized = _normalize(config, _enum_symbols(flatc, fbs_path))
+
+    with tempfile.TemporaryDirectory() as work_dir:
+        # flatc names the output after the input stem; keep the original stem so the
+        # produced ``.bin`` matches the config's name.
+        source = os.path.join(work_dir, os.path.basename(json_path))
+        with open(source, "w", encoding="utf-8") as handle:
+            json.dump(normalized, handle)
+        produced = _run_flatc_binary(flatc, fbs_path, source, work_dir)
```
castler: I would put this into a custom action. If you create `json_to_flatbuffer_rule.bzl` a real rule, you could have two actions in there. One which adjusts the json, one which calls flatc. This will make the code you ahve to write here way simpler and will enable more parallelism (as bazel will schedule each action).

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
```diff
@@ -0,0 +1,145 @@
...
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+"""Convert a mw::com JSON configuration into its FlatBuffer binary via ``flatc``.
+
+The public JSON format uses hyphenated keys (e.g. ``asil-level``) and hyphenated enum
+values (e.g. ``file-permissions-on-empty``), but FlatBuffers identifiers cannot contain
+hyphens, so ``mw_com_config.fbs`` spells them with underscores. This script is a thin,
+*generic* preprocessor: it rewrites ``-`` -> ``_`` in object keys and in enum-valued
```
castler, line 18: Great that it is now generic - I think we should provide it from baselibs as this will be required by the other modules as well.

score/mw/com/impl/configuration/flatbuffers_flags.bzl
```diff
@@ -0,0 +1,39 @@
...
+"""Build flag gating the (experimental) FlatBuffers configuration path.
+
+The FlatBuffers config parsing strategy is not yet wired into the runtime, so the
+generated ``flatbuffer_cc_library`` and related targets are compiled only when the
+flag is set. Enable with ``--config=flatbuffers`` (see .bazelrc).
+"""
+
+load("@bazel_skylib//rules:common_settings.bzl", "bool_flag")
+
+def flatbuffers_flags(name = "flatbuffers_flags"):
```
castler, line 24: Why do we provide a name attribute, if we do not use it?

score/mw/com/impl/configuration/generate_schema.py
```diff
@@ -0,0 +1,294 @@
...
+import os
+import re
+import subprocess
+import sys
+import tempfile
+
+# Table whose def is emitted as ``$defs/serviceVersion`` and referenced (not inlined),
+# because it is shared by both service types and service instances.
+_SHARED_DEF_SUFFIX = "_ServiceVersion"
+_SHARED_DEF_NAME = "serviceVersion"
```
castler: If we get rid of this, then we could have this tool also totaly generic right?

quality/visibility_guard/public_targets.golden
Thomas-Mikhael: I think json_to_flatbuffer_rule.bzl should be public  (it's a build rule meant to be loaded by any test/config package)

score/mw/com/impl/configuration/converter/json_to_flatbuffer_rule.bzl
Thomas-Mikhael: If I understand you correctly, then someone should add this in "baselibs" and with the their coming release this should be available from their package, correct?
If so, can we initially push it here -> trigger that request to baselibs developers -> replace it here once they provide it?

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
Thomas-Mikhael: _"can we initially push it here -> trigger that request to baselibs developers -> replace it here once they provide it?"_

quality/visibility_guard/public_targets.golden
Thomas-Mikhael: After reading your comment about moving that to baselibs:
_"can we initially push it here -> trigger that request to baselibs developers -> replace it here once they provide it?"_

score/mw/com/impl/configuration/flatbuffers_flags.bzl
Thomas-Mikhael: Buildifier warning
_"The macro "flatbuffers_flags" should have a keyword argument called "name".

It is considered a macro because it calls a rule or another macro "native.config_setting" on line 33.

By convention, every public macro needs a "name" argument (even if it doesn't use it).
This is important for tooling and automation."_

score/mw/com/impl/configuration/generate_schema.py
Thomas-Mikhael: I'll work on finding a generic clean solution for it.

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
Thomas-Mikhael: I will try to do it (also the flatc call in enum_symbols function), though it will affect the Round trip test that calls this script, but I'll find a way

score/mw/com/impl/configuration/converter/BUILD
Thomas-Mikhael: Will rework it

score/mw/com/impl/configuration/generate_schema.py
Thomas-Mikhael: Done

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
Thomas-Mikhael: Done

score/mw/com/impl/configuration/converter/BUILD
Thomas-Mikhael: Done

score/mw/com/impl/configuration/flatbuffers_flags.bzl
castler: should we then not just use it?

Thomas-Mikhael: We do not use it, this is just to satisfy the rule
`    _ = name  # buildifier: disable=unused-variable`

score/mw/com/impl/configuration/converter/json_to_flatbuffer.py
castler: I do not fully understand. There is right now quite some infrastructure setup in Baselibs for flatbuffers:
https://github.com/eclipse-score/baselibs/tree/main/score/flatbuffers

IMHO its just wrong to put these generic scripts into mw/com.
What do we gain by it? For me its fine if we first place them here and then move them...but...it sounds more like double work to be honest. @LittleHuba should decide IMHO.

In any case those targets should not be public for now - to avoid usage. Esp. if we want to move them again.

Thomas-Mikhael: I agree with you, but I just need them currently for my coming PR (Flatbuffer Parsing Strategy) because my plan to verify by making the testing of JSON parser common for both with that extra step for the Flatbuffer is to convert the test's json excerpt into flatbuffers binary.
And delaying that until flatbuffers repo push that and release, this will take much of time.

@LittleHuba what do you think?

OliverHeilwagen: The json_to_flatbuffer.py script is mw::com specific and is not required in baselibs.
The rationale for the script is to keep the user facing json schema identical when introducing the additional flatbuffers support, step by step.

The existing schema is not flatc compatible.

Long term it can be decided if the script is kept in order to avoid a breaking changes for the user configurations schema or if the schema is adapted to also meet flatc requirements (e.g. no '-' in keys as the name will be used to generate a read and builder method).

This was also briefly discussed in the initial demo PR / Issue.
https://github.com/eclipse-score/communication/issues/110#issuecomment-3631173732

The actual serialization can be done with the translated json and [serialize_buffer](https://github.com/eclipse-score/baselibs/blob/b99dbd63788ad4345642c26e7c959da6ca99e50f/score/flatbuffers/bazel/tools.bzl#L149) rule or recommended [serialize_versioned_buffer](https://github.com/eclipse-score/baselibs/blob/b99dbd63788ad4345642c26e7c959da6ca99e50f/score/flatbuffers/bazel/tools.bzl#L348) rule. ([examples](https://github.com/eclipse-score/baselibs/blob/main/examples/integration/flatbuffers/config_usecase/BUILD))

The json_to_flatbuffer_rule.bzl and other py scripts seem to use direct invocation of flatc. This should definitely be avoided. 

All flatc interaction shall be wrapped via baselibs. 
Json schema generation is available via [generate_json_schema](https://github.com/eclipse-score/baselibs/blob/b99dbd63788ad4345642c26e7c959da6ca99e50f/score/flatbuffers/bazel/tools.bzl#L551). If additional flatc features are required they should be brought to baselibs e.g. schema evolution check, this is currently not provided as a bazel rule. (I've not checked if this is needed for schema_drift_test.py)

limdor: @OliverHeilwagen can you please elaborate on what makes `json_to_flatbuffer.py` be mw::com specific? I do not see how this could not be reused by any module that could potentially use json and flatbuffers for configuration.

OliverHeilwagen: @limdor 
`Recursively rewrite ``-`` -> ``_`` in object keys and enum-valued strings.`

When starting with a flatbuffer schema, using "_" instead of "-" will already be enforced by flatc as those names will be used for code generation.

It is a choice that needs to be done by mw::com developers if they want to keep this conversion (which is a reasonable thing to do for the user), or dropping it and change to a flatc compatible json format without "-" (breaking change in terms of configuration input files).

In baselibs flatbuffers i would discourage such a conversion feature as it is not required for the intended usage of flatbuffers (use keys and enums with "_" instead of "-").

Other changes in this PR such as generate_schema.py are generic and extend the flatc functionality.

Thomas-Mikhael: @castler I'll create a PR in baselibs and push the flatbuffers changes that was agreed with @OliverHeilwagen in it.
Then migrate the code to use the baselibs/flatbuffers and adjust the scripts to keep only the mw::com relevant code + schema_drift_test
I'll need just to drop the RoundTrip test in schema_drift_test.py; it is not of critical importance anyway, the tests of the new coming flatbuffer parser class will cover the integrity of the same config to be exactly read in both strategies

LittleHuba: The first question here seems to be:
Do we want to make this change in our configuration file visible to the user?

If no, we have two options:
1. Pollute score_baselibs with migration paths that for the moment we only need in score_communication.
2. Add custom logic in score_communication to take care of this piece of the conversion.

If yes, we must provide a good migration path for our users.

The basic rule we should follow:
Only decentralize what is specific to our codebase.

I intentionally did not go deep in the source code. I want to focus on a high-level user perspective here.
A user will likely want a bazel target to get the flatbuffer config from a json config. The rule behind this target could be defined in score_baselibs or in score_communication.
Bazel rules can use subrules, which means we can separate the steps of the conversion.
Having the key fixing in score_communication while having the conversion itself in score_baselibs is feasible.
Breaking the public API has no real benefit at this moment since this is a pure mechanical change with no improvement for the user.

Thus, IMO we should have the core logic of the conversion in score_baselibs (for reuse by others) and only our specific key bending in score_communication. The Bazel rule that we provide to the user would then be a combination of both steps.

If we end up finding a second case where keys must also be bent in configs of another module, we should shift this logic to score_baselibs directly. At that moment it would not be specific to score_communication anymore but a migration pattern for multiple modules.

OliverHeilwagen: Sounds good, the core logic is available in baselibs from my point of view:
- [serialize_buffer](https://github.com/eclipse-score/baselibs/blob/b99dbd63788ad4345642c26e7c959da6ca99e50f/score/flatbuffers/bazel/tools.bzl#L149) or [serialize_versioned_buffer](https://github.com/eclipse-score/baselibs/blob/b99dbd63788ad4345642c26e7c959da6ca99e50f/score/flatbuffers/bazel/tools.bzl#L348)
- generate_cpp
- generate_json_schema

A user from com should be able to invoke something like:
```
# bazel rule defined by com
generate_com_config(
    name = "demo_config",
    data = "demo_config.json",
    output = "demo_config.bin",
)
```

The com specifc rule can then invoke:
- json_to_flatbuffer.py for json format compatibility.
- serialize_buffer or serialize_versioned_buffer rule from baselibs with hardcoded com fbs schema (flatc --binary).

The schema generation support of generate_schema.py can be brought to baselibs, as it is a generic extension of the `flatc --jsonschema` feature.

Thomas-Mikhael: I created the PR
https://github.com/eclipse-score/baselibs/pull/474

Thomas-Mikhael:
Not stale. It's pending on the changes in baselibs to be merged.

Thomas-Mikhael:
@LittleHuba This is now ready for reviewing and merging.

score/mw/com/impl/configuration/converter/json_to_flatbuffer_rule.bzl
```diff
@@ -0,0 +1,107 @@
...
+    )
+
+    # Step 2: flatc --binary (via baselibs) on the normalized JSON.
+    serialize_buffer(
+        name = name,
+        data = ":" + normalized_target,
+        schema = fbs,
+        output = name + ".bin",
+        **kwargs
+    )
```
LittleHuba: Please follow conventions in https://bazel.build/versions/8.7.0/extending/macros#conventions

score/mw/com/impl/configuration/flatbuffers_flags.bzl
```diff
@@ -0,0 +1,41 @@
...
+        name = "experimental_enable_flatbuffers_configuration",
+        build_setting_default = False,
+        visibility = ["//score/mw/com/impl/configuration:__subpackages__"],
+    )
+
+    native.config_setting(
+        name = "flatbuffers_enabled",
+        flag_values = {":experimental_enable_flatbuffers_configuration": "True"},
+        visibility = ["//score/mw/com/impl/configuration:__subpackages__"],
+    )
```
LittleHuba: I don't really see why this is a macro at all...
We should follow this style: https://github.com/eclipse-score/communication/blob/main/score/memory/shared/flags/BUILD
Create a package in score/mw/com/flags and just put bool_flag and config_setting there.

score/mw/com/impl/configuration/BUILD
```diff
@@ -12,9 +12,82 @@
...
+
+# FlatBuffers feature flag + config_setting (enable via --config=flatbuffers).
+flatbuffers_flags()
+
+exports_files(
+    [
+        "mw_com_config.fbs",
+        "mw_com_config_schema.json",
+    ],
+    visibility = ["//score/mw/com/impl/configuration:__subpackages__"],
```
LittleHuba, line 27: Is this visibility not too restrictive? How would users depend on the schemas?

score/mw/com/impl/configuration/generate_schema.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def build_arg_parser():
+    """Argument parser for the CLI entry point (schema_drift_test.py parses --in separately
+    before handing remaining args to unittest)."""
+    here = os.path.dirname(os.path.abspath(__file__))
+    workspace = os.environ.get("BUILD_WORKSPACE_DIRECTORY")
+    default_output = (
+        os.path.join(workspace, _SCHEMA_RELPATH) if workspace else os.path.join(here, "mw_com_config_schema.json")
+    )
```
LittleHuba: Please assign the output file via Bazel. See https://github.com/eclipse-score/tooling/blob/main/manual_analysis/manual_analysis.bzl#L173 as an example.
This allows bazel to understand that this file is an output of that rule.

score/mw/com/impl/configuration/BUILD
Thomas-Mikhael: Users depend on the schema via the mw_com_config_schema filegroup, this is the actual public-facing target (already scoped wider to //score/mw/com:__subpackages_)
Check filegroup mw_com_config_schema in this file

score/mw/com/impl/configuration/generate_schema.py
Thomas-Mikhael: Thanks for the hint. Adjusted now.

score/mw/com/impl/configuration/flatbuffers_flags.bzl
Thomas-Mikhael: Done

score/mw/com/impl/configuration/converter/json_to_flatbuffer_rule.bzl
Thomas-Mikhael: Updated to use macros convention

score/mw/com/impl/configuration/BUILD
LittleHuba: I think we actually need to make that target publicly visible. If I don't have a full knot in my head, users will need that schema on their side to actually validate their config at compile time.
Without access to the schema, they have no chance to actually validate it, no?

I think we should expose both the json schema and also the flatbuffer schema.

LittleHuba (dismissed):
Visibility change can be done in a follow-up. No need to block the other parts for this, as current situation is not worsened.

quality/visibility_guard/public_targets.golden
LittleHuba: Disagree on this. This would in essence make the feature impossible to use even when the user enables the flag. Exposing this does not break anything on user side as long as he does not touch the flag. E.g. qualification efforts would not be impacted on user side, as long as he does not use the feature.

LittleHuba (approved):

LittleHuba (changes requested):
Visibility change is partially wrong

score/mw/com/impl/configuration/BUILD
```diff
@@ -12,10 +12,85 @@
...
 load("//bazel/tools:json_schema_validator.bzl", "validate_json_schema_test")
 load("//quality/unit_testing:unit_testing.bzl", "cc_unit_test")
 load("//score/mw:common_features.bzl", "COMPILER_WARNING_FEATURES")
 
+exports_files(
+    [
+        "mw_com_config.fbs",
+        "mw_com_config_schema.json",
+    ],
+    visibility = ["//visibility:public"],
```
LittleHuba, line 27: Sorry my bad. Having the aliases [here](https://github.com/eclipse-score/communication/pull/738/changes#diff-ddfd00285269d0a8d1275f566736ac678491641b79cf6451b463984571b9cf34) means we don't need the public visiblity here!

Thomas-Mikhael: Updated

LittleHuba (approved):

PR #745, opened 2026-07-22
Remove deprecated score::StringLiteral

crimson11:
Fixed wrong/outdated structural diagrams.

Removed usage of deprecated score::StringLiteral in Runtime.
Replaced with const safecpp::zstring_view on implementation
level. Public/user facing APIs using StringLiteral or char*
have been marked deprecated and overloads taking
safecpp::zstring_view have been introduced.

### DIFF
--- a/score/mw/com/api_surface.lock.json
+++ b/score/mw/com/api_surface.lock.json
@@ -29,7 +29,7 @@
       "name": "InitializeRuntime",
       "qualified_name": "score::mw::com::runtime::InitializeRuntime",
       "kind": "function",
-      "signature": "InitializeRuntime : void (const std::int32_t, score::StringLiteral *)"
+      "signature": "InitializeRuntime : void (const cpp::span<safecpp::zstring_view>)"
     },
     {
       "name": "InitializeRuntime",
@@ -47,7 +47,7 @@
       "name": "RuntimeConfiguration",
       "qualified_name": "score::mw::com::runtime::RuntimeConfiguration::RuntimeConfiguration",
       "kind": "constructor",
-      "signature": "RuntimeConfiguration : void (const std::int32_t, score::StringLiteral *)"
+      "signature": "RuntimeConfiguration : void (cpp::span<safecpp::zstring_view>)"
     },
     {
       "name": "RuntimeConfiguration",
--- a/score/mw/com/design/configuration/structural_view.puml
+++ b/score/mw/com/design/configuration/structural_view.puml
@@ -201,8 +201,7 @@ class "score::mw::com::impl::Runtime" {
   - configuration : Configuration const&
   __
   - Runtime(std::pair<Configuration&&, std::optional<tracing::TracingFilterConfig>&&> configs)
-  {static} + Initialize() : void
-  {static} + Initialize(const score::cpp::span<const score::StringLiteral> arguments) : void
+  {static} + Initialize(const runtime::RuntimeConfiguration& runtime_configuration) : void
   {static} + getInstance() : Runtime&
   + resolve(const InstanceSpecifier&): score::mw::com::InstanceIdentifierContainer
   + GetBindingRuntime(const BindingType binding) const : IBindingRuntime*
--- a/score/mw/com/design/configuration/structural_view_refactored.puml
+++ b/score/mw/com/design/configuration/structural_view_refactored.puml
@@ -129,8 +129,7 @@ class "score::mw::com::impl::Runtime" {
   - configuration_ : Configuration const&
   - parsing_strategy_ : std::unique_ptr<IConfigurationParsingStrategy>
   __
-  {static} + Initialize() : void
-  {static} + Initialize(arguments : score::cpp::span<const score::StringLiteral>) : void
+  {static} + Initialize(const runtime::RuntimeConfiguration& runtime_configuration) : void
   {static} + getInstance() : Runtime&
   + resolve(const InstanceSpecifier&) : InstanceIdentifierContainer
 }
--- a/score/mw/com/design/ipc_tracing/structural_view.puml
+++ b/score/mw/com/design/ipc_tracing/structural_view.puml
@@ -99,9 +99,7 @@ abstract class "mw::com::impl::IRuntime" {
 
 class "mw::com::impl::Runtime" {
   -binding_runtimes_ : std:unordered_map<BindingType, std::unique_ptr<IBindingRuntime>>
-  {static} +Initialize() : void
-  {static} +Initialize(int argc, score::StringLiteral argv) : void
-  {static} +Initialize(std::string const&) : void
+  {static} +Initialize(runtime::RuntimeConfiguration&) : void
   {static} +getInstance() : Runtime&
   +Runtime(Configuration&& config)
   +resolve(const InstanceSpecifier&) : std::vector<InstanceIdentifier>
--- a/score/mw/com/design/runtime/runtime_structural_view.puml
+++ b/score/mw/com/design/runtime/runtime_structural_view.puml
@@ -13,9 +13,7 @@ enum BindingType {
 
 class "mw::com::impl::Runtime" as Runtime {
     -binding_runtimes_ : std::unordered_map<BindingType, std::unique_ptr<IBindingRuntime>>
-    +{static} Initialize() : void
-    +{static} Initialize(int argc, score::StringLiteral argv) : void
-    +{static} Initialize(std::string const&) : void
+    +{static} Initialize(const runtime::RuntimeConfiguration& runtime_configuration) : void
     +{static} getInstance() : Runtime&
     +Runtime(Configuration&& config)
     +resolve(const InstanceSpecifier&) : std::vector<InstanceIdentifier>
--- a/score/mw/com/design/skeleton_proxy/proxy_binding_model.puml
+++ b/score/mw/com/design/skeleton_proxy/proxy_binding_model.puml
@@ -11,7 +11,7 @@ class "ComErrorDomain" {
 }
 
 class "<< Stereotype >>\nglobal function" as GlobalFunction {
-  +MakeError(code : ComErrc, message : score::StringLiteral) : score::result::Error
+  +MakeError(code : ComErrc, message : std::string_view) : score::result::Error
 }
 
 note top of ScoreResultErrorDomain
--- a/score/mw/com/design/skeleton_proxy/skeleton_proxy_binding_model.puml
+++ b/score/mw/com/design/skeleton_proxy/skeleton_proxy_binding_model.puml
@@ -11,7 +11,7 @@ class "ComErrorDomain" {
 }
 
 class "<< Stereotype >>\nglobal function" as GlobalFunction {
-  +MakeError(code : ComErrc, message : score::StringLiteral) : score::result::Error
+  +MakeError(code : ComErrc, message : std::string_view) : score::result::Error
 }
 
 note top of ScoreResultErrorDomain
--- a/score/mw/com/mocking/i_runtime.h
+++ b/score/mw/com/mocking/i_runtime.h
@@ -14,11 +14,9 @@
 #define SCORE_MW_COM_MOCKING_I_RUNTIME_H
 
 #include "score/mw/com/runtime_configuration.h"
-
 #include "score/mw/com/types.h"
 
 #include "score/result/result.h"
-#include "score/string_manipulation/string_literal.h"
 
 #include <score/span.hpp>
 
@@ -34,7 +32,11 @@ class IRuntime
     virtual ~IRuntime() = default;
 
     virtual score::Result<InstanceIdentifierContainer> ResolveInstanceIDs(const InstanceSpecifier model_name) = 0;
-    virtual void InitializeRuntime(const std::int32_t argc, score::cpp::span<const score::StringLiteral> argv) = 0;
+    [[deprecated(
+        "Please use InitializeRuntime(cpp::span<safecpp::zstring_view> command_line_arguments) for guaranteed NULL "
+        "terminated arguments")]]
+    virtual void InitializeRuntime(const std::int32_t argc, score::cpp::span<const char*> argv) = 0;
+    virtual void InitializeRuntime(const cpp::span<safecpp::zstring_view> command_line_arguments) = 0;
     virtual void InitializeRuntime(const runtime::RuntimeConfiguration& runtime_configuration) = 0;
 
   protected:
--- a/score/mw/com/mocking/runtime_mock_test.cpp
+++ b/score/mw/com/mocking/runtime_mock_test.cpp
@@ -18,7 +18,6 @@
 #include "score/mw/com/runtime_configuration.h"
 
 #include "score/filesystem/path.h"
-#include "score/string_manipulation/string_literal.h"
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
@@ -82,6 +81,7 @@ TEST_F(RuntimeMockFixture, ResolveInstanceIdsReturnsErrorWhenMockReturnsError)
     EXPECT_EQ(resolve_instance_ids_result.error(), error_code);
 }
 
+// TODO: Once InitializeRuntime(argc, argv) is removed, this test will also be removed
 TEST_F(RuntimeMockFixture, InitializeRuntimeDispatchesToMockAfterInjectingMock)
 {
     // Given that a mocked runtime has been injected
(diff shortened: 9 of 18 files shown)


limdor (changes requested):
const char * has multiple problems and I do not think this should be the substitute for StringLiteral.
- An `std::string` can be converted implicitly from `const char *` ending with dynamic memory allocation
- `const char*` has the problem that it could be a pointer to characters or a pointer to an array of bytes. `char` meaning is way too overloaded in the standard
- The fact that it is a pointer, it can end up in all the issues with pointer decay
- It has no notion of being null terminated or not

Because of that, my proposal would be to go for better alternatives.
- For main arguments, go with a container of null terminated string views. We can use https://github.com/eclipse-score/baselibs/tree/main/score/language/safecpp/string_view and https://github.com/eclipse-score/baselibs/blob/main/score/string_manipulation/arguments/arguments.h for that.
- For global constant string literals with internal linkage, we can use zstring_view https://github.com/eclipse-score/baselibs/tree/main/score/language/safecpp/string_view
- For the rest, I would propose to take a look one by one, depending if they are owning or not, and depending if they are null terminated or not.

Additional note: I saw that in several places you removed the usage of StringLiteral, but the dependency to the target and the include is still there. If we remove StringLiteral we should remove it completly and also remove the include files. Of course it can happen that we break some code if someone relies on the transitive dependency but imho this is a price the users need to pay if they did not have proper includes.

score/mw/com/runtime_configuration.cpp
```diff
@@ -35,6 +35,25 @@ constexpr auto kDefaultConfigurationPath = "./etc/mw_com_config.json";
 constexpr auto kDeprecatedConfigurationPathCommandLineKey = std::string_view{"-service_instance_manifest"};
 constexpr auto kConfigurationPathCommandLineKey = std::string_view{"--service_instance_manifest"};
 
+void VerifyNullTermination(const std::int32_t argc, const char* argv[])
```
limdor: This is something that you cannot do. The problem of this is that if it is null terminated, you will find it, but if it is not null terminated, you hit undefined behavior.
In the second case, it could even be that you hit some null character in memory that it is totally unrelated.

LittleHuba (commented):
Please use C++ infrastructure. E.g. zstring_view or Arguments helper in baselibs.

crimson11:
OK. I guess, I will close this PR then. The thing is: All the usage of `const char*` stems from the `main()` entry-point! I.e. there you get `const char*`, which are guaranteed to be NULL-terminated.
I don't know, how to fix the "issue" then! We expect the user to hand over the `char*` he gets from the OS in main(). If you now want me to change the signature from our PUBLIC API, where the user currently hands over these `char*` from main() ... I can't do it.

@limdor , @LittleHuba : Maybe I'm wrong, but to me it seems, that all your input means/can only solved via changng the PUBLIC interface. I.e. require the user NOT to hand over the `char*` he gets from OS (cmd-line-args) .... if he then also needs to transform these char* to somethings else what is discussed here ... this might also mean eventually heap-allocation ...?

limdor:
Yes, it means breaking public API. But the first step would be to provide a second overload that allows the user to use a container of zstring_view and deprecate the API that uses `const char *`.
And yes, depending how we do it, it might require allocation at startup for the container, but it can also be an allocation aware type.

score/mw/com/design/configuration/structural_view.puml
```diff
@@ -202,7 +202,7 @@ class "score::mw::com::impl::Runtime" {
   __
   - Runtime(std::pair<Configuration&&, std::optional<tracing::TracingFilterConfig>&&> configs)
   {static} + Initialize() : void
-  {static} + Initialize(const score::cpp::span<const score::StringLiteral> arguments) : void
+  {static} + Initialize(const score::cpp::span<const char*> arguments) : void
```
limdor: Why this change? This seems to already be wrong in the main branch.
Can you fix this in advance in a separate PR? It should only be one Initialize method
https://github.com/eclipse-score/communication/blob/e8abf7262b9e89f6f4bd7310045e1e217f8a6830/score/mw/com/impl/runtime.h#L77

score/mw/com/design/configuration/structural_view_refactored.puml
```diff
@@ -130,7 +130,7 @@ class "score::mw::com::impl::Runtime" {
   - parsing_strategy_ : std::unique_ptr<IConfigurationParsingStrategy>
   __
   {static} + Initialize() : void
-  {static} + Initialize(arguments : score::cpp::span<const score::StringLiteral>) : void
+  {static} + Initialize(arguments : score::cpp::span<const char*>) : void
```
limdor: Same as in https://github.com/eclipse-score/communication/pull/745/changes#r3656356347

score/mw/com/design/ipc_tracing/structural_view.puml
```diff
@@ -99,9 +99,7 @@ abstract class "mw::com::impl::IRuntime" {
 
 class "mw::com::impl::Runtime" {
   -binding_runtimes_ : std:unordered_map<BindingType, std::unique_ptr<IBindingRuntime>>
-  {static} +Initialize() : void
-  {static} +Initialize(int argc, score::StringLiteral argv) : void
-  {static} +Initialize(std::string const&) : void
+  {static} +Initialize(runtime::RuntimeConfiguration&) : void
```
limdor, line 102: Can we move this changes to a separate PR together with https://github.com/eclipse-score/communication/pull/745/changes#r3656356347? That is not really a change in the API, it is just fixing the puml that was wrong.

score/mw/com/design/runtime/runtime_structural_view.puml
```diff
@@ -14,7 +14,7 @@ enum BindingType {
 class "mw::com::impl::Runtime" as Runtime {
     -binding_runtimes_ : std::unordered_map<BindingType, std::unique_ptr<IBindingRuntime>>
     +{static} Initialize() : void
-    +{static} Initialize(int argc, score::StringLiteral argv) : void
+    +{static} Initialize(int argc, const char* argv) : void
```
limdor: Same as https://github.com/eclipse-score/communication/pull/745/changes#r3656356347

score/mw/com/design/skeleton_proxy/proxy_binding_model.puml
```diff
@@ -11,7 +11,7 @@ class "ComErrorDomain" {
 }
 
 class "<< Stereotype >>\nglobal function" as GlobalFunction {
-  +MakeError(code : ComErrc, message : score::StringLiteral) : score::result::Error
+  +MakeError(code : ComErrc, message : const char*) : score::result::Error
```
limdor:
```suggestion
  +MakeError(code : ComErrc, message : const std::string_view) : score::result::Error
```
I believe this should be `std::string_view`. This is what we have in the implementation
https://github.com/eclipse-score/communication/blob/e8abf7262b9e89f6f4bd7310045e1e217f8a6830/score/mw/com/impl/com_error.h#L224
Because this is just fixing the puml, it would be good if we can move it out of this PR, because this is not an API change but fixing a mismatch in the puml.

Another thing about this but it was already there before your change, it seems that this says in the puml that it is a global function but it does not seem to mention the namespace. Not sure how you reflect that in puml, I'm not very familiar with it yet.

score/mw/com/design/skeleton_proxy/skeleton_proxy_binding_model.puml
```diff
@@ -11,7 +11,7 @@ class "ComErrorDomain" {
 }
 
 class "<< Stereotype >>\nglobal function" as GlobalFunction {
-  +MakeError(code : ComErrc, message : score::StringLiteral) : score::result::Error
+  +MakeError(code : ComErrc, message : std::string_view) : score::result::Error
```
limdor, line 14: Same as https://github.com/eclipse-score/communication/pull/745/changes#r3656421146

limdor (commented):
For the moment I reviewed all changes in the design folder. I think it would be better to move them in a separate PR becasue how I see them is that none of them is related to an API change, it is just because the puml was wrong.

Then for the API additions, I did not review yet. But I assume we will have to update some puml files.

score/mw/com/design/skeleton_proxy/proxy_binding_model.puml
```diff
@@ -11,7 +11,7 @@ class "ComErrorDomain" {
 }
 
 class "<< Stereotype >>\nglobal function" as GlobalFunction {
-  +MakeError(code : ComErrc, message : score::StringLiteral) : score::result::Error
+  +MakeError(code : ComErrc, message : const std::string_view) : score::result::Error
```
limdor: The const does not matter a lot here considering that it is just a signature, but there is an inconsistency with score/mw/com/design/skeleton_proxy/skeleton_proxy_binding_model.puml
We should spell it the same way in both places.

score/mw/com/mocking/i_runtime.h
```diff
@@ -34,7 +34,8 @@ class IRuntime
     virtual ~IRuntime() = default;
 
     virtual score::Result<InstanceIdentifierContainer> ResolveInstanceIDs(const InstanceSpecifier model_name) = 0;
-    virtual void InitializeRuntime(const std::int32_t argc, score::cpp::span<const score::StringLiteral> argv) = 0;
+    virtual void InitializeRuntime(const std::int32_t argc, score::cpp::span<const char*> argv) = 0;
```
limdor, line 38:
```suggestion
    [[deprecated("Please use InitializeRuntime(cpp::span<safecpp::zstring_view> command_line_arguments) for guaranteed NULL terminated arguments")]]
    virtual void InitializeRuntime(const std::int32_t argc, score::cpp::span<const char*> argv) = 0;
```
If we add the deprecation attribute also in the interface it also serve us as a reminder that this is deprecated.

score/mw/com/mocking/runtime_mock_test.cpp
```diff
@@ -89,10 +89,10 @@ TEST_F(RuntimeMockFixture, InitializeRuntimeDispatchesToMockAfterInjectingMock)
     // Expecting that InitializeRuntime will be called on the mock with the same argc / argv that is passed to
     // InitializeRuntime (where argv has been converted to an score::cpp::span
     constexpr std::int32_t argc{1U};
-    score::StringLiteral argv[] = {"some_argument"};
```
limdor, line 92: Please add a note to this test mentioning that once the deprecated InitializeRuntime(argc, argv) is removed, the test needs to be removed also. This helps to identify that these char* here are only needed for testing this deprecated API.
```
// TODO: Once InitializeRuntime(argc, argv) is removed, this test will also be removed
TEST_F(RuntimeMockFixture, InitializeRuntimeDispatchesToMockAfterInjectingMock)
```

score/mw/com/mocking/runtime_mock_test.cpp
limdor, line 131: I believe a test is missing for the new added `InitializeRuntime(const cpp::span<safecpp::zstring_view> command_line_arguments)`

score/mw/com/api_surface.lock.json
```diff
@@ -29,7 +29,7 @@
       "name": "InitializeRuntime",
       "qualified_name": "score::mw::com::runtime::InitializeRuntime",
       "kind": "function",
-      "signature": "InitializeRuntime : void (const std::int32_t, score::StringLiteral *)"
+      "signature": "InitializeRuntime : void (const cpp::span<safecpp::zstring_view>)"
```
limdor, line 32: That is interesting, I do not get why you need to change it and you do not have to add it. I would expect that if we have both (the new overload and the deprecated API) then you have to add one and keep also the old.

score/mw/com/runtime.cpp
```diff
@@ -49,12 +49,12 @@ score::Result<InstanceIdentifierContainer> ResolveInstanceIDs(const impl::Instan
 }
 
 // NOLINTNEXTLINE(modernize-avoid-c-arrays):C-style array tolerated for command line arguments
```
limdor:
```suggestion
// NOLINTNEXTLINE(modernize-avoid-c-arrays):C-style array tolerated for command line arguments. This API is deprecated and it will be removed.
```

score/mw/com/mocking/i_runtime.h
limdor:
```suggestion
```
Or we delete it, or we put a comment saying that we need to remove it. With your changes it would only be an issue if someone was relying on the transitive include.

score/mw/com/mocking/runtime_mock_test.cpp
limdor:
```suggestion
```
If you removed them all, I assume this can go away.

score/mw/com/runtime_configuration.cpp
```diff
@@ -60,15 +76,15 @@ const filesystem::Path& RuntimeConfiguration::GetConfigurationPath() const&
...
 std::optional<filesystem::Path> RuntimeConfiguration::ParseConfigurationPath(
-    const score::cpp::span<const score::StringLiteral> command_line_args)
+    const score::cpp::span<safecpp::zstring_view> command_line_args)
 {
     const auto num_args = command_line_args.size();
 
     std::optional<std::string> configuration_path{};
     for (std::uint32_t arg_idx = 0U; arg_idx < num_args; arg_idx++)
     {
         const std::string& command_line_argument_key{
```
limdor, line 85:
```suggestion
        // TODO: Adapt code that we do not need to call `.data()` and that we do not have to rely on life time extension
        const std::string& command_line_argument_key{
```

score/mw/com/runtime_configuration.cpp
```diff
@@ -79,7 +95,7 @@ std::optional<filesystem::Path> RuntimeConfiguration::ParseConfigurationPath(
                     << "\" but no corresponding value. Terminating.";
                 std::terminate();
             }
-            return score::cpp::at(command_line_args, static_cast<std::ptrdiff_t>(index_of_configuration_path));
+            return score::cpp::at(command_line_args, static_cast<std::ptrdiff_t>(index_of_configuration_path)).data();
```
limdor, line 98:
```suggestion
            // TODO: Adapt code that we do not need to call `.data()` and we can provide zstring_view to a filepath
            return score::cpp::at(command_line_args, static_cast<std::ptrdiff_t>(index_of_configuration_path)).data();
```

score/mw/com/runtime_configuration.cpp
```diff
@@ -93,7 +109,7 @@ std::optional<filesystem::Path> RuntimeConfiguration::ParseConfigurationPath(
...
```
limdor, line 113:
```suggestion
            // TODO: Adapt code that we do not need to call `.data()` and we can provide zstring_view to a filepath
            return score::cpp::at(command_line_args, static_cast<std::ptrdiff_t>(index_of_configuration_path)).data();
```

score/mw/com/runtime_configuration_test.cpp
```diff
@@ -123,7 +142,7 @@ TEST(RuntimeConfigurationCommandLineConstructorTest, DeprecatedConfigurationComm
 TEST(RuntimeConfigurationCommandLineConstructorTest, ConfigurationPathContainsDefaultPathIfNoPathKeyInCommandLineArgs)
 {
     // Given command line arguments which do not contain a configuration path key
-    std::vector<score::StringLiteral> arguments = {kDummyApplicationName, kDummyConfigurationPath};
+    std::vector<const char*> arguments = {kDummyApplicationName, kDummyConfigurationPath};
```
limdor: Now most of the adapted tests in this file will call a deprecated function. If we do not fix it right away we will have to do it when we remove the deprecation function or in a separate ticket.
All 3 options are fine for me, but it would make sense to fix them right away to show that the approach works.

score/mw/com/runtime_test.cpp
```diff
@@ -149,7 +149,7 @@ TEST_F(RuntimeTestExternalJsonFixture, ResolveInstanceIdsIsThreadSafe)
 
     // Given a runtime initialised with the path to a configuration file
     auto json_path = get_path("mw_com_config.json");
-    score::StringLiteral test_args[] = {"dummyname", "-service_instance_manifest", json_path.c_str()};
+    const char* test_args[] = {"dummyname", "-service_instance_manifest", json_path.c_str()};
```
limdor: Same here, now this test calls a deprecated function.

limdor (commented):
I added some additional comments.
In general there are several dependency that are still there that could be removed. If you are removing all instances of StringLiteral (that is what I understood) then there should be no more includes of it in our code base and also no more mention of the string literal Bazel target, however there are still quite some, also in places where you removed them.

score/mw/com/api_surface.lock.json
LittleHuba: @castler this looks like the API surface checker has issues with overloads.

score/mw/com/mocking/i_runtime.h
crimson11: It is test-only! If someone relied onj transitive dependency and then his tests explode ... so be it :P

score/mw/com/runtime_configuration_test.cpp
crimson11: Refactored the test -> they do not use deprecated APIs anymore.

limdor (approved):

PR #750, opened 2026-07-23
Add CI design

limdor:


CI.md
```diff
@@ -0,0 +1,208 @@
...
+- SCOM-PG-04: Use resources efficiently
+- SCOM-PG-05: Satisfy all functional requirements
+- SCOM-PG-06: Satisfy all none-functional requirements
+- SCOM-PG-07: Satisfy our processes and enhancement proposals
+
+## Supported configurations
+
+S-CORE Communication supports multiple architectures, compilers, and operating systems. Some of them are only supported for development and
+some are supported for production.
+In the CI we do not test we do not test all combinations in all pipelines, we try to find a combination that helps us to find most of the issues
```
castler: `we do not test` is double

CI.md
```diff
@@ -0,0 +1,208 @@
...
+| x86-64 QNX   | NA*                  | NA*                  | Yes                     | NA*                  |
+| ARM64 QNX    | NA*                  | NA*                  | Yes                     | NA*                  |
+
+Note: Configuration marked as "NA" are considered as not applicable because the architecture/compiler/operating system configuration is in general not supported.
+
+And this is the list of bazel `--config=` for the supported configurations:
+
+|              | GCC 12.2 + libstdc++                          | GCC 15 + libstdc++ | QCC (GCC 12.2) + libc++ | Clang 22 + libc++        |
+|--------------|-----------------------------------------------|--------------------|-------------------------|--------------------------|
+| x86-64 linux | --config=linux_x86_64_score_gcc_12_2_0_posix  | (default)          | NA                      | Config not yet available |
```
castler: clang22 uses libstdc++: https://github.com/eclipse-score/communication/blob/main/MODULE.bazel#L305
And a config is avialable: `--extra_toolchains=@llvm_toolchain//:cc-toolchain-x86_64-linux`

CI.md
```diff
@@ -0,0 +1,208 @@
...
+If the tool used for sanitizers is ever changed, this decision could be revisited.
+
+#### Run C++ unit tests under thread sanitizer (x86-64 linux)
+
+Runs all C++ unit tests under thread sanitizer.
+
+Each unit test target also has a maximum run time [as the job for unit tests without sanitizers](./CI.md#build-everything-and-run-unit-tests-host)
+and the same stability design where a test is rerun 3 times in case of failure.
+
+In case that a test had to be retried, that information is uploaded to some storage and made it visible in Power BI. Such
```
castler: We have no Power BI available

castler (commented):
We are missing the infrastructure jobs (e.g nightly cache recreation) and the nightly docu build

CI.md
limdor: Fixed

CI.md
limdor: Fixed

limdor:
> We are missing the infrastructure jobs (e.g nightly cache recreation) and the nightly docu build

Added. Let me know if you see some additional one missing.

CI.md
limdor: Thanks for making me aware of it, but I do not think we should use libstdc++ with clang on the host. The problem is that if we have only QNX with libc++, then we loose a lot of the oportunities for using the hardened version of the standard library provided by libc++ (see https://libcxx.llvm.org/Hardening.html). We will be limited to the version used in the target that it is not so complete (I think it uses libc++ 16 but not sure about that) and we will be able to run it on QNX.
We will still be able to use the debug mode from libstdc++ (https://gcc.gnu.org/onlinedocs/libstdc++/manual/debug_mode.html) but we will loose the latest hardening from libc++.

When it came to configs, what I meant is that we do not have a `--config=` available. Anyway I updated the table to reflect what we would like to have as a `--config=` regardles of what we have right now.

castler (approved):

PR #756, opened 2026-07-23
mw/com: Add additional Field tests

bemerybmw:
Takes over: https://github.com/eclipse-score/communication/pull/707

Depends-on: https://github.com/eclipse-score/communication/pull/735

### DIFF
--- a/score/mw/com/impl/proxy_field_test.cpp
+++ b/score/mw/com/impl/proxy_field_test.cpp
@@ -19,13 +19,16 @@
 
 #include "score/mw/com/impl/proxy_field.h"
 
-#include "score/mw/com/impl/runtime.h"
-#include "score/mw/com/impl/runtime_mock.h"
-#include "score/mw/com/impl/test/binding_factory_resources.h"
-#include "score/mw/com/impl/test/proxy_resources.h"
+#include "score/mw/com/impl/bindings/mock_binding/proxy_event.h"
+#include "score/mw/com/impl/bindings/mock_binding/proxy_method.h"
+#include "score/mw/com/impl/com_error.h"
+#include "score/mw/com/impl/test/runtime_mock_guard.h"
 
 #include <gmock/gmock.h>
 #include <gtest/gtest.h>
+#include <cstddef>
+#include <memory>
+#include <string_view>
 #include <type_traits>
 
 namespace score::mw::com::impl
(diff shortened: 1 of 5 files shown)


score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -330,5 +338,158 @@ TEST(ProxyFieldNotifierGatingTest, OnlyNotifierApiExistsWhenNoTagsArePresent)
...
+    ProxyField<TestSampleType, WithGetter> CreateFieldWithGetOnly()
+    {
+        return ProxyField<TestSampleType, WithGetter>{
+            kFieldName,
+            std::make_unique<mock_binding::ProxyEvent<TestSampleType>>(),
+            nullptr,
+            std::make_unique<mock_binding::ProxyMethodFacade>(get_method_binding_mock_)};
+    }
+
+    ProxyField<TestSampleType, WithSetter, WithNotifier> CreateFieldWithSet()
```
KrishaDeshkool, line 372: CreateFieldWithSetAndNotifier() was honest name.

KrishaDeshkool: also it is not possible to create a filed only with Set tag, so this name is misleading.

bemerybmw: But for the purposes of the test, it's not relevant that it has a getter or notifier. It's just important that it has a setter. The fixture can take care of the required setup (i.e. choosing whether we have a getter or notifier). There are always details abstracted away e.g. we also don't say CreateFieldWithSetAndNotifierWithValidMockBindings

score/mw/com/impl/proxy_field_test.cpp
```diff
@@ -19,13 +19,16 @@
 
 #include "score/mw/com/impl/proxy_field.h"
 
-#include "score/mw/com/impl/runtime.h"
-#include "score/mw/com/impl/runtime_mock.h"
-#include "score/mw/com/impl/test/binding_factory_resources.h"
-#include "score/mw/com/impl/test/proxy_resources.h"
```
KrishaDeshkool, line 25: does BUILD deps need some cleanup after these header removals?

KrishaDeshkool: agreed to do in a followup

KrishaDeshkool (approved):

score/mw/com/impl/proxy_field_test.cpp
bemerybmw: Will fix in a follow up

LittleHuba (approved):

PR #757, opened 2026-07-23
CodeQL: Have better pack selection (use really the vendored one)

castler:


MODULE.bazel
```diff
@@ -443,8 +436,8 @@ http_archive = use_repo_rule("@bazel_tools//tools/build_defs/repo:http.bzl", "ht
 http_archive(
     name = "codeql_bundle",
     build_file = "//third_party/codeql:codeql.BUILD",
-    sha256 = "0144d4bc415aee0d5638119dfb626a2d8689e2ff21758ba211a3984862b8522c",
-    url = "https://github.com/github/codeql-action/releases/download/codeql-bundle-v2.26.0/codeql-bundle-linux64.tar.gz",
+    sha256 = "a94f674bb3c23ea5e9a2ad06b64847dd0277b15014d2517ecd9c41c88e6caa65",
+    url = "https://github.com/github/codeql-action/releases/download/codeql-bundle-v2.21.4/codeql-bundle-linux64.tar.gz",
```
limdor, line 448: How do we know in the future which bundle can we use together with the coding standard version? Should be put at least a comment with a link that explains how they relate to each other and what is supported with what?

MODULE.bazel
```diff
@@ -446,6 +446,8 @@ git_repository(
     name = "codeql_coding_standards",
     build_file = "//third_party/codeql:codeql_coding_standards.BUILD",
     commit = "06dc6bc32b05152fbe94dbf341a3e854574c9df5",  # v2.61.0
+    patch_args = ["-p1"],
+    patches = ["//third_party/codeql:codeql_coding_standards_misra.patch"],
```
limdor, line 455: Can we also create a PR upstream for that? It seems to be that it should be trivial for them to accept

third_party/codeql/codeql_coding_standards_misra.patch
```diff
@@ -0,0 +1,16 @@
...
+--- a/scripts/reports/utils.py
++++ b/scripts/reports/utils.py
+@@ -183,7 +183,10 @@ def generate_guideline_compliance_summary(output_directory, results_summary):
+                 print(
+                     "**Result**: " + ("Not compliant" if total_guidelines_violated > 0 else "Compliant"))
+                 standard_pretty_name = {
+-                    "cert": "CERT C++ 2016", "autosar": "AUTOSAR C++  R22-11, R21-11, R20-11, R19-11 and R19-03"}
++                    "cert": "CERT C++ 2016",
++                    "autosar": "AUTOSAR C++  R22-11, R21-11, R20-11, R19-11 and R19-03",
++                    "misra": "MISRA C++ 2023",
```
limdor: The official name is "MISRA C++:2023"

MODULE.bazel
```diff
@@ -451,6 +451,22 @@ git_repository(
...
+# codeql-coding-standards release (see codeql_release_pack.bzl). This avoids
+# compiling the queries locally while keeping the analysis hermetic (pinned by
+# sha256) and aligned with the pinned coding_standards version above.
+codeql_release_pack = use_repo_rule(
+    "//third_party/codeql:codeql_release_pack.bzl",
+    "codeql_release_pack",
+)
+
+codeql_release_pack(
+    name = "codeql_coding_standards_compiled",
```
limdor, line 471: With this, we have now `codeql_coding_standards_compiled`, `codeql_coding_standards`, and `codeql_bundle`. I guess all 3 have some restriction between them when it comes to the version. The first 2 I understand they should be the same version. Because `codeql_coding_standards` is a `git_repository` and we mention the hash, it is not so easy to keep in sync with the same version, we need to remember to update one when we update the other. I would believe there is a way that we can use something different than `git_repository` and we can also just rever to the version number, in this way we could extract it in a Bazel variable.
For the `codeql_bundle` it would be good to mention how this version relates to the others.

quality/static_analysis/codeql_lint.py
```diff
@@ -22,28 +22,70 @@
...
+    The pack is vendored by the @codeql_coding_standards_compiled repository
+    (see third_party/codeql/codeql_release_pack.bzl): a pre-compiled pack
+    published with the codeql-coding-standards release, containing the compiled
+    queries (`.qlx`), the default suites and all library dependencies bundled
+    under `.codeql/libraries/`. Analyzing against this pack (via an explicit
+    suite path + `--additional-packs`) runs exactly the pinned ruleset without
+    recompiling the queries and without downloading anything from the registry.
 
+    This vendored pack is the ONLY supported query source: if it cannot be
+    located we raise instead of silently falling back to any other pack (e.g. a
```
limdor:
```suggestion
    located we raise an error instead of silently falling back to any other pack (e.g. a
```

quality/static_analysis/codeql_lint.py
```diff
@@ -22,28 +22,70 @@
...
-    """Resolve the MISRA C++ query pack's own dependencies (like `npm install`)."""
-    pack_root = _find_pack_root()
-    subprocess.run(f"{code_ql_path} pack install {pack_root}", shell=True, check=True)
+    Returns the pack root directory (the one containing qlpack.yml).
+    """
+    from python.runfiles import Runfiles
+
+    runfiles = Runfiles.Create()
+    anchor = runfiles.Rlocation(
+        "codeql_coding_standards_compiled/pack/qlpack.yml")
```
limdor: Nitpick: Not sure if the path would change in the future, but it would be better to extract `"codeql_coding_standards_compiled/pack/qlpack.yml"` into a variable and use it also in the `RuntimeError`

MODULE.bazel
castler: I already did: https://github.com/github/codeql-coding-standards/pull/1163

third_party/codeql/codeql_coding_standards_misra.patch
limdor: I commented in https://github.com/github/codeql-coding-standards/pull/1163/changes#r3643423369
Maybe they intentionally avoid special characters

limdor (approved):

PR #758, opened 2026-07-23
mw/com: Add Field Getter support

bemerybmw:
Takes over: https://github.com/eclipse-score/communication/pull/438

### DIFF
--- a/score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
+++ b/score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
@@ -118,13 +118,15 @@ class SkeletonMethodHandlingFixture : public SkeletonMockedMemoryFixture
         foo_method_ = std::make_unique<SkeletonMethod>(*skeleton_, foo_unique_method_id);
         dumb_method_ = std::make_unique<SkeletonMethod>(*skeleton_, dumb_unique_method_id);
 
-        std::ignore = foo_method_->RegisterHandler([this](std::optional<score::cpp::span<std::byte>> in_args,
+        std::ignore = foo_method_->RegisterHandler([this](QualityType quality_type,
+                                                          std::optional<score::cpp::span<std::byte>> in_args,
                                                           std::optional<score::cpp::span<std::byte>> return_arg) {
-            std::invoke(foo_mock_type_erased_callback_.AsStdFunction(), in_args, return_arg);
+            std::invoke(foo_mock_type_erased_callback_.AsStdFunction(), quality_type, in_args, return_arg);
         });
-        std::ignore = dumb_method_->RegisterHandler([this](std::optional<score::cpp::span<std::byte>> in_args,
+        std::ignore = dumb_method_->RegisterHandler([this](QualityType quality_type,
+                                                           std::optional<score::cpp::span<std::byte>> in_args,
                                                            std::optional<score::cpp::span<std::byte>> return_arg) {
-            std::invoke(dumb_mock_type_erased_callback_.AsStdFunction(), in_args, return_arg);
+            std::invoke(dumb_mock_type_erased_callback_.AsStdFunction(), quality_type, in_args, return_arg);
         });
     }
 
@@ -870,16 +872,16 @@ TEST_F(SkeletonOnServiceMethodsSubscribedFixture, CallingRegistersAMethodCallHan
 
     // Expecting that the type erased callback will be called for each method with InArgs and ReturnArg storage
     // provided if TypeErasedElementInfo for the method in MethodData contains InArgs / a ReturnArg
-    EXPECT_CALL(foo_mock_type_erased_callback_, Call(_, _))
-        .WillOnce(Invoke([](auto in_args_optional, auto result_optional) {
+    EXPECT_CALL(foo_mock_type_erased_callback_, Call(QualityType::kASIL_QM, _, _))
+        .WillOnce(WithArgs<1, 2>(Invoke([](auto in_args_optional, auto result_optional) {
             EXPECT_EQ(in_args_optional.has_value(), kFooTypeErasedElementInfo.in_arg_type_info.has_value());
             EXPECT_EQ(result_optional.has_value(), kFooTypeErasedElementInfo.return_type_info.has_value());
-        }));
-    EXPECT_CALL(dumb_mock_type_erased_callback_, Call(_, _))
-        .WillOnce(Invoke([](auto in_args_optional, auto result_optional) {
+        })));
+    EXPECT_CALL(dumb_mock_type_erased_callback_, Call(QualityType::kASIL_QM, _, _))
+        .WillOnce(WithArgs<1, 2>(Invoke([](auto in_args_optional, auto result_optional) {
             EXPECT_EQ(in_args_optional.has_value(), kDumbTypeErasedElementInfo.in_arg_type_info.has_value());
             EXPECT_EQ(result_optional.has_value(), kDumbTypeErasedElementInfo.return_type_info.has_value());
-        }));
+        })));
 
     // Expecting that a method call handler is registered for both methods which calls the handler directly with the
     // largest possible queue index for that method
@@ -907,6 +909,53 @@ TEST_F(SkeletonOnServiceMethodsSubscribedFixture, CallingRegistersAMethodCallHan
     score::cpp::ignore = std::invoke(captured_method_subscribed_handler_qm_.value(),
                                      proxy_instance_identifier_qm_,
                                      test::kAllowedQmMethodConsumer,
+                                     kDummyPid);
+}
+
+TEST_F(SkeletonOnServiceMethodsSubscribedFixture,
+       CallingAsilBRegistersAMethodCallHandlerPerMethodWithInfoFromMethodData)
+{
+    GivenAnAsilBSkeletonWithTwoMethods().WhichCapturesRegisteredMethodSubscribedHandlers().WhichIsOffered();
+
+    // Expecting that the type erased callback will be called for each method with InArgs and ReturnArg storage
+    // provided if TypeErasedElementInfo for the method in MethodData contains InArgs / a ReturnArg
+    EXPECT_CALL(foo_mock_type_erased_callback_, Call(QualityType::kASIL_B, _, _))
+        .WillOnce(WithArgs<1, 2>(Invoke([](auto in_args_optional, auto result_optional) {
+            EXPECT_EQ(in_args_optional.has_value(), kFooTypeErasedElementInfo.in_arg_type_info.has_value());
+            EXPECT_EQ(result_optional.has_value(), kFooTypeErasedElementInfo.return_type_info.has_value());
+        })));
+    EXPECT_CALL(dumb_mock_type_erased_callback_, Call(QualityType::kASIL_B, _, _))
+        .WillOnce(WithArgs<1, 2>(Invoke([](auto in_args_optional, auto result_optional) {
+            EXPECT_EQ(in_args_optional.has_value(), kDumbTypeErasedElementInfo.in_arg_type_info.has_value());
+            EXPECT_EQ(result_optional.has_value(), kDumbTypeErasedElementInfo.return_type_info.has_value());
+        })));
+
+    // Expecting that a method call handler is registered for both methods which calls the handler directly with the
+    // largest possible queue index for that method with quality type ASIL-B
+    EXPECT_CALL(message_passing_mock_,
+                RegisterMethodCallHandler(QualityType::kASIL_B, foo_proxy_method_identifier_b_, _, _))
+        .WillOnce(WithArgs<2>(Invoke([this](auto method_call_handler) {
+            std::invoke(method_call_handler, test::kFooMethodQueueSize - 1U);
+            return MethodCallRegistrationGuardFactory::Create(message_passing_mock_,
+                                                              kDummyQualityType,
+                                                              foo_proxy_method_identifier_b_,
+                                                              method_call_registration_guard_scope_);
+        })));
+    EXPECT_CALL(message_passing_mock_,
+                RegisterMethodCallHandler(QualityType::kASIL_B, dumb_proxy_method_identifier_b_, _, _))
+        .WillOnce(WithArgs<2>(Invoke([this](auto method_call_handler) {
+            std::invoke(method_call_handler, test::kDumbMethodQueueSize - 1U);
+            return MethodCallRegistrationGuardFactory::Create(message_passing_mock_,
+                                                              kDummyQualityType,
+                                                              dumb_proxy_method_identifier_b_,
+                                                              method_call_registration_guard_scope_);
+        })));
+
+    // When calling the asil-b registered method subscribed handler
+    ASSERT_TRUE(captured_method_subscribed_handler_b_.has_value());
+    score::cpp::ignore = std::invoke(captured_method_subscribed_handler_b_.value(),
+                                     proxy_instance_identifier_b_,
+                                     test::kAllowedAsilBMethodConsumer,
                                      kDummyPid);
 }
 
--- a/score/mw/com/impl/configuration/BUILD
+++ b/score/mw/com/impl/configuration/BUILD
@@ -39,7 +39,9 @@ cc_library(
     hdrs = ["quality_type.h"],
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
-    visibility = ["//score/mw/com/impl/bindings/lola/messaging:__pkg__"],
+    visibility = [
+        "//score/mw/com/impl:__subpackages__",
+    ],
 )
 
 # A common library for configuration validation and parsing. It shall be used by both strategies (JSON & Flatbuffers) parsers.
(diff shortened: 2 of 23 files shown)


crimson11:
If you are done with this PR, don't forget to update the tutorial: [Chapter 9](https://eclipse-score.github.io/communication/latest/tutorial/chapter_9/README.html)
I.e. the WARNING has to be removed, because after this PR is merged Field-Getter should work.

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -23,6 +24,39 @@
 
 namespace score::mw::com::impl
 {
+
+/// \brief Enum to indicate whether a method handler callable is expected to have QualityType as its first argument.
+enum class WithQuality
+{
+    TRUE,
+    FALSE
+};
```
Tejveerpratap2803, line 33: I think the enum value should be something like kYes or kNo or something like kWithQualityType/kWithoutQualityType instead of true or false.

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -85,52 +93,81 @@ struct get_method_in_args<std::tuple<TupleArgs...>>
 template <typename T>
 using get_method_in_args_t = typename get_method_in_args<T>::type;
 
-template <FailureMode FailureMode, typename CallableArgs>
+template <FailureMode FailureMode, WithQuality WithQualityType, typename CallableArgs>
```
Tejveerpratap2803: I think for this non-type template parametre should be lowercase as to avoid same template typename and parametre(for example: failure_mode and with_quality_type.)

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -23,6 +23,32 @@
 
 namespace score::mw::com::impl
 {
+
+enum class FailureMode
+{
+    COMPILE_TIME,
+    RUNTIME
+};
```
Tejveerpratap2803, line 39: Suggestion: I think it should be named as kCompileTime and kRunTime.

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -107,6 +131,21 @@ class SkeletonMethod<ReturnType(ArgTypes...)> final : public SkeletonMethodBase
     /// \return score::cpp::blank on success and ComErrc code specified by the binding on failiure
     template <typename Callable>
     Result<void> RegisterHandler(Callable&& callback);
+
+  private:
+    /// \brief Overload for method handlers which accept callables which follow the same signature requirements as
+    /// RegisterHandler except that the first argument of the callable is QualityType.
+    ///
+    /// Currently, this is only used by SkeletonField setters so that they can be called with the QualityType of the
```
Tejveerpratap2803: It should be this is only used by  SkeletonField  getters.

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -124,15 +163,41 @@ SkeletonMethod<ReturnType(ArgTypes...)>::SkeletonMethod(SkeletonBase& skeleton_b
...
+                                               WithQuality::TRUE,
+                                               CallableWithQuality,
+                                               ReturnType,
+                                               ArgTypes...>();
+
+    return RegisterHandlerImpl<WithQuality::TRUE>(std::forward<CallableWithQuality>(user_callback));
+}
+
+template <typename ReturnType, typename... ArgTypes>
+template <WithQuality WithQuality, typename Callable>
```
Tejveerpratap2803, line 201: I think for this non-type template parametre should be lowercase as to avoid same template typename and parametre(for example: with_quality_type.)

score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -107,6 +131,21 @@ class SkeletonMethod<ReturnType(ArgTypes...)> final : public SkeletonMethodBase
     /// \return score::cpp::blank on success and ComErrc code specified by the binding on failiure
     template <typename Callable>
     Result<void> RegisterHandler(Callable&& callback);
+
+  private:
+    /// \brief Overload for method handlers which accept callables which follow the same signature requirements as
+    /// RegisterHandler except that the first argument of the callable is QualityType.
+    ///
+    /// Currently, this is only used by SkeletonField setters so that they can be called with the QualityType of the
+    /// ProxyMethod which made the call. This function can be accessed by the SkeletonField since it's a friend of this
```
Tejveerpratap2803: It should be ProxyField which made call instead of ProxyMethod which made call.

score/mw/com/impl/proxy_field.h
```diff
@@ -309,7 +309,28 @@ class ProxyFieldImpl : public ProxyFieldBase
...
+
+        // If the method call itself succeeded but executing the getter itself on skeleton side failed (e.g. because
+        // reading the field value failed), then we return the error code to the user as well.
+        auto getter_return_type_ptr = std::move(method_call_result).value();
+        if (!getter_return_type_ptr->has_value())
+        {
+            return MakeUnexpected<MethodReturnTypePtr<T>>(std::move(*getter_return_type_ptr).error());
+        }
+
+        // If the method call and executing the setter itself succeeded, then we return a pointer to the set value to
```
Tejveerpratap2803: it should be getter instead of setter. And in the commit message you have mistakenly written wetter instead of getter.

score/mw/com/impl/proxy_field.h
```diff
@@ -309,7 +309,28 @@ class ProxyFieldImpl : public ProxyFieldBase
               typename = std::enable_if_t<is_tag_enabled<T, SampleDataType, WithGetter, Tags...>::value>>
     score::Result<MethodReturnTypePtr<T>> Get() noexcept
     {
-        return proxy_method_get_dispatch_->operator()();
+        // If the method call itself fails, then we return the error code to the user here.
+        auto method_call_result = proxy_method_get_dispatch_->operator()();
```
Tejveerpratap2803, line 313: Suggestion: Get() and Set() share the same three-step unwrap logic check method-call  -> check inner getter/setter -> wrap the value in a  MethodReturnTypePtr<T>. We can extract this logic into a private helper function.

score/mw/com/impl/methods/method_traits_checker.h
bemerybmw: This is an enum class, which means that you always have to reference a value with the enum name. E.g. you can't do something like `WithQuality val = TRUE;` but rather must be `WithQuality val = WithQuality::TRUE;`. So I think it definitely shouldn't be kWithQualityType because then there would be reduntant information: `WithQuality val = WithQuality::kWithQualityType;`.  You're right about the naming though. I checked the coding guidelines and it says that enum names should be like kGlobalConstants.

Tejveerpratap2803 (approved):

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -59,33 +85,56 @@ struct get_method_in_args<std::tuple<TupleArgs...>>
 template <typename T>
 using get_method_in_args_t = typename get_method_in_args<T>::type;
 
-}  // namespace detail
+template <FailureMode FailureMode, typename CallableArgs>
+void AssertCallableHasNoInArgsOrReturn()
```
crimson11: `AssertCallableHasNoInArgsOrReturn` is imho misleading to your comments/usage.

What this "checker" should imho do:
- check that there are no call-args at all.

Why not simply name it
`AssertCallableHasNoArgs`
this is, what it does?

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -59,33 +85,56 @@ struct get_method_in_args<std::tuple<TupleArgs...>>
...
+{
+    CompileOrRuntimeAssert<std::is_same_v<CallableArgs, std::tuple<>>, FailureMode>(
+        "Registered method callable must not have any arguments since the method signature does not specify "
+        "any in arguments or return type!");
+}
 
-enum class FailureMode
+template <FailureMode FailureMode, typename CallableArgs, typename ReturnType, typename... ArgTypes>
+void AssertCallableHasInArgsAndReturn()
```
crimson11: see above ... the naming is imho confusing... I know, from where you are comming (the root signature from the service-.interface) ... but here it makes no sense to have this "background knowledge" and encode into the name?
`AssertCallableHasCorrectInArgsAndOutArg`

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -59,33 +85,56 @@ struct get_method_in_args<std::tuple<TupleArgs...>>
...
+        "following the return type, but with const reference semantics!");
+}
 
-// TODO: We currently use FailureMode as a workaround to allow testing AssertMethodHandlerSupportsMethodSignature at
-// runtime since we don't currently have infrastructure for compile time testing. When compile time testing is enabled
-// in SWP-46885, we can remove FailureMode and the runtime tests and replace them with compile time tests.
-template <bool condition, FailureMode failure_mode>
-void CompileOrRuntimeAssert([[maybe_unused]] const char* message)
+template <FailureMode FailureMode, typename CallableArgs, typename ReturnType>
+void AssertCallableHasReturnTypeOnly()
```
crimson11: `AssertCallableHasCorrectOutArgOnly`

score/mw/com/impl/methods/method_traits_checker.h
```diff
@@ -59,33 +85,56 @@ struct get_method_in_args<std::tuple<TupleArgs...>>
...
+    using ExpectedReturnType = ReturnType&;
+
+    CompileOrRuntimeAssert<std::is_same_v<MethodReturnType, ExpectedReturnType>, FailureMode>(
+        "Registered method callable must have the method return type as only argument as non-const "
+        "reference "
+        "!");
+}
+
+template <FailureMode FailureMode, typename CallableArgs, typename... ArgTypes>
+void AssertCallableHasInArgsOnly()
```
crimson11: `AssertCallableHasCorrectInArgsOnly`

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
```diff
@@ -870,13 +872,13 @@ TEST_F(SkeletonOnServiceMethodsSubscribedFixture, CallingRegistersAMethodCallHan
 
     // Expecting that the type erased callback will be called for each method with InArgs and ReturnArg storage
     // provided if TypeErasedElementInfo for the method in MethodData contains InArgs / a ReturnArg
-    EXPECT_CALL(foo_mock_type_erased_callback_, Call(_, _))
-        .WillOnce(Invoke([](auto in_args_optional, auto result_optional) {
+    EXPECT_CALL(foo_mock_type_erased_callback_, Call(_, _, _))
+        .WillOnce(Invoke([](auto /*quality_type*/, auto in_args_optional, auto result_optional) {
```
crimson11: check, that `quality_type` is QM? Should be as it gets fired originating from `captured_method_subscribed_handler_qm_` ?

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
```diff
@@ -870,13 +872,13 @@ TEST_F(SkeletonOnServiceMethodsSubscribedFixture, CallingRegistersAMethodCallHan
...
             EXPECT_EQ(in_args_optional.has_value(), kFooTypeErasedElementInfo.in_arg_type_info.has_value());
             EXPECT_EQ(result_optional.has_value(), kFooTypeErasedElementInfo.return_type_info.has_value());
         }));
-    EXPECT_CALL(dumb_mock_type_erased_callback_, Call(_, _))
-        .WillOnce(Invoke([](auto in_args_optional, auto result_optional) {
+    EXPECT_CALL(dumb_mock_type_erased_callback_, Call(_, _, _))
+        .WillOnce(Invoke([](auto /*quality_type*/, auto in_args_optional, auto result_optional) {
```
crimson11: see above

score/mw/com/impl/configuration/BUILD
```diff
@@ -39,7 +39,10 @@ cc_library(
     hdrs = ["quality_type.h"],
     features = COMPILER_WARNING_FEATURES,
     tags = ["FFI"],
-    visibility = ["//score/mw/com/impl/bindings/lola/messaging:__pkg__"],
+    visibility = [
+        "//score/mw/com/impl/bindings/lola/messaging:__pkg__",
+        "//score/mw/com/impl/methods:__pkg__",
```
crimson11: this looks "weird".
We are using `QualityType` everywhere in `impl` ... according to this visibility it should NOT be the case/possible. But it is the case, because this BUILD file exports hundreds other targets, which have a `deps` (NOT `implementation_deps`) to `quality_type`!

So - either we clean up this BUILD file:
-> making `quality_type` an `implementation_deps` everywhere here (to avoid "unwanted transient leakage" and then list here package by package the visibility of `quality_type` 
or we are simply fu..ing pragmatic here:
-> and put simply `//score/mw/com/impl/__subpackages_` here.

score/mw/com/impl/proxy_field.h
```diff
@@ -309,7 +309,28 @@ class ProxyFieldImpl : public ProxyFieldBase
...
+        if (!method_call_result.has_value())
+        {
+            return MakeUnexpected<MethodReturnTypePtr<T>>(std::move(method_call_result).error());
+        }
+
+        // If the method call itself succeeded but executing the getter itself on skeleton side failed (e.g. because
```
crimson11: >If the method call itself succeeded but executing the getter itself ...

what does this mean?
Do you mean:
>If the message_passing `SendWaitReply()` succeeded, but executing the getter-implementation on the skeleton side failed ...

score/mw/com/impl/proxy_field.h
```diff
@@ -309,7 +309,28 @@ class ProxyFieldImpl : public ProxyFieldBase
...
+        // reading the field value failed), then we return the error code to the user as well.
+        auto getter_return_type_ptr = std::move(method_call_result).value();
+        if (!getter_return_type_ptr->has_value())
+        {
+            return MakeUnexpected<MethodReturnTypePtr<T>>(std::move(*getter_return_type_ptr).error());
+        }
+
+        // If the method call and executing the getter itself succeeded, then we return a pointer to the set value to
```
crimson11: > set value

get value?

score/mw/com/impl/skeleton_field.h
```diff
@@ -289,6 +289,16 @@ class SkeletonFieldImpl : public SkeletonFieldBase
         }
     }
 
+    /// \brief Registers a get handler with the skeleton set method dispatch which will be called when a connected
```
crimson11: > set method dispatch

get method dispatch?

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -981,18 +984,18 @@ TEST(ProxyFieldNotifierGatingTest, RegisterSetHandlerDoesNotExistWhenNoTagPresen
                   "RegisterSetHandler must be SFINAE-removed on a SkeletonField without WithSetter");
 }
 
-class SkeletonFieldSetHandlerTest : public SkeletonFieldTestFixture
+class SkeletonFieldGetSetHandlerTest : public SkeletonFieldTestFixture
 {
   public:
-    SkeletonFieldSetHandlerTest& GivenASkeletonWithSetterEnabled()
+    SkeletonFieldGetSetHandlerTest& GivenASkeletonWithSetterAndGetterEnabled()
```
crimson11: GivenASkeletonWithSetterAndGetterEnabled ->
GivenASkeletonFieldWithSetterAndGetterEnabled
or
GivenAFieldWithSetterAndGetterEnabled

score/mw/com/impl/skeleton_field_test.cpp
```diff
@@ -84,7 +87,7 @@ class MyDummySkeleton : public SkeletonBase
     using SkeletonBase::SkeletonBase;
 
     // WithSetter is intentionally absent here since setter-related behaviour is tested via MySetterSkeleton below.
```
crimson11: >// WithSetter is intentionally absent here since setter-related behaviour is tested via MySetterSkeleton below.

->
// WithSetter/WithGetter is intentionally absent here since setter/getter-related behaviour is tested via MySetterAndGetterSkeleton below.

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/skeleton_method_handling_test.cpp
bemerybmw: Also added a test CallingAsilBRegistersAMethodCallHandlerPerMethodWithInfoFromMethodData

score/mw/com/impl/configuration/BUILD
bemerybmw: Made it visible everywhere in impl

score/mw/com/impl/proxy_field.h
bemerybmw: We are on the binding independent level here and message passing is lola specific. Maybe "If the middleware method communication succeeded but calling the get / set handler returned an error..."?

score/mw/com/impl/proxy_field.h
bemerybmw: This is used for both setter / getter. Updated the comment accordingly.

crimson11 (approved):

PR #760, opened 2026-07-23
impl/lola: Added shm-size calculation by analysis for DATA/CONTROL

crimson11:
Currently we determine shm-sizes of the
CTRL and DATA shm-objects by a simulation run.
I.e. we initialize the content of both section
1st within a heap-allocated resource. At the end
we use the sizes to correctly size the shm-objects.
This "SIMULATION" method is exact but has high runtime costs
and might consume lots of memory during startup.

This change now redesigns the containers/dynamic data
types being used within the DATA/CTRL sections, to use only
classes, which we are in control of and where we exactly
know based on our configuration, how much size they will
need. Thus the whole simulation can be skipped and we calculate
the size correctly from the configuration settings for the
service instance.

We re-introduce therefore the ANALYSIS mode in parallel
to the SIMULATION mode. For now we keep both in parallel.
SIMULATION is still the default, when not otherwise stated
in the config global section. When the ANALYSIS mode is
field-proven, we might later decide to remove SIMULATION
to simplify code!

### DIFF
--- a/score/mw/com/impl/bindings/lola/linear_search_map.h
+++ b/score/mw/com/impl/bindings/lola/linear_search_map.h
@@ -1,0 +1,292 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#ifndef SCORE_MW_COM_IMPL_BINDINGS_LOLA_LINEAR_SEARCH_MAP_H
+#define SCORE_MW_COM_IMPL_BINDINGS_LOLA_LINEAR_SEARCH_MAP_H
+
+#include "score/containers/non_relocatable_vector.h"
+#include "score/memory/shared/managed_memory_resource.h"
+#include "score/memory/shared/polymorphic_offset_ptr_allocator.h"
+
+#include <score/assert.hpp>
+
+#include <cstddef>
+#include <functional>
+#include <scoped_allocator>
+#include <tuple>
+#include <utility>
+
+namespace score::mw::com::impl::lola
+{
+
+/// \brief A fixed-capacity, associative container with a std::map-like interface backed by a contiguous array.
+///
+/// \details LinearSearchMap replaces the previously used std::map / boost::interprocess::map inside the shared-memory
+/// data structures (e.g. ServiceDataStorage). Those map-types allocate memory dynamically on each insertion
+/// (the exact amount depends on the implementation) which makes it impossible to calculate the exact size of the
+/// surrounding shared-memory object up-front (without a costly simulation run). LinearSearchMap in contrast allocates
+/// all of its storage exactly once on construction (via score::containers::NonRelocatableVector) with a capacity equal
+/// to the maximum number of elements. This capacity is always known in advance (it equals the number of
+/// service-elements of a service-instance). Hence the memory needed by a LinearSearchMap can be calculated
+/// deterministically.
+///
+/// Since the number of elements is small (number of events + fields of a single service-instance) and the container is
+/// filled once during initialization and only searched afterwards, a simple linear search is sufficient.
+///
+/// LinearSearchMap is designed to be placeable in shared-memory: it uses the provided (shared-memory capable) allocator
+/// and NonRelocatableVector guarantees that no reallocation (which would invalidate offset-pointers) ever takes place.
+///
+/// \note The complexity for lookup (find/at) is currently O(n). Since we typically use LinearSearchMap for a small
+/// number of elements (number of events + fields of a single service-instance), this is acceptable. If the number of
+/// elements grows, we could sort the underlying vector and then use std::binary_search for O(log n) lookup.
+/// This should be done if necessary.
+///
+/// \attention Adding more elements to it as defined in the ctor via max_number_of_elements will lead to termination.
+///
+/// \tparam Key key type used to identify an element. Must be copy-constructible and comparable via KeyEqual.
+/// \tparam MappedType mapped value type.
+/// \tparam KeyEqual binary predicate used to compare two keys for equality. Defaults to std::equal_to<Key> (i.e.
+///         operator==). A custom predicate can be supplied for keys without a suitable operator== or for alternative
+///         equality semantics (analogous to the KeyEqual parameter of std::unordered_map). Be careful: When placing
+///         the LinearSearchMap in shared-memory, the KeyEqual predicate must be stateless and trivially copyable
+///         (i.e. it must not contain any non-offset pointers or references).
+/// \tparam Allocator allocator used for the underlying storage. Defaults to the shared-memory capable
+///         PolymorphicOffsetPtrAllocator.
+template <typename Key,
+          typename MappedType,
+          typename KeyEqual = std::equal_to<Key>,
+          typename Allocator =
+              std::scoped_allocator_adaptor<memory::shared::PolymorphicOffsetPtrAllocator<std::pair<Key, MappedType>>>>
+class LinearSearchMap
+{
+  public:
+    using key_type = Key;
+    using mapped_type = MappedType;
+    using value_type = std::pair<Key, MappedType>;
+    using size_type = std::size_t;
+    using key_equal = KeyEqual;
+    using iterator = value_type*;
+    using const_iterator = const value_type*;
+
+    /// \brief Constructs an empty LinearSearchMap which can hold up to max_number_of_elements elements.
+    /// \param max_number_of_elements maximum number of elements the map can hold (its fixed capacity).
+    /// \param resource memory resource used to allocate the underlying storage.
+    /// \param key_equal binary predicate used to compare keys for equality.
+    LinearSearchMap(const size_type max_number_of_elements,
+                    memory::shared::ManagedMemoryResource& resource,
+                    const KeyEqual& key_equal = KeyEqual{});
+
+    iterator begin() noexcept;
+    const_iterator begin() const noexcept;
+    const_iterator cbegin() const noexcept;
+
+    iterator end() noexcept;
+    const_iterator end() const noexcept;
+    const_iterator cend() const noexcept;
+
+    size_type size() const noexcept;
+
+    size_type capacity() const noexcept;
+
+    bool empty() const noexcept;
+
+    iterator find(const Key& key) noexcept;
+
+    const_iterator find(const Key& key) const noexcept;
+
+    /// \brief Returns the key-equality predicate used by this map.
+    key_equal key_eq() const;
+
+    mapped_type& at(const Key& key);
+
+    const mapped_type& at(const Key& key) const;
+
+    /// \brief Inserts a new element constructed in-place from the given piecewise tuples (std::map-like).
+    /// \details If an element with the same key already exists, no insertion takes place.
+    /// \attention emplacing more elements as specified in ctor max_number_of_elements will lead to termination.
+    /// \return pair of an iterator to the (existing or newly inserted) element and a bool that is true if an insertion
+    ///         took place.
+    template <typename... KeyArgs, typename... MappedArgs>
+    std::pair<iterator, bool> emplace(std::piecewise_construct_t,
+                                      std::tuple<KeyArgs...> key_args,
+                                      std::tuple<MappedArgs...> mapped_args);
+
+    /// \brief Inserts a new element with the given key and mapped value (std::map-like).
+    /// \details If an element with the same key already exists, no insertion takes place.
+    /// \attention emplacing more elements as specified in ctor max_number_of_elements will lead to termination.
+    template <typename MappedArg>
+    std::pair<iterator, bool> emplace(const Key& key, MappedArg&& mapped_value);
+
+  private:
+    score::containers::NonRelocatableVector<value_type, Allocator> storage_;
+    KeyEqual key_equal_;
+};
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::LinearSearchMap(const size_type max_number_of_elements,
+                                                                       memory::shared::ManagedMemoryResource& resource,
+                                                                       const KeyEqual& key_equal)
+    : storage_{max_number_of_elements, Allocator{resource}}, key_equal_{key_equal}
+{
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::begin() noexcept
+{
+    return storage_.begin();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::const_iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::begin() const noexcept
+{
+    return storage_.begin();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::const_iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::cbegin() const noexcept
+{
+    return storage_.cbegin();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::end() noexcept
+{
+    return storage_.end();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::const_iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::end() const noexcept
+{
+    return storage_.end();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::const_iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::cend() const noexcept
+{
+    return storage_.cend();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::size_type
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::size() const noexcept
+{
+    return storage_.size();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::size_type
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::capacity() const noexcept
+{
+    return storage_.capacity();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+bool LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::empty() const noexcept
+{
+    return storage_.size() == 0U;
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::find(const Key& key) noexcept
+{
+    for (auto it = begin(); it != end(); ++it)
+    {
+        if (key_equal_(it->first, key))
+        {
+            return it;
+        }
+    }
+    return end();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::const_iterator
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::find(const Key& key) const noexcept
+{
+    for (auto it = cbegin(); it != cend(); ++it)
+    {
+        if (key_equal_(it->first, key))
+        {
+            return it;
+        }
+    }
+    return cend();
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::key_equal
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::key_eq() const
+{
+    return key_equal_;
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::mapped_type&
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::at(const Key& key)
+{
+    const auto it = find(key);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(it != end(), "LinearSearchMap::at: key not found.");
+    return it->second;
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+const typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::mapped_type&
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::at(const Key& key) const
+{
+    const auto it = find(key);
+    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(it != cend(), "LinearSearchMap::at: key not found.");
+    return it->second;
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+template <typename... KeyArgs, typename... MappedArgs>
+std::pair<typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::iterator, bool>
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::emplace(std::piecewise_construct_t,
+                                                               std::tuple<KeyArgs...> key_args,
+                                                               std::tuple<MappedArgs...> mapped_args)
+{
+    const Key key = std::make_from_tuple<Key>(key_args);
+    const auto existing = find(key);
+    if (existing != end())
+    {
+        return {existing, false};
+    }
+    auto& inserted = storage_.emplace_back(std::piecewise_construct, std::move(key_args), std::move(mapped_args));
+    return {&inserted, true};
+}
+
+template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
+template <typename MappedArg>
+std::pair<typename LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::iterator, bool>
+LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::emplace(const Key& key, MappedArg&& mapped_value)
+{
+    const auto existing = find(key);
+    if (existing != end())
+    {
+        return {existing, false};
+    }
+    auto& inserted = storage_.emplace_back(std::piecewise_construct,
+                                           std::forward_as_tuple(key),
+                                           std::forward_as_tuple(std::forward<MappedArg>(mapped_value)));
+    return {&inserted, true};
+}
+
+}  // namespace score::mw::com::impl::lola
+
+#endif  // SCORE_MW_COM_IMPL_BINDINGS_LOLA_LINEAR_SEARCH_MAP_H
--- a/score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
+++ b/score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
@@ -1,0 +1,269 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/impl/bindings/lola/linear_search_map.h"
+
+#include "score/memory/shared/new_delete_delegate_resource.h"
+
+#include <gtest/gtest.h>
+
+#include <cstdint>
+#include <tuple>
+#include <utility>
+
+namespace score::mw::com::impl::lola
+{
+namespace
+{
+
+const std::uint64_t kMemoryResourceId{42U};
+constexpr std::int32_t kSomeKey{1};
+constexpr std::int32_t kSomeValue{100};
+
+using TestMap = LinearSearchMap<std::int32_t, std::int32_t>;
+
+class LinearSearchMapFixture : public ::testing::Test
+{
+  protected:
+    memory::shared::NewDeleteDelegateMemoryResource memory_{kMemoryResourceId};
+};
+
+TEST_F(LinearSearchMapFixture, IsEmptyAfterConstruction)
+{
+    // When constructing a LinearSearchMap with a capacity of 4
+    TestMap unit{4U, memory_};
+
+    // Then it is empty, has size 0 and reports the requested capacity
+    EXPECT_TRUE(unit.empty());
+    EXPECT_EQ(unit.size(), 0U);
+    EXPECT_EQ(unit.capacity(), 4U);
+}
+
+TEST_F(LinearSearchMapFixture, BeginAndEndEqualAfterConstruction)
+{
+    // When constructing a LinearSearchMap with a capacity of 4
+    TestMap unit{4U, memory_};
+
+    // Then begin and end iterators are equal (no elements to iterate)
+    EXPECT_EQ(unit.begin(), unit.end());
+}
+
+TEST_F(LinearSearchMapFixture, EmplaceSucceedsAndReturnsCorrectIterator)
+{
+    // Given an empty LinearSearchMap
+    TestMap unit{4U, memory_};
+
+    // When emplacing a new key/value pair
+    const auto result = unit.emplace(kSomeKey, kSomeValue);
+
+    // Then the insertion succeeds and the element is retrievable via the returned iterator
+    EXPECT_TRUE(result.second);
+    ASSERT_NE(result.first, unit.end());
+    EXPECT_EQ(result.first->first, kSomeKey);
+    EXPECT_EQ(result.first->second, kSomeValue);
+    EXPECT_EQ(unit.size(), 1U);
+    EXPECT_FALSE(unit.empty());
+}
+
+TEST_F(LinearSearchMapFixture, EmplaceWithPiecewiseConstructSucceedsAndReturnsCorrectIterator)
+{
+    // Given an empty LinearSearchMap
+    TestMap unit{4U, memory_};
+
+    // When emplacing a new element using the piecewise-construct overload
+    const auto result =
+        unit.emplace(std::piecewise_construct, std::forward_as_tuple(kSomeKey), std::forward_as_tuple(kSomeValue));
+
+    // Then the insertion succeeds and the element is retrievable via the returned iterator
+    EXPECT_TRUE(result.second);
+    ASSERT_NE(result.first, unit.end());
+    EXPECT_EQ(result.first->first, kSomeKey);
+    EXPECT_EQ(result.first->second, kSomeValue);
+}
+
+TEST_F(LinearSearchMapFixture, EmplaceWithDuplicateKeyDoesNotInsert)
+{
+    // Given a LinearSearchMap that already contains the key 1
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+
+    // When emplacing another element with the same key and different value
+    const auto result = unit.emplace(kSomeKey, kSomeValue + 5);
+
+    // Then no insertion takes place and the originally inserted value is preserved
+    EXPECT_FALSE(result.second);
+    EXPECT_EQ(unit.size(), 1U);
+    EXPECT_EQ(result.first->second, kSomeValue);
+}
+
+TEST_F(LinearSearchMapFixture, FindReturnsInsertedElement)
+{
+    // Given a LinearSearchMap containing several elements
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+    std::ignore = unit.emplace(kSomeKey + 1, kSomeValue + 1);
+    std::ignore = unit.emplace(kSomeKey + 2, kSomeValue + 2);
+
+    // When searching for an existing key
+    const auto it = unit.find(kSomeKey);
+
+    // Then the corresponding element is returned
+    ASSERT_NE(it, unit.end());
+    EXPECT_EQ(it->second, kSomeValue);
+}
+
+TEST_F(LinearSearchMapFixture, FindReturnsEndForMissingKey)
+{
+    // Given a LinearSearchMap containing a single element
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+
+    // When searching for a key that is not present
+    // Then the end iterator is returned
+    EXPECT_EQ(unit.find(99), unit.end());
+}
+
+TEST_F(LinearSearchMapFixture, ConstFindReturnsInsertedElement)
+{
+    // Given a const reference to a LinearSearchMap containing one element
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+    const TestMap& const_unit = unit;
+
+    // When searching via the const overload of find
+    const auto it = const_unit.find(kSomeKey);
+
+    // Then an existing key is found and the value matches.
+    ASSERT_NE(it, const_unit.cend());
+    EXPECT_EQ(it->second, kSomeValue);
+}
+
+TEST_F(LinearSearchMapFixture, ConstFindReturnsEndForMissingKey)
+{
+    // Given a const reference to a LinearSearchMap containing one element
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+    const TestMap& const_unit = unit;
+
+    // When searching via the const overload of find for a non existent key
+    const auto it = const_unit.find(kSomeKey + 5);
+
+    // Then it yields the const end iterator
+    EXPECT_EQ(it, const_unit.cend());
+}
+
+TEST_F(LinearSearchMapFixture, AtReturnsMappedValue)
+{
+    // Given a LinearSearchMap containing one element
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+
+    // When accessing the mapped value via at
+    // Then the stored value is returned
+    EXPECT_EQ(unit.at(kSomeKey), kSomeValue);
+
+    // and can be modified through the returned reference
+    unit.at(kSomeKey) = kSomeValue + 1;
+    EXPECT_EQ(unit.at(kSomeKey), kSomeValue + 1);
+}
+
+TEST_F(LinearSearchMapFixture, IterationVisitsAllElements)
+{
+    // Given a LinearSearchMap containing two elements
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+    std::ignore = unit.emplace(kSomeKey + 1, kSomeValue + 1);
+
+    // When iterating over all elements and accumulating keys and values
+    std::int32_t sum_keys{0};
+    std::int32_t sum_values{0};
+    for (const auto& element : unit)
+    {
+        sum_keys += element.first;
+        sum_values += element.second;
+    }
+
+    // Then every element is visited exactly once
+    EXPECT_EQ(sum_keys, kSomeKey + (kSomeKey + 1));
+    EXPECT_EQ(sum_values, kSomeValue + (kSomeValue + 1));
+}
+
+// A custom key-equality predicate that considers two keys equal if they have the same absolute value.
+struct AbsoluteValueEqual
+{
+    bool operator()(const std::int32_t lhs, const std::int32_t rhs) const noexcept
+    {
+        return (lhs < 0 ? -lhs : lhs) == (rhs < 0 ? -rhs : rhs);
+    }
+};
+
+using CustomEqualMap = LinearSearchMap<std::int32_t, std::int32_t, AbsoluteValueEqual>;
+
+TEST_F(LinearSearchMapFixture, CustomKeyEqualIsUsedForFind)
+{
+    // Given a LinearSearchMap using a custom absolute-value key-equality predicate, containing the key kSomeKey
+    CustomEqualMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+
+    // When searching for -kSomeKey, which the custom predicate considers equal to kSomeKey
+    const auto it = unit.find(-kSomeKey);
+
+    // Then the element stored under key kSomeKey is found
+    ASSERT_NE(it, unit.end());
+    EXPECT_EQ(it->second, kSomeValue);
+}
+
+TEST_F(LinearSearchMapFixture, CustomKeyEqualPreventsDuplicateInsertion)
+{
+    // Given a LinearSearchMap using a custom absolute-value key-equality predicate, containing the key kSomeKey
+    CustomEqualMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+
+    // When emplacing -kSomeKey, which the custom predicate treats as a duplicate of kSomeKey
+    const auto result = unit.emplace(-kSomeKey, kSomeValue + 1);
+
+    // Then no insertion takes place and the original value is preserved
+    EXPECT_FALSE(result.second);
+    EXPECT_EQ(unit.size(), 1U);
+    EXPECT_EQ(result.first->second, kSomeValue);
+}
+
+TEST_F(LinearSearchMapFixture, KeyEqAccessorReturnsPredicate)
+{
+    // Given a LinearSearchMap using a custom absolute-value key-equality predicate
+    CustomEqualMap unit{4U, memory_};
+
+    // When retrieving the predicate via key_eq
+    const auto predicate = unit.key_eq();
+
+    // Then the returned predicate exhibits the custom equality semantics
+    EXPECT_TRUE(predicate(-7, 7));
+    EXPECT_FALSE(predicate(-7, 8));
+}
+
+using LinearSearchMapDeathTest = LinearSearchMapFixture;
+TEST_F(LinearSearchMapDeathTest, AddMoreElementsThanCapacity)
+{
+    // Given a LinearSearchMap which contains capacity count elements
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+    std::ignore = unit.emplace(kSomeKey + 1, kSomeValue + 1);
+    std::ignore = unit.emplace(kSomeKey + 2, kSomeValue + 2);
+    std::ignore = unit.emplace(kSomeKey + 3, kSomeValue + 3);
+
+    // When adding another element
+    // Then the program terminates
+    EXPECT_DEATH(unit.emplace(kSomeKey + 4, kSomeValue + 4), ".*");
+}
+
+}  // namespace
+}  // namespace score::mw::com::impl::lola
--- a/score/mw/com/impl/bindings/lola/service_data_control.cpp
+++ b/score/mw/com/impl/bindings/lola/service_data_control.cpp
@@ -11,3 +11,99 @@
  * SPDX-License-Identifier: Apache-2.0
  ********************************************************************************/
 #include "score/mw/com/impl/bindings/lola/service_data_control.h"
+
+#include "score/mw/com/impl/bindings/lola/application_id_pid_mapping_entry.h"
+#include "score/mw/com/impl/bindings/lola/event_data_control.h"
+#include "score/mw/com/impl/bindings/lola/transaction_log.h"
+#include "score/mw/com/impl/bindings/lola/transaction_log_set.h"
+
+#include "score/memory/data_type_size_info.h"
+#include "score/memory/shared/pointer_arithmetic_util.h"
+
+#include <cstddef>
+#include <vector>
+
+namespace score::mw::com::impl::lola
+{
+
+std::size_t CalculateServiceDataControlShmSize(
+    const score::cpp::span<const ServiceElementControlSizeInfo> events_and_fields_size_info)
+{
+    // The number of events + fields determines the (fixed) capacity of the event_controls_ LinearSearchMap within the
+    // ServiceDataControl. It equals the number of sizing entries handed over.
+    const auto number_of_events_and_fields = events_and_fields_size_info.size();
+
+    // The real construction of a ServiceDataControl and its contained EventControls performs a fixed, deterministic
+    // sequence of allocations from the (strictly monotonic) shared-memory resource, which itself always starts
+    // allocating at a std::max_align_t aligned location. Since we know the exact size/alignment of every single one
+    // of these allocations and the exact order in which they happen, we can reconstruct the exact sequence here and
+    // let memory::shared::CalculateAlignedSizeOfSequence() compute the exact (not just worst-case) total size, taking
+    // into account the exact alignment-padding between consecutive allocations.
+    std::vector<score::memory::DataTypeSizeInfo> allocation_sequence{};
+
+    // (1) The ServiceDataControl object itself (including the inline bookkeeping of its event_controls_ LinearSearchMap
+    // and of its application_id_pid_mapping_).
+    allocation_sequence.emplace_back(sizeof(ServiceDataControl), alignof(ServiceDataControl));
+
+    // (2) The allocated array of the event_controls_ LinearSearchMap (allocated once, with capacity ==
+    // number_of_events_and_fields).
+    allocation_sequence.emplace_back(
+        number_of_events_and_fields * sizeof(decltype(ServiceDataControl::event_controls_)::value_type),
+        alignof(decltype(ServiceDataControl::event_controls_)::value_type));
+
+    // (3) The backing array of the application_id_pid_mapping_ (a fixed-capacity DynamicArray with a capacity of
+    // kMaxApplicationIdPidMappings). It is allocated once during ServiceDataControl construction, independent of the
+    // number of events + fields.
+    allocation_sequence.emplace_back(static_cast<std::size_t>(ServiceDataControl::kMaxApplicationIdPidMappings) *
+                                         sizeof(ApplicationIdPidMappingEntry),
+                                     alignof(ApplicationIdPidMappingEntry));
+
+    // (4) For each event/field: the (deeply) nested fixed-capacity DynamicArrays contained within its EventControl.
+    // The EventControl object itself is stored inline within the event_controls_ backing array (accounted for in (2));
+    // only the allocated arrays of its nested DynamicArrays allocate separately and are accounted for here.
+    for (const auto& service_element : events_and_fields_size_info)
+    {
+        const std::size_t number_of_slots = service_element.number_of_slots;
+        const std::size_t max_subscribers = service_element.max_subscribers;
+
+        // (4a) EventControl::data_control (EventDataControl): its state_slots_ is a DynamicArray<ControlSlotType>
+        // with a capacity of number_of_slots.
+        allocation_sequence.emplace_back(number_of_slots * sizeof(EventDataControl::EventControlSlots::value_type),
+                                         alignof(EventDataControl::EventControlSlots::value_type));
+
+        // (4b) EventControl::transaction_log_set_ (TransactionLogSet):
+        //   proxy_transaction_logs_(max_subscribers, TransactionLogNode{number_of_slots, resource}, resource) is
+        //   constructed via the DynamicArray fill-constructor. The exact allocation order is:
+        //     * one reference_count_slots_ array for the temporary prototype TransactionLogNode, which is
+        //       constructed (as a function argument) before the DynamicArray constructor itself even runs. This
+        //       constucts the TransactionLog for the given TransactionLogNode, which contains a DynamicArray of
+        //       TransactionLogSlots representing the TransactionLog for slot-referencing.
+        const std::size_t transaction_log_slots_array_raw_size =
+            number_of_slots * sizeof(TransactionLog::TransactionLogSlots::value_type);
+        const std::size_t transaction_log_slots_array_alignment =
+            alignof(TransactionLog::TransactionLogSlots::value_type);
+        allocation_sequence.emplace_back(transaction_log_slots_array_raw_size, transaction_log_slots_array_alignment);
+
+        //     * then the backing array of proxy_transaction_logs_ itself (max_subscribers elements of
+        //       TransactionLogNode), allocated in a single allocation.
+        allocation_sequence.emplace_back(max_subscribers * sizeof(TransactionLogSet::TransactionLogNode),
+                                         alignof(TransactionLogSet::TransactionLogNode));
+
+        //     * then, one at a time (in order), each of the max_subscribers elements of proxy_transaction_logs_ gets
+        //       copy-constructed from the prototype, each copy-construction allocating its own, separate
+        //       reference_count_slots_ array.
+        for (std::size_t i = 0U; i < max_subscribers; ++i)
+        {
+            allocation_sequence.emplace_back(transaction_log_slots_array_raw_size,
+                                             transaction_log_slots_array_alignment);
+        }
+
+        //   - finally, the inline skeleton_tracing_transaction_log_ member is constructed, allocating one more
+        //     reference_count_slots_ array.
+        allocation_sequence.emplace_back(transaction_log_slots_array_raw_size, transaction_log_slots_array_alignment);
+    }
+
+    return score::memory::shared::CalculateAlignedSizeOfSequence(allocation_sequence);
+}
+
+}  // namespace score::mw::com::impl::lola
--- a/score/mw/com/impl/bindings/lola/service_data_control.h
+++ b/score/mw/com/impl/bindings/lola/service_data_control.h
@@ -16,9 +16,13 @@
 #include "score/mw/com/impl/bindings/lola/application_id_pid_mapping.h"
 #include "score/mw/com/impl/bindings/lola/element_fq_id.h"
 #include "score/mw/com/impl/bindings/lola/event_control.h"
+#include "score/mw/com/impl/bindings/lola/linear_search_map.h"
 
-#include "score/memory/shared/map.h"
 #include "score/memory/shared/polymorphic_offset_ptr_allocator.h"
+
+#include <score/span.hpp>
+
+#include <cstddef>
 
 namespace score::mw::com::impl::lola
 {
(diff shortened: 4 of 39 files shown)


score/mw/com/impl/configuration/shm_size_calc_mode.h
```diff
@@ -22,6 +22,7 @@ namespace score::mw::com::impl
 enum class ShmSizeCalculationMode : std::uint8_t
 {
     kSimulation,
+    kEstimation,
```
castler: This is no longer an estimation - this is now an calculation! We should be clear in the wording.

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+#include <tuple>
+#include <utility>
+
+namespace score::mw::com::impl::lola
+{
+
+/// \brief A fixed-capacity, associative container with a std::map-like interface backed by a contiguous array.
+///
+/// \details LinearSearchMap replaces the previously used std::map / boost::interprocess::map inside the shared-memory
+/// data structures (e.g. ServiceDataStorage). Those map-types allocate memory dynamically on each insertion which makes
```
bemerybmw:
```suggestion
/// data structures (e.g. ServiceDataStorage). Those map-types allocate memory dynamically on each insertion (the exact amount depends on the implementation) which makes
```

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+///         operator==). A custom predicate can be supplied for keys without a suitable operator== or for alternative
+///         equality semantics (analogous to the KeyEqual parameter of std::unordered_map). Be careful: When placing
+///         the LinearSearchMap in shared-memory, the KeyEqual predicate must be stateless and trivially copyable
+///         (i.e. it must not contain any non-offset pointers or references).
+/// \tparam Allocator allocator used for the underlying storage. Defaults to the shared-memory capable
+///         PolymorphicOffsetPtrAllocator.
+template <typename Key,
+          typename MappedType,
+          typename KeyEqual = std::equal_to<Key>,
+          typename Allocator = memory::shared::PolymorphicOffsetPtrAllocator<std::pair<Key, MappedType>>>
```
bemerybmw: Should have scoped_allocator_adaptor?

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+        return cend();
+    }
+
+    /// \brief Returns the key-equality predicate used by this map.
+    key_equal key_eq() const
+    {
+        return key_equal_;
+    }
+
+    mapped_type& at(const Key& key)
```
bemerybmw:
```suggestion
    mapped_type& at(const Key& key)&
```
I think this should have the trailing & so that it can't be called on a temporary(would lead to a dangling reference).

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+    using size_type = std::size_t;
+    using key_equal = KeyEqual;
+    using iterator = value_type*;
+    using const_iterator = const value_type*;
+
+    /// \brief Constructs an empty LinearSearchMap which can hold up to max_number_of_elements elements.
+    /// \param max_number_of_elements maximum number of elements the map can hold (its fixed capacity).
+    /// \param resource memory resource used to allocate the underlying storage.
+    /// \param key_equal binary predicate used to compare keys for equality.
+    LinearSearchMap(const size_type max_number_of_elements,
```
bemerybmw, line 84: IMO, the definition of these functions should all be below. The declaration of the class should ideally fit in one screen (or a little bit of scrolling). But as it is now, it's quite hard to read the interface of this class.

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+                    memory::shared::ManagedMemoryResource& resource,
+                    const KeyEqual& key_equal = KeyEqual{})
+        : storage_{max_number_of_elements, Allocator{resource}}, key_equal_{key_equal}
+    {
+    }
+
+    iterator begin() noexcept
```
bemerybmw: Shouldn't be noexcept (unless this is about be compliant with some standard container specification)

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(it != cend(), "LinearSearchMap::at: key not found.");
+        return it->second;
+    }
+
+    /// \brief Inserts a new element constructed in-place from the given piecewise tuples (std::map-like).
+    /// \details If an element with the same key already exists, no insertion takes place.
+    /// \return pair of an iterator to the (existing or newly inserted) element and a bool that is true if an insertion
+    ///         took place.
+    template <typename... KeyArgs, typename... MappedArgs>
+    std::pair<iterator, bool> emplace(std::piecewise_construct_t,
```
bemerybmw, line 119: Why not follow the unordered_map interface which is `std::pair<iterator, bool> emplace( Args&&... args );` i.e. basically takes a std::pair?

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+            return {existing, false};
+        }
+        auto& inserted = storage_.emplace_back(std::piecewise_construct, std::move(key_args), std::move(mapped_args));
+        return {&inserted, true};
+    }
+
+    /// \brief Inserts a new element with the given key and mapped value (std::map-like).
+    /// \details If an element with the same key already exists, no insertion takes place.
+    template <typename MappedArg>
+    std::pair<iterator, bool> emplace(const Key& key, MappedArg&& mapped_value)
```
bemerybmw: Why do we need these 2 emplace overloads?

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
...
+    {
+        return storage_.capacity();
+    }
+
+    bool empty() const noexcept
+    {
+        return storage_.size() == 0U;
+    }
+
+    iterator find(const Key& key) noexcept
```
bemerybmw: Should we support a templated find i.e. so we can lookup a string key with a string_view?

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -0,0 +1,213 @@
+/********************************************************************************
```
bemerybmw, line 1: We need a cpp file for linear_search_map

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+class LinearSearchMapFixture : public ::testing::Test
+{
+  protected:
+    memory::shared::NewDeleteDelegateMemoryResource memory_{kMemoryResourceId};
+};
+
+TEST_F(LinearSearchMapFixture, IsEmptyAfterConstruction)
+{
+    TestMap unit{4U, memory_};
```
bemerybmw, line 43: Given, when, then

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+using TestMap = LinearSearchMap<std::int32_t, std::int32_t>;
+
+class LinearSearchMapFixture : public ::testing::Test
+{
+  protected:
+    memory::shared::NewDeleteDelegateMemoryResource memory_{kMemoryResourceId};
+};
+
+TEST_F(LinearSearchMapFixture, IsEmptyAfterConstruction)
```
bemerybmw, line 40: I guess the tests should test different allocators? You can check the typed tests in https://github.com/eclipse-score/baselibs/blob/main/score/containers/non_relocatable_vector_test.cpp. And we've already got some helpers in here: https://github.com/eclipse-score/baselibs/blob/main/score/containers/test/allocator_test_type_helpers.h

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+{
+    TestMap unit{4U, memory_};
+
+    EXPECT_TRUE(unit.empty());
+    EXPECT_EQ(unit.size(), 0U);
+    EXPECT_EQ(unit.capacity(), 4U);
+    EXPECT_EQ(unit.begin(), unit.end());
```
bemerybmw, line 57: This test is testing a few different things. IMO, this should be split up. I made the tests very small in https://github.com/eclipse-score/baselibs/blob/main/score/containers/non_relocatable_vector_special_member_funcs_test.cpp. You can follow that or keep them a bit bigger since they're quite simple, but I'd at least make one test for size / capacity and one test that unit.begin() == unit.end() because those are completely different functionalities.

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+}
+
+TEST_F(LinearSearchMapFixture, EmplaceInsertsElement)
```
bemerybmw: The name should be explicit about what the test is testing.

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+{
```
bemerybmw, line 61: Please fix these comments that I've already mentioned for all the tests, then I'll review the rest of them. Because a lot of the comments are the same.

score/mw/com/impl/configuration/mw_com_config_schema.json
```diff
@@ -501,8 +501,9 @@
                 },
                 "shm-size-calc-mode": {
                     "title": "Calculation mode for shared memory size",
-                    "description": "How shall the sizes of the shm-objects get calculated (Currently we only support SIMULATION): Via an initial simulation run against a heap-memory resource. It has the advantage of perfect accuracy, but adds a minimal runtime overhead at startup plus some initial temporary heap allocation.",
+                    "description": "How shall the sizes of the shm-objects get calculated: Via an initial simulation run against a heap-memory resource. It has the advantage of perfect accuracy, but adds a minimal runtime overhead at startup plus some initial temporary heap allocation. Alternatively via an analysis based on the config/deployment. It has the advantage of no runtime overhead and no temporary heap allocation, but may lead to slightly overestimation of the required shm-size, because of conservative alignment usage. The default is simulation.",
```
bemerybmw:
```suggestion
                    "description": "How shall the sizes of the shm-objects get calculated: Via an initial simulation run against a heap-memory resource. It has the advantage of perfect accuracy, but adds a minimal runtime overhead at startup plus some initial temporary heap allocation. Alternatively via an analysis based on the config/deployment. It has the advantage of no runtime overhead and no temporary heap allocation, but may lead to a slight overestimation of the required shm-size, because of conservative alignment usage. The default is simulation.",
```

score/mw/com/impl/configuration/configuration_json_parsing_strategy.cpp
```diff
@@ -228,6 +229,10 @@ auto ParseShmSizeCalcMode(const score::json::Object& json_map) -> std::optional<
         {
             return ShmSizeCalculationMode::kSimulation;
         }
+        else if (shm_size_calc_mode_value == kShmSizeCalcModeAnalysis)
+        {
+            return ShmSizeCalculationMode::kAnalysis;
```
bemerybmw, line 234: We should have config parser tests for this.

score/mw/com/impl/bindings/lola/service_data_control.h
```diff
@@ -35,10 +37,15 @@ class ServiceDataControl
     /// \details ServiceDataControl is designed to be located in shared memory, therefore the explicit
     ///          MemoryResourceProxy argument! (Yes one could come up with a MemoryResourceProxy pointing to a local
     ///          memory resource, but this would be "uncommon")
-    /// \param proxy MemoryResourceProxy pointing to the memory-resource to be used
+    /// \param number_of_service_elements the (fixed) number of service-elements (events + fields) this service-instance
+    ///        provides. It is used as the fixed capacity of the event_controls_ container. Since event_controls_ uses a
+    ///        fixed-capacity container (LinearSearchMap), its capacity must be known at construction time.
+    /// \param resource MemoryResourceProxy pointing to the memory-resource to be used
```
bemerybmw: This is a resource, not a proxy

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -17,22 +17,38 @@
...
 namespace score::mw::com::impl::lola
 {
 
 class ServiceDataStorage
 {
   public:
-    explicit ServiceDataStorage(score::memory::shared::ManagedMemoryResource& resource)
-        : events_(resource),
-          events_metainfo_(resource),
+    /// \brief associative container mapping a service-element (event/field) to the raw storage of its event-data slots.
```
bemerybmw, line 37: Can you add what type offsetptr is pointing to and why it's type erased.

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -17,22 +17,38 @@
...
+    using EventDataStorageMap = LinearSearchMap<ElementFqId, score::memory::shared::OffsetPtr<void>>;
+    /// \brief associative container mapping a service-element (event/field) to its (type-erased) meta-information.
+    using EventMetaInfoMap = LinearSearchMap<ElementFqId, EventMetaInfo>;
+
+    /// \brief Ctor for the ServiceDataStorage to place it in the given shared memory resource.
```
bemerybmw: I think is a bit misleading. This constructor doesn't put ServiceDataStorage in shared memory, it constructs it wherever the caller constructs it and the containers allocate in the provided resource. In tests for example, it prob won't be created in shm

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -17,22 +17,38 @@
...
+    /// \details ServiceDataStorage no longer uses dynamically allocating map-types. Instead it uses fixed-capacity
+    ///          containers (LinearSearchMap) whose capacity has to be provided at construction time. The capacity
+    ///          equals the number of service-elements (events + fields) of the service-instance, which is known
+    ///          up-front. This makes the memory footprint of ServiceDataStorage deterministic and calculable without a
+    ///          simulation run.
+    /// \param number_of_service_elements maximum number of service-elements (events + fields) that will be stored.
```
bemerybmw:
```suggestion
    /// \param number_of_service_elements maximum number of service-elements (events + fields) that will be stored in events_ and events_metainfo_.
```

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -183,6 +184,42 @@ class SkeletonMemoryManager final
...
+    /// \details In contrast to the simulation based approach this function does NOT allocate any (heap) memory and does
+    /// not create a ServiceDataStorage. Instead it calculates the exactly needed (worst-case) size based on the
+    /// handed-over event/field bindings and the deployment configuration. This is possible because ServiceDataStorage
+    /// uses fixed-capacity containers (see LinearSearchMap / EventDataStorage) whose sizes are deterministic once the
+    /// number of service-elements and their number of slots / sample-sizes are known.
+    /// \return needed size (in bytes) for the data shm-object.
+    std::size_t CalculateDataShmResourceStorageSize(SkeletonBinding::SkeletonEventBindings& events,
+                                                    SkeletonBinding::SkeletonFieldBindings& fields) const;
+
+    /// \brief Calculates the needed size for a control shm-object (holding a ServiceDataControl) analytically.
```
bemerybmw, line 196:
```suggestion
    /// \brief Calculates the needed size for a control shm-object (holding a ServiceDataControl[s]) analytically.
```

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -183,6 +184,42 @@ class SkeletonMemoryManager final
...
+    /// of slots and their number of max-subscribers are known.
+    /// The same size applies to the QM and (if present) the ASIL-B control shm-object, as both hold a
+    /// ServiceDataControl created with the very same configuration.
+    /// \return needed size (in bytes) for a single control shm-object.
+    std::size_t CalculateControlShmResourceStorageSize(SkeletonBinding::SkeletonEventBindings& events,
+                                                       SkeletonBinding::SkeletonFieldBindings& fields) const;
+
+    /// \brief Looks up the configured number of sample-slots for the given service-element.
+    /// \param service_element_name name of the event/field.
+    /// \param is_field true if the service-element is a field, false if it is an event.
```
bemerybmw: We shouldn't add documentation for the sake of it. I'd remove the line about is_field and the return

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -183,6 +184,42 @@ class SkeletonMemoryManager final
...
+    /// \return the configured number of sample-slots.
+    std::size_t GetNumberOfSampleSlotsFromConfig(const std::string_view service_element_name,
+                                                 const bool is_field) const;
+
+    /// \brief Looks up the configured maximum number of subscribers for the given service-element.
+    /// \param service_element_name name of the event/field.
+    /// \param is_field true if the service-element is a field, false if it is an event.
```
bemerybmw: We shouldn't add documentation for the sake of it. I'd remove the line about is_field and the return

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -225,6 +262,11 @@ class SkeletonMemoryManager final
     ServiceDataControl* control_qm_;
     ServiceDataControl* control_asil_b_;
 
+    /// \brief Number of service-elements (events + fields) of the service-instance.
+    /// \details Determines the fixed capacity of the containers within ServiceDataStorage. Set before the
```
bemerybmw:
```suggestion
    /// \details Determines the fixed capacity of the containers within ServiceDataStorage and ServiceDataControl. Set before the
```

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -118,6 +125,7 @@ SkeletonMemoryManager::SkeletonMemoryManager(QualityType quality_type,
       storage_{nullptr},
       control_qm_{nullptr},
       control_asil_b_{nullptr},
+      number_of_service_elements_{0U},
```
bemerybmw: This should be a std::optional to ensure that we don't access it before it's set in CreateSharedMemory

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,194 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
     return ShmResourceStorageSizes{control_data_size, control_qm_size, control_asil_b_size};
 }
 
+namespace
```
bemerybmw: The anonymous namespace should go at the top of the file (to match the rest of the code base)

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,194 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+{
+
+// Rounds size up to the next multiple of alignof(std::max_align_t).
+// The SharedMemoryResource is a monotonic (bump) allocator: the bytes it accounts for a single allocation are the
+// requested size plus the padding needed to bring the current offset to the requested alignment. All alignments
+// occurring within the data shm-object are <= alignof(std::max_align_t). By rounding every individual allocation up to
+// alignof(std::max_align_t) we keep the running offset max-aligned and therefore obtain a size that is guaranteed to be
+// sufficient (it is exact whenever the allocated sizes are multiples of the involved alignment, which is the common
+// case).
+constexpr std::size_t RoundUpToMaxAlign(const std::size_t size) noexcept
```
bemerybmw: This is a duplicate of CalculateAlignedSize?

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,194 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+
+std::size_t SkeletonMemoryManager::GetNumberOfSampleSlotsFromConfig(const std::string_view service_element_name,
+                                                                    const bool is_field) const
+{
+    const std::string name{service_element_name};
+    if (is_field)
+    {
+        const auto field_it = lola_service_instance_deployment_.fields_.find(name);
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(field_it != lola_service_instance_deployment_.fields_.cend(),
+                                                    "Could not find field in deployment configuration.");
```
bemerybmw: The message should not just be repeating what the check is in words. It should either be explaining why the assertion is checked / expected or just omit it. It's not adding any new information here. Same for the other assertions

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,194 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+            total_size += number_of_transaction_log_slot_arrays * transaction_log_slots_array_size;
+        }
+    };
+    accumulate_service_elements(events, false);
+    accumulate_service_elements(fields, true);
+
+    return total_size;
+}
+
+std::size_t SkeletonMemoryManager::GetNumberOfSampleSlotsFromConfig(const std::string_view service_element_name,
```
bemerybmw, line 595: What about something like this to avoid code duplication?

```
template <ServiceElementType ServiceElementType>
std::size_t SkeletonMemoryManager::GetNumberOfSampleSlotsFromConfig(const std::string_view service_element_name) const
{
    SCORE_LANGUAGE_FUTURECPP_PRECONDITION(ServiceElementType == ServiceElementType::EVENT ||
                                          ServiceElementType == ServiceElementType::FIELD);
    const auto& lola_service_element_deployment = [this, service_element_name]() {
        if constexpr (ServiceElementType == ServiceElementType::EVENT)
        {
            return lola_service_instance_deployment_.events_.at(std::string{service_element_name});
        }
        else
        {
            return lola_service_instance_deployment_.fields_.at(std::string{service_element_name})
                .lola_event_instance_deployment_;
        }
    }();
    const std::string name{service_element_name};

    const auto field_it = lola_service_element_deployment.fields_.find(name);
    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(field_it != lola_service_instance_deployment_.fields_.cend(),
                                                "Could not find field in deployment configuration.");
    const auto number_of_slots = field_it->second.lola_event_instance_deployment_.GetNumberOfSampleSlots();
    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(number_of_slots.has_value(),
                                                "Number of sample slots not specified for field.");
    return number_of_slots.value();
}
```

bemerybmw: Same for GetMaxSubscribersFromConfig

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/linear_search_map.h
crimson11: At least map/unordered_map have `noexcept` spec ... so I took it over.

score/mw/com/impl/bindings/lola/linear_search_map.h
crimson11: No - imho the user shouldthen use `StringComparisonAdaptor` instead of `string` as a key! This is imho now our score solution if you want "transparent" compare between various string-like representations ...

score/mw/com/impl/bindings/lola/linear_search_map.h
crimson11: Because of the constraints we have here with our "fixed size" underlying storage!
With the signature you propose, we are forced to completely construct the key-value-pair 1st from the variadic-template args! Where shall we create this pair? In our fixed-size underlying array we don't have a guaranteed spare slot! Creating the whole pair on temp stack does also not work! MappedType might be LARGE! std::map et all can handle it, because they are anyhow node-based! The create the complete pair in a heap-allocated node and then either discard afterwards (if key already exists) or re-link the node!

We can't do all this - we would need to eventually move/shuffle around slots, which we can't do if e.g. MappedType is non-moveable ....

score/mw/com/impl/bindings/lola/linear_search_map.h
crimson11: It is just there for comfort reasons, for the typical/standard case, where you have single arg mapped_type construction! It saves you the `std::piecewise_construct` syntax of the 1st overload! Technically it isn't needed. Maybe I should document this ;)

score/mw/com/impl/bindings/lola/service_data_storage.h
crimson11: Will rephrase.

score/mw/com/impl/bindings/lola/linear_search_map.h
crimson11: Hm. I'm not convinced ;) Why?
- we resemble a std::map interface here! map doesn't lvalue qualify at() neither. This confuses imho.
- then I would need to be consistent! Also would have to add it o plenty of other methods! Like begin()/end() etc. they are all returning iterators/pointers which may be dangling in case of a temporary ...

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
crimson11: Makes no real sense for me ;) We are completly relying on `NonRelocatableVector` in the LinearSearchMap. This imho would just repeat the existing test you already have in `NonRelocatableVector`. This looks to me like no added value - just bloating the tests ...

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
crimson11: Yes! Somewhat! I replaced with `CalculateAlignedSize(size, kMaxAlignedSize)`

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
crimson11: Hm. I'm not convinced. Just compare the src-code sizes. It is not really much less source-code and the original was also "easy" to grasp imho. Now you are templating it and in reality we will have larger code-size, because we have now the impl. instantiated twice ... and as I said - the new impl. still has a cosnstexpr if-else, which makes the code not really simpler?

If you still insist -> I can change ;)

score/mw/com/impl/bindings/lola/linear_search_map.h
bemerybmw: I don't think this is / should be the score solution. The issue with StringComparisonAdaptor is that it's not clear if it owns the string or not. LinearSearchMap<std::string, int> and LinearSearchMap<std::string_view, int> are very different types! But  LinearSearchMap<StringComparisonAdaptor, int> could be either.

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+class LinearSearchMapFixture : public ::testing::Test
+{
+  protected:
+    memory::shared::NewDeleteDelegateMemoryResource memory_{kMemoryResourceId};
+};
+
+TEST_F(LinearSearchMapFixture, IsEmptyAfterConstruction)
+{
+    // Given a freshly constructed LinearSearchMap with a capacity of 4
```
bemerybmw:
```suggestion
```

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+
+    // When constructing the map
```
bemerybmw: When constructing a LinearSearchMap with a capacity of 4

bemerybmw: Same for the test below

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+    EXPECT_EQ(unit.begin(), unit.end());
+}
+
+TEST_F(LinearSearchMapFixture, EmplaceSucceedsAndReturnsCorrectIterator)
+{
+    // Given an empty LinearSearchMap
+    TestMap unit{4U, memory_};
+
+    // When emplacing a new key/value pair
+    const auto result = unit.emplace(1, 100);
```
bemerybmw: We should minimise the use of magic numbers which are reused. The key and value should be variables which are reused in the Then section.

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+{
+    // Given a const reference to a LinearSearchMap containing one element
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(5, 500);
+    const TestMap& const_unit = unit;
+
+    // When searching via the const overload of find
+    const auto it = const_unit.find(5);
+
+    // Then an existing key is found and a missing key yields the const end iterator
```
bemerybmw:
```suggestion
    // Then an existing key is found
```

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,254 @@
...
+    // Given a LinearSearchMap using a custom absolute-value key-equality predicate
+    CustomEqualMap unit{4U, memory_};
+
+    // When retrieving the predicate via key_eq
+    const auto predicate = unit.key_eq();
+
+    // Then the returned predicate exhibits the custom equality semantics
+    EXPECT_TRUE(predicate(-7, 7));
+    EXPECT_FALSE(predicate(-7, 8));
+}
```
bemerybmw, line 251: What do we want to happen when we try to emplace more elements than we reserved capacity for? I guess that it would have to terminate because otherwise emplace would return false but what iterator to return? I think this should be specified in the header docs and also probably should have a test here. I know the behaviour is implemented in NonRelocatableVector, but IMO, this is an important variant of this class which should be documented and tested.

score/mw/com/impl/bindings/lola/service_data_control.h
```diff
@@ -30,15 +34,19 @@ class ServiceDataControl
...
+    /// \brief Ctor for the ServiceDataControl with a given memory resource to be used..
     /// \details ServiceDataControl is designed to be located in shared memory, therefore the explicit
-    ///          MemoryResourceProxy argument! (Yes one could come up with a MemoryResourceProxy pointing to a local
-    ///          memory resource, but this would be "uncommon")
-    /// \param proxy MemoryResourceProxy pointing to the memory-resource to be used
+    ///          ManagedMemoryResource argument!
+    /// \param number_of_service_elements the (fixed) number of service-elements (events + fields) this service-instance
+    ///        provides. It is used as the fixed capacity of the event_controls_ container. Since event_controls_ uses a
+    ///        fixed-capacity container (LinearSearchMap), its capacity must be known at construction time.
+    /// \param resource ManagedMemoryResource pointing to the memory-resource to be used for allocating underlying
```
bemerybmw:
```suggestion
    /// \param resource ManagedMemoryResource used for allocating underlying
```

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
bemerybmw: Sorry, this was actually wrong by me. I thought it was EventDataControl. But there's only one ServiceDataControl per region.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
bemerybmw: My point wasn't really about number of lines, it was more about avoiding code duplication. We're currently duplicating the "algorithm" to extract number of slots from a deployment. So if we for example changed the structure of the deployments, we'd have to change it (correctly) in 2 places instead of 1. But it's such a small function, I'm fine if you leave it as is.

score/mw/com/impl/bindings/lola/linear_search_map.h
bemerybmw: There was some misra rule I believe about this. But I'm fine if you follow the std::map interface and we can revisit this if the warning pops up

score/mw/com/impl/bindings/lola/service_data_storage.cpp
```diff
@@ -11,3 +11,63 @@
...
+    std::size_t total_size = memory::shared::CalculateAlignedSize(sizeof(ServiceDataStorage), kMaxAlign);
+
+    // (2) The two backing arrays of the LinearSearchMaps (allocated once, with capacity == number_of_service_elements).
+    total_size += memory::shared::CalculateAlignedSize(
+        number_of_service_elements * sizeof(ServiceDataStorage::EventDataStorageMap::value_type), kMaxAlign);
+    total_size += memory::shared::CalculateAlignedSize(
+        number_of_service_elements * sizeof(ServiceDataStorage::EventMetaInfoMap::value_type), kMaxAlign);
+
+    // The size of the EventDataStorage control structure (a DynamicArray) is independent of the concrete sample-type
+    // (it only holds an offset-pointer, an allocator and two size_t members).
```
bemerybmw: We're making a lot of assumptions here about the internals of a DynamicArray. Would it make sense that this is added to the DynamicArray itself?

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,106 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+    return CalculateServiceDataControlShmSize(service_elements);
+}
+
+std::size_t SkeletonMemoryManager::GetNumberOfSampleSlotsFromConfig(const std::string_view service_element_name,
+                                                                    const bool is_field) const
+{
+    const std::string name{service_element_name};
+    if (is_field)
+    {
+        const auto field_it = lola_service_instance_deployment_.fields_.find(name);
```
bemerybmw: You can actually use GetServiceElementInstanceDeployment from score/mw/com/impl/configuration/lola_service_instance_deployment.h here.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,106 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(field_it != lola_service_instance_deployment_.fields_.cend(),
+                                                    "Could not find field in deployment configuration.");
+        const auto number_of_slots = field_it->second.lola_event_instance_deployment_.GetNumberOfSampleSlots();
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(number_of_slots.has_value(),
+                                                    "Number of sample slots not specified for field.");
+        return number_of_slots.value();
+    }
+    const auto event_it = lola_service_instance_deployment_.events_.find(name);
```
bemerybmw: You can actually use GetServiceElementInstanceDeployment from score/mw/com/impl/configuration/lola_service_instance_deployment.h here.

score/mw/com/impl/bindings/lola/service_data_control.h
```diff
@@ -71,6 +79,31 @@ class ServiceDataControl
...
+///          derived purely from the (fixed) container capacities. The result mirrors exactly what the real construction
+///          allocates on the (monotonic) shared-memory resource. Eventually slightly more as we have to assume, that
+///          memory allocation starts from a worst-case aligned situation.
+///          The same size applies to the QM and (if present) the ASIL-B control shm-object, as both hold a
+///          ServiceDataControl created with the very same configuration.
+/// \param service_elements_size_info per service-element sizing information. Its size equals the number of
+///        service-elements (events + fields), which is the fixed capacity the event_controls_ container is constructed
+///        with.
+/// \return needed size (in bytes) for a single control shm-object.
+std::size_t CalculateServiceDataControlShmSize(
```
bemerybmw, line 103: We're missing tests for this?

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -42,15 +65,40 @@ class ServiceDataStorage
...
+///          size algorithm stays coupled to the data structure it reasons about. It does NOT allocate any memory nor
+///          construct a ServiceDataStorage; the size is derived purely from the (fixed) container capacities.
+///          The result mirrors exactly what the real construction allocates on the (monotonic) shared-memory resource.
+///          Eventually slightly more as we have to assume, that memory allocation starts from a worst-case aligned
+///          situation.
+/// \param service_elements_size_info per service-element sizing information. Its size equals the number of
+///        service-elements (events + fields), which is the fixed capacity the ServiceDataStorage containers are
+///        constructed with.
+/// \return needed size (in bytes) for the data shm-object.
+std::size_t CalculateServiceDataStorageShmSize(
```
bemerybmw, line 99: We're missing tests for this?

score/mw/com/impl/bindings/lola/linear_search_map.h
crimson11: You are right! Currently we are (in my PR) only using `LinearSearchMap` with an `ElementFqId` as key. So right now we don't need such "optimizations"! Let's add them, if we need them.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
crimson11: will leave it.

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
crimson11: Added test and doc.

score/mw/com/impl/bindings/lola/service_data_storage.cpp
crimson11: Is this our discussion regarding splitting up/making a recursive approach for `CalculateServiceDataStorageShmSize` ? I.e. also implement a corresponding calculate-method in `DynamicArray` ?
This we could do, but I would like to postpone it after we have switched to type-erased storage for the binding ... as it will change a lot again.

score/mw/com/impl/bindings/lola/service_data_control.h
crimson11: Added tests

score/mw/com/impl/bindings/lola/service_data_storage.h
crimson11: Added tests

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +538,102 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+                                                               const bool is_field) const
+{
+    const std::string name{service_element_name};
+    if (is_field)
+    {
+        const auto deployment =
+            GetServiceElementInstanceDeployment<ServiceElementType::FIELD>(lola_service_instance_deployment_, name);
+        SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(
+            deployment.lola_event_instance_deployment_.max_subscribers_.has_value(),
+            "Number of sample slots need to be specified for event on provider side!");
```
bemerybmw:
```suggestion
            "Max subscribers need to be specified for field on provider side!");
```

score/mw/com/impl/bindings/lola/linear_search_map_test.cpp
```diff
@@ -0,0 +1,269 @@
...
+{
+    // Given a const reference to a LinearSearchMap containing one element
+    TestMap unit{4U, memory_};
+    std::ignore = unit.emplace(kSomeKey, kSomeValue);
+    const TestMap& const_unit = unit;
+
+    // When searching via the const overload of find
+    const auto it = const_unit.find(kSomeKey);
+
+    // Then an existing key is found and
```
bemerybmw:
```suggestion
    // Then an existing key is found
```

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -89,14 +90,18 @@ struct ServiceElementDataStorageSizeInfo
...
-std::size_t CalculateServiceDataStorageShmSize(
+/// \return a pair of (in bytes) needed sizes for the data shm-object:
+///         - first: the minimal size, i.e. the size that would be needed if none of the individual allocations
+///           performed by the real construction needed any alignment padding at all.
+///         - second: the worst-case size, i.e. the size that is needed if every individual allocation performed by
+///           the real construction started off in the worst-case (i.e. maximally unaligned) situation and therefore
+///           needed the maximum possible alignment padding. This mirrors what the real construction allocates on the
+///           (monotonic) shared-memory resource, eventually slightly overestimating it, as we have to assume that
+///           memory allocation starts from a worst-case aligned situation.
+std::pair<std::size_t, std::size_t> CalculateServiceDataStorageShmSize(
```
bemerybmw: Better to return a struct than a pair so that the values are named rather than accessing via .first / .second.

score/mw/com/impl/bindings/lola/service_data_storage.cpp
```diff
@@ -13,71 +13,67 @@
...
     // The number of service-elements determines the (fixed) capacity of the two LinearSearchMaps within the
     // ServiceDataStorage. It equals the number of sizing entries handed over.
     const auto number_of_service_elements = service_elements_size_info.size();
 
+    std::size_t minimal_size{0U};
+    std::size_t worst_case_size{0U};
+
     // (1) The ServiceDataStorage object itself (including the inline bookkeeping of its two LinearSearchMaps).
-    std::size_t total_size = CalculateWorstCaseAllocationSize(sizeof(ServiceDataStorage), alignof(ServiceDataStorage));
+    AccumulateAllocationSize(sizeof(ServiceDataStorage), alignof(ServiceDataStorage), minimal_size, worst_case_size);
```
bemerybmw: Why do we need to calculate the size using worst case alignment for every allocation? I would have thought that we would do this for the first allocation, but after that, the size / alignment calculations would be deterministic for all other allocations, assuming that we calculate them all in the same order as they're allocated in production code (and obviously don't miss any allocations).

And if we actually do the first allocation in shared memory using max_align_t then wouldn't the simulation run then calculate the same size as the analysis run?

score/mw/com/impl/bindings/lola/test/skeleton_component_test.cpp
```diff
@@ -878,6 +816,130 @@ TEST_F(SkeletonComponentTestFixture,
...
+    // and that the size calculated by analysis is only as much greater than the size resulting from simulation as
+    // the maximum alignment-padding overshoot CalculateServiceDataControlShmSize() itself reports for this exact
+    // configuration (the delta between its worst-case and its minimal size).
+    const auto control_shm_size_analysis = GetSize(control_shm);
+    EXPECT_GE(control_shm_size_analysis, control_shm_size_simulation);
+
+    const std::vector<ServiceElementControlSizeInfo> service_elements_size_info{
+        ServiceElementControlSizeInfo{kEventProperties.GetTotalNumberOfSlots(), kEventProperties.max_subscribers},
+        ServiceElementControlSizeInfo{kEventProperties.GetTotalNumberOfSlots(), kEventProperties.max_subscribers}};
+    const auto [minimal_size, worst_case_size] = CalculateServiceDataControlShmSize(
```
bemerybmw: I don't really like that we are modifying the production function just to use the minimal_size here in a test. I'd like to understand this point better and then see if we can come up with a better solution: https://github.com/eclipse-score/communication/pull/760/changes/13d59fa31253a687bbed7c29f94174b16b08494a..c60cea9e544f63cf221586f77d524dd3f08b1716#r3747958827

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/service_data_storage.h
crimson11: outdated.

score/mw/com/impl/bindings/lola/service_data_storage.cpp
crimson11: fixed.

score/mw/com/impl/bindings/lola/test/skeleton_component_test.cpp
crimson11: Fixed.

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -75,27 +75,35 @@ class ServiceDataStorage
...
-    /// \brief Number of event-data slots the service-element provides.
-    std::size_t number_of_slots;
-    /// \brief Size (in bytes) of a single event-data slot, already padded to the slot's alignment. For typed events
-    ///        this equals sizeof(SampleType); for generic (type-erased) events it is the sample-size padded to the
-    ///        sample-alignment.
-    std::size_t aligned_slot_size;
+    /// \brief Exact size (in bytes) of the raw slot-array that will be allocated for this service-element.
+    std::size_t slot_array_size;
+    /// \brief Alignment (in bytes) required for the raw slot-array allocation.
+    std::size_t slot_array_alignment;
```
bemerybmw: Why is this not the memory::DataTypeSizeInfo?

score/mw/com/impl/bindings/lola/service_data_storage.cpp
```diff
@@ -11,3 +11,60 @@
...
+    // always starts allocating at a std::max_align_t aligned location. Since we know the exact size/alignment of
+    // every single one of these allocations and the exact order in which they happen, we can reconstruct the exact
+    // sequence here and let memory::shared::CalculateAlignedSizeOfSequence() compute the exact (not just worst-case)
+    // total size, taking into account the exact alignment-padding between consecutive allocations.
+    std::vector<score::memory::DataTypeSizeInfo> allocation_sequence{};
+
+    // (1) The ServiceDataStorage object itself (including the inline bookkeeping of its two LinearSearchMaps).
+    allocation_sequence.emplace_back(sizeof(ServiceDataStorage), alignof(ServiceDataStorage));
+
+    // (2) The two backing arrays of the LinearSearchMaps (allocated once, with capacity == number_of_service_elements).
```
bemerybmw:
```suggestion
    // (2) The two allocated arrays of the LinearSearchMaps (allocated once, with capacity == number_of_service_elements).
```

score/mw/com/impl/bindings/lola/service_data_storage.cpp
```diff
@@ -11,3 +11,60 @@
...
+    allocation_sequence.emplace_back(
+        number_of_service_elements * sizeof(ServiceDataStorage::EventDataStorageMap::value_type),
+        alignof(ServiceDataStorage::EventDataStorageMap::value_type));
+    allocation_sequence.emplace_back(
+        number_of_service_elements * sizeof(ServiceDataStorage::EventMetaInfoMap::value_type),
+        alignof(ServiceDataStorage::EventMetaInfoMap::value_type));
+
+    // The size/alignment of the EventDataStorage control structure (a DynamicArray) is independent of the concrete
```
bemerybmw: Nitpick, this should be after the comment (3) For each...? Because it's related to EventDataStorage.

score/mw/com/impl/bindings/lola/service_data_control.h
```diff
@@ -30,15 +34,18 @@ class ServiceDataControl
     ///       calculation based on config settings and then hand over this calculated number ...
     static constexpr std::uint16_t kMaxApplicationIdPidMappings = 50U;
 
-    /// \brief Ctor for the ServiceDataControl to place it in given shared memory resource identified via given
-    ///        memory resource proxy.
+    /// \brief Ctor for the ServiceDataControl with a given memory resource to be used..
```
bemerybmw:
```suggestion
    /// \brief Ctor for the ServiceDataControl with a given memory resource to be used.
```

score/mw/com/impl/bindings/lola/service_data_control.h
```diff
@@ -30,15 +34,18 @@ class ServiceDataControl
...
     /// \details ServiceDataControl is designed to be located in shared memory, therefore the explicit
-    ///          MemoryResourceProxy argument! (Yes one could come up with a MemoryResourceProxy pointing to a local
-    ///          memory resource, but this would be "uncommon")
-    /// \param proxy MemoryResourceProxy pointing to the memory-resource to be used
+    ///          ManagedMemoryResource argument!
+    /// \param number_of_service_elements the (fixed) number of service-elements (events + fields) this service-instance
```
bemerybmw: I guess we should be careful with the term service elements and try to use it conistently to mean events, fields and methods. I would call this number_of_events_and_fields to be clear.

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -17,22 +17,45 @@
...
+    /// \brief Ctor for the ServiceDataStorage with a given memory resource to be used for internal storage allocation.
+    /// \details ServiceDataStorage no longer uses dynamically allocating map-types. Instead it uses fixed-capacity
+    ///          containers (LinearSearchMap) whose capacity has to be provided at construction time. The capacity
+    ///          equals the number of service-elements (events + fields) of the service-instance, which is known
+    ///          up-front. This makes the memory footprint of ServiceDataStorage deterministic and calculable without a
+    ///          simulation run.
+    /// \param number_of_service_elements maximum number of service-elements (events + fields) that will be stored in
+    ///        events_ and events_metainfo_.
+    /// \param resource memory-resource to be used for the (single, up-front) allocation of the containers.
+    ServiceDataStorage(const std::size_t number_of_service_elements,
```
bemerybmw: I guess we should be careful with the term service elements and try to use it conistently to mean events, fields and methods. I would call this number_of_events_and_fields to be clear.

bemerybmw: Maybe you can do a quick search in your pr for usages of "\*service_elements\*".

score/mw/com/impl/bindings/lola/service_data_control.cpp
```diff
@@ -11,3 +11,99 @@
...
+    // of these allocations and the exact order in which they happen, we can reconstruct the exact sequence here and
+    // let memory::shared::CalculateAlignedSizeOfSequence() compute the exact (not just worst-case) total size, taking
+    // into account the exact alignment-padding between consecutive allocations.
+    std::vector<score::memory::DataTypeSizeInfo> allocation_sequence{};
+
+    // (1) The ServiceDataControl object itself (including the inline bookkeeping of its event_controls_ LinearSearchMap
+    // and of its application_id_pid_mapping_).
+    allocation_sequence.emplace_back(sizeof(ServiceDataControl), alignof(ServiceDataControl));
+
+    // (2) The backing array of the event_controls_ LinearSearchMap (allocated once, with capacity ==
```
bemerybmw:
```suggestion
    // (2) The allocated array of the event_controls_ LinearSearchMap (allocated once, with capacity ==
```

score/mw/com/impl/bindings/lola/service_data_control.cpp
```diff
@@ -11,3 +11,99 @@
...
+
+    // (3) The backing array of the application_id_pid_mapping_ (a fixed-capacity DynamicArray with a capacity of
+    // kMaxApplicationIdPidMappings). It is allocated once during ServiceDataControl construction, independent of the
+    // number of service-elements.
+    allocation_sequence.emplace_back(static_cast<std::size_t>(ServiceDataControl::kMaxApplicationIdPidMappings) *
+                                         sizeof(ApplicationIdPidMappingEntry),
+                                     alignof(ApplicationIdPidMappingEntry));
+
+    // (4) For each event/field: the (deeply) nested fixed-capacity DynamicArrays contained within its EventControl.
+    // The EventControl object itself is stored inline within the event_controls_ backing array (accounted for in (2));
```
bemerybmw, line 62:
```suggestion
    // The EventControl object itself is stored inline within the event_controls_ allocated array (accounted for in (2));
```

score/mw/com/impl/bindings/lola/service_data_control.cpp
```diff
@@ -11,3 +11,99 @@
...
+    // only the backing arrays of its nested DynamicArrays allocate separately and are accounted for here.
```
bemerybmw:
```suggestion
    // only the allocated arrays of its nested DynamicArrays allocate separately and are accounted for here.
```

score/mw/com/impl/bindings/lola/service_data_control.cpp
bemerybmw: This is my preference as I think it's clearer and less open to interpretation, but I don't have a very strong opinion so you can decide.

score/mw/com/impl/bindings/lola/service_data_control.cpp
```diff
@@ -11,3 +11,99 @@
...
+    for (const auto& service_element : service_elements_size_info)
+    {
+        const std::size_t number_of_slots = service_element.number_of_slots;
+        const std::size_t max_subscribers = service_element.max_subscribers;
+
+        // (4a) EventControl::data_control (EventDataControl): its state_slots_ is a DynamicArray<ControlSlotType>
+        // with a capacity of number_of_slots.
+        allocation_sequence.emplace_back(number_of_slots * sizeof(EventDataControl::EventControlSlots::value_type),
```
bemerybmw, line 71: We have a lot of these cases in which we're calculating the size of a dynamic array. As already discussed, I'd like to see a function (probably alongside DynamicArray) which does this. To avoid us for example accidentally doing something like this:
```
allocation_sequence.emplace_back(number_of_slots * sizeof(EventDataControl::EventControlSlots::value_type),
alignof(TransactionLog::TransactionLogSlots::value_type));
```
Doesn't have to be done now, but keep it in mind.

score/mw/com/impl/bindings/lola/service_data_control.cpp
```diff
@@ -11,3 +11,99 @@
...
+                                         alignof(EventDataControl::EventControlSlots::value_type));
+
+        // (4b) EventControl::transaction_log_set_ (TransactionLogSet):
+        //   proxy_transaction_logs_(max_subscribers, TransactionLogNode{number_of_slots, resource}, resource) is
+        //   constructed via the DynamicArray fill-constructor. The exact allocation order is:
+        //     * one reference_count_slots_ array for the temporary prototype TransactionLogNode, which is
```
bemerybmw, line 77: Make it clear that the allocation is because the TransactionLogNode itself contains a TransactionLog which containg a DynamicArray.

score/mw/com/impl/bindings/lola/service_data_storage.cpp
```diff
@@ -11,3 +11,60 @@
...
+    constexpr std::size_t event_data_storage_object_size = sizeof(EventDataStorage<std::max_align_t>);
+    constexpr std::size_t event_data_storage_object_alignment = alignof(EventDataStorage<std::max_align_t>);
+
+    // (3) For each event/field (in the exact order it gets registered/offered): the EventDataStorage object plus its
+    // raw slot-array. The exact size/alignment of the slot-array (which differs between typed and generic
+    // events/fields, see SkeletonMemoryManager::CreateEventDataInCreatedSharedMemory() resp.
+    // CreateGenericEventDataInCreatedSharedMemory()) is provided by the caller.
+    for (const auto& service_element : service_elements_size_info)
+    {
+        allocation_sequence.emplace_back(event_data_storage_object_size, event_data_storage_object_alignment);
```
bemerybmw, line 64: Would it maybe be clearer to do number_of_service_elements * event_data_storage_object_size outside the loop, since it's the same for all iterations and can then go directly with its comment above (and also matches the rest of the code).

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -129,6 +157,12 @@ auto SkeletonMemoryManager::CreateSharedMemory(
     SkeletonBinding::SkeletonFieldBindings& fields,
     std::optional<SkeletonBinding::RegisterShmObjectTraceCallback> register_shm_object_trace_callback) -> Result<void>
 {
+    // The number of service-elements determines the (fixed) capacity of the containers within ServiceDataStorage and
+    // ServiceDataControl. It has to be known before either of them is constructed (in CreateSharedMemoryForData /
+    // CreateSharedMemoryForControl below), independent of how the shm-object sizes are determined (simulation,
+    // analytic estimation, or manually specified sizes).
+    number_of_service_elements_ = events.size() + fields.size();
```
bemerybmw:
```suggestion
    number_of_events_and_fields_ = events.size() + fields.size();
```

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +558,126 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+    // mixes both kinds - see AreServiceElementBindingsGeneric() for details). This distinction between generic/typed
+    // bindings is necessary as the memory allocation taking place is slightly different currently. When we switch to
+    // complete type-erasure at some point for the binding layer, this distinction will go away!
+    const bool are_bindings_generic = AreServiceElementBindingsGeneric(events, fields);
+
+    // Collect the per service-element sizing information from the deployment configuration and the (typed/generic)
+    // event bindings. The layout-dependent size algorithm itself lives next to ServiceDataStorage
+    // (CalculateServiceDataStorageShmSize), so that the data-structure and the algorithm reasoning about its memory
+    // footprint stay closely coupled.
+    std::vector<ServiceElementDataStorageSizeInfo> service_elements{};
```
bemerybmw: Since this is basically an out arg, I would put it as an argument to the lambda instead of capturing it to make it more explicit in the calling code.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +558,126 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+    const auto collect_service_elements = [this, &service_elements, are_bindings_generic](
+                                              SkeletonBinding::SkeletonEventBindings& bindings, const bool are_fields) {
```
bemerybmw:
```suggestion
                                              auto& bindings, const bool are_fields) {
```
This currently only works because SkeletonBinding::SkeletonEventBindings == SkeletonBinding::SkeletonFieldBindings?

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +558,126 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+            // (see SkeletonMemoryManager::CreateGenericEventDataInCreatedSharedMemory()): the raw byte count gets
+            // rounded up to a whole number of std::max_align_t elements and the array is aligned to
+            // alignof(std::max_align_t). A typed event/field allocates its slot-array as an
+            // EventDataStorage<SampleType> (see SkeletonMemoryManager::CreateEventDataInCreatedSharedMemory()):
+            // exactly number_of_slots * sizeof(SampleType) bytes, aligned to alignof(SampleType) - no rounding.
+            if (are_bindings_generic)
+            {
+                const auto& generic_event_binding = static_cast<const GenericSkeletonEventBinding&>(event_binding);
+                const auto [sample_size, sample_alignment] = generic_event_binding.GetSizeInfo();
+                const auto aligned_sample_size = memory::shared::CalculateAlignedSize(sample_size, sample_alignment);
```
bemerybmw: If GetSizeInfo returns memory::DataTypeSizeInfo then we wouldn't need CalculateAlignedSize? And once the type erasing stuff is done and this hack for the broken generic skeleton is fixed, then we could just put the memory::DataTypeSizeInfo returned by CalculateAlignedSize directly into the vector?

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -504,6 +558,126 @@ SkeletonMemoryManager::ShmResourceStorageSizes SkeletonMemoryManager::CalculateS
...
+}
+
+std::size_t SkeletonMemoryManager::CalculateControlShmResourceStorageSize(
+    SkeletonBinding::SkeletonEventBindings& events,
+    SkeletonBinding::SkeletonFieldBindings& fields) const
+{
+    // Collect the per service-element sizing information from the deployment configuration. The layout-dependent size
+    // algorithm itself lives next to ServiceDataControl (CalculateServiceDataControlShmSize), so that the
+    // data-structure and the algorithm reasoning about its memory footprint stay closely coupled.
+    std::vector<ServiceElementControlSizeInfo> service_elements{};
```
bemerybmw: Since this is basically an out arg, I would put it as an argument to the lambda instead of capturing it to make it more explicit in the calling code.

score/mw/com/impl/bindings/lola/test/skeleton_component_test.cpp
```diff
@@ -549,11 +551,109 @@ TEST_F(SkeletonComponentTestFixture,
     EXPECT_TRUE(prepare_offer_result.has_value());
 }
 
+/// \brief Verifies that the data shared-memory object size can be calculated analytically (mode kAnalysis), i.e.
+///        WITHOUT relying on the value obtained from the simulation dry-run, and that the analytically sized shared
+///        memory is large enough to hold the actually registered events/fields.
+TEST_F(SkeletonComponentTestFixture, ShmObjectSizeCalc_Analysis_Data_QM)
+{
+    RecordProperty("Verifies", "SCR-5899126");
```
bemerybmw, line 559: Is there any reason to have this test if we have ShmObjectSizeCalc_Analysis_QM_Control_SizeMatchesSimulationResult below?

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -64,6 +64,9 @@ class SkeletonEventBindingBase
     /// allocations)
     virtual std::size_t GetMaxSize() const noexcept = 0;
 
+    /// \brief Alignment requirement (in bytes) of the underlying event-type's sample data.
+    virtual std::size_t GetAlignment() const noexcept = 0;
```
bemerybmw, line 68: Would it make sense to have a single function which returns memory::DataTypeSizeInfo? I think we should just try to use that everywhere so that we get the benefit of the checks and also we can pass it around easier. Can be done in a follow up.

score/mw/com/impl/bindings/lola/service_data_control_test.cpp
```diff
@@ -0,0 +1,181 @@
...
+    memory::shared::NewDeleteDelegateMemoryResource resource{kMemoryResourceId};
+    const auto actual_allocated_bytes = ConstructServiceDataControlAndGetAllocatedBytes({}, resource);
+
+    EXPECT_EQ(calculated_size, actual_allocated_bytes);
+}
+
+TEST(ServiceDataControlShmSizeTest, SingleServiceElementCalculatedSizeMatchesActualAllocation)
+{
+    // Given the sizing information of a single service-element (event/field)
+    const std::vector<ServiceElementControlSizeInfo> service_elements_size_info{ServiceElementControlSizeInfo{5U, 3U}};
```
bemerybmw: This ServiceElementControlSizeInfo is invalid? Size should be multiple of alignment?

score/mw/com/impl/bindings/lola/service_data_control_test.cpp
```diff
@@ -0,0 +1,181 @@
...
+                                  static_cast<EventControl::SubscriberCountType>(service_element.max_subscribers),
+                                  kEnforceMaxSamples,
+                                  resource));
+        ++element_id;
+    }
+
+    return resource.GetUserAllocatedBytes();
+}
+
+TEST(ServiceDataControlShmSizeTest, EmptyServiceElementsYieldsSizeOfServiceDataControlWithoutEventControls)
```
bemerybmw: These 3 tetsts should be paramaterised tests since they're identical but just depend on the ServiceElementControlSizeInfo provided e.g. score/memory/shared/pointer_arithmetic_util_calculate_aligned_size_test.cpp

score/mw/com/impl/bindings/lola/service_data_storage_test.cpp
```diff
@@ -53,6 +58,22 @@ class ServiceDataStorageFixture : public ::testing::Test
...
+/// \brief Fixture for tests that need to actually construct a real ServiceDataStorage (and not just call
+/// CalculateServiceDataStorageShmSize()). In addition to mocking the binding-runtime (like ServiceDataStorageFixture
+/// does), it also stubs the PID/UID retrieval performed by ServiceDataStorage's constructor, since the concrete
+/// values are irrelevant for the shm-layout/size.
+class ServiceDataStorageShmSizeFixture : public ServiceDataStorageFixture
+{
+  public:
+    ServiceDataStorageShmSizeFixture()
+    {
+        ON_CALL(lola_runtime_mock_, GetPid()).WillByDefault(::testing::Return(pid_t{42}));
```
bemerybmw: Why  create a new fixture? Why not just add these to ServiceDataStorageFixture?

score/mw/com/impl/bindings/lola/service_data_storage_test.cpp
```diff
@@ -123,11 +112,185 @@ TEST_F(ServiceDataStorageFixture, GetsUidFromRuntimAndStoresItOnConstruction)
...
+    const auto size_with_fewer_slots = CalculateServiceDataStorageShmSize(
+        score::cpp::span<const ServiceElementDataStorageSizeInfo>{service_elements_with_fewer_slots});
+    const auto size_with_more_slots = CalculateServiceDataStorageShmSize(
+        score::cpp::span<const ServiceElementDataStorageSizeInfo>{service_elements_with_more_slots});
+
+    // Then the calculated size for the service-element with the bigger raw slot-array is bigger.
+    EXPECT_GT(size_with_more_slots, size_with_fewer_slots);
+}
+
+TEST(ServiceDataStorageShmSizeTest, IncreasingAlignedSlotSizeOfAServiceElementIncreasesCalculatedSize)
```
bemerybmw: How is this test different to the one before?

score/mw/com/impl/bindings/lola/service_data_storage_test.cpp
```diff
@@ -123,11 +112,185 @@ TEST_F(ServiceDataStorageFixture, GetsUidFromRuntimAndStoresItOnConstruction)
...
+            default:
+                ADD_FAILURE() << "Unsupported slot_array_alignment in test: " << service_element.slot_array_alignment;
+                break;
+        }
+    }
+
+    return resource.GetUserAllocatedBytes();
+}
+
+TEST_F(ServiceDataStorageShmSizeFixture, EmptyServiceElementsYieldsSizeOfServiceDataStorageWithoutEventDataStorages)
```
bemerybmw: Same here that these should be typed tests e.g. score/memory/shared/pointer_arithmetic_util_calculate_aligned_size_test.cpp

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/service_data_storage.h
crimson11: good find! The whole struct `ServiceElementDataStorageSizeInfo` is basically a "replica" of `memory::DataTypeSizeInfo`! Refactored to use memory::DataTypeSizeInfo!

score/mw/com/impl/bindings/lola/service_data_storage.h
crimson11: OK! Fixed it in various other places .... I like, that you are now also very concise with naming! :P

score/mw/com/impl/bindings/lola/service_data_control.cpp
crimson11: I would like to shift it! I'm with you, but my "upcoming" type-erased-binding PR will get rid of these DynamicArrays anyhow ... so this would be waste now.

score/mw/com/impl/bindings/lola/service_data_control.cpp
crimson11: Re-phrased it!

score/mw/com/impl/bindings/lola/service_data_storage.cpp
crimson11: No, because it would NOT represent the exact order! And therefore we could have different padding/alignment!

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
crimson11: Touché! You are right! This CalculateAlignedSize() is UNNEEDED here! Because `sample_size` and `sample_alignment` already come from a `memory::DataTypeSizeInfo` instance! And therefore `sample_size` is already the correctly aligned size! (This changed by my previous PR, wherte I introduced `memory::DataTypeSizeInfo` to be used internally!

And YES -> all this special code vanishes soonish, when my next PR (type erased binding) gets merged. Teaser: I vastly simplifies our code in so many places! :D

score/mw/com/impl/bindings/lola/test/skeleton_component_test.cpp
crimson11: This is a very good question! I would leave this test in (for now): It does a semantic check, which should be always true! If this test fails, it would tell us, that something with kSimulation AND kAnalysis is BROKEN

score/mw/com/impl/skeleton_event_binding.h
crimson11: Let me do it with my type-erasure!

score/mw/com/impl/bindings/lola/service_data_control_test.cpp
crimson11: Nope ;)
You are confusing it with the old/now replaced `ServiceElementDataStorageSizeInfo`

score/mw/com/impl/bindings/lola/service_data_storage_test.cpp
crimson11: Yep - integrated into existing fixture!

score/mw/com/impl/bindings/lola/service_data_storage_test.cpp
crimson11: This was a left-over! Removed duplicate and renamed the other accordingly.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
bemerybmw: but for now, generic_event_binding.GetSizeInfo() returns the size/alignment container which doesn't check the invariants. should we leave the CalculateAlignedSize until GetSizeInfo is changed to return DataTypeSizeInfo?

crimson11: See my comment above: The implementation of `GetSizeInfo()` we currently have, creates the returned pair from `DataTypeSizeInfo`! So - currently we alread make SURE, that the invariants are fulfilled!
I will change the signature of generic_event_binding.GetSizeInfo() in a later PR.

bemerybmw (approved):

PR #762, opened 2026-07-24
Support for loading multiple configurations as addons

SebSparrowHawk:


score/mw/com/impl/configuration/example/mw_com_config_to_merge_second.json
```diff
@@ -0,0 +1,102 @@
...
+      "B-receiver": 5,
+      "B-sender": 12
+    },
+    "shm-size-calc-mode": "SIMULATION"
+  },
+  "tracing": {
+    "enable": true,
+    "applicationInstanceID": "ara_com_example",
+    "traceFilterConfigPath": "./mw_com_trace_filter.json"
+  }
```
castler, line 101: These are parts of the "AddOn" configuration that we should not allow IMHO. A library provider cannot know global values!

score/mw/com/impl/runtime.h
```diff
@@ -131,16 +145,28 @@ class Runtime final : public IRuntime
...
+    mutable std::mutex configuration_mutex_;
+
     /// \brief flag, whether the runtime is now locked in the sense, that the singleton has been already created
     ///        with a given configuration.
     static bool runtime_initialization_locked_;
 
+    /// \brief flag, to identify if an add-on configuration has already been loaded, which has been handled as the
+    /// initial configuration. Used to identify that such a configuration has been loaded before the user explicitly
+    /// loaded a configuration.
+    static bool addon_configuration_loaded_;
```
castler: I would not do that. IMHO we should never load an "AddOn" configuration as initial configuration. This can only lead to problems.

I would clearly state, that AddOn configurations can only be loaded after mw::com was correctly initalized.

Or we make it completly interexchangeable, but then we should _not_ use this variable.

score/mw/com/impl/runtime.cpp
```diff
@@ -28,6 +28,7 @@
 #include <score/assert.hpp>
 #include <score/utility.hpp>
 
+#include <exception>
```
castler, line 31: Why is this include required?

score/mw/com/impl/runtime.cpp
```diff
@@ -218,6 +249,7 @@ Runtime::~Runtime() noexcept
 
 std::vector<InstanceIdentifier> Runtime::resolve(const InstanceSpecifier& specifier) const
 {
+    std::lock_guard<std::mutex> lock{configuration_mutex_};
```
castler, line 252: So you added here locks for what reason? To ensure that we have no race condition of using the configuration while extending it?

How would we handle configuration usages from e.g. Proxy / Skeleton creations.

SebSparrowHawk:
After requirements and expectations have been changed for this feature, the actual changes have been merged with https://github.com/eclipse-score/communication/pull/930

PR #763, opened 2026-07-24
Extend CPP API checker with further cases

castler:


quality/api_surface/tests/golden/constexpr_members.lock.json
hoe-jo, line 1: All Files in Folder are autogenerated? So mark them as linguist file via gitattributes?

quality/api_surface/tests/test_api_extraction_validation.py
hoe-jo, line 1: - No Negative test cases at all
- What about test cases where private attributes change?

quality/api_surface/tests/test_extract_api.py
hoe-jo, line 1: Those unit tests are hard to read, and to understand if they really now cover everything. What about e2e tests?

hoe-jo, line 1: Also not covered by any build file

hoe-jo (commented):
Some Comments, also (not directly related to the PR) why use system clang and not llvm toolchain bundled one?
Also a specification/description would be very helpful on what gets validated

quality/api_surface/tests/golden/constexpr_members.lock.json
castler: Discussed and decided not to mark these as `linguist-generated`. Although the `golden/*.lock.json` files are produced by the `*_api_test.update` targets, they are committed on purpose as review-relevant fixtures: a change to one of them signals an actual public-API-surface change that reviewers should see in the diff. Marking them `linguist-generated` would collapse them by default and hide exactly the signal we want visible. They are already JSON, so they don't skew language stats meaningfully. Happy to revisit if you feel strongly.

quality/api_surface/tests/test_api_extraction_validation.py
castler: Added in 47e4da3f:
- **Private attributes change**: `private_changes_v2.h` declares the same public `test::StableApi` as `private_changes.h` but with a completely different private implementation (different data members and helpers). New `TestPrivateAttributeChanges` asserts the extracted public surface is *identical* across the two, and that none of the (differing) private members leak.
- **Negative cases**: `all_private.h` covers a class with only private members plus a public class inside an internal `test::detail` namespace; `TestNegativeCases` asserts none of them are exported. A missing-symbol lookup is asserted to return `None`.
- The e2e test additionally proves the checker "has teeth": a renamed public method *is* detected as a surface change (`TestPrivateChangeStability.test_public_change_is_detected`).

quality/api_surface/tests/test_extract_api.py
castler: These *are* the e2e tests: `test_extract_api.py` drives the full `clang -> extract_api.py` pipeline end to end for every case (runs clang to produce the AST dump, then feeds it to the extractor and asserts on the result), whereas `test_api_extraction_validation.py` only reads the JSON that Bazel already generated. I clarified this in the module docstring in 47e4da3f and grouped the new negative/stability scenarios explicitly. Each test class maps 1:1 to a small fixture header so coverage is auditable per construct. Happy to split/rename further if specific cases are still unclear.

castler: Fixed in 47e4da3f: wired `test_extract_api.py` into the build as the `extract_api_test` py_test. It runs under Bazel using the hermetic LLVM toolchain clang (`@llvm_toolchain//:bin/clang-cpp`) passed via `CLANG_PATH`; `get_clang_binary` now resolves that runfiles-relative path against `TEST_SRCDIR`. Verified green: `bazel test //quality/api_surface/tests:extract_api_test`.

hoe-jo (approved):

PR #773, opened 2026-07-24
Dependency compatibility checker

castler:


quality/dependency_compatibility_checker/README.md
```diff
@@ -0,0 +1,116 @@
...
+-->
+
+# Dependency Compatibility Checker
+
+A config-driven tool that tests combinations of external dependency versions (and
+Bazel versions) against the root `MODULE.bazel`, classifies each combination as
+green / orange / red, and produces an HTML + JSON report.
+
+See the design spec for the full rationale:
+`docs/superpowers/specs/2026-07-24-dependency-compatibility-checker-design.md`.
```
castler: Remove

.github/workflows/test_dependency_version_combinations.yml
hoe-jo, line 1: Not maintainable / really complex to understand what is done here.
What about move it to a python script which we could also test?

hoe-jo, line 1: Also I think we should include it in our docs build (valuable addition)

quality/dependency_compatibility_checker/module_rewriter.py
hoe-jo, line 1: This Regex stuff is very brittle, either replace by some parser or enhance the test suite, both positive and negative cases, currently I guess a bracket inside a comment would break the tool

.github/workflows/test_dependency_version_combinations.yml
castler: What exactly is to complex? What exactly you want to extract into a python script?

We want to utilize Github Actions matrix builds, if you test all combinations within one action, it will take ages and not run in a parallel?

castler: We can at some point, goal of this is to get this kind of checks starting.

hoe-jo (approved):

PR #777, opened 2026-07-27
Rust::com Method Rust APIs design and Example usage

bharatGoswami8:
edit -

Since the Field and Method changes are related, and the Field implementation also uses the Method interface while both share the same macro for the interface and type-state pattern, I believe it would be better to review the overall design in a single PR. This will help ensure that any suggestions or feedback can be addressed more effectively and consistently across both implementations.
Therefore, I am closing this PR. Please review the PR - https://github.com/eclipse-score/communication/pull/818

---------------------------------
- Created Method APIs design interface and example of usage.
- Updated supporting macro like interface macro.
- Runtime side created impl block.

https://github.com/eclipse-score/communication/issues/579


score/mw/com/impl/rust/com-api/com-api-runtime-lola/method.rs
```diff
@@ -0,0 +1,151 @@
...
+    unsafe fn assume_init(self) -> ZeroCopyArgs<LolaMethodInArgPtr<T>> {
+        todo!("Implement assume_init for Lola shared-memory slot");
+    }
+}
+
+// Lola placeholder allocator.
+pub struct LolaMethodInArgAllocator;
+
+impl MethodInArgAllocator for LolaMethodInArgAllocator {
+    type MethodInArgPtr<T: CommData> = LolaMethodInArgPtr<T>;
```
bharatGoswami8, line 146: Optimization ?

bharatGoswami8: No

bharatGoswami8:
https://github.com/eclipse-score/communication/pull/818

One combine PR created for Field and Method, and Field is using Method interface also both share same macro for interface and type-state pattern so i feel we should review on one PR both design so that suggestion or feedback can be address in better way.

closing this PR.

Please review mentioned PR.

PR #798, opened 2026-07-28
Introduce Flaky-Test Detection

castler:


.github/actions/00_infrastructure/setup_qnx_environment/main.js
```diff
@@ -0,0 +1,49 @@
...
+
+const { execSync } = require("node:child_process");
+const { appendFileSync } = require("node:fs");
+
+function run(command) {
+  execSync(command, { stdio: "inherit" });
+}
+
+function main() {
+  const licenseDir = process.env["INPUT_LICENSE-DIR"] || "/opt/score_qnx/license";
```
limdor, line 22: Nitpick: We are defining the default here but also in the yml file.

.github/actions/00_infrastructure/setup_qnx_environment/post.js
```diff
@@ -0,0 +1,22 @@
...
+// terms of the Apache License Version 2.0 which is available at
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// SPDX-License-Identifier: Apache-2.0
+// *******************************************************************************
+
+const { execSync } = require("node:child_process");
+
+function main() {
+  const licenseDir = process.env.STATE_LICENSE_DIR || "/opt/score_qnx/license";
```
limdor, line 17: Nitpick: And as in https://github.com/eclipse-score/communication/pull/798/changes/1cb242a69d908d839193266c9598c36169b1dd18#r3680485623, we also define the default one more time

.github/workflows/_nightly_flaky_detection_runner.yml
```diff
@@ -0,0 +1,182 @@
...
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      actions: write
+    outputs:
+      flaky_count: ${{ steps.summary.outputs.flaky_count }}
+      failed_count: ${{ steps.summary.outputs.failed_count }}
+      test_exit_code: ${{ steps.run-tests.outputs.test_exit_code }}
+    steps:
+      - uses: actions/checkout@v6.0.2
```
limdor, line 78: Just note, nothing to do: This version is consistent with what we have in the others, but at some point we should consider updating it. There ia already a new patch version (6.0.3), a new minor version (6.0.1), and a new major version (7.0.1)

.github/workflows/_nightly_flaky_detection_runner.yml
```diff
@@ -0,0 +1,182 @@
...
+            --bep-json "${RUNNER_TEMP}/bep_${{ inputs.config-name }}.json" \
+            --raw-log "${RUNNER_TEMP}/bazel_${{ inputs.config-name }}.log" \
+            --output-dir "${REPORT_DIR}" \
+            --runs-per-test "${{ inputs.runs-per-test }}" \
+            --test-exit-code "${{ steps.run-tests.outputs.test_exit_code || 99 }}" \
+            --github-output "${GITHUB_OUTPUT}"
+
+      - name: Upload per-config flaky report
+        if: always()
+        uses: actions/upload-artifact@v4
```
limdor, line 173: Just note, nothing to do: This version is consistent with what we have in the others, but at some point we should consider updating it. There ia already a v5, v6, and v7

.github/workflows/_nightly_flaky_detection_runner.yml
```diff
@@ -0,0 +1,182 @@
...
+          bazel "${ARGS[@]}" 2>&1 | tee "${RAW_LOG}"
+          EXIT_CODE=${PIPESTATUS[0]}
+          set -e
+          echo "test_exit_code=${EXIT_CODE}" >> "${GITHUB_OUTPUT}"
+
+      - name: Collect flaky summary
+        id: summary
+        if: always()
+        run: |
+          REPORT_DIR="${RUNNER_TEMP}/nightly-flaky/${{ inputs.config-name }}"
```
limdor, line 161: This `nightly-flaky` appears in several places. It would be good if we could add a variable somewhere for the id and then reuse it.

.github/workflows/_nightly_flaky_detection_runner.yml
```diff
@@ -0,0 +1,182 @@
...
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: nightly-flaky-${{ inputs.config-name }}-${{ github.run_id }}
+          path: |
+            ${{ runner.temp }}/nightly-flaky/${{ inputs.config-name }}/summary.json
+            ${{ runner.temp }}/nightly-flaky/${{ inputs.config-name }}/summary.md
+            ${{ runner.temp }}/bazel_${{ inputs.config-name }}.log
+            ${{ runner.temp }}/bep_${{ inputs.config-name }}.json
+          if-no-files-found: ignore
```
limdor, line 181: Shouldn't this always contain a file? What would be the reason to not contain one?

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,176 @@
...
+# *******************************************************************************
+
+name: Nightly Flaky Test Detection
+
+on:
+  schedule:
+    - cron: "30 0 * * *"
+  workflow_dispatch:
+    inputs:
+      runs_per_test:
```
limdor, line 21: `unit_tests_runs_per_test` would be better to match the description. The other option is that if we want to use it for some other tests that are not unit tests neither integration tests, we then juss change the description.

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,192 @@
...
+
+jobs:
+  flaky-gcc15-unit:
+    uses: ./.github/workflows/_nightly_flaky_detection_runner.yml
+    with:
+      config-name: gcc15-unit
+      bazel-config: ""
+      target-pattern: //...
+      test-tag-filters: unit
+      runs-per-test: ${{ fromJSON(github.event.inputs.runs_per_test || '300') }}
```
limdor, line 47: Here we have also duplicated default. 300 is already before in another place

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,192 @@
...
+      UBUNTU_SNAPSHOT_MIRROR_URL: ${{ secrets.UBUNTU_SNAPSHOT_MIRROR_URL }}
+
+  flaky-gcc15-integration:
+    uses: ./.github/workflows/_nightly_flaky_detection_runner.yml
+    with:
+      config-name: gcc15-integration
+      bazel-config: ""
+      target-pattern: //...
+      test-tag-filters: integration-test
+      runs-per-test: ${{ fromJSON(github.event.inputs.integration_runs_per_test || '100') }}
```
limdor, line 61: Same as https://github.com/eclipse-score/communication/pull/798/changes#r3703498674

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,192 @@
...
+      UBUNTU_SNAPSHOT_MIRROR_URL: ${{ secrets.UBUNTU_SNAPSHOT_MIRROR_URL }}
+
+  flaky-asan-ubsan-lsan:
+    uses: ./.github/workflows/_nightly_flaky_detection_runner.yml
+    with:
+      config-name: asan-ubsan-lsan
+      bazel-config: asan_ubsan_lsan
+      target-pattern: //...
+      test-tag-filters: unit
+      runs-per-test: ${{ fromJSON(github.event.inputs.runs_per_test || '300') }}
```
limdor, line 75: Same as https://github.com/eclipse-score/communication/pull/798/changes#r3703498674
It is also in other places, but I will not add comments to every instance to not blow up the PR comments.

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,192 @@
...
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+      actions: read
+      issues: write
+    steps:
+      - uses: actions/checkout@v6.0.2
+
+      - name: Download per-config artifacts
+        continue-on-error: true
```
limdor, line 151: Why would we want to continue on error?

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,192 @@
...
+        uses: actions/download-artifact@v4
+        with:
+          pattern: nightly-flaky-*-${{ github.run_id }}
+          path: /tmp/nightly_flaky
+          merge-multiple: false
+
+      - name: Build consolidated report
+        id: merge
+        run: |
+          python3 quality/scripts/merge_flaky_reports.py \
```
limdor, line 161: Is it not possible to run the bazelized Python or we do not want to do that? The risk we have is that we end up using multiple Python versions and it is not clear which one is used for what.

.github/workflows/nightly_flaky_detection.yml
```diff
@@ -0,0 +1,192 @@
...
+      - name: Sync flaky test issues
+        if: always()
+        env:
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          if [[ ! -f /tmp/nightly_flaky/merged/summary.json ]]; then
+            echo "No merged summary found; skipping flaky issue sync."
+            exit 0
+          fi
+          python3 quality/scripts/sync_flaky_issues.py \
```
limdor, line 180: Same as https://github.com/eclipse-score/communication/pull/798/changes#r3703536802

quality/scripts/BUILD
```diff
@@ -0,0 +1,51 @@
...
+        "collect_flaky_tests.py",
+        "merge_flaky_reports.py",
+        "sync_flaky_issues.py",
+    ],
+)
+
+py_binary(
+    name = "collect_flaky_tests",
+    srcs = ["collect_flaky_tests.py"],
+    main = "collect_flaky_tests.py",
```
limdor, line 29:
```suggestion
```
Why do we need main?

quality/scripts/BUILD
```diff
@@ -0,0 +1,51 @@
...
+    deps = [":flaky_reports_lib"],
+)
+
+py_binary(
+    name = "merge_flaky_reports",
+    srcs = ["merge_flaky_reports.py"],
+    main = "merge_flaky_reports.py",
```
limdor, line 36:
```suggestion
```
Why do we need main?

quality/scripts/BUILD
```diff
@@ -0,0 +1,51 @@
...
+    deps = [":flaky_reports_lib"],
+)
+
+py_binary(
+    name = "sync_flaky_issues",
+    srcs = ["sync_flaky_issues.py"],
+    main = "sync_flaky_issues.py",
```
limdor, line 43:
```suggestion
```
Why do we need main?

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+# This program and the accompanying materials are made available under the
+# terms of the Apache License Version 2.0 which is available at
+# https://www.apache.org/licenses/LICENSE-2.0
+#
+# SPDX-License-Identifier: Apache-2.0
+# *******************************************************************************
+
+"""Collect flaky/failed test targets from Bazel BEP and raw logs."""
+
+from __future__ import annotations
```
limdor, line 17: Why do we need `__future__` annotations here? This will become an error at some point and it would be better if we do not need it and we can stick to standard Python without `__future__`. If we need it, then lets add a comment saying what Python version we need before we can get rid of it.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+import re
+from pathlib import Path
+from typing import Sequence
+
+
+FAILED_STATES = {"FAILED", "TIMEOUT", "INCOMPLETE", "REMOTE_FAILURE", "FAILED_TO_BUILD"}
+PRIORITY = {"PASSED": 1, "FAILED": 2, "FLAKY": 3}
+
+
+def _merge_status(current: str | None, incoming: str) -> str:
```
limdor, line 30: What is this function supposed to do? I see the merge being done just pure in priority but I would have expected that if current is `PASSED` and incoming is `FAILED` then the result is `FLAKY`

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+    if current is None:
+        return incoming
+    return incoming if PRIORITY.get(incoming, 0) > PRIORITY.get(current, 0) else current
+
+
+def _parse_bep(path: Path, target_stats: dict[str, dict]) -> None:
+    if not path.is_file():
```
limdor, line 37: Why would we silently want to return `None` instead of just raising an exception? It seems to me a bigger issue if someone ends up calling `_parse_bep` on something that it is not a file.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+        return
+
+    with path.open(encoding="utf-8") as stream:
+        for line in stream:
+            stripped = line.strip()
+            if not stripped:
+                continue
+            try:
+                event = json.loads(stripped)
+            except json.JSONDecodeError:
```
limdor, line 47: Why do we want to silent a json decoding error? If there is a real issue it might be harder to fine. I would expect that in general there are not decode errors.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+                continue
+
+            event_id = event.get("id", {})
```
limdor, line 50: I see several gets with a default, but most of them I would expect them to have a value unless something went really wrong. If something went really wrong and we just put a default I'm not sure if this will make it harder to find the issue.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+PRIORITY = {"PASSED": 1, "FAILED": 2, "FLAKY": 3}
+
+
+def _merge_status(current: str | None, incoming: str) -> str:
+    if current is None:
+        return incoming
+    return incoming if PRIORITY.get(incoming, 0) > PRIORITY.get(current, 0) else current
+
+
+def _parse_bep(path: Path, target_stats: dict[str, dict]) -> None:
```
limdor, line 36: The fact that `target_stats` is an input/output parameter, makes it a bit harder to understand. Considering that this function returns nothing, I would have prefered if the input could have been a read only dict and the output the modified dict. No need to change, but you can think about.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+            if not stripped:
+                continue
+            try:
+                event = json.loads(stripped)
+            except json.JSONDecodeError:
+                continue
+
+            event_id = event.get("id", {})
+            summary_id = event_id.get("testSummary", {})
+            label = summary_id.get("label")
```
limdor, line 52: `label` is obtained here, but only 3 lines later it checks if it is available or not an otherwise continues. Considering that `summary` and `overall` are local variables, it would be better to move the `if not label` just after getting the label.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+            summary = event.get("testSummary", {})
+            overall = str(summary.get("overallStatus", "")).upper()
+            if not label:
+                continue
+            failed_runs_value = summary.get("failedRunCount")
+            if failed_runs_value is None:
+                failed_runs_value = len(summary.get("failed", []))
```
limdor, line 59: Isn't this the same as calling `get` with a default parameter where the default is `len(summary.get("failed", []))`?

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+            total_runs_value = summary.get("totalRunCount")
+            if total_runs_value is None:
+                total_runs_value = summary.get("runCount", 0)
```
limdor, line 62: Same here, it looks like a get with default

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+            failed_runs = int(failed_runs_value)
+            total_runs = int(total_runs_value)
+            if total_runs <= 0:
+                total_runs = len(summary.get("failed", [])) + len(summary.get("passed", []))
+
+            current = target_stats.get(label, {"status": None, "failed_runs": 0, "total_runs": 0})
```
limdor, line 68: This `current` comes from the BEP? I do not get what this contains and why we have to merge it. Maybe we can discuss it offline. If you have a link that explain that content of the BEP regarding test execution it might help.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+            elif overall in FAILED_STATES:
+                current["status"] = _merge_status(current["status"], "FAILED")
+            elif overall == "PASSED":
+                current["status"] = _merge_status(current["status"], "PASSED")
+            current["failed_runs"] = max(int(current["failed_runs"]), failed_runs)
+            current["total_runs"] = max(int(current["total_runs"]), total_runs)
+            target_stats[label] = current
+
+
+def _parse_raw_log(path: Path, target_stats: dict[str, dict]) -> None:
```
limdor, line 80: Same comments like `_parse_bep` when it comes to input/output parameter and silently returning if the path is not a file.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    parser = argparse.ArgumentParser(description="Collect flaky tests from Bazel outputs.")
+    parser.add_argument("--config-name", required=True)
+    parser.add_argument("--bep-json", required=True)
+    parser.add_argument("--raw-log", required=True)
+    parser.add_argument("--output-dir", required=True)
+    parser.add_argument("--test-exit-code", required=True, type=int)
+    parser.add_argument("--runs-per-test", required=False, type=int, default=1000)
```
limdor, line 136: This is only for collecting and generating a report. Do we really want to provide a default here? This will not match most probably with the default that it was executed.

quality/scripts/collect_flaky_tests.py
```diff
@@ -0,0 +1,197 @@
...
+    if args.github_output:
+        output_path = Path(args.github_output)
+        with output_path.open("a", encoding="utf-8") as stream:
+            stream.write(f"flaky_count={summary['flaky_count']}\n")
+            stream.write(f"failed_count={summary['failed_count']}\n")
+    return 0
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
```
limdor, line 197: Why not exit with sys.exit? First time I see we raise an exception directly. Is there an advantage?

limdor (approved):
I started doing a review like if this would be tooling code that is quite critical, that is why at the beginning there are quite some comments.

However, this is something that it is not critical and we know that it brings us value. Part of the value comes from the ration effort spend / value provided. Considering that most of it has been done by AI + human driven (with the human checking the overall direction), the effort spend is not so high. We already detected flaky tests that we were not aware, this means that the value provided is already quite important.

In the moment that we start goingh through a strict review process for something that it is just nice to have, the effort / value decreases quite a lot (the author will have to spend much more time and also the reviwer will have to spend much more time).
Because of that, I decided to stop the review mid way and just focus on the overall direction.

I think the direction is good and that is something we want.

The risk of not doing a full review or leaving the code as it and not addressing some of the comments provided, is that if we have a bug in this tooling it might be hard to fix/find.
The main issues I find with the code is that it might be hard to understand for a human when going deeper (not from a high level point of view). If at some point we figure out that we have too many bugs in this code, we can throw it away and do it proper/better.

Until then, I think the direction is good enough and the value important enough that we should proceed with this even if the quality in some parts might be questionable.

PR #799, opened 2026-07-28
fix(message_passing): make cerr logger allocation-free and shutdown-safe

castler:
The console fallback logger returned by GetCerrLogger() serialised its per-item output with a function-local `static std::mutex cerr_mutex`. Such a static is registered for destruction via __cxa_atexit and is therefore destroyed during process shutdown (__cxa_finalize).

Background dispatch threads can still be running at that point. For example a ClientConnection retrying its connection logs "TryOpenClientConnection ..." from the DispatchThreadRunner thread. If that logging happens after cerr_mutex has already been destroyed, std::mutex::lock() calls pthread_mutex_lock() which returns EINVAL on the dead mutex. libc++ then throws std::system_error, and in the no-exception QNX build __cxa_allocate_exception aborts the process (SIGABRT).

Remove the shared mutex entirely: assemble the whole line into a fixed-size, stack-allocated buffer and emit it with a single std::cerr.write(). This:
  - performs no dynamic memory allocation (a bad_alloc could itself abort in a no-exception build), formatting integers/pointers via std::to_chars;
  - keeps concurrent log calls from interleaving character-by-character, since each line is written in one call;
  - has no shared state that must survive shutdown. std::cerr is guaranteed by [iostream.objects] to remain valid for the whole program execution, so there is no destruction race.

Content exceeding the buffer is truncated, which is acceptable for a debug console fallback logger.

### DIFF
--- a/score/message_passing/log/logging_callback_test.cpp
+++ b/score/message_passing/log/logging_callback_test.cpp
@@ -1,0 +1,308 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/message_passing/log/logging_callback.h"
+
+#include "score/message_passing/log/log.h"
+
+#include <gmock/gmock.h>
+#include <gtest/gtest.h>
+
+#include <unistd.h>
+
+#include <array>
+#include <atomic>
+#include <chrono>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+#include <cstdlib>
+#include <iostream>
+#include <set>
+#include <sstream>
+#include <streambuf>
+#include <string>
+#include <string_view>
+#include <thread>
+#include <vector>
+
+namespace score::message_passing
+{
+namespace
+{
+
+using namespace ::testing;
+
+// A std::streambuf that records every write into a fixed, preallocated buffer without ever allocating memory.
+// It also counts how many times the underlying stream buffer receives a write (xsputn/overflow). This lets a test
+// prove that a whole log line is emitted in a single atomic write - the property that replaced the removed
+// std::mutex and prevents concurrent log lines from interleaving character-by-character.
+class RecordingStreamBuf final : public std::streambuf
+{
+  public:
+    std::size_t WriteCalls() const noexcept
+    {
+        return write_calls_;
+    }
+    std::string_view View() const noexcept
+    {
+        return std::string_view{buffer_.data(), length_};
+    }
+    void Reset() noexcept
+    {
+        write_calls_ = 0U;
+        length_ = 0U;
+    }
+
+  protected:
+    std::streamsize xsputn(const char* s, std::streamsize n) override
+    {
+        ++write_calls_;
+        const std::size_t count = std::min(static_cast<std::size_t>(n), buffer_.size() - length_);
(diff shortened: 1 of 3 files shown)


score/message_passing/log/logging_callback_test.cpp
```diff
@@ -0,0 +1,377 @@
...
+// while main() calls std::exit(0), running static destruction concurrently. With the fix the logger holds no static
+// state that gets destroyed, so the child must terminate normally (exit code 0) instead of aborting.
+//
+// Note on portability: a literal SIGABRT reproduction of the *old* code is platform-dependent - glibc's
+// pthread_mutex_lock on a destroyed plain mutex succeeds instead of returning EINVAL, so on the host this test
+// passes even against the buggy implementation (verified empirically); only the QNX no-exception runtime aborts.
+// The deterministic, host-observable regression guard for the root cause is therefore
+// EmitsEachLineInASingleAtomicWrite above (the single-write invariant that removed the fatal static mutex); this
+// test additionally exercises the exact runtime scenario and asserts the invariant the fix guarantees.
+TEST(GetCerrLoggerShutdownDeathTest, LoggingThreadSurvivesStaticDestructionAtExit)
```
crimson11, line 290: Hm. I guess this test is not able to reveal the initial bug ....
Because in your test - even if `GetCerrLogger()` would still access the static `mutex` - there is NO static order issue! Because in your test the `std::thread worker` you create is NOT created in static context. This means: The destruction, which gets initiated with `std::exit` destroys the `worker` 1st BEFORE it would destroy the `static mutex`. Thus, you would never see the issue.

I.e. imho you need to create the `worker` in static context BEFORE you call `GetCerrLogger()`
In this case the `worker` gets created in static context 1st. Then the `nstatic mutex` in `GetCerrLogger()` will be created 2nd. Destruction order is then reversed ... and this would then show the issue.

crimson11 (changes requested):

score/message_passing/log/logging_callback_test.cpp
castler: Thanks — I dug into this carefully, and I believe the test does establish the ordering you're after; the worker is not torn down first. The key is the exact meaning of the standard's rule.

The rule "the last destructor for thread-local objects is sequenced-before the first destructor for a static object" is scoped to the **current (exiting) thread's `thread_local` objects only**. The same `std::exit` page also states:

> "No destructors are called for objects with **automatic storage duration** or for **thread-local objects associated with threads other than the current one**."

Mapping the three entities in the test:

| Entity | Storage duration | What `std::exit` does |
|---|---|---|
| `std::thread worker` (the handle) | **automatic** | nothing — no dtor is run for automatic objects (no stack unwinding) |
| the worker's **OS thread** | (a thread, not an object) | nothing — not joined, not stopped; keeps running |
| `static std::mutex` in the logger | **static** | destroyed in the static-destruction phase |

So:
- `std::exit` does **not** unwind the stack ⇒ the `worker` handle's destructor never runs.
- `std::exit` does **not** join/stop other threads ⇒ the detached OS thread runs concurrently through static destruction.
- The mutex is a function-local **`static`** (shared across threads), **not** a `thread_local`, so the thread-local-before-static ordering rule doesn't apply to it. It's destroyed while the worker is still logging.

That's precisely the use-after-static-destruction the standard warns about (the reason threads must be joined before `exit`) — i.e. the bug under test. Making `worker` a `static` doesn't add the ordering (it's already guaranteed) and would introduce a spurious `std::terminate()` from a joinable `static std::thread`'s own destructor.

I verified this empirically with an instrumented stand-in (a `static` Guard whose dtor flips a flag; a detached worker recording whether it runs afterward): every run, ~1M logger calls occurred **after** the static's destructor ran (`used_after_destroy=1`). The only reason it doesn't `abort()` on the host against the buggy code is glibc's `pthread_mutex_lock` returning success on a destroyed-but-mapped plain mutex (EINVAL only on QNX), which the comment already documents.

I've pushed a commit adding this reasoning as a comment in the test so it's clear to future readers.

crimson11 (approved):

PR #803, opened 2026-07-29
fix(coverage): disable clang default hardening feature for coverage builds

AAmbuj:
Disable clang default hardening feature for coverage builds


LittleHuba (changes requested):
Unsetting all our default flags is not okay. Please only unset the specific flags that cause problems.

AAmbuj:
Done using this PR: https://github.com/eclipse-score/communication/pull/808

PR #808, opened 2026-07-29
## Fix: Separate fortify_source feature and resolve coverage build failures

AAmbuj:
Extract -D_FORTIFY_SOURCE=2 into its own cc_feature to enable selective disabling during coverage builds without affecting other hardening flags.

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -297,6 +297,7 @@ llvm.toolchain(
     cxx_standard = {"": "c++17"},
     extra_enabled_features = [
         "//quality/compiler_warnings/clang:default_flags",
+        "//quality/compiler_warnings:fortify_source",
         "//quality/compiler_warnings/clang:minimal_warnings",
     ],
     extra_known_features = [
--- a/quality/compiler_warnings/BUILD
+++ b/quality/compiler_warnings/BUILD
@@ -27,9 +27,6 @@ cc_args(
         "-fPIC",
         "-fstack-protector-strong",
         "-fdiagnostics-color=always",
-        # Fortify libc calls (buffer-overflow checks) used by message (de)serialization.
-        # NOTE: effective only at -O1+; requires the non-executable stack link hardening below.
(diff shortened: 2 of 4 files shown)


quality/compiler_warnings/BUILD
```diff
@@ -46,6 +43,26 @@ cc_args(
     visibility = ["//quality/compiler_warnings:__subpackages__"],
 )
 
+cc_args(
+    name = "fortify_source_args",
+    actions = [
+        "@rules_cc//cc/toolchains/actions:compile_actions",
+    ],
+    args = [
+        "-D_FORTIFY_SOURCE=2",
```
LittleHuba, line 53: Please prepend with `-U_FORTIFY_SOURCE`. Some toolchains define this already and throw an error if it is defined twice without undefining inbetween.

quality/compiler_warnings/BUILD
```diff
@@ -61,7 +78,6 @@ cc_args(
         "-Wl,-as-needed",
         "-ldl",
         "-Wl,--pop-state",
-        "-fuse-ld=gold",
```
LittleHuba, line 64: This should be moved to gcc only. Keeping it there is legit.

MODULE.bazel
```diff
@@ -298,6 +298,7 @@ llvm.toolchain(
     extra_enabled_features = [
         "//quality/compiler_warnings/clang:default_flags",
         "//quality/compiler_warnings/clang:minimal_warnings",
+        "//quality/compiler_warnings:fortify_source",
```
LittleHuba, line 300: Order matters here. Please put it after default_flags.

AAmbuj: Updated

quality/compiler_warnings/BUILD
AAmbuj: Updated

LittleHuba (approved):

AAmbuj:
test-qnx

PR #810, opened 2026-07-29
mw/com: Add single source of truth for getter / setter signatures

bemerybmw:
Also use the correct signature in the binding factories to address the bug ticket: SWP-275041.

### DIFF
--- a/score/mw/com/impl/field_getter_setter_signatures.h
+++ b/score/mw/com/impl/field_getter_setter_signatures.h
@@ -1,0 +1,33 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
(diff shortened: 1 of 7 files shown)


sahithi-nukala (approved):

score/mw/com/impl/field_getter_setter_signatures.h
```diff
@@ -0,0 +1,33 @@
...
+
+#include "score/result/result.h"
+
+namespace score::mw::com::impl
+{
+
+/// \brief Provides the method signatures for field getter and setter methods.
+///
+/// Since the setter / getter functions can fail within the middleware code (e.g. when trying to update the field
+/// value fails), we need to return a result from the setter and getter.
```
KrishaDeshkool, line 24: " we need to return a result from the setter and getter."
do we return errors from getters also?
does not look like it from the API.

LittleHuba (dismissed):

score/mw/com/impl/field_getter_setter_signatures.h
bemerybmw: Yes, but it's added in the getter pr

KrishaDeshkool:
This can be re-approved then.

LittleHuba (approved):

PR #813, opened 2026-07-30
mw/com/impl: Internally use memory::DataTypeSizeInfo

crimson11:
The type DataTypeMetaInfo, which we use on our
public interface is too weak! It doesn't assure
invariants on size/alignment, which C++ demands:
size has to be always a power of two and be an
integer multiple of tzhe alignment.

This commit switches to internal use of memory::DataTypeSizeInfo which forces this. On the public interface we keep the weak DataTypeMetaInfo for now to avoid a breaking change! In the public layer we transform DataTypeMetaInfo to DataTypeSizeInfo. In case this isn't possible because DataTypeMetaInfo is invalid, we return an error in the public API.

### DIFF
--- a/score/mw/com/gateway/gateway_application/BUILD
+++ b/score/mw/com/gateway/gateway_application/BUILD
@@ -112,6 +112,7 @@ cc_unit_test(
         "//score/mw/com/impl/test:dummy_instance_identifier_builder",
         "//score/mw/com/impl/test:runtime_mock_guard",
         "@googletest//:gtest",
+        "@score_baselibs//score/memory:data_type_size_info",
         "@score_baselibs//score/result",
     ],
 )
--- a/score/mw/com/impl/bindings/lola/generic_skeleton_event.h
+++ b/score/mw/com/impl/bindings/lola/generic_skeleton_event.h
@@ -18,9 +18,10 @@
 #include "score/mw/com/impl/bindings/lola/event_data_storage.h"
 #include "score/mw/com/impl/bindings/lola/skeleton_event_common.h"
 #include "score/mw/com/impl/bindings/lola/skeleton_event_properties.h"
-#include "score/mw/com/impl/data_type_meta_info.h"
 #include "score/mw/com/impl/generic_skeleton_event_binding.h"
 #include "score/mw/com/impl/sample_allocatee_guard.h"
+
+#include "score/memory/data_type_size_info.h"
 
 #include <optional>
 
@@ -38,7 +39,7 @@ class GenericSkeletonEvent : public GenericSkeletonEventBinding
                          const std::string_view event_name,
                          const SkeletonEventProperties& event_properties,
                          const ElementFqId& element_fq_id,
-                         const DataTypeMetaInfo& size_info,
+                         const memory::DataTypeSizeInfo& size_info,
                          impl::tracing::SkeletonEventTracingData tracing_data = {});
 
     Result<void> Send(score::mw::com::impl::SampleAllocateePtr<void> sample) noexcept override;
@@ -59,7 +60,7 @@ class GenericSkeletonEvent : public GenericSkeletonEventBinding
 
     std::size_t GetMaxSize() const noexcept override
     {
-        return size_info_.size;
+        return size_info_.Size();
     }
 
     /// \brief Set callback, to get notified, when either the 1st event-notification has been registered or the last
@@ -80,7 +81,7 @@ class GenericSkeletonEvent : public GenericSkeletonEventBinding
     }
 
   private:
-    DataTypeMetaInfo size_info_;
+    memory::DataTypeSizeInfo size_info_;
     std::uint8_t* event_data_storage_;
     SkeletonEventCommon<void> skeleton_event_common_;
 };
--- a/score/mw/com/impl/data_type_meta_info.h
+++ b/score/mw/com/impl/data_type_meta_info.h
@@ -13,6 +13,9 @@
 #ifndef SCORE_MW_COM_IMPL_BINDINGS_LOLA_DATA_TYPE_META_INFO_H
 #define SCORE_MW_COM_IMPL_BINDINGS_LOLA_DATA_TYPE_META_INFO_H
 
+#include "score/memory/data_type_size_info.h"
+#include "score/result/result.h"
+
 #include <cstddef>
 #include <cstdint>
 
@@ -27,6 +30,17 @@ struct DataTypeMetaInfo
     std::size_t size;
     std::size_t alignment;
 };
+
+/// \brief Validates the given public DataTypeMetaInfo and converts it into the internal
+/// score::memory::DataTypeSizeInfo.
+///
+/// DataTypeSizeInfo enforces (via assertions) that the alignment is a non-zero power of two and that the size is an
+/// integer multiple of the alignment. This function checks these invariants up-front so that invalid meta-info handed
+/// over via the public API results in an error Result instead of a contract violation/abort.
+///
+/// \param meta_info The (public) meta-info to validate and convert.
+/// \return The converted DataTypeSizeInfo on success, or a ComErrc error if the invariants are violated.
+score::Result<memory::DataTypeSizeInfo> MakeDataTypeSizeInfo(const DataTypeMetaInfo& meta_info);
 
 }  // namespace score::mw::com::impl
 
--- a/score/mw/com/impl/data_type_meta_info_test.cpp
+++ b/score/mw/com/impl/data_type_meta_info_test.cpp
@@ -1,0 +1,92 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/impl/data_type_meta_info.h"
+
+#include "score/mw/com/impl/com_error.h"
+
+#include <gtest/gtest.h>
+
+namespace score::mw::com::impl
+{
+namespace
+{
+
+TEST(MakeDataTypeSizeInfoTest, ConvertsSizeMultipleOfAlignment)
+{
+    // Given a valid meta-info (size multiple of a power-of-two alignment)
+    const DataTypeMetaInfo meta_info{16U, 8U};
+
+    // When converting it
+    const auto result = MakeDataTypeSizeInfo(meta_info);
+
+    // Then the conversion succeeds and preserves size and alignment
+    ASSERT_TRUE(result.has_value());
+    EXPECT_EQ(result.value().Size(), 16U);
+    EXPECT_EQ(result.value().Alignment(), 8U);
+}
+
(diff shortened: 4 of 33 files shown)


LittleHuba (changes requested):
I'd like to discuss different approaches to this topic.
This is not a hard no, I just want to brainstorm for ideas.

crimson11:
> I'd like to discuss different approaches to this topic. This is not a hard no, I just want to brainstorm for ideas.

Sure! I'm curious for your proposals! Low level I want to "extend" a type used in the public API ... and I have no clue how to do this minimal invasive - opposed to creating a NEW type and deprecate all APIs with the old type and introduce new/overloads with the new Type! In this case that would even cascade as `DataTypeMetaInfo` is enclosed in other public types ... so this would be a crazy escalation!

score/mw/com/impl/data_type_meta_info_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include <gtest/gtest.h>
+
+namespace score::mw::com::impl
+{
+namespace
+{
+
+TEST(MakeDataTypeSizeInfoTest, ConvertsValidMetaInfo)
+{
+    RecordProperty("Description",
```
bemerybmw: We shouldn't have these properties if we're not linking a requirement imo. They can easily get out of sync with the code and the tests should be clear from the name and given/when/then documentation.

score/mw/com/impl/data_type_meta_info_test.cpp
```diff
@@ -0,0 +1,115 @@
...
+#include "score/mw/com/impl/com_error.h"
+
+#include <gtest/gtest.h>
+
+namespace score::mw::com::impl
+{
+namespace
+{
+
+TEST(MakeDataTypeSizeInfoTest, ConvertsValidMetaInfo)
```
bemerybmw: ConvertsSizeMultipleOfAlignment

score/mw/com/impl/generic_skeleton_test.cpp
```diff
@@ -205,6 +206,64 @@ TEST_F(GenericSkeletonTest, CreateWithEventsInitializesEventBindings)
     EXPECT_NE(events.find(event_name), events.cend());
 }
 
+TEST_F(GenericSkeletonTest, CreateWithInvalidDataTypeMetaInfoAlignmentFails)
+{
+    RecordProperty("Description",
```
bemerybmw: Same thing, remove recordproperty

score/mw/com/impl/generic_skeleton_test.cpp
```diff
@@ -205,6 +206,64 @@ TEST_F(GenericSkeletonTest, CreateWithEventsInitializesEventBindings)
...
+    const std::string event_name = "test_event";
+    const DataTypeMetaInfo meta_info{16, 6};
+
+    std::vector<EventInfo> event_storage;
+    event_storage.push_back({event_name, meta_info});
+
+    GenericSkeletonServiceElementInfo params;
+    params.events = event_storage;
+
+    // The event factory must never be called for invalid meta-info
```
bemerybmw:
```suggestion
    // Expecting that the event factory is never be called for invalid meta-info
```

score/mw/com/impl/generic_skeleton.cpp
```diff
@@ -100,8 +101,21 @@ Result<GenericSkeleton> GenericSkeleton::Create(const InstanceIdentifier& identi
...
+        // alignment not a power of two, or size not a multiple of alignment) results in an error instead of a later
+        // contract violation. Unfortunately our public interface DataTypeMetaInfo does not enforce these invariants,
+        // so we have to check them here. @ToDo: At some point we should already enforce this on DataTypeMetaInfo,
+        // but this would require a break in the public API!
+        auto data_type_size_info_result = MakeDataTypeSizeInfo(info.data_type_meta_info);
+        if (!data_type_size_info_result.has_value())
+        {
+            score::mw::log::LogError("GenericSkeleton")
+                << "Invalid data type meta-info provided for event: " << info.name;
+            return MakeUnexpected(ComErrc::kInvalidConfiguration);
```
bemerybmw, line 114: Should this really be an invalidConfiguration error? I associate kInvalidConfiguration with the mw_com_config. But this is an error coming from C++.  And do we want to return an error here or simply terminate? Because they can only be providing a data_type_meta_info which is invalid if they are doing something weird and not simply creating with sizeof / alignof.

score/mw/com/impl/generic_skeleton_test.cpp
bemerybmw: same for the other test

score/mw/com/impl/generic_skeleton_test.cpp
bemerybmw: same for the other test

score/mw/com/impl/data_type_meta_info.h
```diff
@@ -28,6 +31,17 @@ struct DataTypeMetaInfo
...
+/// \brief Validates the given public DataTypeMetaInfo and converts it into the internal
+/// score::memory::DataTypeSizeInfo.
+///
+/// DataTypeSizeInfo enforces (via assertions) that the alignment is a non-zero power of two and that the size is an
+/// integer multiple of the alignment. This function checks these invariants up-front so that invalid meta-info handed
+/// over via the public API results in an error Result instead of a contract violation/abort.
+///
+/// \param meta_info The (public) meta-info to validate and convert.
+/// \return The converted DataTypeSizeInfo on success, or a ComErrc error if the invariants are violated.
+score::Result<score::memory::DataTypeSizeInfo> MakeDataTypeSizeInfo(const DataTypeMetaInfo& meta_info) noexcept;
```
bemerybmw: Should not be noexcept

score/mw/com/impl/plumbing/i_generic_skeleton_event_binding_factory.h
bemerybmw, line 1: Why is this file here and not in plumbing?

score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
```diff
@@ -13,14 +13,14 @@
 #ifndef SCORE_MW_COM_IMPL_PLUMBING_SKELETON_SERVICE_ELEMENT_BINDING_FACTORY_IMPL_H
 #define SCORE_MW_COM_IMPL_PLUMBING_SKELETON_SERVICE_ELEMENT_BINDING_FACTORY_IMPL_H
 
+#include "score/memory/data_type_size_info.h"
```
bemerybmw: The first block of includes are includes from the same package/ component, so this should be in the second block i.e. with logging

score/mw/com/impl/plumbing/generic_skeleton_event_binding_factory_mock.h
```diff
@@ -13,7 +13,7 @@
 #ifndef SCORE_MW_COM_IMPL_PLUMBING_GENERIC_SKELETON_EVENT_BINDING_FACTORY_MOCK_H
 #define SCORE_MW_COM_IMPL_PLUMBING_GENERIC_SKELETON_EVENT_BINDING_FACTORY_MOCK_H
 
-#include "score/mw/com/impl/data_type_meta_info.h"
+#include "score/memory/data_type_size_info.h"
```
bemerybmw, line 18: The first block of includes are includes from the same package/ component, so this should be in a separate second block


```suggestion
#include "score/mw/com/impl/i_generic_skeleton_event_binding_factory.h"

#include "score/memory/data_type_size_info.h"
```

score/mw/com/impl/plumbing/generic_skeleton_event_binding_factory_impl.h
```diff
@@ -13,7 +13,7 @@
 #ifndef SCORE_MW_COM_IMPL_PLUMBING_GENERIC_SKELETON_EVENT_BINDING_FACTORY_IMPL_H
 #define SCORE_MW_COM_IMPL_PLUMBING_GENERIC_SKELETON_EVENT_BINDING_FACTORY_IMPL_H
 
-#include "score/mw/com/impl/data_type_meta_info.h"
+#include "score/memory/data_type_size_info.h"
 #include "score/mw/com/impl/i_generic_skeleton_event_binding_factory.h"
```
bemerybmw:
```suggestion
#include "score/mw/com/impl/i_generic_skeleton_event_binding_factory.h"

#include "score/memory/data_type_size_info.h"
```

score/mw/com/impl/bindings/lola/generic_skeleton_event.h
```diff
@@ -14,11 +14,11 @@
 #ifndef SCORE_MW_COM_IMPL_BINDINGS_LOLA_GENERIC_SKELETON_EVENT_H_
 #define SCORE_MW_COM_IMPL_BINDINGS_LOLA_GENERIC_SKELETON_EVENT_H_
 
+#include "score/memory/data_type_size_info.h"
```
bemerybmw:
```suggestion
#include "score/mw/com/impl/bindings/lola/element_fq_id.h"
#include "score/mw/com/impl/bindings/lola/event_data_storage.h"
#include "score/mw/com/impl/bindings/lola/skeleton_event_common.h"
#include "score/mw/com/impl/bindings/lola/skeleton_event_properties.h"
#include "score/mw/com/impl/generic_skeleton_event_binding.h"
#include "score/mw/com/impl/sample_allocatee_guard.h"

#include "score/memory/data_type_size_info.h"
```

bemerybmw (changes requested):

score/mw/com/gateway/gateway_application/BUILD
```diff
@@ -112,6 +112,7 @@ cc_unit_test(
         "//score/mw/com/impl/test:dummy_instance_identifier_builder",
         "//score/mw/com/impl/test:runtime_mock_guard",
         "@googletest//:gtest",
+        "@score_baselibs//score/memory:data_type_size_info",
```
bemerybmw, line 115: Should now be referencing target in the com repo?

crimson11: nope. //score/memory hasn't been moved.

score/mw/com/impl/generic_skeleton.cpp
crimson11: `kInvalidConfiguration` was the only available/existing error-code, which looked not completely "off" :)
... and the `DataTypeMetaInfo` can be imho sen as the configuration of a GenericSkeleton ...
I could introduce a new error code, if you really want:
`kInvalidDataTypeSize` ?

In this case, I would "prefer" not to terminate ... (it is always a philosophical question): This is not a somewhat internal broken state, where we can't recover from. It is a clear call/argument violation. And if you look at how `GenericSkeleton` is being used, it makes sense NOT to terminate! The typical users (like these data-collectors) get some info from the backend! They are not doing sizeof/alignof! If they could do sizeof/alignof, thzen there would be no need for a GenericSkeleton ... so imho terminate is too harsh here!

score/mw/com/impl/generic_skeleton_test.cpp
crimson11: OK!

score/mw/com/impl/plumbing/i_generic_skeleton_event_binding_factory.h
crimson11: good question! Will fix in a separate commit.

bemerybmw (approved):

LittleHuba (approved):
Discussed with @crimson11 offline. Agreed to proceed.

PR #814, opened 2026-07-30
mw/com: Pass the GetNewSamplesCallback to WaitForSamples

bemerybmw:
This change allows the ProxyEventReceiver to be used with different callbacks called each time WaitForSamples is called.


bemerybmw:
To be merged as part of: https://github.com/eclipse-score/communication/pull/816

PR #816, opened 2026-07-31
mw/com: General fixes for field integration tests

bemerybmw:


score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -178,7 +179,7 @@ void run_set_and_notifier_consumer(const score::cpp::stop_token& stop_token)
 
     // Step 7. Verify transformed value received via field notification
     std::cout << "\nConsumer: Step 7 - Verify transformed value via field notification" << std::endl;
-    if (!field_receiver.WaitForSamples(stop_token, 1U))
+    if (!field_receiver.WaitForSamples(stop_token, 1U, value_callback))
```
Tejveerpratap2803: There is inconsistency intially you moved the value and here you are not moving it. I will suggest you remove move from everywhere as in WaitForSamples() we are using reference of get_new_samples_callback.

score/mw/com/test/fields/set_and_notifier/consumer.cpp
```diff
@@ -100,7 +88,19 @@ void run_notifier_consumer(const score::cpp::stop_token& stop_token)
 
     // Step 6. Wait for all expected samples
     std::cout << "\nConsumer: Step 6 - Wait for all expected samples" << std::endl;
-    if (!field_receiver.WaitForSamples(stop_token, kMaxNumSamples))
+    std::size_t sample_index{0U};
```
Tejveerpratap2803, line 92: Suggestion: run_notifier_consumer and run_set_and_notifier_consumer both define an almost identical value_callback. can we create a helper function for that to avoid duplication?

score/mw/com/test/fields/set_and_notifier/common_resources.cpp
```diff
@@ -0,0 +1,70 @@
...
+{
+    if (mode == "notifier")
+    {
+        return TestMode::kNotifier;
+    }
+    if (mode == "set_and_notifier")
+    {
+        return TestMode::kSetAndNotifier;
+    }
+    FailTest("Provider: unsupported --mode value: ", mode);
```
Tejveerpratap2803: Provider is hardcoded in multiple places in this file. As this function is shared by both Consumer and Provider, it should be made generic instead of being provider-specific.

score/mw/com/test/fields/set_and_notifier/config/mw_com_config.json
Tejveerpratap2803, line 1: I think we should consider renaming the file from set_notifier to set_and_get_and_notifier, since the serviceTypeName is set_and_get_and_notifier. It seems like this file is intended to handle  set, get and notify operations. Is that the direction we're planning to take?

score/mw/com/test/fields/set_and_notifier/main_consumer_and_provider.cpp
```diff
@@ -0,0 +1,44 @@
...
+        std::cerr << "Unable to set signal handler for SIGINT and/or SIGTERM, cautiously continuing\n";
+    }
+
+    auto provider_future = std::async(score::mw::com::test::run_provider, stop_source.get_token(), config.mode);
+    auto consumer_future = std::async(score::mw::com::test::run_consumer, stop_source.get_token(), config.mode);
+
+    provider_future.get();
+    consumer_future.get();
+
+    std::cout << "SetAndNotifierSameProcessTest: Provider and Consumer completed successfully" << std::endl;
```
Tejveerpratap2803: IMO we cannot hard code it as SetAndNotifierSameProcessTest, it's been utilized by both Notifier and SetAndNotifier.

score/mw/com/test/fields/set_and_notifier/consumer.cpp
bemerybmw: Yep, it's a good point. I have already minimised some duplicate code in the follow up pr which adds the getter integration tests. Once we get to that point, you can have a look and tell me if you think it's enough or we should try to extract more.

score/mw/com/test/fields/set_and_notifier/consumer.cpp
bemerybmw: I changed it to move to be in line with the other usages

score/mw/com/test/fields/set_and_notifier/config/mw_com_config.json
bemerybmw: You're exactly right :) https://github.com/eclipse-score/communication/pull/821/changes/0838ee3d39a5a48bd206576fa744c287eaf8549e

score/mw/com/test/fields/set_and_notifier/consumer.cpp
Tejveerpratap2803: I noticed a few other instances of duplication too. We can clean those up in the next PR after you add the getter integration tests.

Tejveerpratap2803 (approved):

crimson11 (approved):

PR #834, opened 2026-08-04
ci/quality: Add QNX CodeQL analysis with platform-specific delta reporting

castler:
Extends the CodeQL nightly workflow with an optional QNX build pass controlled by the new `run-qnx` input flag. When enabled, the workflow:

- Runs `codeql_lint` for both Linux and `--config=qnx` targets.
- Merges the two result sets via the new `merge_codeql_findings` script, producing a deduplicated Linux∪QNX union and a QNX-only delta.
- Publishes union findings under the `codeql-nightly` Code Scanning category and QNX-specific findings under `codeql-nightly-qnx-delta`.
- Uploads separate CSV and MISRA compliance report artifacts per platform.

The quality dashboard (`generate_dashboard.py`) gains awareness of the QNX-only delta: it tags each finding with `qnx_only` and displays per-severity QNX-only counts in both the rendered HTML and the GitHub step summary. The `load_codeql_csv` / `load_qnx_delta_keys` helpers mirror the deduplication key used by `merge_codeql_findings.py` so that QNX-delta rows are matched correctly.


.github/workflows/_codeql.yml
```diff
@@ -49,38 +69,112 @@ jobs:
...
+          retention-days: 30
+          if-no-files-found: warn
+
+      - name: Upload MISRA compliance reports (QNX)
+        if: inputs.fetch-only != 'true' && always() && inputs.run-qnx == 'true'
+        uses: actions/upload-artifact@v4
+        with:
+          name: codeql-misra-reports-qnx
+          path: /tmp/codeql-results/qnx/analysis_reports/
           retention-days: 30
```
castler: We upload this but never use it

quality/dashboard/generate_dashboard.py
```diff
@@ -224,11 +228,71 @@ def load_linter_findings(path: pathlib.Path) -> dict | None:
...
+        severity = "warning"
+    else:
+        # Treat everything else as recommendation (including empty or unknown).
+        severity = "recommendation"
+    return {
+        "severity": severity,
+        "name": row.get("name") or row.get("Name") or row.get("rule_id") or row.get("Rule") or "",
+        "message": row.get("message") or row.get("Message") or row.get("description") or row.get("Description") or "",
+        "path": row.get("path") or row.get("Path") or row.get("file") or row.get("File") or "",
+        "line": row.get("start:line") or row.get("Line") or row.get("line_number") or "",
```
castler: Why all these conditions, should the format not be specific and we should fail if the format is not as expected?

quality/static_analysis/BUILD
```diff
@@ -35,3 +35,11 @@ filegroup(
     name = "config",
     srcs = ["config.yaml"],
 )
+
+py_binary(
+    name = "merge_codeql_findings",
+    srcs = ["merge_codeql_findings.py"],
+    main = "merge_codeql_findings.py",
+    target_compatible_with = ["@platforms//os:linux"],
+    visibility = ["//visibility:public"],
```
castler: This should not be public

quality/static_analysis/BUILD
```diff
@@ -35,3 +35,11 @@ filegroup(
     name = "config",
     srcs = ["config.yaml"],
 )
+
+py_binary(
+    name = "merge_codeql_findings",
+    srcs = ["merge_codeql_findings.py"],
+    main = "merge_codeql_findings.py",
```
castler: main is not required.

quality/static_analysis/merge_codeql_findings.py
```diff
@@ -0,0 +1,243 @@
...
+    if isinstance(msg, dict):
+        message = msg.get("text", "")
+    uri = start_line = start_col = ""
+    locations = result.get("locations") or []
+    if locations:
+        phys = (locations[0] or {}).get("physicalLocation") or {}
+        uri = ((phys.get("artifactLocation") or {}).get("uri") or "")
+        region = phys.get("region") or {}
+        start_line = str(region.get("startLine", ""))
+        start_col = str(region.get("startColumn", ""))
```
castler, line 64: Should we not fail if the schema is not as expected, instead of giving a lot of default values?

.github/workflows/_codeql.yml
```diff
@@ -24,9 +24,23 @@ on:
         description: "When true, runs --nobuild to fetch dependencies only."
         type: string
         default: "false"
+      run-qnx:
+        description: "When true, additionally analyzes the --config=qnx build and
+          publishes the deduplicated union plus a QNX-only delta. Requires the
+          SCORE_QNX_* secrets. Defaults to false so cache-warming callers keep the
+          Linux-only behavior."
```
LittleHuba: This workflow is not part of checks or gate. And codeql must run uncached to provide correct results.
We may now decide to still do repository-caching. But then we should do this for both configs, since the caches could differ based on config.
If we decide to forego repository-caching as well, since this is a nightly job that can take 2mins longer to build up the repositories, then we should not talk about cache warming here at all.

.github/workflows/_codeql.yml
```diff
@@ -49,38 +69,103 @@ jobs:
...
+
+          bazel run //quality/static_analysis:merge_codeql_findings -- \
+            --base-prefix /tmp/codeql-results/linux/codeql-nightly \
+            --overlay-prefix /tmp/codeql-results/qnx/codeql-nightly-qnx \
+            --out-prefix /tmp/codeql-results/codeql-merged
+
+      # Choose the canonical published outputs: the deduplicated union when the
+      # QNX analysis ran, otherwise the plain Linux results. Downstream steps and
+      # nightly_quality.yml always consume codeql-nightly.{csv,sarif}, keeping
+      # cache-warming (Linux-only) callers unchanged.
```
LittleHuba:
```suggestion
      # Choose the canonical published outputs: the deduplicated union when the
      # QNX analysis ran, otherwise the plain Linux results.
```

.github/workflows/_codeql.yml
```diff
@@ -49,38 +69,103 @@ jobs:
...
-      - name: Upload MISRA compliance reports
+      - name: Upload MISRA compliance reports (Linux)
         if: inputs.fetch-only != 'true' && always()
         uses: actions/upload-artifact@v4
         with:
           name: codeql-misra-reports
-          path: /tmp/codeql-results/analysis_reports/
+          path: /tmp/codeql-results/linux/analysis_reports/
           retention-days: 30
           if-no-files-found: warn
```
LittleHuba, line 144: Why not upload the QNX ones as well?

.github/skills/codeql-analysis/SKILL.md
```diff
@@ -153,6 +156,53 @@ join -1 2 -2 2 -a1 -a2 -e0 -o '0,1.1,2.1' \
...
+  --base-prefix    /tmp/codeql-results/linux/codeql-nightly \
+  --overlay-prefix /tmp/codeql-results/qnx/codeql-nightly-qnx \
+  --out-prefix     /tmp/codeql-results/codeql-merged
+```
+Writes `codeql-merged-union.{csv,sarif}` (base ∪ overlay) and `codeql-merged-qnx-delta.{csv,sarif}`
+(overlay \ base). Dedup key: `(ruleId, file, startLine, startColumn, message)`. Invariant to
+sanity-check: `union == linux + qnx_unique` and `delta == qnx \ linux`. CI uploads the union under
+Code Scanning category `codeql-nightly` and the delta under `codeql-nightly-qnx-delta`; the union
+CSV feeds the dashboard KPI, and the QNX-only delta CSV is passed alongside it
+(`--codeql-qnx-delta-csv`) so the dashboard flags which findings are QNX-specific.
```
LittleHuba: In S-CORE Linux is also a target platform. So should we do the same as for QNX and provide the only-Linux-findings?

quality/static_analysis/merge_codeql_findings.py
```diff
@@ -0,0 +1,252 @@
+# *******************************************************************************
```
LittleHuba, line 1: I'm honestly leaning more and more to make SARIF the sole output of CodeQL.
SARIF is a very simple format. There are a lot of tools out there that can do the conversion from SARIF to CSV and I trust that there is at least one that does this reliably.

Removing the CSV path would allow us to do diffing/merging/filtering based on established tools like https://github.com/microsoft/sarif-tools

Doing this by hand is at least equally error-prone as using an existing tool. And it raises the maintenance cost on our side.

Note: did not review this file further

quality/dashboard/generate_dashboard.py
```diff
@@ -224,52 +228,114 @@ def load_linter_findings(path: pathlib.Path) -> dict | None:
...
+                    f"(expected {list(_CODEQL_FIELDS)!r})"
+                )
+            for row in reader:
+                keys.add(_codeql_dedup_key(_codeql_row_fields(row)))
+    except (OSError, csv.Error, ValueError) as e:
+        raise ValueError(f"Error parsing CodeQL QNX-delta CSV {path}: {e}") from e
+    return keys
+
+
+def load_codeql_csv(path: pathlib.Path, qnx_delta_keys: set | None = None) -> dict | None:
```
LittleHuba: Use https://pypi.org/project/simple-sarif/ to reduce the manual parsing efforts here.

.github/workflows/_codeql.yml
```diff
@@ -49,38 +75,85 @@ jobs:
...
         run: |
           set -euo pipefail
 
           bazel run //quality/static_analysis:codeql_lint -- \
-            --output-dir /tmp/codeql-results \
+            --output-dir /tmp/codeql-results/linux \
             --output-prefix codeql-nightly \
             --target //score/message_passing //score/mw/com
 
+      - name: Run CodeQL analysis (QNX, --config=qnx)
```
LittleHuba, line 133: Making this a step will cause the two runs to be sequential.
Since they don't acutally share cache, I don't think this has any benefits.
The drawback is clearly extended runtimes.

Better would be to split this in separate jobs. One for linux, one for qnx and finally one waiting an all OS-specific jobs that merges their results and reports to the user.

.github/workflows/_codeql.yml
LittleHuba: Still open https://github.com/eclipse-score/communication/pull/834#discussion_r3719479967

castler: This should be a second step - because the question is how we can "merge" this and what would be the best user experience for it.

.github/workflows/_codeql.yml
```diff
@@ -27,15 +27,44 @@ on:
...
+            echo "available=true" >> "$GITHUB_OUTPUT"
+          else
+            echo "available=false" >> "$GITHUB_OUTPUT"
+          fi
+
+  # Linux analysis runs on its own runner so it can proceed concurrently
+  # with the QNX analysis below rather than waiting on it.
+  codeql-linux:
+    name: Run CodeQL analysis (Linux)
+    needs: [determine-qnx]
```
LittleHuba, line 63: Nit-pick: Unnecessary

.github/workflows/_codeql.yml
```diff
@@ -49,38 +78,154 @@ jobs:
...
+  # union SARIF. Both tools are invoked directly (no custom wrapper
+  # script): the official Microsoft SARIF SDK CLI (@sarif_multitool, the
+  # same tool used by .github/actions/00_infrastructure/merge_sarif_reports
+  # for rules_lint) merges/dedups the SARIF, then the `sarif` CLI from
+  # sarif-tools (https://github.com/microsoft/sarif-tools) derives the
+  # CSV. This job always runs (even Linux-only, i.e. when codeql-qnx was
+  # skipped) since it is now the only place that produces the published CSV.
+  merge:
+    name: Merge & deduplicate findings (Linux ∪ QNX union)
+    needs: [determine-qnx, codeql-linux, codeql-qnx]
```
LittleHuba, line 166: Nit-pick: `determine-qnx` not required here.

LittleHuba (approved):

PR #835, opened 2026-08-04
memory/shared: eliminate flaky race in ConcurrentAccess test

ahmed0mousa:
The test's inserted_ bookkeeping flag was updated by the writer as a
separate step from the actual UpdateKnownRegion()/RemoveKnownRegion()
call, racing against the reader's plausibility check. A reader could
be paused between its own steps long enough for the writer to fully
insert and remove the same region, making the flag read false on both
sides of a bounds-check that returned true.

Use one shared_mutex per region: the writer holds it exclusively for
its whole insert-or-remove transaction, and the reader holds it shared
across its whole before-flag/bounds-check/after-flag triplet. This
makes the region's state provably frozen during a reader's check while
different regions remain fully concurrent, so the map's lock-free
behaviour is still genuinely exercised.

### DIFF
--- a/score/memory/shared/memory_region_map_test.cpp
+++ b/score/memory/shared/memory_region_map_test.cpp
@@ -18,8 +18,11 @@
 
 #include "gtest/gtest.h"
 
+#include <atomic>
 #include <chrono>
+#include <condition_variable>
 #include <cstdint>
+#include <mutex>
 #include <random>
 #include <thread>
 #include <vector>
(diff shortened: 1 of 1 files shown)


LittleHuba (changes requested):
IMHO this will not remove the flakiness.
Scenario:
1. Reader updates inserted_before
2. Writer UpdatesKnownRegion + sets inserted
3. Reader updates bounds
4. Writer RemoveKnownRegion + sets inserted
5. Reader updates inserted_after

Result: `bounds.has_value() == true`, but both `inserted_before` and `inserted_after` are `false`.

score/memory/shared/memory_region_map_test.cpp
```diff
@@ -368,37 +369,56 @@ TEST_F(MemoryRegionMapTest, ConcurrentAccess)
             false);
     }
 
+    // One lock per region. A reader holds its region's lock (shared) across its whole before-flag/bounds-check/
+    // after-flag triplet. The writer holds it (exclusive) across its whole insert-or-remove transaction for that same
+    // region. This prevents the writer from fully inserting AND removing a region while a reader is merely paused
+    // between its own steps (e.g. due to scheduling delays) - which would otherwise let the reader's bounds-check see
+    // the region as present while both of its flag samples still read "false". Different regions remain fully
+    // concurrent with each other, so the map's actual lock-free behaviour is still exercised.
```
LittleHuba: I'm not really happy with this concept. You may test the synchronization between readers but not between reader and writer. And between reader and writer is more crucial IMHO.

Proposal:
Create a synchronization point between the inserting and removing loops. In the reader, create two loops one checking for the correct inserting and one for the correct removal.

Logically this splits the test in two distinct phases which should clear up the race conditions.

LittleHuba (dismissed):

LittleHuba (approved):

PR #856, opened 2026-08-05
mw/com/test: Fix deprecated Initialize

crimson11:
Refactored all usages in tests of deprecated
Runtime::Initialize(argc, argv) to use the
safecpp::zstring_view based overload instead.

### DIFF
--- a/score/mw/com/BUILD
+++ b/score/mw/com/BUILD
@@ -67,6 +67,7 @@ cc_library(
         ":types",
         "//score/mw/com/impl:runtime",
         "//score/mw/com/mocking:i_runtime",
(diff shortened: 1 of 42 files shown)


castler (changes requested):
We should add a helper that does that once, instead of copying it in all places

limdor:
> We should add a helper that does that once, instead of copying it in all places

@crimson11 I created https://github.com/eclipse-score/baselibs/blob/main/score/string_manipulation/arguments/arguments.h exactly for this use case. If it does not work let me know.

castler (approved):

PR #857, opened 2026-08-06
mw/com: Add test for disabling field set/get in config

bemerybmw:


score/mw/com/test/fields/getter_setter_if_available/main_consumer_and_provider.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    score::cpp::stop_source stop_source{};
+    const bool sig_term_handler_setup_success = score::mw::com::SetupStopTokenSigTermHandler(stop_source);
+    if (!sig_term_handler_setup_success)
+    {
+        std::cerr << "Unable to set signal handler for SIGINT and/or SIGTERM, cautiously continuing\n";
+    }
+
+    auto provider_future = std::async(score::mw::com::test::run_provider, stop_source.get_token(), config.mode);
+    auto consumer_future = std::async(score::mw::com::test::run_consumer, config.mode);
```
LittleHuba: Default launch policy is `deferred invocation` which could lead to a deadlock here. Provide an explicit launch policy here.

```suggestion
    auto provider_future = std::async(std::launch::async, score::mw::com::test::run_provider, stop_source.get_token(), config.mode);
    auto consumer_future = std::async(std::launch::async, score::mw::com::test::run_consumer, config.mode);
```

bemerybmw: Fixed and added a new commit fixing it in a few places.

LittleHuba (approved):

PR #861, opened 2026-08-06
mw/com: Mark skeleton_recreation_test as flaky

gdadunashvili:
Test fails regularly when running on qnx-8 due to a bug in inotify in which we receive spurious inotify events. To be reevaluated once the bug is fixed or service discovery is reimplemented to not rely on inotify events.

### DIFF
--- a/score/mw/com/test/methods/edge_cases_test/integration_test/BUILD
+++ b/score/mw/com/test/methods/edge_cases_test/integration_test/BUILD
@@ -52,8 +52,8 @@ integration_test(
     filesystem = ":filesystem",
     # Test fails regularly when running on qnx-8 due to a bug in
     # inotify in which we receive spurious inotify events.
-    # To be enabled on qnx once the bug is fixed or
+    # To be reevaluated once the bug is fixed or
     # service discovery is reimplemented to not rely on
     # inotify events (SWP-214271).
-    target_compatible_with = ["@platforms//os:linux"],
+    flaky = True,
 )


score/mw/com/test/methods/edge_cases_test/integration_test/BUILD
```diff
@@ -52,8 +52,8 @@ integration_test(
     filesystem = ":filesystem",
     # Test fails regularly when running on qnx-8 due to a bug in
     # inotify in which we receive spurious inotify events.
-    # To be enabled on qnx once the bug is fixed or
+    # To be reevaluated once the bug is fixed or
     # service discovery is reimplemented to not rely on
     # inotify events (SWP-214271).
-    target_compatible_with = ["@platforms//os:linux"],
+    flaky = True,
```
limdor, line 58: I guess it is not possible to do a select statement in the `flaky` flag? With this change we are kind of accepting that in linux it could be flaky. On the other side without this change, we would not notice a reliably failing test on qnx.
If you coult try a select statement on `flaky` and see what Bazel says it would be great. Probably it is not a configurable flag and it will not work, but we could consider asking it as a feature request to Bazel if that is the case.

bemerybmw: IMO, we shouldn't make this change anyway. I'm pretty sure that this test was previously marked as flaky but still occasionally failed in the CI. I don't think that anything has changed that would make the test more reliable so I think we'd still be prone to intermittent CI failures. If there's a way to run the test multiple times in the CI then I'd be more comfortable merging it.

castler: that is exactly what the flaky tag is about.

PR #868, opened 2026-08-07
Restructure access to members of Configuration

SebSparrowHawk:
Currently a Configuration provides direct access to its member variables (of type std::unordered_map) for service types and service instances. 
With this PR the access to these members would be limited, but on the other side the public API of Configuration has been extended so that the necessary information can be retrieved directly and the Configuration can validate itself. (Consumers of a configuration can now call these methods directly, instead of accessing the underlying data and extracting the relevant information. )
With this stricter encapsulation we want to gain the possibility to update configurations after they have been initially loaded.


score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -60,4 +73,258 @@
...
+{
+    auto deployment_info_visitor = score::cpp::overload(
+        [](const LolaServiceTypeDeployment&) {
+            return true;
+        },
+        [](const score::cpp::blank&) noexcept {
+            return false;
+        });
+
+    for (auto& service_type : service_types_)
```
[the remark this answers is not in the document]
SebSparrowHawk: Done

score/mw/com/impl/plumbing/binding_runtime_factory.cpp
```diff
@@ -31,46 +31,25 @@
...
-            break;
+            const auto number_of_needed_tracing_slots = tracing_filter_config->GetNumberOfTracingSlots(configuration);
+            lola_tracing_runtime =
+                std::make_unique<lola::tracing::TracingRuntime>(number_of_needed_tracing_slots, configuration);
         }
-        // LCOV_EXCL_BR_STOP
+
+        auto lola_runtime = std::make_unique<score::mw::com::impl::lola::Runtime>(
+            configuration, long_running_threads, std::move(lola_tracing_runtime));
+        const auto pair = result.emplace(std::make_pair(BindingType::kLoLa, std::move(lola_runtime)));
```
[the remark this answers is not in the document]
SebSparrowHawk: Done

score/mw/com/runtime.cpp
```diff
@@ -48,13 +47,14 @@
     return instance_identifier_container;
 }
 
-// NOLINTNEXTLINE(modernize-avoid-c-arrays):C-style array tolerated for command line arguments
-void InitializeRuntime(const std::int32_t argc, score::StringLiteral argv[])
+// NOLINTNEXTLINE(modernize-avoid-c-arrays):C-style array tolerated for command line arguments. This API is deprecated
+// and it will be removed.
+void InitializeRuntime(const std::int32_t argc, const char* argv[])
```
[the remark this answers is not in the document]
SebSparrowHawk: Change was actually not part of this PR, fixed with rebase

score/mw/com/runtime_configuration.cpp
```diff
@@ -45,10 +43,26 @@
...
+RuntimeConfiguration::RuntimeConfiguration(const std::int32_t argc, const char* argv[]) : configuration_path_{}
+{
+    // We convert the const char* arguments into safecpp::zstring_views to ensure that they are null-terminated and safe
+    // to use.
+    // This is a deprecated API for a reason! If the given argv[] is not null-terminated, it can lead to undefined
+    // behavior. But this isn't introduced newly by this conversion!
+    std::vector<safecpp::zstring_view> command_line_arguments{};
+    for (std::int32_t arg_idx = 0U; arg_idx < argc; arg_idx++)
+    {
+        auto argument = std::string_view{argv[arg_idx]};
```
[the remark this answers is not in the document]
SebSparrowHawk: Change was actually not part of this PR, fixed with rebase

score/mw/com/runtime_configuration.cpp
```diff
@@ -45,10 +43,26 @@
...
+        command_line_arguments.push_back(safecpp::zstring_view{argument.data(), argument.size()});
```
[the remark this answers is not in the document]
SebSparrowHawk: Change was actually not part of this PR, fixed with rebase

score/mw/com/impl/configuration/configuration.h
```diff
@@ -84,7 +80,113 @@ class Configuration final
         return tracing_configuration_;
     }
 
+    ServiceTypeDeployment GetServiceTypeDeployment(const ServiceIdentifierType& service_identifier_type) const noexcept
```
crimson11, line 83: Why a copy here? Why not `ServiceTypeDeployment&` ? We said, that ptrs/references to VALUES stored within the map remain stable ... (oposed to iterators)

score/mw/com/impl/configuration/configuration.h
```diff
@@ -84,7 +80,113 @@ class Configuration final
...
+    {
+        std::lock_guard lock_instances{service_types_mutex_};
+        return service_types_.at(service_identifier_type);
+    }
+
+    ServiceInstanceDeployment GetServiceInstanceDeployment(const InstanceSpecifier& instance_specifier) const noexcept
```
crimson11, line 89: See above: Why a copy and not a ref?

score/mw/com/impl/configuration/configuration.h
```diff
@@ -84,7 +80,113 @@ class Configuration final
...
+    {
+        std::lock_guard lock_instances{service_instances_mutex_};
+        return service_instances_.at(instance_specifier);
+    }
+
+    std::optional<std::reference_wrapper<const ServiceTypeDeployment>> FindServiceTypeDeployment(
```
crimson11, line 95: Here you are using refs! Fine! :D

score/mw/com/impl/configuration/configuration.h
```diff
@@ -84,7 +80,113 @@ class Configuration final
...
+        std::lock_guard lock_instances{service_instances_mutex_};
+        const auto it = service_instances_.find(specifier);
+        if (it == service_instances_.end())
+        {
+            return std::nullopt;
+        }
+        return std::cref(it->second);
+    }
+
+    size_t GetSizeOfServiceTypes() const noexcept
```
crimson11, line 119: Do we need this? If so - "Size" sounds "weird" .... better: `GetNumberOfServiceTypes()`

score/mw/com/impl/configuration/configuration.h
```diff
@@ -84,7 +80,113 @@ class Configuration final
...
+        return service_types_.size();
+    }
+
+    bool IsServiceTypesEmpty() const noexcept
+    {
+        std::lock_guard lock_instances{service_types_mutex_};
+        return service_types_.empty();
+    }
+
+    size_t GetSizeOfServiceInstances() const noexcept
```
crimson11, line 131: See above ... "size" is odd

score/mw/com/impl/configuration/configuration.h
```diff
@@ -84,7 +80,113 @@ class Configuration final
...
+    }
+
+    /// \brief Public interface to trigger a validation of this configuration.
+    score::Result<void> Validate() const noexcept;
+
+    /// \brief Determine if any service with a LoLa binding is defined in this configuration
+    score::Result<bool> HasLolaServiceDeployment() const noexcept;
+
+    /// \brief Returns the list of names (ToString()) of all configured ServiceIdentifierTypes
+    std::set<std::string_view> GetListOfNamesOfConfiguredServices() const noexcept;
```
crimson11, line 150: Why not simply:
GetServiceInstanceNames()?

score/mw/com/impl/configuration/configuration.h
```diff
@@ -95,6 +197,9 @@ class Configuration final
     ServiceInstanceDeployments service_instances_;
     GlobalConfiguration global_configuration_;
     TracingConfiguration tracing_configuration_;
+
+    mutable std::mutex service_instances_mutex_;
```
crimson11, line 201: do we really need mutexes so fine-grained? Isn't one configuration mutex "good" enough?

score/mw/com/impl/configuration/config_validate.h
crimson11, line 14: Makes no sense for me, to keep this stand-alone validation! You should move ALL these validations/checks into the `Configuration`! Having some validation in Configuration and some here ... makes no sense imho

crimson11 (dismissed):

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: I renamed it. It's currently used by some tests. (which probably could also be rewritten.)

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: I renamed it. It's currently used by some tests. (which probably could also be rewritten.)

score/mw/com/impl/configuration/config_validate.h
SebSparrowHawk: I didn't move the remaining parts because they didn't work on the configuration object. 
But yes, having a "half" configuration validation does not make that much sense. I therefore moved these validation to the using code configuration_json_parsing_strategy.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: I would assume that having a coarse grained configuration mutex should be enough. I switched to that, and we can maybe verify if that is enough.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: Changed to GetServiceTypeNames(), thank you!

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: You're right. There was some kind of confussion. Switched to a reference now.

score/mw/com/impl/configuration/configuration.h
```diff
@@ -90,6 +82,60 @@ class Configuration final
...
+        const ServiceIdentifierType& service_identifier_type) const noexcept
+    {
+        return service_types_.at(service_identifier_type);
+    }
+
+    const ServiceInstanceDeployment& GetServiceInstanceDeployment(
+        const InstanceSpecifier& instance_specifier) const noexcept
+    {
+        return service_instances_.at(instance_specifier);
+    }
```
castler: We should not have this, we should only have Find or Get, but not both!

score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -32,9 +34,18 @@ Configuration::Configuration(ServiceTypeDeployments service_types,
...
+      service_instances_{std::move(other.service_instances_)},
+      global_configuration_{std::move(other.global_configuration_)},
+      tracing_configuration_{std::move(other.tracing_configuration_)}
+{
+}
+
 ServiceTypeDeployment* Configuration::AddServiceTypeDeployment(ServiceIdentifierType service_identifier_type,
                                                                ServiceTypeDeployment service_type_deployment) noexcept
 {
+    std::lock_guard lock_types{configuration_mutex_};
```
castler: I don't think we can do this - and I do not get why we should do that? I believe this will be a performance nightmare

score/mw/com/impl/configuration/BUILD
```diff
@@ -63,7 +63,6 @@ cc_library(
     hdrs = ["configuration_json_parsing_strategy.h"],
     features = COMPILER_WARNING_FEATURES,
     implementation_deps = [
-        ":config_validate",
```
castler, line 68: This commit needs to be dropped, as this stuff was extracted in a step for the common validation for the flatbuffers configuration validation...

castler (changes requested):

castler (approved):

PR #884, opened 2026-08-10
memory/shared: fix static destruction order fiasco via nifty-counter idiom

ahmed0mousa:
MemoryResourceRegistry and SharedMemoryFactoryImpl were plain Meyer singletons (MeyerSingleton<T>::GetInstance()). ~SharedMemoryResource() accesses MemoryResourceRegistry::getInstance(), so if a client stores a std::shared_ptr<ISharedMemoryResource> in a static/long-lived context created before the registry's first use, that static could be destroyed after the registry, causing UB/crashes at shutdown.

Both singletons now use score::os::StaticDestructionGuard (the nifty-counter idiom), matching the pattern already used by score::os::Mman. The guard for MemoryResourceRegistry is pulled into the public i_shared_memory_resource.h so any client holding a static shared_ptr<ISharedMemoryResource> is protected; the guard for SharedMemoryFactoryImpl stays internal to shared_memory_factory_impl.h to keep the public shared_memory_factory.h lightweight.

### DIFF
--- a/score/memory/shared/memory_resource_registry.cpp
+++ b/score/memory/shared/memory_resource_registry.cpp
@@ -16,7 +16,8 @@
 #include "score/memory/shared/shared_memory_error.h"
 
 #include "score/mw/log/logging.h"
-#include "score/utils/meyer_singleton/meyer_singleton.h"
+
+#include "score/utils/static_destruction_guard.h"
 
 #include <score/utility.hpp>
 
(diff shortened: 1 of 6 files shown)


score/memory/shared/memory_resource_registry.cpp
```diff
@@ -25,15 +26,20 @@
...
+    // std::shared_ptr<ISharedMemoryResource> stored in some other static/long-lived
+    // context could otherwise be destroyed after this registry, and ~SharedMemoryResource() accesses this registry.
+    // Suppress “AUTOSAR_Cpp14_A5_2_4” rule finding: “Reinterpret_cast shall not be used.”
+    // Rationale: Same idiom as score::os::Mman::instance() (baselibs): reinterpret_cast is used to treat the guarded
+    // aligned storage as MemoryResourceRegistry, which is safe as the storage only ever holds a MemoryResourceRegistry
+    // object.
+    // NOLINTBEGIN(cppcoreguidelines-pro-type-reinterpret-cast): safe usage of reinterpret_cast
+    // coverity[autosar_cpp14_a5_2_4_violation]
+    return reinterpret_cast<MemoryResourceRegistry&>(
+        ::score::utils::StaticDestructionGuard<MemoryResourceRegistry>::GetStorage());
```
LittleHuba: I dislike these reinterpret casts in our codebase. They should be internals of StaticDestructionGuard.
Can you please create a PR in baselibs to fix this?

ahmed0mousa: A PR has been created in baselibs https://github.com/eclipse-score/baselibs/pull/476/ to fix this

ahmed0mousa: A PR has been created to bump the baselibs commit https://github.com/eclipse-score/communication/pull/970

LittleHuba (approved):

PR #896, opened 2026-08-11
bump score_baselibs commit hash

ahmed0mousa:
bump score_baselibs commit hash


module_integration_test/MODULE.bazel
```diff
@@ -8,7 +8,7 @@ bazel_dep(name = "rules_cc", version = "0.2.17")
 bazel_dep(name = "score_baselibs", version = "0.2.9")
 git_override(
     module_name = "score_baselibs",
-    commit = "8b26bb6faa1cf639adeeef73121edf02aae53aa1",
+    commit = "54dc5d43498d93239c27370c8733550cd599fd16",
```
limdor:
```suggestion
    commit = "54d67260077a0e7cdf37faf4366ca3a72f35739f",
```
I believe we should take the merge commit, to make sure we are taking the right content. If not we might end up with the content of the branch with the state it was.
This would also be the latest commit from baselibs.

MODULE.bazel
```diff
@@ -30,7 +30,7 @@ bazel_dep(name = "score_qnx_unit_tests", version = "0.2.0")
 bazel_dep(name = "score_baselibs", version = "0.2.9")
 git_override(
     module_name = "score_baselibs",
-    commit = "8b26bb6faa1cf639adeeef73121edf02aae53aa1",
+    commit = "54dc5d43498d93239c27370c8733550cd599fd16",
```
limdor:
```suggestion
    commit = "54d67260077a0e7cdf37faf4366ca3a72f35739f",
```

castler (approved):

PR #897, opened 2026-08-11
Mf type erased binding

crimson11:


crimson11:
This PR is huge! No chance to break it down further. Thus, we will see conflicts popping up during review-phase. This is expected. Conflicts will be resolved after 1st review phase is done and then (obviously) before merge!

score/mw/com/impl/bindings/lola/skeleton_event.h
```diff
@@ -80,190 +86,108 @@ class SkeletonEvent final : public SkeletonEventBinding<SampleType>
...
+    Result<void> UnsetReceiveHandlerRegistrationChangedHandler() noexcept override;
 
-    auto guard = skeleton_event_common_.AllocateGetterGuard();
-    if (!guard.has_value())
+  private:
+    Skeleton& parent_;
+    std::string_view event_name_;
+    ElementFqId element_fq_id_;
+    EventDataStorage* event_data_storage_;
+    memory::DataTypeSizeInfo event_sample_size_info;
```
SebSparrowHawk: minor change, but the "_" is missing for this member variable name.

score/mw/com/impl/bindings/lola/skeleton_event.cpp
```diff
@@ -11,3 +11,361 @@
...
+
+Result<void> SkeletonEvent::Notify() noexcept
+{
+    return NotifyConsumersIfHandlersRegistered();
+}
+
+Result<void> SkeletonEvent::SetReceiveHandlerRegistrationChangedHandler(
+    ReceiveHandlerRegistrationChangedCallback callback) noexcept
+{
+    static_assert(std::is_same_v<decltype(callback), ReceiveHandlerRegistrationChangedCallback>,
```
SebSparrowHawk: Error message should be adapted because not specific to GenericSkeleton anymore

score/mw/com/impl/skeleton_event.h
```diff
@@ -222,10 +222,10 @@ Result<void> SkeletonEvent<SampleDataType>::Send(SampleAllocateePtr<EventType> s
         return MakeUnexpected(ComErrc::kNotOffered);
     }
 
-    auto tracing_handler =
-        impl::tracing::CreateTracingSendWithAllocateCallback<SampleDataType>(tracing_data_, *binding_);
+    auto tracing_handler = impl::tracing::CreateTracingSendWithAllocateCallback(
+        tracing_data_, memory::DataTypeSizeInfo{sizeof(SampleDataType), alignof(SampleDataType)}, *binding_);
 
-    const auto send_result = GetTypedEventBinding()->Send(std::move(sample), std::move(tracing_handler));
+    const auto send_result = binding_->Send(std::move(sample), std::move(tracing_handler));
```
SebSparrowHawk, line 246: Since we are calling std::memcpy in the binding's Send() (https://github.com/eclipse-score/communication/pull/897/changes/c76dd003fcf1861400d4c7cf4ca3950583927f98#diff-2eb9fffb321ae1e77bafa81061b828344308ae48da9296307d48b9bac9650b50R50), should we add the same static_assert(std::is_trivially_copyable_v) for the sample type as we do for methods in  https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/util/type_erased_storage.h#L138?

SebSparrowHawk (changes requested):

score/mw/com/impl/bindings/lola/event_data_storage.cpp
```diff
@@ -11,3 +11,79 @@
...
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(
+        (event_sample_size_info.Size() == 0U) ||
+            (number_of_slots <= (std::numeric_limits<std::size_t>::max() / event_sample_size_info.Size())),
+        "Overflow while calculating the total size of the raw event-data slot-array.");
+    const auto storage_bytes_needed = number_of_slots * event_sample_size_info.Size();
+
+    // The alignment used here must match exactly the alignment CalculateServiceDataStorageShmSize() assumes for this
+    // allocation (see service_data_storage.cpp), i.e. event_sample_size_info.Alignment(). Using a different (e.g.
+    // hardcoded, stricter) alignment here would make the analytically calculated shm-size wrong (too small).
+    void* const type_erased_data_slots_start =
```
Tejveerpratap2803, line 45: Suggestion: This coupling is only documented in a comment, not enforced. If this alignment logic ever changes here without a matching change in CalculateServiceDataStorageShmSize(), the shm segment gets allocated too small, causing an out-of-bounds write in shared memory. Can we add a shared helper, a runtime size-check, or a test comparing actual vs. estimated shm size   instead of relying on the comment. What do you think?

Tejveerpratap2803:
I think we should consider updating the PUML diagram on the skeleton side as well.

Tejveerpratap2803 (changes requested):

score/mw/com/impl/bindings/lola/skeleton_event.cpp
crimson11: You are right! But I just removed the whole assert ... as the assert is always true! With our current signature!

score/mw/com/impl/skeleton_event.h
crimson11: YES. Good point!
And we seem to need to update our AoUs!

score/mw/com/impl/bindings/lola/event_data_storage.cpp
crimson11: Good point! I now also did provide a specific size-calc method for EventDataStorage, which now gets calles from CalculateServiceDataStorageShmSize() to have the size/alignment decisions closer together.
I had to add some todos as we wouold need to change the call-chain a bit to have it even stronger verified ...

crimson11:
> I think we should consider updating the PUML diagram on the skeleton side as well.

Yep! But I think we should adapt doc/design/PUML, when we also have adapted the Proxy side AND Brendan is happy! Always do doc/design update, when we are really done with code to avoid double effort!

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -60,12 +61,8 @@ class SkeletonEventBindingBase
...
     virtual void PrepareStopOffer() noexcept = 0;
 
-    /// \brief Calculate the necessary memory for the underlying event-type (including possible dynamic memory
-    /// allocations)
-    virtual std::size_t GetMaxSize() const noexcept = 0;
-
-    /// \brief Alignment requirement (in bytes) of the underlying event-type's sample data.
-    virtual std::size_t GetAlignment() const noexcept = 0;
+    /// \brief Get size for the underlying event-type (including possible dynamic memory allocations) and its alignment
+    virtual memory::DataTypeSizeInfo GetSizeInfo() const noexcept = 0;
```
bemerybmw: Remove noexcept

score/mw/com/impl/generic_skeleton_event_binding.h
bemerybmw, line 1: Review note: This should be removed?

score/mw/com/impl/initialize_sample_callback.h
```diff
@@ -0,0 +1,30 @@
...
+#include <score/callback.hpp>
+
+namespace score::mw::com::impl
+{
+
+/// \brief Callback type for initializing a sample. The callback takes a pointer to the type-erased sample to be
+/// initialized as an argument.
+/// \details The callback is handed over from our strongly typed (binding independent) layer, which has the strong
+/// type definition to the type-erased (binding) layers, in cases, where type-erased storage needs to be correctly
+/// initialized.
```
bemerybmw, line 25: Do you think it would help to have a "possible implementation" in the docs here? e.g.
```
/// Possible implementation:
/// auto initialize_sample_callback = [](void* type_erased_ptr_to_initialize){
    // Default initialize sample
    typed_ptr_to_initialize = new (EventType);
}
```
Not sure if you need a cast in there as well.

score/mw/com/impl/skeleton_event_binding.h
bemerybmw: I think the calling code looks a bit weird that you"re getting the size info and then you get the alignment out of the returned object. What about something like GetEventTypeInfo?

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,61 @@
...
+#include "score/mw/com/impl/initialize_sample_callback.h"
 
-#include <scoped_allocator>
+#include <cstddef>
 
 namespace score::mw::com::impl::lola
 {
 
-/// \brief Container for storing the actual data of a LoLa Event
+/// \brief Container for storing the actual data of a LoLa event (resp. field) within shared-memory.
```
bemerybmw: of a LoLa event or field

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,61 @@
...
 ///
```
bemerybmw, line 29: Maybe add a comment about why it's type erased and that the data type info is provided on construction? Also that the memory is allocated once on construction.

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,61 @@
...
-using EventDataStorage = score::containers::
-    DynamicArray<SampleType, std::scoped_allocator_adaptor<memory::shared::PolymorphicOffsetPtrAllocator<SampleType>>>;
+class EventDataStorage final
+{
+  public:
+    EventDataStorage(memory::shared::ManagedMemoryResource& resource,
+                     SlotIndexType number_of_slots,
+                     memory::DataTypeSizeInfo event_sample_size_info);
+
+    ~EventDataStorage();
```
bemerybmw, line 54: Rule of 5 ;)

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,61 @@
...
+
+    /// \brief Initializes the (type erased) slots, by calling the callback for each slot.
+    /// \details EventDataStorage itself is type-erased. But in the end strongly typed elements are stored. Since
+    ///          EventDataStorage itself has no type information, it cannot initialize the slots itself. Therefore, an
+    ///          upper layer with type knowledge calls this API handing over a callback, which does the correct
+    ///          initialization.
+    /// \param callback The callback to be called for each slot, which has a type erased signature.
+    void InitializeSlots(InitializeSampleCallback callback);
```
bemerybmw: Why isn't the callback provided to the constructor and this function is called internally? The slots will be in an invalid state in between the constructor and this function being called.

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,61 @@
...
+
+    /// \brief Returns a pointer to the type-erased data slot at the given index.
+    /// \details This access also does a complete bounds-check to verify that the returned raw-pointer is within the
+    ///          bounds as well as the end-address (returned pointer plus data_size).
+    /// \param data_size The size of the data slot. This is used to verify, that the callers size expectation matches
+    ///        the size of the event data type, the EventDataStorage was constructed with.
+    /// @return A pointer to the type-erased data slot.
+    void* GetTypeErasedDataSlot(SlotIndexType index, size_t data_size) const;
```
bemerybmw, line 72: I guess we should add nodiscard to new apis? Or do we just want to do it when returning a result?

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,61 @@
...
+
+  private:
+    SlotIndexType number_of_slots_;
+    memory::DataTypeSizeInfo sample_size_info_;
+    memory::shared::ManagedMemoryResource& memory_resource_;
+
+    memory::shared::OffsetPtr<std::byte> type_erased_data_slots_;
+
+    /// size of type_erased_data_slots_ storage in bytes. This is equal to number_of_slots_ * sample_size_info_.Size()
+    std::size_t type_erased_data_slots_storage_size_;
```
bemerybmw: type_erased_data_slots_storage_size_in_bytes_?

score/mw/com/impl/bindings/lola/event_data_storage.cpp
```diff
@@ -11,3 +11,94 @@
...
+      sample_size_info_(event_sample_size_info),
+      memory_resource_(resource),
+      type_erased_data_slots_(nullptr),
+      type_erased_data_slots_storage_size_(0)
+{
+    // Guard against an overflow when calculating the total number of bytes needed for the raw slot-array. Without
+    // this check, an overflowing multiplication would silently wrap around to a much smaller value than what is
+    // actually required, leading to an undersized allocation and, subsequently, out-of-bounds accesses when the
+    // slots are used (see GetTypeErasedDataSlot()).
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(
```
bemerybmw: Use safe math?

score/mw/com/impl/bindings/lola/event_data_storage.cpp
```diff
@@ -11,3 +11,94 @@
...
+
+EventDataStorage::~EventDataStorage()
+{
+    if (type_erased_data_slots_ != nullptr)
+    {
+        memory_resource_.deallocate(type_erased_data_slots_.get(), type_erased_data_slots_storage_size_);
+    }
+}
+
+void EventDataStorage::InitializeSlots(InitializeSampleCallback callback)
```
bemerybmw: InitializeSampleCallback initialization_callback?

score/mw/com/impl/bindings/lola/proxy_event.h
```diff
@@ -218,26 +195,17 @@ inline Result<std::size_t> ProxyEvent<SampleType>::GetNewSamplesImpl(Callback&&
...
-        // The current fix avoids interpreting GenericSkeleton-created storage as EventDataStorage<SampleType>, since
-        // the DynamicArray element count may not match the typed proxy sample type.
-        // NOLINTBEGIN(cppcoreguidelines-pro-bounds-pointer-arithmetic) The pointer event_slots_raw_array_ points to
-        // the first byte of the type-erased event sample storage in shared memory. Samples may originate from either a
-        // typed SkeletonEvent or a GenericSkeletonEvent, therefore slot lookup must use the stable EventMetaInfo raw
-        // storage address and SampleType stride instead of interpreting the shared-memory DynamicArray object type.
-        const auto* const object_start_address = &event_slots_raw_array_[aligned_sample_size_ * (*slot_index_it)];
-        // NOLINTEND(cppcoreguidelines-pro-bounds-pointer-arithmetic)
+        const void* type_erased_sample_ptr =
+            event_data_storage_.GetTypeErasedDataSlot(*slot_index_it, sizeof(SampleType));
```
bemerybmw, line 202: This will lead to bounds checking. As discussed, I think it would make sense that EventDataStorage has a private function (accessed via friend) to get the pointer to the start of the type erased array. This function returns a span<std::byte> and does bounds checking for the full array (i.e. start and end address). We then have an EventDataStorageLocalView which is owned by Proxy/SKeleton Events which is constructed with a reference EventDataStorage and does the pointer arithmetic to return a span to the individual slot.

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,71 @@
...
+    /// \param callback The callback to be called for each slot, which has a type erased signature.
+    void InitializeSlots(const InitializeSampleCallback& callback);
+
+    /// \brief Returns a pointer to the type-erased data slot at the given index.
+    /// \details This access also does a complete bounds-check to verify that the returned raw-pointer is within the
+    ///          bounds as well as the end-address (returned pointer plus data_size).
+    /// \param data_size The size of the data slot. This is used to verify, that the callers size expectation matches
+    ///        the size of the event data type, the EventDataStorage was constructed with.
+    /// @return A pointer to the type-erased data slot.
+    void* GetTypeErasedDataSlot(SlotIndexType index, size_t data_size) const;
```
bemerybmw, line 72: General point: I think that we should represent type erased pointers as span<std::byte> instead of void* so that we always have the expected size in bytes.

score/mw/com/impl/generic_skeleton_event_binding.h
crimson11: ? This file was intentionally deleted.

score/mw/com/impl/skeleton_event_binding.h
crimson11: I did rename to `GetDataTypeSizeInfo()` -> this reflects the type it returns.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.h
```diff
@@ -95,25 +95,16 @@ class SkeletonMemoryManager final
...
     /// \param element_properties Properties of the event.
-    /// \param sample_size The size of a single data sample.
-    /// \param sample_alignment The alignment of the data sample.
-    /// \return A raw pointer to the first byte of the generic event sample storage.
-    auto CreateGenericEventDataInCreatedSharedMemory(const ElementFqId element_fq_id,
-                                                     const SkeletonEventProperties& element_properties,
-                                                     size_t sample_size,
-                                                     size_t sample_alignment) -> void*;
+    /// \param sample_size_info Information about the size and alignment of the data sample.
+    /// \return A raw pointer to the EventDataStorage.
```
bemerybmw: I think the return comment can be removed. If you want to keep it, it's a reference not a raw pointer.

score/mw/com/impl/skeleton_event_binding.h
bemerybmw: I still think it should mention Event. Because it should be clear that the DataTypeSizeInfo is of the Event type.

score/mw/com/impl/generic_skeleton_event_binding.h
bemerybmw: This was just a note to me because I was just reviewing the first commit. I wanted to double check that it was removed in the second commit.

score/mw/com/impl/bindings/lola/event_meta_info.h
```diff
@@ -23,22 +22,18 @@ namespace score::mw::com::impl::lola
 /// \details Normally proxies/skeletons or "user code" dealing with an event, know its properties. This info is
 ///          provided and placed into shared-memory for the GenericProxy use-case, where a proxy connects to a provided
 ///          service based on only deployment info, NOT having any knowledge about the exact data type of the event.
+///          Currently, the only "meta-info" needed is DataTypeSizeInfo. However, we wrap it into EventMetaInfo to be
+///          prepared for future extensions.
 class EventMetaInfo
```
bemerybmw, line 27: Do we still need this class? Can we just use DataTypeSizeInfo directly?

score/mw/com/impl/bindings/lola/generic_proxy_event.cpp
bemerybmw, line 132: I guess this will all be removed anyway when we do the type erasure on the proxy side?

score/mw/com/impl/bindings/lola/proxy.h
```diff
@@ -143,6 +143,12 @@ class Proxy : public ProxyBinding
     /// Terminates if the event control structure cannot be found.
     TransactionLogSet& GetTransactionLogSet(const ElementFqId element_fq_id);
 
+    /// Retrieves a reference to the event data storage area for a given ElementFqId.
+    ///
+    /// \param element_fq_id The Event ID.
```
bemerybmw: Remove these redundant comments (param / return)

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -624,6 +624,29 @@ TransactionLogSet& Proxy::GetTransactionLogSet(const ElementFqId element_fq_id)
...
+{
+    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD_MESSAGE(
+        data_ != nullptr, "Proxy::GetEventDataStorage: Managed memory data pointer is Null");
+    auto& service_data_storage = detail_proxy::GetServiceDataStorage(*data_);
+    const auto event_entry = service_data_storage.events_.find(element_fq_id);
+    if (event_entry == service_data_storage.events_.end())
+    {
+        score::mw::log::LogFatal("lola") << __func__ << __LINE__
+                                         << "Unable to find data storage for given event instance. Terminating.";
+        std::terminate();
```
bemerybmw: Should be an assertion so that we can do precondition violation instead of death tests.

score/mw/com/impl/bindings/lola/service_data_storage.h
```diff
@@ -36,16 +37,15 @@ class ServiceDataStorage
...
     /// \brief associative container mapping a service-element (event/field) to the raw storage of its event-data slots.
-    /// \details The value-type of the map is a type-erased pointer to the raw storage of the event-data slots.
-    ///          The OffsetPtr points to a EventDataStorage<SampleType>, which gets created by events/fields, when
-    ///          calling Skeleton::Register()!
-    ///
-    using EventDataStorageMap = LinearSearchMap<ElementFqId, score::memory::shared::OffsetPtr<void>>;
+    /// \details The value-type of the map is a pointer to the storage of the type-erased event-data slots.
+    ///          The OffsetPtr points to a EventDataStorage, which gets created by events/fields, when calling
+    ///          Skeleton::Register()!
+    using EventDataStorageMap = LinearSearchMap<ElementFqId, score::memory::shared::OffsetPtr<EventDataStorage>>;
```
bemerybmw, line 43: Doesn't have to be done now, but what do you think about adding an OffsetReference / OffsetReferenceWrapper or something similar? Basically a thin wrapper around an OffsetPtr but has reference semantics? Because it's weird that the called has to check that the OffsetPtr here is not null when getting an EventDataStorage.

score/mw/com/impl/bindings/lola/skeleton.h
```diff
@@ -122,31 +114,18 @@ class Skeleton final : public SkeletonBinding
...
-
-    /// \brief Enables dynamic registration of Events at the Skeleton.
-    /// \tparam SampleType The type of the event
-    /// \param element_fq_id The full qualified of the element (event or field) that shall be registered
-    /// \param element_properties properties of the element, which are currently event specific properties.
+    /// \param data_type_size_info The size and alignment of a single data sample in bytes.
     /// \return The registered data structures within the Skeleton
-    /// (first: where to store data, second: control data
-    ///         access) If PrepareOffer created the shared memory, then will create an EventDataControl (for QM and
+    ///         (first: where to store data, second: control data access)
```
bemerybmw: What is first / second here? Are you referring to elements of RegistrationResult? Would it make more sense to add that documentation to RegistrationResult itself? Otherwise, it's kind of defeating the purpose of the abstraction and requires us to keep this comment in sync with the code defined elsewhere.

score/mw/com/impl/bindings/lola/skeleton_memory_manager.cpp
```diff
@@ -226,73 +207,54 @@ auto SkeletonMemoryManager::CreateEventControlsInCreatedSharedMemory(const Eleme
...
-    const auto aligned_sample_size =
-        memory::shared::CalculateAlignedSize(sample_size, static_cast<std::size_t>(sample_alignment));
-    const auto total_event_slots_size = safe_math::Multiply<safe_math::ReturnMode::kAbortOnError>(
-        aligned_sample_size, element_properties.GetTotalNumberOfSlots());
+    // Suppress "AUTOSAR C++14 A5-3-2": Don't dereference null pointers.
+    // Justification: The "event_data_storage_it" variable is an iterator of interprocess map returned by the
+    // "find_element" method. A check is made that the iterator is not equal to map.cend(). Therefore, the call to
+    // "event_data_storage_it->" does not return nullptr.
+    // coverity[autosar_cpp14_a5_3_2_violation]
+    auto* const type_erased_event_data_storage_ptr = event_data_storage_it->second.get();
```
bemerybmw: Couldn't you dereference it within find_element() above and return a reference to the pair? That way, you don't have to dereference the iterator here and can avoid the suppression

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
```diff
@@ -235,7 +233,8 @@ SlotIndexType LolaProxyEventResources::PutData(const std::uint32_t value,
     auto slot_result = provider_event_data_control_local_->AllocateNextSlot();
     EXPECT_TRUE(slot_result.has_value());
     auto slot_index = slot_result.value();
-    event_data_storage_->at(slot_index) = value;
+    auto* storage_slot = event_data_storage_->GetTypeErasedDataSlot(slot_index, sizeof(value));
+    memcpy(storage_slot, &value, sizeof(value));
```
bemerybmw, line 237: If we introduce the local view which returns a span, then I guess you should be able to use std::copy instead of memcpy?

score/mw/com/impl/bindings/lola/test/skeleton_component_test.cpp
```diff
@@ -150,8 +151,10 @@ class SkeletonComponentTestFixture : public ::testing::Test
         ON_CALL(lola_runtime_mock_, GetLolaMessaging()).WillByDefault(ReturnRef(message_passing_service_mock_));
         ON_CALL(runtime_mock_, GetTracingRuntime()).WillByDefault(Return(&tracing_runtime_mock_));
 
-        ON_CALL(mock_event_binding_, GetMaxSize()).WillByDefault(Return(sizeof(TestSampleType)));
-        ON_CALL(mock_field_binding_, GetMaxSize()).WillByDefault(Return(sizeof(TestSampleType)));
+        ON_CALL(mock_event_binding_, GetSizeInfo())
+            .WillByDefault(Return(memory::DataTypeSizeInfo{sizeof(TestSampleType), alignof(TestSampleType)}));
```
bemerybmw: Return(kSampleTypeSizeInfo)?

score/mw/com/impl/bindings/lola/event_data_storage_test.cpp
```diff
@@ -12,28 +12,190 @@
...
+TEST(EventDataStorageDeathTest, GetTypeErasedDataSlotTerminatesOnDataSizeMismatch)
+{
+    // Given an EventDataStorage constructed for a std::uint32_t sample type
+    memory::shared::NewDeleteDelegateMemoryResource memory_resource{kMemoryResourceId};
+    const memory::DataTypeSizeInfo sample_size_info{sizeof(std::uint32_t), alignof(std::uint32_t)};
+    EventDataStorage unit{memory_resource, kNumberOfSlots, sample_size_info};
+
+    // When requesting a data slot with a data_size that does not match the sample type's actual size
+    // Then the program terminates, since the caller's size expectation doesn't match the storage's sample size.
+    EXPECT_DEATH(unit.GetTypeErasedDataSlot(0U, sizeof(std::uint32_t) + 1U), ".*");
```
bemerybmw: These should be precondition violation tests instead of death tests.

score/mw/com/impl/bindings/lola/event_data_storage.h
crimson11: I tend to rather add it, when the return code also could hint at an error (e.g. with Result), because the "error case" we want to avoid a discard ...

score/mw/com/impl/bindings/lola/event_data_storage.h
crimson11: I guess, it depends, where it is used! In the cases, where I hand back such a type-erased ptr to a layer, which HAS the needed type-info and is expected to do a cast (`static_cast`) to some type it is imho crucial to use a void*! Because this is perfect C++ semantics! With a other pointers (like pointers to a `std::span<std::byte>`) you force the other layer to do `reinterprete_cast`! No way!" This is just bad and aginst the idea of type-erasure imho!

score/mw/com/impl/bindings/lola/proxy_event.h
crimson11: Checked that. Before the haxx we did for the genetic-skeleton, we also had an access to the underlying `DynamicArray<SampleType>` , which contained bounds-checking. 

So I created https://jira.cc.bmwgroup.net/browse/SWP-281506 for this change.

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/event_meta_info.h
crimson11: See my in code documentation:

/// Currently, the only "meta-info" needed is DataTypeSizeInfo. However, we wrap it into EventMetaInfo to be
///          prepared for future extensions.

For now I would like to keep it. We can refactor later, when we are sure, we never have additional meta-info to the size/alignment

score/mw/com/impl/bindings/lola/generic_proxy_event.cpp
crimson11: Yes! Then there is no GenericProxy on the binding level anymore!

score/mw/com/impl/bindings/lola/service_data_storage.h
crimson11: I guess that makes sense ... if done correctly it will really be an entity "beside" `OffsetPtr` and not just a simple/small wrapper around, which handles `nullptr` representations ...

score/mw/com/impl/bindings/lola/test/proxy_event_test_resources.cpp
crimson11: Yep, I guess so.

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -35,78 +38,81 @@ namespace score::mw::com::impl
...
     /// de-initialization)
     virtual void PrepareStopOffer() noexcept = 0;
 
-    /// \brief Calculate the necessary memory for the underlying event-type (including possible dynamic memory
-    /// allocations)
-    virtual std::size_t GetMaxSize() const noexcept = 0;
-
-    /// \brief Alignment requirement (in bytes) of the underlying event-type's sample data.
-    virtual std::size_t GetAlignment() const noexcept = 0;
+    /// \brief Get size and alignment for the underlying event-type (including possible dynamic memory allocations).
```
bemerybmw: (including possible dynamic memory allocations). -> can you elaborate? How would we reflect the dynamic allocations here?

score/mw/com/impl/bindings/lola/event_data_storage.h
```diff
@@ -13,24 +13,92 @@
...
-using EventDataStorage = score::containers::
-    DynamicArray<SampleType, std::scoped_allocator_adaptor<memory::shared::PolymorphicOffsetPtrAllocator<SampleType>>>;
+/// The EventDataStorage is type-erased. Because it is located at the binding level. A binding just stores/moves
+/// byte-streams. It doesn't need type-information.
+/// The storage slots are allocated and (optionally) initialized at construction. The calling layer, which knows, how to
+/// type-correctly initialize a slot can hand over a callable, which does the correct initialization.
+class EventDataStorage final
+{
+  public:
+    /// \param initialize_sample_callback Optional callback used to initialize (default-construct) every slot right
```
bemerybmw: It doesn't technically have to default construct?

score/mw/com/impl/bindings/lola/event_data_storage.cpp
```diff
@@ -11,3 +11,110 @@
...
+                                   SlotIndexType number_of_slots,
+                                   memory::DataTypeSizeInfo event_sample_size_info,
+                                   const std::optional<InitializeSampleCallback>& initialize_sample_callback)
+    : number_of_slots_(number_of_slots),
+      sample_size_info_(event_sample_size_info),
+      memory_resource_(resource),
+      type_erased_data_slots_(nullptr),
+      type_erased_data_slots_storage_size_in_bytes_(0)
+{
+    const auto storage_bytes_needed_result =
```
bemerybmw, line 34: safe_math::Multiply<safe_math::ReturnMode::kAbortOnError>(

this way you don't have to check the return value.

score/mw/com/impl/bindings/lola/event_data_storage.cpp
```diff
@@ -11,3 +11,110 @@
...
+    const auto last_slot_offset = sample_size_info_.Size() * (number_of_slots_ - 1U);
+    auto last_slot_ptr = type_erased_data_slots_ + decltype(type_erased_data_slots_)::difference_type(last_slot_offset);
+    auto* last_slot_raw_ptr = last_slot_ptr.get();
+
+    // This is our low-level data storage, where we work on type-erased data, thus pointer arithmetic can't be avoided.
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) see above
+    for (auto* current_slot_raw_ptr = first_slot_raw_ptr; current_slot_raw_ptr <= last_slot_raw_ptr;
+         current_slot_raw_ptr += sample_size_info_.Size())
+    {
+        initialization_callback(current_slot_raw_ptr);
```
bemerybmw:
```suggestion
        std::invoke(initialization_callback, current_slot_raw_ptr);
```

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
 template <typename SampleType>
 class SampleAllocateePtr;
 
-class SkeletonEventBindingBase
+/// \brief The SkeletonEventBinding represents the interface that _every_ binding has to provide, if it wants to support
+/// events. It will be used by a concrete SkeletonEvent to perform any binding specific operation.
```
bemerybmw: by a concrete impl::SkeletonEvent

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
...
-    SkeletonEventBindingBase(SkeletonEventBindingBase&&) noexcept = delete;
-    SkeletonEventBindingBase& operator=(const SkeletonEventBindingBase&) & = delete;
-    SkeletonEventBindingBase& operator=(SkeletonEventBindingBase&&) & noexcept = delete;
+    SkeletonEventBinding(const SkeletonEventBinding&) = delete;
+    SkeletonEventBinding(SkeletonEventBinding&&) noexcept = delete;
+    SkeletonEventBinding& operator=(const SkeletonEventBinding&) & = delete;
+    SkeletonEventBinding& operator=(SkeletonEventBinding&&) & noexcept = delete;
+
+    /// \brief SampleType is allocated by the user and provided to the middleware to send
+    /// \return On failure, returns an error code.
```
bemerybmw: remove \return

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
...
+    virtual Result<void> Send(const void*, std::optional<SendTraceCallback>, SampleAllocateeGuard) noexcept = 0;
+
+    /// \brief SampleType is previously allocated by middleware and provided by the user to indicate that he is finished
+    /// filling the provided pointer with live.
+    /// \return On failure, returns an error code.
```
bemerybmw: remove \return

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
 template <typename SampleType>
 class SampleAllocateePtr;
 
-class SkeletonEventBindingBase
+/// \brief The SkeletonEventBinding represents the interface that _every_ binding has to provide, if it wants to support
+/// events. It will be used by a concrete SkeletonEvent to perform any binding specific operation.
+class SkeletonEventBinding
```
bemerybmw, line 46: I think we should have an explanation here about why we have void* / SampleAllocateePtr<void> and where the type info comes from i.e. the impl::SkeletonEvent is typed.

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
...
+    /// handed over from the binding independent (strongly typed) layer, in case the binding layer needs to initialize
+    /// binding specific (type erased) storage. As only the binding independent layer has strong type knowledge, it is
+    /// the one to provide such an initializer.
+    /// The callback is optional: callers that don't have type knowledge (e.g. GenericSkeletonEvent, which is
+    /// already type-erased on the binding independent layer) or that don't need type-correct initialization hand over
+    /// an empty optional. In that case, the binding shall skip initializing the type-erased storage. The callback (if
+    /// any) is taken by const reference (instead of by value) and must only be invoked synchronously within
+    /// PrepareOffer(), i.e. it must not be stored/moved for later use. This allows the caller (SkeletonEventBase) to
+    /// keep ownership of the callback and reuse it across multiple PrepareOffer() calls over the lifetime of the event
+    /// (e.g. offer -> stop-offer -> offer again). \param initialize_sample_callback Optional callback to initialize a
```
bemerybmw: \param on a new line

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
...
 
-/// \brief The SkeletonEventBinding represents the interface that _every_ binding has to provide, if it wants to support
-/// events. It will be used by a concrete SkeletonEvent to perform any binding specific operation.
-template <typename SampleType>
-class SkeletonEventBinding : public SkeletonEventBindingBase
-{
-  public:
-    using SendTraceCallback = score::cpp::callback<void(SampleAllocateePtr<SampleType>&), 64U>;
+    /// \brief Trigger notification of potential registered receive handlers.
+    /// \details This is a specific API for the gateway use-case!
```
bemerybmw, line 104: and will only be publicly exposed via a GenericSkeletonEvent?

score/mw/com/impl/skeleton_event_binding.h
```diff
@@ -36,69 +38,81 @@ namespace score::mw::com::impl
...
-    /// \brief Allocates memory for SampleType for the user to fill it. This is especially necessary for Zero-Copy
-    /// implementations.
-    virtual Result<SampleAllocateePtr<SampleType>> Allocate(SampleAllocateeGuard guard) noexcept = 0;
-
-    /// \brief Retrieves the latest sample, intended to support the getter of a SkeletonField.
-    virtual Result<SamplePtr<SampleType>> GetLatestSample(QualityType quality_type) = 0;
+    /// \brief Sets a callback that will be called when the first ReceiveHandler of a GenericEvent
+    /// will get registered or the last ReceiveHandler will be removed.
+    /// \details This is a specific API for the gateway use-case!
+    virtual Result<void> SetReceiveHandlerRegistrationChangedHandler(
```
bemerybmw, line 112: As discussed, we should have GenericSkeletonEventBinding which inherits from SkeletonEventBinding with these GenericSkeletonEvent only functions. The lola::SkeletonEvent would implement the  GenericSkeletonEvent binding. Can be done as a follow up (need a ticket)

score/mw/com/impl/plumbing/sample_allocatee_ptr.h
```diff
@@ -68,6 +68,15 @@ class SampleAllocateePtr
     /// \brief Constructs a SampleAllocateePtr by transferring ownership from other to *this.
     SampleAllocateePtr(SampleAllocateePtr<SampleType>&& other) noexcept;
 
+    /// \brief Constructs a SampleAllocateePtr by transferring ownership from other to *this, even if other has a
+    /// different SampleType. This is a rebind-move-ctor used by the binding independent layer!
+    /// \details The binding independent layer exchanges SampleAllocateePtr<void> with the binding layer. I.e. bindings
+    /// are expected to store type-erased pointers in the variant member (internal_). But towards the user API/layer,
+    /// the binding independent layer uses SampleAllocateePtr<SampleType> to provide type-safety. This constructor
+    /// allows to move from a SampleAllocateePtr<void> to a SampleAllocateePtr<SampleType>.
+    template <typename OtherSampleType>
```
bemerybmw, line 77: As discussed, should either have an enable_if or AsVoid / ToVoid or something similar.

score/mw/com/impl/plumbing/sample_ptr.h
```diff
@@ -78,6 +79,45 @@ class SamplePtr final
...
     SamplePtr& operator=(SamplePtr<SampleType>&& other) & noexcept = default;
 
+    /// \brief Interim rebind ctor converting a type-erased SamplePtr<void> into a concrete SamplePtr<SampleType>.
+    ///
+    /// \details This is needed because, as part of a stepwise refactoring, the skeleton-side binding API
+    /// already returns type-erased SamplePtr<void> instances (see SkeletonEventBinding::GetLatestSample()), while
+    /// this class - unlike SampleAllocateePtr - still is (and, for now, needs to remain) a class template, since it
+    /// is also used unchanged by the (not yet refactored) proxy-side (e.g. GetNewSamples()).
+    ///
+    /// \attention This is an INTERIM solution only! Once lola::SamplePtr and mock_binding::SamplePtr get converted
```
bemerybmw, line 89: Link the tickefor the proxy side changes.

score/mw/com/impl/plumbing/sample_ptr.h
```diff
@@ -78,6 +79,45 @@ class SamplePtr final
...
+    /// variant type differs between SamplePtr<void> and SamplePtr<SampleType>. Instead, the currently active
+    /// alternative of \p other is visited and rebuilt as the corresponding alternative of this instance.
+    ///
+    /// \tparam OtherSampleType source SampleType; only enabled for OtherSampleType == void and SampleType != void, as
+    ///                         the rebind direction is always from type-erased to concrete.
+    /// \param other type-erased SamplePtr<void> to rebind into a SamplePtr<SampleType>. Left in an invalid
+    ///              (blank) state afterwards.
+    template <typename OtherSampleType,
+              typename = std::enable_if_t<std::is_void<OtherSampleType>::value && !std::is_void<SampleType>::value>>
+    explicit SamplePtr(SamplePtr<OtherSampleType>&& other) noexcept
```
bemerybmw, line 105: Couldn't this just be explicit SamplePtr(SamplePtr<void>&& other) noexcept

score/mw/com/impl/generic_skeleton_event.cpp
```diff
@@ -53,9 +52,10 @@ Result<void> GenericSkeletonEvent::Send(SampleAllocateePtr<void> sample) noexcep
     }
 
     SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(binding_ != nullptr, "Binding is not initialized!");
-    auto* const binding = static_cast<GenericSkeletonEventBinding*>(binding_.get());
+    auto* const binding = binding_.get();
```
bemerybmw: This could be removed and simply call binding_->Send(... ?

score/mw/com/impl/generic_skeleton_event.cpp
```diff
@@ -12,34 +12,33 @@
...
 {
 
 GenericSkeletonEvent::GenericSkeletonEvent(SkeletonBase& skeleton_base,
                                            const std::string_view event_name,
-                                           std::unique_ptr<GenericSkeletonEventBinding> binding)
-    : SkeletonEventBase(event_name, std::move(binding))
+                                           std::unique_ptr<SkeletonEventBinding> binding)
+    // GenericSkeletonEvent is already type-erased on the binding independent layer (it has no concrete SampleType),
+    // so it cannot provide a type-correct InitializeSampleCallback and hands over an empty optional instead.
+    : SkeletonEventBase(event_name, std::nullopt, std::move(binding))
```
bemerybmw: What about having a global constant called kEmptyInitializeSampleCallback or something similar?

score/mw/com/impl/skeleton_event.h
```diff
@@ -175,14 +182,16 @@ SkeletonEvent<SampleDataType>::SkeletonEvent(SkeletonBase& skeleton_base,
...
-    : SkeletonEventBase{event_name, std::move(binding)}, skeleton_event_mock_{nullptr}
+                                             std::unique_ptr<SkeletonEventBinding> binding)
+    : SkeletonEventBase{event_name, MakeInitializeSampleCallback(), std::move(binding)}, skeleton_event_mock_{nullptr}
 {
 }
 
 template <typename SampleDataType>
 Result<void> SkeletonEvent<SampleDataType>::Send(const EventType& sample_value) noexcept
 {
+    static_assert(std::is_trivially_copyable_v<SampleDataType>,
```
bemerybmw: To be removed as discussed

score/mw/com/impl/skeleton_event.h
```diff
@@ -250,23 +260,26 @@ Result<SampleAllocateePtr<SampleDataType>> SkeletonEvent<SampleDataType>::Alloca
...
-}
 
-template <typename SampleDataType>
-auto SkeletonEvent<SampleDataType>::GetTypedEventBinding() const noexcept -> SkeletonEventBinding<SampleDataType>*
-{
-    auto* const typed_binding = dynamic_cast<SkeletonEventBinding<EventType>*>(binding_.get());
-    SCORE_LANGUAGE_FUTURECPP_ASSERT_PRD_MESSAGE(typed_binding != nullptr,
-                                                "Downcast to SkeletonEventBinding<EventType> failed!");
-    return typed_binding;
+    // Currently the underlying (binding managed) slots are initialized once via call to binding_->PrepareOffer().
```
bemerybmw, line 279: I'm not sure that we should leave this comment in the code. I don't think it's clear that we should take this route. e.g. I think we discussed in the path that it might make sense to have an Allocate and a Construct call, in which case, the current Allocate functionality might make sense. But the point is, we still have to think about this and design a solution, so I wouldn't add a todo to the code. (Maybe the first part is fine, but I wouldn't add what we want to do in the future.)

score/mw/com/impl/skeleton_event.h
```diff
@@ -275,14 +288,23 @@ class SkeletonEventView
...
-        return GetBinding()->GetLatestSample(quality_type);
+        auto latest_sample_result = GetBinding()->GetLatestSample(quality_type);
+        if (!latest_sample_result.has_value())
+        {
+            return MakeUnexpected<SamplePtr<SampleType>>(latest_sample_result.error());
+        }
+        // Interim rebind: GetLatestSample() is (skeleton-side) already type-erased at the binding level and returns
+        // SamplePtr<void>, while this (proxy-facing) API still needs to hand out a concrete SamplePtr<SampleType>.
+        // See the doxygen comment on SamplePtr's rebind ctor (score/mw/com/impl/plumbing/sample_ptr.h) for context;
+        // this rebind step goes away once SamplePtr no longer needs to be a class template.
```
bemerybmw, line 306: Link the proxy side ticket and add a todo

score/mw/com/impl/bindings/lola/event_data_storage.cpp
```diff
@@ -72,6 +72,8 @@ void EventDataStorage::InitializeSlots(const InitializeSampleCallback& initializ
     auto last_slot_ptr = type_erased_data_slots_ + decltype(type_erased_data_slots_)::difference_type(last_slot_offset);
     auto* last_slot_raw_ptr = last_slot_ptr.get();
 
+    // This is our low-level data storage, where we work on type-erased data, thus pointer arithmetic can't be avoided.
+    // NOLINTNEXTLINE(cppcoreguidelines-pro-bounds-pointer-arithmetic) see above
```
bemerybmw, line 78: As discussed, the underlying byte array should be a span? Then I think that we could avoid this suppression? Because it would be doing arithmetic on indices and then accessing the span which should not violate this rule?

score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
```diff
@@ -25,55 +25,33 @@
...
-    template <typename T>
     // coverity[autosar_cpp14_a11_3_1_violation] see above
     friend class SampleAllocateePtrMutableView;
 
   public:
-    using pointer = SampleType*;
-    using const_pointer = SampleType* const;
-    using element_type = SampleType;
+    using pointer = void*;
+    using const_pointer = void* const;
```
bemerybmw: Shouldn't this be const void*?

score/mw/com/impl/bindings/lola/service_data_storage.cpp
```diff
@@ -53,15 +53,12 @@ std::size_t CalculateServiceDataStorageShmSize(
     // (3) For each event/field (in the exact order it gets registered/offered): the EventDataStorage object plus its
     // data-slot-array (type_erased_data_slots_). The exact size/alignment of the slot-array (see
     // SkeletonMemoryManager::CreateEventDataInCreatedSharedMemory()) is provided by the caller.
+    // \ToDo see also comment in AddEventDataStorageShmSizeAllocation(): We should eventually hand down number_of_slots/
```
bemerybmw, line 56: Create / link a ticket

score/mw/com/impl/bindings/lola/skeleton_event.h
```diff
@@ -80,190 +86,109 @@ class SkeletonEvent final : public SkeletonEventBinding<SampleType>
...
+    /// "pending" type erased sample pointers which are created in PrepareOffer() and destroyed in
+    /// PrepareStopOffer()
+    /// - optional as only needed when tracing is enabled and when they haven't been cleaned up via a call to
+    /// PrepareStopOffer().
+    std::optional<TransactionLogRegistrationGuard> transaction_log_registration_guard_qm_{};
+    std::optional<TransactionLogRegistrationGuard> transaction_log_registration_guard_asil_b_{};
+    std::optional<tracing::TypeErasedSamplePtrsGuard> type_erased_sample_ptrs_guard_{};
+    std::optional<ReceiveHandlerRegistrationChangedCallback> receive_handler_registration_changed_callback_;
+
+    void UpdateCurrentTimestamp();
```
bemerybmw: Private functions should go before members

score/mw/com/impl/bindings/lola/skeleton_event.h
```diff
@@ -80,190 +86,109 @@ class SkeletonEvent final : public SkeletonEventBinding<SampleType>
...
+    void SetQmNotificationsRegistered(bool value);
+    void SetAsilBNotificationsRegistered(bool value);
+    void ResetGuards() noexcept;
+    EventDataControlComposite<>& GetEventDataControlComposite()
```
bemerybmw: Do we really need this function?

score/mw/com/impl/bindings/lola/skeleton_event.h
```diff
@@ -80,190 +86,109 @@ class SkeletonEvent final : public SkeletonEventBinding<SampleType>
...
+
+    /// \brief Atomic flags indicating whether any receive handlers are currently registered for this event
+    ///        at each quality level (QM and ASIL-B).
+    /// \details These flags are updated via callbacks from MessagePassingServiceInstance when handler
+    ///          registration status changes. They allow Send() to skip the NotifyEvent() call when no
+    ///          handlers are registered for a specific quality level, avoiding unnecessary lock overhead
+    ///          in the main path. Uses memory_order_relaxed as the flags are optimisation hints - false
+    ///          positives (thinking handlers exist when they don't) are harmless, and false negatives
+    ///          (missing handlers) are prevented by the callback mechanism.
+    std::atomic<bool> qm_event_update_notifications_registered_{false};
```
bemerybmw: All members should be initialised in initialiser list

score/mw/com/impl/bindings/lola/skeleton_event.cpp
```diff
@@ -11,3 +11,361 @@
...
+    if (event_data_control_composite_.has_value())
+    {
+        transaction_log_registration_guard_asil_b_.reset();
+        transaction_log_registration_guard_qm_.reset();
+    }
+}
+
+Result<void> SkeletonEvent::NotifyConsumersIfHandlersRegistered() noexcept
+{
+    // Only call NotifyEvent if there are any registered receive handlers for each quality level.
```
bemerybmw, line 318: Why doesn't the contents of this function go directly in Notify and we remove this function?

score/mw/com/impl/bindings/lola/skeleton_event.cpp
```diff
@@ -11,3 +11,361 @@
...
+    // This avoids the expensive lock operation in the common case where no handlers are registered.
+    // Using memory_order_relaxed is safe here as this is an optimisation, if we miss a very recent
```
bemerybmw: We're not using memory_order_relaxed?

score/mw/com/impl/bindings/mock_binding/sample_ptr.h
```diff
@@ -31,7 +33,36 @@ using CustomDeleter = score::cpp::callback<void(SampleType*)>;
...
+/// \note The original (void-based) deleter is heap-allocated behind a shared_ptr rather than captured by value in
+/// the new deleter lambda: SamplePtrCustomDeleter itself already uses (almost) the whole small-buffer-optimized
+/// capacity of score::cpp::callback, so capturing it by value would exceed the wrapping callback's capacity.
+template <typename SampleType, typename = std::enable_if_t<!std::is_void<SampleType>::value>>
+SamplePtr<SampleType> RebindSamplePtr(SamplePtr<void>&& other) noexcept
+{
+    auto deleter = std::make_shared<SamplePtrCustomDeleter<void>>(std::move(other.get_deleter()));
+    auto* const raw_pointer = static_cast<SampleType*>(other.release());
+    return SamplePtr<SampleType>{raw_pointer, [deleter](SampleType* pointer) mutable noexcept {
+                                     (*deleter)(static_cast<void*>(pointer));
```
bemerybmw: use std::invoke

bemerybmw (dismissed):

score/mw/com/impl/bindings/lola/skeleton_event.cpp
crimson11: I just did copy over this code from the GenericSkeletonEvent, where this was before - so I did not write this! ;)
But you are right - will remove.

score/mw/com/impl/bindings/lola/skeleton_event.cpp
crimson11: Which Notify? The `NotifyEvent` from our message_passing facade? Then no! This isn't possible for performance reasons - this was btw the reason this API has been introduced here ... if you check the original PR for it :P

score/mw/com/impl/skeleton_event_binding.h
crimson11: Was a "take-over" from the old sig. Fixed.

score/mw/com/impl/bindings/lola/event_data_storage.cpp
crimson11: No. We can't imho. a std::span always uses/relies on raw-pointers ... this is not shared-memory compatible ...
This change can be done later, when we have the EventDataSTorageLocalView.

score/mw/com/impl/skeleton_event_binding.h
crimson11: We have this ticket: https://jira.cc.bmwgroup.net/browse/SWP-281633

score/mw/com/impl/plumbing/sample_allocatee_ptr.h
crimson11: I decided for a static_assert, which checks, that at least sourcxe or target of the rebind is void!
I feel, this is good enough!

score/mw/com/impl/skeleton_event_binding.h
crimson11: method has been removed

score/mw/com/impl/skeleton_event_binding.h
crimson11: Added todo/ref to SWP-281633

score/mw/com/impl/plumbing/sample_ptr.h
crimson11: Imho no. Since we restrict intentionally the supported rebind-direction for our current use cases. So we need to distinguish between SampleType and OtherSampleType ....

score/mw/com/impl/skeleton_event.h
crimson11: fine for me. At least it is in our two heads ;)

score/mw/com/impl/bindings/lola/sample_allocatee_ptr.h
crimson11: You are basically right! It was already wrong in the beginning.
I changed that.
Sidenote: const_pointer is never used and it ius also NOT part of c++ pointer_traits! only allocator_traits have it ... so maybe we should remove at some point in time?

score/mw/com/impl/bindings/lola/skeleton_event.h
crimson11: It is called/used within prod.code. So I will leave it for now ... but it isn't used consistently in the codebase - sometimes in the code the same check, which this func does is repeated instead of using this func ... but this was all there already ... I can't fix every old bollocks now ;)

score/mw/com/impl/skeleton_event.h
bemerybmw: We're no longer doing a memcpy.

score/mw/com/impl/bindings/lola/skeleton_event.cpp
bemerybmw: I meant SkeletonEvent::Notify() which currently just dispatches to this function.

score/mw/com/impl/bindings/lola/event_data_storage.cpp
crimson11: Decided against that! Because then I need to do DEATH-test, which we wanted to avoid!

score/mw/com/impl/bindings/lola/skeleton_event.cpp
crimson11: No. It is als called by SkeletonEvent::Send() .... so we need this "extra" function.

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
```diff
@@ -359,19 +371,19 @@ using SkeletonEventSingleSlotComponentTestFixture = SkeletonEventComponentTestTe
...
     ASSERT_TRUE(prepare_offer_result.has_value());
 
     // Allocate a slot so that there are no free slots remaining
     const auto slot_result = skeleton_event_.Allocate(SampleAllocateeGuard{});
     ASSERT_TRUE(slot_result.has_value());
 
-    // When sending by value
-    const auto send_result = skeleton_event_.Send(5, {}, SampleAllocateeGuard{});
+    // When trying to allocate a further slot for sending by value
+    const auto allocate_result = skeleton_event_.Allocate(SampleAllocateeGuard{});
```
SebSparrowHawk, line 382: Maybe the test case could be renamed, since we are not calling Send anymore. Something like "AllocationReturnsErrorIfNoFreeSlotsRemaining"

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
```diff
@@ -300,14 +302,19 @@ TEST_F(SkeletonEventComponentTestFixture, CanSendByValue)
...
-    const auto prepare_offer_result = skeleton_event_.PrepareOffer();
+    const auto prepare_offer_result = skeleton_event_.PrepareOffer(std::nullopt);
     ASSERT_TRUE(prepare_offer_result.has_value());
 
     // store the number of free slots before sending ...
     auto free_slots_before = GetFreeSampleSlots();
 
-    // When  sending by value
-    std::ignore = skeleton_event_.Send(5, {}, SampleAllocateeGuard{});
+    // When sending by value
```
SebSparrowHawk, line 311: I think you want to use the new helper method "Result<void> SendValueByCopy(...)" from skeleton_event_test_resources.h here.

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
```diff
@@ -345,11 +352,16 @@ TEST_F(SkeletonEventComponentDeathTest, CallingSendWillTerminateWhenLolaRuntimeD
             .WillOnce(::testing::Return(nullptr));
 
         // Given that an event has been offered
-        const auto prepare_offer_result = skeleton_event_.PrepareOffer();
+        const auto prepare_offer_result = skeleton_event_.PrepareOffer(std::nullopt);
         ASSERT_TRUE(prepare_offer_result.has_value());
 
         // When sending by value
-        score::cpp::ignore = skeleton_event_.Send(5, {}, SampleAllocateeGuard{});
+        SkeletonEventSampleType new_value{5U};
```
SebSparrowHawk, line 359: I think you want to use the new helper method "Result<void> SendValueByCopy(...)" from skeleton_event_test_resources.h here.

score/mw/com/impl/bindings/lola/proxy.cpp
```diff
@@ -624,6 +624,30 @@ TransactionLogSet& Proxy::GetTransactionLogSet(const ElementFqId element_fq_id)
     return event_entry->second.transaction_log_set_;
 }
 
+const EventDataStorage& Proxy::GetEventDataStorage(const ElementFqId element_fq_id) const
```
SebSparrowHawk, line 627: There is no test for this new method in proxy_test.cpp

SebSparrowHawk (changes requested):

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
crimson11: Nope. But thanks to your comment I did check the test! It makes no sense anymore, since on binding layer we don't have a "SendByValue" anymore! This only exists on binding-independent layer, where the binding independent layer then transforms it to an AllocateAndSend, which is tested by the test case above.

The SendValueByCopy helper gets used in other tests (tracing). Here we want directly test lola::SkeletonEvent, so this "helper", which simulates old behaviour makes no sense.

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
crimson11: I did completely remove this test now! Reason: We don't need a "component test" here, because the exact same test exist in skeleton_event_test.cpp (SkeletonEventAllocateFixture, AllocateErrorLeadsToErrorResult=)

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
crimson11: Nope. But thanks for hinting at this test. I removed it ;)
Why, because it was WRONG since ages! This test (checked it) only terminatzes/dies, because of the PrepareOffer call being done, while GetBindingRuntime() return a nullptr! Not what the test says/did intend to test!

score/mw/com/impl/bindings/lola/proxy.cpp
crimson11: Added two tests for it.

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
SebSparrowHawk: Yes, that makes sense. It's the same test.

score/mw/com/impl/bindings/lola/test/skeleton_event_component_test.cpp
SebSparrowHawk: Okay, and that is tested in skeleton_test.cpp.

SebSparrowHawk (approved):

LittleHuba (approved):

PR #899, opened 2026-08-11
Add review checklists workflow

LittleHuba:
Adds a workflow to require acknowledgement of checklists by reviewers.

Currently no checklist is enforced.


BUILD
```diff
@@ -42,6 +42,7 @@ copyright_checker(
     name = "copyright",
     srcs = [
         ".github",
+        "actions",
```
castler: Why this change?

.github/workflows/review_checklists_apply.yml
```diff
@@ -0,0 +1,173 @@
...
+    types: [completed]
+
+permissions:
+  pull-requests: write
+  statuses: write
+  actions: read
+
+concurrency:
+  group: >-
+    review-checklists-apply-${{ github.event.workflow_run.head_sha || github.event.workflow_run.id }}
```
castler, line 84: I think we are missing here a "cancel-in-progress" option? Or do we say this is so short running, that it does not make sense?

tools/review-checklists/action.yml
```diff
@@ -0,0 +1,109 @@
...
+      run: python3 -m pip install -r "$GITHUB_ACTION_PATH/requirements.txt"
+
+    - name: Run post_checklists
+      if: inputs.action == 'post'
+      shell: bash
+      env:
+        GITHUB_TOKEN: ${{ inputs.github-token }}
+        GITHUB_REPOSITORY: ${{ github.repository }}
+        PR_NUMBER: ${{ inputs.pr-number }}
+      run: python3 "$GITHUB_ACTION_PATH/scripts/post_checklists.py" --config-path "${{ inputs.config-path }}"
```
castler, line 93: We should make a comment here that we we do not on purpose start this with Bazel, and what was our reasoning for it.

tools/review-checklists/action.yml
```diff
@@ -0,0 +1,109 @@
+# *******************************************************************************
```
castler, line 1: Wondering if we do not want to place it also under actions/00_infrastructure or similar?

tools/review-checklists/README.md
```diff
@@ -0,0 +1,244 @@
...
+   restricts `pull_request_target` to `branches: [main]`. Change this if
+   your default branch has a different name.
+4. **Require the commit status, not a workflow check run, in branch
+   protection.** Because the apply workflow only ever runs via
+   `workflow_run`, GitHub cannot bind it as a "required workflow" check
+   directly on the PR. Instead, configure your branch protection rule /
+   ruleset to require the commit status context **`review-checklists`**
+   (the one this action sets via `set_commit_status`). Requiring the
+   workflow job itself will not gate merges correctly.
+5. **Require at least one approving review** in branch protection. The
```
castler, line 142: Are we sure this is possible with Otterdog? As per https://otterdog.readthedocs.io/en/latest/reference/organization/repository/status-check/ I do not believe so?

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+    Raises:
+        FileNotFoundError: If the config file cannot be located.
+    """
+    # 1. Bazel runfiles via the bazel-runfiles library (Rlocation API).
+    try:
+        from runfiles import Runfiles  # type: ignore[import-untyped]
+
+        r = Runfiles.Create()
+        if r:
+            candidate = r.Rlocation(f"_main/{config_relpath}")
```
castler: IMHO, "_main" should be removed as it is a Bazel implementation detail that can change any point in time.

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+
+
+def get_approving_reviewers(pr: PullRequest) -> list[str]:
+    """Return a list of usernames who have an active APPROVED review."""
+    approvers = set()
+    for review in pr.get_reviews():
+        if review.state == "APPROVED":
+            approvers.add(review.user.login)
+        elif review.state in ("CHANGES_REQUESTED", "DISMISSED"):
+            approvers.discard(review.user.login)
```
castler, line 273: This is unnecessary, no? As the approvers is an empty set from the beginning. And it is not possible that the same person has approved and requested changes at the same point in time.

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+    threaded conversations where reviewers can reply with OK.
+    """
+    result = {}
+    for comment in pr.get_review_comments():
+        body = comment.body or ""
+        prefix = "<!-- review-checklist:"
+        if prefix in body:
+            start = body.index(prefix) + len(prefix)
+            end = body.index(" -->", start)
+            cid = body[start:end]
```
castler: Do we not have to check that `end` also exists? Otherwise we will get weird behavior if the markers are incomplete?

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+        if reply_to != checklist_comment_id:
+            continue
+        body = (comment.body or "").strip()
+        if body.upper() == OK_KEYWORD:
+            ok_replies.append(comment)
+    return ok_replies
+
+
+def _collect_acknowledgement_details(
+    pr: PullRequest, existing_comments: dict[str, Any], relevant_ids: list[str]
```
castler: `relevant_ids` is a bad naming. What is it relevant for? Is it the checklist comments, the OK comments? what other id?

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+
+def set_commit_status(
+    repo: Any,
+    sha: str,
+    state: str,
+    description: str,
+    context: str = "review-checklists",
+) -> None:
+    """Set a commit status on the given SHA."""
+    desc = description[:140]
```
castler: magic number

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+        "<details>",
+        "<summary>Checklist Report (do not modify)</summary>",
+        "",
+        "## Review Checklist Evidence",
+        "",
+        f"**Last updated:** {datetime.now(timezone.utc).isoformat()}",
+        "",
+    ]
+
+    for cl in relevant:
```
castler: for reading in this whole file - it would be beneficial to write `checklist` instead of `cl`.
Maybe I am getting to old, but the brain capacity to think each time what `cl` means is quite exhausting.

Maybe for future PRs.

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+        cid = cl["id"]
```
castler: `cid` is it the comment or checklist id?

I understand that it is here the checklist id, but just wanted to give my previous point some more evidence.

tools/review-checklists/scripts/helpers.py
```diff
@@ -0,0 +1,529 @@
...
+        result.get("data", {})
+        .get("repository", {})
+        .get("pullRequest", {})
+        .get("isInMergeQueue")
+    )
+    if isinstance(is_in_queue, bool):
+        return is_in_queue
+
+    print("GraphQL merge-queue lookup returned no boolean state")
+    return False
```
castler, line 465: Should we not throw here and rather catch that. Are we sure that IF we would be in the merge queue but the lookup fails for what ever reason (e.g. API error on the graphQL call) that `FALSE` would not bring us problems?

tools/review-checklists/scripts/check_acknowledgements.py
```diff
@@ -0,0 +1,261 @@
...
+        set_commit_status(repo, head_sha, state, f"Merge queue: {description}")
+        if state != "success":
+            print(f"Merge-queue checklist evidence validation failed: {description}")
+            sys.exit(1)
+        print("Merge-queue checklist evidence validated ✅")
+        return
+
+    repo, pr = get_repo_and_pr(gh)
+
+    if is_pr_in_merge_queue(pr):
```
castler, line 196: This I do not get, why do we not execute this in the if statement above?

> if event_name == "merge_group":

then the PR needs to be in the merge_queue, no?

tools/review-checklists/scripts/check_acknowledgements.py
```diff
@@ -0,0 +1,261 @@
...
+from __future__ import annotations
+
+import argparse
+import json
+import os
+import sys
+from typing import Any
+
+from helpers import (
+    _collect_acknowledgement_details,
```
castler: I personally think its weird to "import" private functions.

tools/review-checklists/scripts/check_acknowledgements.py
```diff
@@ -0,0 +1,261 @@
...
+    point.
+    """
+    changed_files = get_changed_files(pr)
+    relevant = match_checklists(checklists, changed_files)
+
+    if not relevant:
+        return "success", "No checklists applicable"
+
+    existing = find_existing_checklist_comments(pr)
+    relevant_ids = [cl["id"] for cl in relevant if cl["id"] in existing]
```
castler: As per my current understanding, this would filter out only checklists where a comment has been made. Thus, if for a checklist no comment has been made (yet), we would skip it here.

Thus we would set success later, if for another checklist is succesful. Do we not rather have to "pend" on that all comments have been posted?

tools/review-checklists/scripts/check_acknowledgements.py
```diff
@@ -0,0 +1,261 @@
...
+            continue
+
+        cid = cl_comment_ids[reply_to]  # type: ignore[index]
+        body = (comment.body or "").strip()
+        user = comment.user.login
+
+        if body.upper() == OK_KEYWORD:
+            acks[cid].add(user)
+
+    return acks
```
castler: This functionality is duplicated.

See for example: `def _find_ok_comments_for_checklist`

I think we can extract it into an helper. (I am aware that we store different information, but I think we can make a super set of information returned?

castler (commented):
We need to extend the CI design also to explain why and for what we have these workflows.

Due to the size, let's discuss f2f which of the findings we tackle, tackle now or in a follow up.

BUILD
LittleHuba: Fixed, code was moved and this was not updated.

.github/workflows/review_checklists_apply.yml
LittleHuba: Adding the cancel-in-progress does not hurt. So I added it. But in general this is quite fast (given GitHub schedules the runs fast).

tools/review-checklists/action.yml
LittleHuba: Added comment.

tools/review-checklists/action.yml
LittleHuba: Not quite sure about that. Separating the code over the repo is not perfect, since this file contains logic that is closely related. GitHub actually has no convention where actions must be placed. So keeping it here is perfectly fine.

tools/review-checklists/README.md
LittleHuba: Commit status checks are viable with `any:` prefix. I extended the documentation. Of course this is subject to trials.

tools/review-checklists/scripts/helpers.py
LittleHuba: Fixed

tools/review-checklists/scripts/helpers.py
LittleHuba: No this is actually necessary. You get the full history. If a reviewer first acknowledged and then requested changes, both reviews will be in this list.

tools/review-checklists/scripts/helpers.py
LittleHuba: Fixed.

tools/review-checklists/scripts/helpers.py
LittleHuba: Fixed

tools/review-checklists/scripts/helpers.py
LittleHuba: Fixed.

tools/review-checklists/scripts/helpers.py
LittleHuba: This branch is only used to provide the warning comment if a PR is modified after entering the merge queue. The worst happening here would be that the comment is wrongly emitted.
Still this omission would not modify the evidence.

tools/review-checklists/scripts/check_acknowledgements.py
LittleHuba: We want to add this warning comment when the PR is modified after it enters the merge queue. This means we react to events like pull_request_comment, pull_request_review.
Hence, we do not enter the first if condition at all but must still verify whether the PR is in the merge queue or not.

tools/review-checklists/scripts/check_acknowledgements.py
LittleHuba: Fixed.

tools/review-checklists/scripts/check_acknowledgements.py
LittleHuba: Fixed

tools/review-checklists/scripts/check_acknowledgements.py
LittleHuba: Fixed

tools/review-checklists/scripts/helpers.py
LittleHuba: Fixed

tools/review-checklists/scripts/helpers.py
LittleHuba: Fixed

LittleHuba:
I'll update the CI design tomorrow.

castler (approved):

PR #905, opened 2026-08-11
fix: Allow sized C-like enums in Rust

NEOatNHNG:
The equivalent of enum class: size_type on C++

### DIFF
--- a/score/mw/com/rust/score_com_macros/lib.rs
+++ b/score/mw/com/rust/score_com_macros/lib.rs
@@ -223,7 +223,8 @@ fn check_attrs(fields: &Fields) -> Result<(), syn::Error> {
 ///
 /// # Validation performed by this macro
 ///
-/// - **`#[repr(C)]`** must be present on the type.
+/// - **`#[repr(C)]`** must be present on the type. For C-like enums, a primitive
+///   integer repr (e.g. `#[repr(u8)]`) is accepted as well.
 /// - It must be a struct, tuple struct, or C-like enum (no unions or non-C-like enums).
 ///
 /// These checks are intentionally kept in `Reloc` rather than `CommData` so that each macro
@@ -237,7 +238,10 @@ pub fn derive_reloc(input: TokenStream) -> TokenStream {
     if !has_repr_c(&input_args.attrs) {
         return syn::Error::new_spanned(
             ident_name,
-            "The #[derive(Reloc)] macro requires #[repr(C)] on the type",
+            concat!(
+                "The #[derive(Reloc)] macro requires #[repr(C)] or a primitive integer repr ",
+                "(e.g. #[repr(u8)]) on the type"
+            ),
         )
         .to_compile_error()
         .into();
@@ -260,8 +264,10 @@ pub fn derive_reloc(input: TokenStream) -> TokenStream {
         Err(()) => {
             return syn::Error::new_spanned(
                 ident_name,
-                "The #[derive(Reloc)] macro is supported only for enums(C like), structs and \
-                 tuple structs",
+                concat!(
+                    "The #[derive(Reloc)] macro is supported only for enums(C like), structs ",
+                    "and tuple structs"
+                ),
             )
             .to_compile_error()
             .into()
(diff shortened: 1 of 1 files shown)


score/mw/com/rust/score_com_macros/lib.rs
```diff
@@ -260,8 +264,10 @@ pub fn derive_reloc(input: TokenStream) -> TokenStream {
         Err(()) => {
             return syn::Error::new_spanned(
                 ident_name,
-                "The #[derive(Reloc)] macro is supported only for enums(C like), structs and \
-                 tuple structs",
+                concat!(
+                    "The #[derive(Reloc)] macro is supported only for enums(C like), structs ",
+                    "and tuple structs"
+                ),
             )
```
bharatGoswami8, line 271: why changed ?

score/mw/com/rust/score_com_macros/lib.rs
```diff
@@ -293,16 +299,45 @@ fn create_bounds_with_reloc(mut generics: Generics) -> Generics {
     generics
 }
 
-/// Check for #[repr(C)] existence
+/// Check whether the type is representable as C struct or enum.
 fn has_repr_c(attrs: &[syn::Attribute]) -> bool {
```
bharatGoswami8, line 303:
```suggestion
fn has_repr_c_and_primitive(attrs: &[syn::Attribute]) -> bool {
```

score/mw/com/rust/score_com_macros/lib.rs
```diff
@@ -293,16 +299,45 @@ fn create_bounds_with_reloc(mut generics: Generics) -> Generics {
...
+            }
+
+            // Unrecognized modifier (e.g. align(4), packed(2)); consume any
+            // parenthesized arguments so parsing succeeds, then ignore it.
+            if meta.input.peek(syn::token::Paren) {
+                let content;
+                syn::parenthesized!(content in meta.input);
+                let _ = content;
             }
+            Ok(())
```
bharatGoswami8, line 336: I think padding should be handled by ABI, possibly through some kind of code generation tool. If we allow padding to be managed on the Rust side, but other applications or components do not apply the exact same rules, it could lead to compatibility issues.

I would recommend making #`[repr(packed)] and #[repr(packed(N))]` a compilation error when user put this.

bharatGoswami8: I have same opinion about `align ` as well

score/mw/com/rust/score_com_macros/lib.rs
NEOatNHNG: Because otherwise the error log contains strange white spaces. This was just "by-catch", I can remove it if you want.

score/mw/com/rust/score_com_macros/lib.rs
NEOatNHNG: OK, I can also only allow `repr(C)` and `repr(inttype)` and forbid everything else if that is what we want. That effectively blocks a user from using align and packed on the data structure if using Rust while still being allowed on the C++ side. And this did work in the code before this PR so it would remove a "feature".

score/mw/com/rust/score_com_macros/lib.rs
bharatGoswami8: No , it is fine, i was just asking what was the issue.

score/mw/com/rust/score_com_macros/lib.rs
bharatGoswami8: My concern is that alignment and padding should be managed from a single source of truth for both C++ and Rust. If you look at the Rust examples, we write the corresponding C++ structure as well for FFI layer translation. If a user applies packing or custom alignment only on the Rust side and forgets to apply the same settings in C++, it can easily result in layout mismatches and misalignment issues.

So, if we want to support this feature, we need to ensure that both the Rust and C++ definitions are generated or configured with identical alignment requirements.

We can enable this feature, but we need a mechanism to guarantee consistency across both sides. A Rust macro can validate attributes present on the Rust type, but currently we do not have a way to verify that the equivalent C++ structure has been defined with the same alignment characteristics. Without such validation, we risk introducing subtle interoperability bugs.

bharatGoswami8: On the second thought, I think when an `IDL` tool is introduced, this issue will be resolved. So maybe we can go with this, but we need to update the documentation and introduce one type like `Temperature` in the example with alignment and padding, so that we can demonstrate the assumptions of use on both the C++ and Rust sides. 
What is your opinion on this?

bharatGoswami8:
As discussed in last week's score communication weekly meeting, the changes in this PR require corresponding documentation and example updates.

PR #921, opened 2026-08-13
feat(gateway/transport_layer): qemu transport layer

Rahul-Sutariya:
Adds a second `Transport` implementation for the LoLa gateway, selectable via
`transport-layer.id: "qemu_hypervisor"` in `mw_com_gateway_config.json`. It
reuses the sample transport's TCP-based `BidirectionalTransport` for the
message control plane and adds a QEMU `ivshmem-plain` shared-memory BAR for
the zero-copy data plane, so cross-VM service data no longer needs to be
copied over the socket.

Transport layer (score/mw/com/gateway/transport_layer/qemu/):
- `QemuHypervisorTransport` — orchestrates the message layer and the ivshmem
  memory layer. On `ProvideService`, binds the peer's inter-VM shm at the BAR
  offset it published, before handing off to `GatewayCore`.
- `ResolveInterVmShmPaths` / `GetInterVmShmSizes` — resolve and size the
  `/intervm-shared-shmem/...` paths that `SharedMemoryFactory` routes through
  the new inter-VM memory provider.

ivshmem support (.../qemu/ivshmem/):
- `IvshmemTypedMemoryProvider` — a `TypedMemory` implementation backing named
  shm objects with sub-ranges of the ivshmem BAR. Maintains a BAR-resident
  allocation directory (name -> offset/size) visible to both VMs, so either
  side can allocate (`AllocateNamedTypedMemory`) or bind at a peer-discovered
  offset (`AllocateNamedTypedMemoryAtOffset`) without a partitioning scheme.
  Directory entries are hashed (FNV-1a) with the name stored alongside for
  collision safety, and access is serialized with a cross-VM spinlock.
- `DiscoverIvshmemBar` — locates the QEMU `ivshmem-plain` PCI device
  (vendor 0x1af4 / device 0x1110) via QNX pci-server and returns the physical
  address/size of the configured BAR (default BAR2, configurable).

Configuration (.../qemu/configuration/):
- `QemuTransportConfiguration` / `ParseQemuTransportConfig` — parse the
  transport config file's `ivshmem.preferred-bar-num` on top of the existing
  hypervisor-socket fields shared with the sample transport.

Wiring:
- `TransportFactory::Create` gains the `qemu_hypervisor` branch: parses the
  QEMU config, discovers the BAR, constructs the `IvshmemTypedMemoryProvider`,
  registers it as the process's inter-VM memory provider, and builds the
  transport with a real `BidirectionalTransport`.
- BUILD/visibility updates so `transport_layer:transport_factory` and
  `gateway_application` can depend on the new qemu/ivshmem targets.
- README updated to document the `qemu_hypervisor` transport id and its
  `ivshmem.preferred-bar-num` config field.

Tests:
- Unit tests for the transport (message dispatch, ProvideService/OfferService/
  notification round-trips, PreCreateInterVmSharedMemory offset binding),
  the ivshmem provider (allocation, directory read/write, BAR-exhaustion,
  offset conflicts), and the BAR discovery/config parser.
- `quality/coverage/coverage_justifications.yaml` entries for QNX-hardware-only
  paths (real PCI/ivshmem device access, x86_64-only shm_ctl_special fallback)
  that cannot be exercised by host-side coverage runs.

Platform support: full ivshmem functionality is QNX-only; on other platforms
`DiscoverIvshmemBar`/the provider's allocation paths return ENOSYS/false so the
transport still builds and links but is non-functional.

### DIFF
--- a/score/mw/com/gateway/transport_layer/qemu/configuration/qemu_transport_config_parser_test.cpp
+++ b/score/mw/com/gateway/transport_layer/qemu/configuration/qemu_transport_config_parser_test.cpp
@@ -1,0 +1,87 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/gateway/transport_layer/qemu/configuration/qemu_transport_config_parser.h"
+
+#include "score/json/json_parser.h"
+
+#include <gtest/gtest.h>
+
+#include <utility>
+
+namespace score::mw::com::gateway::qemu
+{
+namespace
+{
+
+using score::json::operator""_json;
+
+TEST(QemuTransportConfigParserTest, UsesDefaultIvshmemBarNumWhenMissing)
+{
+    auto json = R"JSON(
+    {
+      "hypervisor-socket": {
+        "remote-ip": "10.0.2.2",
+        "local-port": 45001,
+        "remote-port": 45002
+      }
+    }
+    )JSON"_json;
+
+    const auto config = ParseQemuTransportConfig(std::move(json));
+    EXPECT_EQ(config.GetPreferredIvshmemBarNum(), ivshmem::kDefaultIvshmemBarNum);
+}
+
+TEST(QemuTransportConfigParserTest, ReadsPreferredIvshmemBarNum)
+{
+    auto json = R"JSON(
+    {
+      "hypervisor-socket": {
+        "remote-ip": "10.0.2.2",
+        "local-port": 45001,
+        "remote-port": 45002
+      },
+      "ivshmem": {
+        "preferred-bar-num": 3
+      }
+    }
+    )JSON"_json;
+
+    const auto config = ParseQemuTransportConfig(std::move(json));
+    EXPECT_EQ(config.GetPreferredIvshmemBarNum(), 3U);
+}
+
+TEST(QemuTransportConfigParserTest, UsesDefaultWhenPreferredBarNumIsNotInteger)
+{
+    // Given a config where preferred-bar-num is a non-integer value
+    auto json = R"JSON(
+    {
+      "hypervisor-socket": {
+        "remote-ip": "10.0.2.2",
+        "local-port": 45001,
+        "remote-port": 45002
+      },
+      "ivshmem": {
+        "preferred-bar-num": "not-a-number"
+      }
+    }
+    )JSON"_json;
+
+    // When parsing the config
+    const auto config = ParseQemuTransportConfig(std::move(json));
+
+    // Then the parser silently falls back to the default bar num instead of crashing or throwing
+    EXPECT_EQ(config.GetPreferredIvshmemBarNum(), ivshmem::kDefaultIvshmemBarNum);
+}
+
+}  // namespace
+}  // namespace score::mw::com::gateway::qemu
--- a/score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
+++ b/score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
@@ -1,0 +1,145 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
+ ********************************************************************************/
+#include "score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.h"
+
+#include "score/mw/log/logging.h"
+
+#include <cerrno>
+#include <cstring>
+#include <vector>
+
+#if defined(__QNXNTO__)
+extern "C" {
+#include <pci/pci.h>
+}
+#include <sys/neutrino.h>  // ThreadCtl, _NTO_TCTL_IO
+#endif
+
+namespace score::mw::com::gateway::qemu::ivshmem
+{
+
+std::optional<IvshmemBar> SelectIvshmemBar(const std::vector<IvshmemBar>& bars, std::uint32_t required_bar_num) noexcept
+{
+    for (const auto& bar : bars)
+    {
+        if (bar.is_memory && bar.bar_num == required_bar_num)
+        {
+            return bar;
+        }
+    }
+    return std::nullopt;
+}
+
+#if defined(__QNXNTO__)
+
+// COV_JUSTIFIED_START ivshmem-bar-discovery-hardware-path
+bool DiscoverIvshmemBar(std::uint64_t& paddr, std::uint64_t& size, std::uint32_t required_bar_num) noexcept
+{
+    constexpr pci_vid_t kVendor = 0x1af4U;
+    constexpr pci_did_t kDevice = 0x1110U;
+
+    if (::ThreadCtl(_NTO_TCTL_IO, nullptr) == -1)
+    {
+        ::score::mw::log::LogError() << "ThreadCtl(_NTO_TCTL_IO) failed: " << std::strerror(errno);
+        return false;
+    }
+
+    const pci_bdf_t bdf = pci_device_find(0U, kVendor, kDevice, PCI_CCODE_ANY);
+    if (bdf == PCI_BDF_NONE)
+    {
+        ::score::mw::log::LogError() << "ivshmem PCI device " << static_cast<unsigned int>(kVendor) << ":"
+                                     << static_cast<unsigned int>(kDevice) << " not found";
+        return false;
+    }
+
+    pci_err_t err = PCI_ERR_OK;
+    const pci_devhdl_t hdl = pci_device_attach(bdf, pci_attachFlags_OWNER, &err);
+    if (hdl == nullptr || err != PCI_ERR_OK)
+    {
+        ::score::mw::log::LogError() << "pci_device_attach failed (err=" << static_cast<int>(err) << ")";
+        return false;
+    }
+
+    constexpr pci_cmd_t kMemSpaceEnable = static_cast<pci_cmd_t>(1U << 1);
+    constexpr pci_cmd_t kBusMasterEnable = static_cast<pci_cmd_t>(1U << 2);
+    pci_cmd_t cmd = 0U;
+    err = pci_device_read_cmd(bdf, &cmd);
+    if (err != PCI_ERR_OK)
+    {
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+    pci_cmd_t cmd_set = 0U;
+    err = pci_device_write_cmd(hdl, static_cast<pci_cmd_t>(cmd | kMemSpaceEnable | kBusMasterEnable), &cmd_set);
+    if (err != PCI_ERR_OK)
+    {
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+
+    pci_ba_t ba[7];
+    int_t nba = static_cast<int_t>(sizeof(ba) / sizeof(ba[0]));
+    err = pci_device_read_ba(hdl, &nba, ba, pci_reqType_e_UNSPECIFIED);
+    if (err != PCI_ERR_OK)
+    {
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+
+    std::vector<IvshmemBar> bars;
+    bars.reserve(static_cast<std::size_t>(nba));
+    for (int_t i = 0; i < nba; ++i)
+    {
+        bars.push_back(IvshmemBar{static_cast<std::uint32_t>(ba[i].bar_num),
+                                  static_cast<std::uint64_t>(ba[i].addr),
+                                  static_cast<std::uint64_t>(ba[i].size),
+                                  ba[i].type == pci_asType_e_MEM});
+    }
+    for (const auto& bar : bars)
+    {
+        ::score::mw::log::LogInfo() << "ivshmem: BAR" << static_cast<int>(bar.bar_num)
+                                    << " type=" << (bar.is_memory ? "MEM" : "IO") << " addr=0x"
+                                    << static_cast<std::uint64_t>(bar.addr) << " size=0x"
+                                    << static_cast<std::uint64_t>(bar.size);
+    }
+
+    const auto shared = SelectIvshmemBar(bars, required_bar_num);
+    if (!shared.has_value())
+    {
+        ::score::mw::log::LogError() << "required ivshmem BAR" << static_cast<unsigned int>(required_bar_num)
+                                     << " not found on the ivshmem device";
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+
+    ::score::mw::log::LogInfo() << "ivshmem: BAR" << static_cast<int>(shared->bar_num) << " paddr=0x"
+                                << static_cast<std::uint64_t>(shared->addr) << " size=0x"
+                                << static_cast<std::uint64_t>(shared->size);
+
+    paddr = static_cast<std::uint64_t>(shared->addr);
+    size = static_cast<std::uint64_t>(shared->size);
+    return true;
+}
+// COV_JUSTIFIED_STOP
+
+#else
+
+bool DiscoverIvshmemBar(std::uint64_t& /*paddr*/, std::uint64_t& /*size*/, std::uint32_t /*required_bar_num*/) noexcept
+{
+    ::score::mw::log::LogError() << "DiscoverIvshmemBar is only supported on QNX";
+    return false;
+}
+
+#endif
+
+}  // namespace score::mw::com::gateway::qemu::ivshmem
--- a/score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
+++ b/score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
@@ -1,0 +1,489 @@
+/********************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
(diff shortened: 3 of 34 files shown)


score/mw/com/gateway/transport_layer/qemu/configuration/qemu_transport_config_parser_test.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+namespace score::mw::com::gateway::qemu
+{
+namespace
+{
+
+using score::json::operator""_json;
+
+TEST(QemuTransportConfigParserTest, UsesDefaultIvshmemBarNumWhenMissing)
+{
```
SebSparrowHawk, line 29: You could add the //Given... //When... //Then comments to the first two test cases

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
```diff
@@ -0,0 +1,145 @@
...
+
+#if defined(__QNXNTO__)
+
+// COV_JUSTIFIED_START ivshmem-bar-discovery-hardware-path
+bool DiscoverIvshmemBar(std::uint64_t& paddr, std::uint64_t& size, std::uint32_t required_bar_num) noexcept
+{
+    constexpr pci_vid_t kVendor = 0x1af4U;
+    constexpr pci_did_t kDevice = 0x1110U;
+
+    if (::ThreadCtl(_NTO_TCTL_IO, nullptr) == -1)
```
SebSparrowHawk, line 42: Please add a comment what this command is doing

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
```diff
@@ -0,0 +1,145 @@
...
+    }
+    pci_cmd_t cmd_set = 0U;
+    err = pci_device_write_cmd(hdl, static_cast<pci_cmd_t>(cmd | kMemSpaceEnable | kBusMasterEnable), &cmd_set);
+    if (err != PCI_ERR_OK)
+    {
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+
+    pci_ba_t ba[7];
```
SebSparrowHawk: Please explain where the magic number 7 is coming from.

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
```diff
@@ -0,0 +1,145 @@
...
+    {
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+
+    std::vector<IvshmemBar> bars;
+    bars.reserve(static_cast<std::size_t>(nba));
+    for (int_t i = 0; i < nba; ++i)
+    {
+        bars.push_back(IvshmemBar{static_cast<std::uint32_t>(ba[i].bar_num),
```
SebSparrowHawk: I'm wondering if instead of collecting all bars and later on filtering for the relevant one with "SelectIvshmemBar(...)" later, you could simply filter in this loop and only keep the one that is relevant? This could simplify the code.

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
```diff
@@ -0,0 +1,489 @@
...
+    const int rc_special = ::shm_ctl_special(
+        fd, SHMCTL_PHYS | SHMCTL_HAS_SPECIAL, sub_paddr, alloc_size, static_cast<unsigned>(X86_64_PTE_MTYPE_WB));
+    if (rc_special != -1)
+    {
+        std::ignore = ::close(static_cast<int>(fd));
+        return {};
+    }
+    // COV_JUSTIFIED_STOP
+#endif
+    const auto shm_ctl_result = mman_qnx_->shm_ctl(fd, SHMCTL_PHYS, sub_paddr, alloc_size);
```
SebSparrowHawk, line 335: You could maybe add a short comment what this command is doing, but maybe it's just me ;)

score/mw/com/gateway/transport_layer/qemu/test/app2_main.cpp
```diff
@@ -0,0 +1,364 @@
...
+/// Here we use a minimal version: the skeleton writes data, then increments event_count.
+/// The proxy polls event_count to know when data is ready.
+struct ServiceControl
+{
+    volatile std::uint32_t event_count;  // incremented by provider after writing DATA
+    volatile std::uint32_t verified;     // set by consumer after successful read
+};
+
+/// Polls a volatile uint32 until it reaches the expected value or times out.
+bool WaitForCtrl(const volatile std::uint32_t& field, std::uint32_t expected, int timeout_ms = 60000)
```
SebSparrowHawk: this function and the test stub definitions that follow are duplicates of the ones in app1_main.cpp

score/mw/com/gateway/transport_layer/qemu/test/app1_main.cpp
```diff
@@ -0,0 +1,366 @@
...
+                         field,
+                         expected);
+            std::fflush(stderr);
+        }
+    }
+    return field >= expected;
+}
+
+/// Test IBidirectionalTransport stub — captures the message handler for injection.
+class TestBidirectionalTransport final : public IBidirectionalTransport
```
SebSparrowHawk: Is there a reason why you defined stubs here instead of using the socket based transport?

SebSparrowHawk (changes requested):

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
Rahul-Sutariya: Good point, refactored it — I kept the collect-then-select split intentionally: the code logs every discovered BAR for diagnostics before filtering, now handled batter, Thanks.

score/mw/com/gateway/transport_layer/qemu/test/app1_main.cpp
Rahul-Sutariya: Good point, this integration test is actually validating: cross-VM ivshmem shared-memory visibility (offset discovery via the BAR directory, binding CTRL/DATA at the right physical sub-range, and correct payload read-back). but for E2E test i think it's better i use real BidirectionalTransport, will update the pr with the new changes, Thank you.

score/mw/com/gateway/transport_layer/qemu/configuration/qemu_transport_config_parser_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+      "ivshmem": {
+        "preferred-bar-num": "not-a-number"
+      }
+    }
+    )JSON"_json;
+
+    // When parsing the config
+    const auto config = ParseQemuTransportConfig(std::move(json));
+
+    // Then the parser silently falls back to the default bar num instead of crashing or throwing
```
SebSparrowHawk, line 90: I'm wondering if there should at least be a warning in this case that the value will be ignored (the log does not have to be tested here). 
If this wasn't a configuration for tests only, I would not use the default value in this case but terminate, but I think it could be okay for this integration test configuration.

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
```diff
@@ -0,0 +1,144 @@
...
+    {
+        (void)pci_device_detach(hdl);
+        return false;
+    }
+
+    // Scan for the required BAR
+    std::optional<IvshmemBar> shared;
+    for (int_t i = 0; i < nba; ++i)
+    {
+        const IvshmemBar bar{static_cast<std::uint32_t>(ba[i].bar_num),
```
SebSparrowHawk, line 98: With the new (simpler) code structure I'm having my doubts if we really need the IvshmemBar structure anymore and create instances of it. Because the structure is only used within this method and only short to hold the value, of which you are only interested in one. 
You could simply check against the bar.type and its number and then store paddr and size directly. What do you think?

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
```diff
@@ -0,0 +1,527 @@
...
+                                             static_cast<std::int64_t>(dir_paddr));
+    if (!mmap_result.has_value())
+    {
+        return nullptr;
+    }
+    directory_map_ = mmap_result.value();
+    return mmap_result.value();
+}
+
+std::optional<std::uint64_t> LookupOffsetInDirectoryLocked(const std::string& shm_name,
```
SebSparrowHawk, line 170: small change: I would name that method LookupOffsetInLockedDirectory

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
```diff
@@ -0,0 +1,840 @@
...
+
+    // Expecting shm_open to be called and return error
+    EXPECT_CALL(*raw, shm_open(::testing::_, ::testing::_, ::testing::_))
+        .WillOnce(::testing::Return(score::cpp::make_unexpected(score::os::Error::createFromErrno(ENOMEM))));
+
+    // When calling BindShmToBar
+    const auto result = p.BindShmToBar(4096U, "/test_shm", 0U);
+
+    // Then the operation fails
+    EXPECT_FALSE(result.has_value());
```
SebSparrowHawk, line 193: If the error codes are different, checking for error codes makes sense

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
```diff
@@ -0,0 +1,840 @@
...
+
+TEST(IvshmemTypedMemoryProviderBindShmToBarTest, SucceedsWhenShmOpenAndShmCtlSucceed)
+{
+    // Given mock where shm_open and shm_ctl succeed
+    auto mman_mock = std::make_unique<::testing::StrictMock<score::os::qnx::MmanQnxMock>>();
+    auto* const raw = mman_mock.get();
+    TestableIvshmemTypedMemoryProvider p{0x100000U, 1024U * 1024U, std::move(mman_mock)};
+
+    // Expecting shm_open and shm_ctl calls to succeed
+    EXPECT_CALL(*raw, shm_open(::testing::_, ::testing::_, ::testing::_)).WillOnce(::testing::Return(std::int32_t{-2}));
```
SebSparrowHawk: that negative return value is actually confusing. I would return a positive fd value if successful.

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
```diff
@@ -0,0 +1,840 @@
...
+                     ::testing::_,
+                     ::testing::_))
+        .WillOnce(::testing::Return(static_cast<void*>(buf.data())));
+
+    // When attempting to allocate
+    const auto result =
+        p.AllocateNamedTypedMemory(4096U, "/test_shm", score::memory::shared::permission::WorldWritable{});
+
+    // Then allocation fails due to insufficient space
+    EXPECT_FALSE(result.has_value());
```
SebSparrowHawk, line 542: Test case name contains error code but not checked here

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
```diff
@@ -0,0 +1,840 @@
...
+        p.AllocateNamedTypedMemory(4096U, "/test_shm", score::memory::shared::permission::WorldWritable{});
+
+    // Then allocation succeeds and is cached for future lookups
+    EXPECT_TRUE(result.has_value());
+    EXPECT_TRUE(p.GetAllocationOffset("/test_shm").has_value());
+}
+
+// This covers WriteDirectoryEntry's count-limit guard; the allocation still succeeds via BindShmToBar.
+
+TEST(IvshmemQnxAllocateNamedTest, FullDirectoryStillSucceedsViaBindShmToBar)
```
SebSparrowHawk, line 577: I have the feeling the test case name is not really explaining what it is actually testing.

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
```diff
@@ -0,0 +1,527 @@
...
+        return;
+    }
+
+    auto* header = AsDirectoryHeader(dir);
+    auto* entries = GetDirectoryEntries(header);
+
+    const std::uint32_t count = header->count;
+    if (count >= kMaxDirectoryEntries)
+    {
+        ::score::mw::log::LogError() << "IvshmemTypedMemoryProvider: directory full (" << count << " entries)";
```
SebSparrowHawk, line 209: I'm having my doubts about not returninig an error value in this case (or in general from the whole method). 
Callers should know that the entry could not be written, but for example AllocateNamedTypedMemory(...) continues even in an error case. But the other VM could never detect this entry in that case.

score/mw/com/gateway/transport_layer/qemu/qemu_hypervisor_transport.cpp
```diff
@@ -0,0 +1,450 @@
...
+        {
+            ::score::mw::log::LogError() << "PreCreateInterVmSharedMemory: failed to bind CTRL shm to BAR for "
+                                         << specifier.ToString() << " at offset " << ctrl_offset.value()
+                                         << ", OS error " << ctrl_result.error().GetOsDependentErrorCode();
+            return false;
+        }
+    }
+    else if (shm_control_size > 0U)
+    {
+        ::score::mw::log::LogWarn() << "PreCreateInterVmSharedMemory: CTRL offset not found in directory for "
```
SebSparrowHawk, line 415: From my perspective this warning is not enough and the method should in this case also return false. Because no further operation makes sense if you could not allocate the memory.

score/mw/com/gateway/transport_layer/qemu/qemu_hypervisor_transport.cpp
```diff
@@ -0,0 +1,450 @@
...
+        {
+            ::score::mw::log::LogError() << "PreCreateInterVmSharedMemory: failed to bind DATA shm to BAR for "
+                                         << specifier.ToString() << " at offset " << data_offset.value()
+                                         << ", OS error " << data_result.error().GetOsDependentErrorCode();
+            return false;
+        }
+    }
+    else if (shm_data_size > 0U)
+    {
+        ::score::mw::log::LogWarn() << "PreCreateInterVmSharedMemory: DATA offset not found in directory for "
```
SebSparrowHawk, line 437: Same here

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
```diff
@@ -0,0 +1,144 @@
...
+    const pci_bdf_t bdf = pci_device_find(0U, kVendor, kDevice, PCI_CCODE_ANY);
+    if (bdf == PCI_BDF_NONE)
+    {
+        ::score::mw::log::LogError() << "ivshmem PCI device " << static_cast<unsigned int>(kVendor) << ":"
+                                     << static_cast<unsigned int>(kDevice) << " not found";
+        return false;
+    }
+
+    pci_err_t err = PCI_ERR_OK;
+    const pci_devhdl_t hdl = pci_device_attach(bdf, pci_attachFlags_OWNER, &err);
```
SebSparrowHawk, line 57: for an actual Lola test case we would need pci_attachFlags_MULTI_OWNER instead of pci_attachFlags_OWNER, because more than one process needs to access it.

SebSparrowHawk (changes requested):

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
Rahul-Sutariya: Made it MULTI_OWNER, Thanks

score/mw/com/gateway/transport_layer/qemu/qemu_hypervisor_transport.cpp
Rahul-Sutariya: Adapted the return, Thanks

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
Rahul-Sutariya: Adapted the error value rerun and corrected AllocateNamedTypedMemory flow, Thanks

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
Rahul-Sutariya: corrected test name, Thanks

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
Rahul-Sutariya: Corrected the test, Thanks

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider_test.cpp
Rahul-Sutariya: Corrected the test, Thanks

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_typed_memory_provider.cpp
Rahul-Sutariya: Thanks for suggestion, updated the method name

score/mw/com/gateway/transport_layer/qemu/ivshmem/ivshmem_bar_discovery.cpp
Rahul-Sutariya: Thanks, removed IvshmemBar structure, added bar.type and its number check

score/mw/com/gateway/transport_layer/qemu/configuration/qemu_transport_config_parser_test.cpp
Rahul-Sutariya: Thanks for suggestion, added waring logs and terminate()

SebSparrowHawk (approved):

PR #922, opened 2026-08-13
Bump dependencies to drop overrides

LittleHuba:


limdor (changes requested):
Can we not ask for a new baselib release? This brings us quite some commits back and some changes are needed for follow-up work.

score/mw/com/gateway/gateway_application/gateway_error_test.cpp
```diff
@@ -61,17 +61,21 @@ TEST_F(GatewayErrorTest, MessageForUnknownServiceElement)
...
+}
+
 TEST_F(GatewayErrorTest, MessageForReceiveHandlerRegistrationFailed)
 {
     TestErrorMessage(
         GatewayErrorc::kReceiveHandlerRegistrationFailed,
         "Gateway couldn't register event-receive-handler at its generic proxy to forward event update notifications.");
 }
 
-TEST_F(GatewayErrorTest, MessageForNonWhitelistedService)
```
limdor, line 71: Why is one test case being dropped here?

limdor (approved):

PR #925, opened 2026-08-14
Fix AoU linkage

limdor:


score/mw/com/design/methods/README.md
```diff
@@ -31,7 +31,7 @@ mw::com::impl::ProxyMethod<int(double, std::array<char, 10>)>
 ```
 
 Note, that the signature does not include `const` or reference qualifiers and just support data types, which are
-generally supported by the `LoLa` communication framework. (see [AoU](broken_link_c/issue/5835119)
+generally supported by the `LoLa` communication framework. (see ScoreReq.AoU OnlyLoLaSupportedTypes).
```
gdadunashvili, line 34: ✅ I agree with this change of the AoU. 

The original link 5835119 was referring to a wrong AoU `NoApisFromImplementationNamespace`. Which makes no sense in this context.
The currently linked AoU `OnlyLoLaSupportedTypes` is the one suitable for this section.

LittleHuba (approved):

PR #928, opened 2026-08-14
docs: fix broken sphinx include/toctree paths from prefix mismatch

LittleHuba:
sphinx_docs_library targets used a prefix that placed files at an arbitrary location inside the merged Sphinx source tree, unrelated to where the consuming .rst files expected them. This caused:

- docs/sphinx/lola.rst's `.. include:: README.md` to silently resolve to empty content (no warning), losing the entire mw::com README page from the built/deployed docs.
- The `tutorial/README` and `mw_com_index/index` toctree entries in lola.rst, and the `CONTRIBUTING` toctree entry in index.rst, to be orphaned from site navigation.

Fix: adopt the convention that `prefix` mirrors the target's own source package path, so files keep their real repo-relative path inside the merged Sphinx tree. This makes include/toctree references predictable and avoids collisions between same-named files from different packages. Updated the affected BUILD targets and the lola.rst references accordingly.

### DIFF
--- a/docs/sphinx/utils/extract_api_items.py
+++ b/docs/sphinx/utils/extract_api_items.py
@@ -28,6 +28,15 @@ from typing import Dict, List, Optional
 class APITagExtractor: # pylint: disable=too-few-public-methods
     """Extracts API-tagged items from Doxygen XML output."""
 
+    # Matches the compound (namespace/class/struct) portion of a Doxygen
+    # refid, e.g. "namespacescore_1_1mw_1_1com_1ad1234" -> group(1) is
+    # "namespacescore_1_1mw_1_1com". Shared by `_get_member_kind_from_xml`
+    # (to locate the compound's XML file) and `_get_enclosing_scope`
+    # (to detect/dedupe direct namespace/class-scope members).
+    _COMPOUND_REFID_RE = (
+        r'((?:namespace|class|struct)[^_]+(?:_1_1[^_]+)*?)_1[a-f0-9]'
+    )
+
     def __init__(self, xml_file_path: str):
         """Initialize the API extractor.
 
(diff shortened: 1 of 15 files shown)


limdor:
@LittleHuba will this affect the URL path of the different subpages?

LittleHuba:
That is quite hard to tell, since most of the pages currently are not showing up in our documentation at all.
But yes, the prefixes end up in part in the URL. It depends on how they are included in Sphinx.

The main LoLa readme you pointed out originally stays in the current path at least.
The tutorial appears under `./score/mw/com/doc/tutorial/README.html`

limdor:
These are the links that we have internally
image

LittleHuba:
Since the fixes are now also ready and found/fixes some more things, I directly added them here.

I also checked links in our repository and adapted them to the changes.

docs/sphinx/utils/extract_api_items.py
```diff
@@ -78,8 +87,71 @@ def extract_api_items(self) -> Dict[str, List[Dict[str, str]]]:
...
+        if not documented_scopes:
+            return
+
+        api_items['members'] = [
+            member for member in api_items['members']
+            if self._get_enclosing_scope(member['id'])
+            not in documented_scopes
+        ]
+
+    def _get_enclosing_scope(self, refid: str) -> Optional[str]:
```
limdor, line 129: I start having the feeling that we reach the point where we should start having tests for such utilities. However I do not want to blow up the PR.
Is there a way that we can treat the sphinx warnings as errors with the fixes you did in this PR but without the changes in `docs/sphinx/utils/extract_api_items.py` by ignoring the duplicate warnings onlly?

docs/sphinx/utils/extract_api_items.py
```diff
@@ -78,8 +87,71 @@ def extract_api_items(self) -> Dict[str, List[Dict[str, str]]]:
             sys.exit(1)
 
         self._extract_from_api_references(api_items)
+        self._drop_members_covered_by_enclosing_scope(api_items)
         return api_items
 
+    def _drop_members_covered_by_enclosing_scope(
```
limdor, line 93: I do not fully get what we try to achieve here and what we loose if we do remove the duplicates.
Is this not just bypassing the warning by just removing duplicates but we still have the issue that we document an api in two places and one wins and the other gets silently dropped?
It talks about api_members.rst but this is not a file that we have in the codebase.

LittleHuba:
Superseded by https://github.com/eclipse-score/communication/pull/933
Keeping this PR to retain the deduplication logic should cleaning the tagging not be feasible.

PR #930, opened 2026-08-14
Support loading add-on configurations

SebSparrowHawk:
Enables users to load add-on mw::com configurations after loading the actual application's configuration. 
Only service definitions of these add-on configurations will be considered and merged into the existing configuration if there is no conflict with the already existing configuration. 
Restructured the Configuration's internal data structures for service types and service instances so that the use of mutex/locks to protect the configuration can be limited to a minimum.

### DIFF
--- a/score/mw/com/impl/configuration/configuration.cpp
+++ b/score/mw/com/impl/configuration/configuration.cpp
@@ -25,48 +25,174 @@ Configuration::Configuration(ServiceTypeDeployments service_types,
                              ServiceInstanceDeployments service_instances,
                              GlobalConfiguration global_configuration,
                              TracingConfiguration tracing_configuration) noexcept
-    : service_types_{std::move(service_types)},
-      service_instances_{std::move(service_instances)},
+    : service_types_{std::make_shared<std::list<std::shared_ptr<ServiceTypeDeployments>>>()},
+      service_instances_{std::make_shared<std::list<std::shared_ptr<ServiceInstanceDeployments>>>()},
       global_configuration_{std::move(global_configuration)},
       tracing_configuration_{std::move(tracing_configuration)}
+{
+    service_types_->push_front(std::make_shared<ServiceTypeDeployments>(std::move(service_types)));
+    service_instances_->push_front(std::make_shared<ServiceInstanceDeployments>(std::move(service_instances)));
+}
+
+Configuration::Configuration(Configuration&& other) noexcept
+    // Use lock_guard to guard against a concurrent MergeServiceEntries() call on `other` racing with this move.
+    : Configuration(other, std::lock_guard<std::mutex>(other.merge_mutex_))
+{
+}
+
+Configuration::Configuration(Configuration& other, const std::lock_guard<std::mutex>&) noexcept
+    : service_types_{std::move(other.service_types_)},
+      service_instances_{std::move(other.service_instances_)},
+      global_configuration_{std::move(other.global_configuration_)},
+      tracing_configuration_{std::move(other.tracing_configuration_)}
+// merge_mutex_ itself is not movable and is left default-constructed in the new object.
 {
 }
 
 ServiceTypeDeployment* Configuration::AddServiceTypeDeployment(ServiceIdentifierType service_identifier_type,
                                                                ServiceTypeDeployment service_type_deployment) noexcept
 {
-    const auto emplace_result =
-        service_types_.emplace(std::move(service_identifier_type), std::move(service_type_deployment));
-    if (!emplace_result.second)
+    // If this service type has not yet been added (not in any map of the list), this service type will be added to the
+    // latest map of service types.
+
+    const auto current_list = GetListOfServiceTypeMaps();
+
+    const bool type_found = CheckServiceTypeExists(service_identifier_type, current_list);
+
+    if (!type_found)
     {
-        ::score::mw::log::LogFatal("lola")
-            << "Could not insert service type deployment into Configuration map. Terminating";
-        std::terminate();
+        const auto last_map_entry = current_list->back();
+        const auto emplace_result =
+            last_map_entry->emplace(std::move(service_identifier_type), std::move(service_type_deployment));
+
+        if (emplace_result.second)
+        {
+            return &emplace_result.first->second;
+        }
     }
-    return &emplace_result.first->second;
+
+    ::score::mw::log::LogFatal("lola")
+        << "Could not insert service type deployment into Configuration map. Terminating";
+    std::terminate();
 }
 
 ServiceInstanceDeployment* Configuration::AddServiceInstanceDeployments(
     InstanceSpecifier instance_specifier,
     ServiceInstanceDeployment service_instance_deployment) noexcept
 {
-    const auto emplace_result =
-        service_instances_.emplace(std::move(instance_specifier), std::move(service_instance_deployment));
-    if (!emplace_result.second)
+    // If this service instance has not yet been added (not in any map of the list), this service instance will be added
+    // to the latest map of service instances.
+
+    const auto current_list = GetListOfServiceInstanceMaps();
+
+    const bool instance_found = CheckServiceInstanceExists(instance_specifier, current_list);
+
+    if (!instance_found)
     {
-        ::score::mw::log::LogFatal("lola")
-            << "Could not insert service instance deployment into Configuration map. Terminating";
-        std::terminate();
+        const auto last_map_entry = current_list->back();
+        const auto emplace_result =
+            last_map_entry->emplace(std::move(instance_specifier), std::move(service_instance_deployment));
+
+        if (emplace_result.second)
+        {
+            return &emplace_result.first->second;
+        }
     }
-    return &emplace_result.first->second;
+
+    ::score::mw::log::LogFatal("lola")
+        << "Could not insert service instance deployment into Configuration map. Terminating";
+    std::terminate();
+}
+
+Result<void> Configuration::MergeServiceEntries(const Configuration& additional_configuration) noexcept
+{
+    std::lock_guard<std::mutex> lock(merge_mutex_);
+
+    // Construct a new map of service type deployments and add all service_type_deployments of
+    // additional_configuration to it. Add this map as a front entry in the list of those entries, so that it will
+    // be returned as the latest version.
+    auto new_type_map = std::make_shared<ServiceTypeDeployments>();
+    bool new_type_element_inserted = false;
+
+    const auto current_type_list = std::atomic_load_explicit(&service_types_, std::memory_order_acquire);
+    if (current_type_list != nullptr)
+    {
+        for (const auto& service_type : additional_configuration.GetServiceTypes())
+        {
+            const bool type_found = CheckServiceTypeExists(service_type.first, current_type_list);
+
+            if (!type_found)
+            {
+                new_type_map->emplace(service_type.first, service_type.second);
+                new_type_element_inserted = true;
+            }
+            else
+            {
+                return Unexpected(MakeError(configuration_errc::configuration_merge_duplicate_service_type));
+            }
+        }
+    }
+    else
+    {
+        return Unexpected(MakeError(configuration_errc::configuration_merged_invalid_configuration_state));
+    }
+
+    if (new_type_element_inserted)
+    {
+        auto updated_list = std::make_shared<std::list<std::shared_ptr<ServiceTypeDeployments>>>(*service_types_);
+        updated_list->push_front(new_type_map);
+
+        std::atomic_store_explicit(&service_types_, updated_list, std::memory_order_release);
+    }
+
+    // Update service instances
+
+    auto new_instance_map = std::make_shared<ServiceInstanceDeployments>();
+    bool new_instance_element_inserted = false;
+
+    const auto current_instance_list = std::atomic_load_explicit(&service_instances_, std::memory_order_acquire);
+    if (current_instance_list != nullptr)
+    {
+        for (const auto& service_instance : additional_configuration.GetServiceInstances())
+        {
+            const bool instance_found = CheckServiceInstanceExists(service_instance.first, current_instance_list);
+
+            if (!instance_found)
+            {
+                new_instance_map->emplace(service_instance.first, service_instance.second);
+                new_instance_element_inserted = true;
+            }
+            else
+            {
+                return Unexpected(MakeError(configuration_errc::configuration_merge_duplicate_service_instance));
+            }
+        }
+    }
+    else
+    {
+        return Unexpected(MakeError(configuration_errc::configuration_merged_invalid_configuration_state));
+    }
+
+    if (new_instance_element_inserted)
+    {
+        auto updated_list =
+            std::make_shared<std::list<std::shared_ptr<ServiceInstanceDeployments>>>(*service_instances_);
+        updated_list->push_front(new_instance_map);
+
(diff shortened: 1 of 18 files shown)


score/mw/com/impl/instance_identifier.h
```diff
@@ -179,15 +180,19 @@ class InstanceIdentifier final
      * Whenever an InstanceIdentifier is created from another serialized InstanceIdentifier, the ServiceTypeDeployment /
      * ServiceInstanceDeployment held by the serialized InstanceIdentifier needs to be added to the maps within the
      * global configuration object. The newly created InstanceIdentifier will then store pointers to these structs.
+     *
+     * Pointer to Configuration is an atomic pointer because SetConfiguration() might be called from different threads,
+     * while it is read from Create() concurrently. Using std::atomic avoids the data race that could occur in this
+     * situation.
```
castler: This does not change with the addon configuration feature - does it?

score/mw/com/impl/runtime.cpp
```diff
@@ -245,6 +277,17 @@ std::vector<InstanceIdentifier> Runtime::resolve(const InstanceSpecifier& specif
...
 }
 
+Result<void> Runtime::MergeAdditionalConfiguration(const Configuration& additional_configuration) noexcept
+{
+    const auto merge_result = configuration_.MergeServiceEntries(std::move(additional_configuration));
+    if (!merge_result.has_value())
+    {
+        return merge_result;
+    }
+    InstanceIdentifier::SetConfiguration(&configuration_);
```
castler: Why should we do that? We are just changing the content of the original configuration object, its not that we create a new one. Thus, this change is not needed! You can then remove also the changes in the `InstanceIdentifier` no?

score/mw/com/impl/configuration/configuration.h
```diff
@@ -53,6 +57,11 @@ class Configuration final
     using ServiceInstanceDeployments = std::unordered_map<InstanceSpecifier, ServiceInstanceDeployment>;
     using BindingInformation = std::variant<LolaServiceTypeDeployment, score::cpp::blank>;
 
+    // Suppress "AUTOSAR C++14 A11-3-1", The rule declares: "Friend declarations shall not be used".
+    // Test only use to check internal state of configuration after (merge) operations.
+    // coverity[autosar_cpp14_a11_3_1_violation]
+    friend class ConfigurationFixture;
```
castler, line 61: What internal state is that? Should not every change be publicly visible?

score/mw/com/impl/instance_identifier.h
```diff
@@ -23,6 +23,7 @@
 
 #include "score/result/result.h"
 
+#include <atomic>
```
castler: This include should not be needed.

score/mw/com/impl/configuration/configuration.h
```diff
@@ -22,9 +22,13 @@
 
 #include "score/result/result.h"
 
+#include "score/mw/log/logging.h"
```
castler: You are not performing any logging in the header.

If you add this here, this will pollute the dependencies, thus better remove it.

score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -25,48 +25,185 @@ Configuration::Configuration(ServiceTypeDeployments service_types,
...
+    service_instances_->push_front(std::make_shared<ServiceInstanceDeployments>(std::move(service_instances)));
+}
+
+Configuration::Configuration(Configuration&& other) noexcept
+    : service_types_{std::move(other.service_types_)},
+      service_instances_{std::move(other.service_instances_)},
+      global_configuration_{std::move(other.global_configuration_)},
+      tracing_configuration_{std::move(other.tracing_configuration_)}
 {
 }
```
castler, line 50: Why do we define a custom move-constructor here, and do not just keep the default as it was previously?

score/mw/com/impl/configuration/configuration.h
```diff
@@ -63,7 +72,7 @@ class Configuration final
      * \brief Class is moveable but not copyable
      */
     Configuration(const Configuration& other) = delete;
-    Configuration(Configuration&& other) noexcept = default;
+    Configuration(Configuration&& other) noexcept;
```
castler, line 73: Should stay default IMHO

score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -290,19 +421,101 @@ bool Configuration::AggregateAllowedUsers(std::set<uid_t>& aggregated_allowed_us
...
 
+std::optional<std::reference_wrapper<const ServiceTypeDeployment>> Configuration::GetServiceTypeDeployment(
+    const ServiceIdentifierType& service_identifier_type) const noexcept
+{
+    const auto current_list = std::atomic_load_explicit(&service_types_, std::memory_order_acquire);
+    if ((current_list == nullptr) || current_list->empty())
+    {
+        return std::nullopt;
+    }
+    for (auto& element : *current_list.get())
```
castler: `const auto&`

score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -290,19 +421,101 @@ bool Configuration::AggregateAllowedUsers(std::set<uid_t>& aggregated_allowed_us
...
+
+std::optional<std::reference_wrapper<const ServiceInstanceDeployment>> Configuration::GetServiceInstanceDeployment(
+    const InstanceSpecifier& specifier) const noexcept
+{
+    const auto current_list = std::atomic_load_explicit(&service_instances_, std::memory_order_acquire);
+    if ((current_list == nullptr) || current_list->empty())
+    {
+        return std::nullopt;
+    }
+    for (auto& element : *current_list.get())
```
castler: `const auto&`

score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -290,19 +421,101 @@ bool Configuration::AggregateAllowedUsers(std::set<uid_t>& aggregated_allowed_us
...
+    return std::nullopt;
+}
+
+size_t Configuration::GetNumberOfServiceTypes() const noexcept
+{
+    const auto current_list = std::atomic_load_explicit(&service_types_, std::memory_order_acquire);
+    if (current_list != nullptr && !current_list->empty())
+    {
+        size_t number_of_types = 0;
+        for (auto& element : *current_list.get())
```
castler: `const auto&`

score/mw/com/impl/configuration/configuration.cpp
```diff
@@ -290,19 +421,101 @@ bool Configuration::AggregateAllowedUsers(std::set<uid_t>& aggregated_allowed_us
...
+    return 0;
+}
+
+size_t Configuration::GetNumberOfServiceInstances() const noexcept
+{
+    const auto current_list = std::atomic_load_explicit(&service_instances_, std::memory_order_acquire);
+    if (current_list != nullptr && !current_list->empty())
+    {
+        size_t number_of_instances = 0;
+        for (auto& element : *current_list.get())
```
castler: `const auto&`

score/mw/com/impl/configuration/configuration.h
```diff
@@ -83,49 +92,35 @@ class Configuration final
...
     {
-        return service_instances_.empty();
+        return GetNumberOfServiceInstances() == 0;
     }
 
     /// \brief Public interface to trigger a validation of this configuration.
-    score::Result<void> Validate() const noexcept;
+    /// \param addon_configuration If true, the configuration is an add-on configuration and will be validated
+    /// accordingly.
+    score::Result<void> Validate(bool addon_configuration = false) const noexcept;
```
castler: What is the benefit of checking a partial configuration?

Should we not rather have the following workflow:

1. create initial configuration
2. validate it
3. Add addon confgiuration
4. validate it

For me the main question here - what should the user / we do if this is invalid. Will we then terminate? Because otherwise the configuration is in an undefined state?

score/mw/com/impl/configuration/configuration.h
```diff
@@ -156,6 +151,110 @@ class Configuration final
...
+            {
+                for (auto& entry : *element.get())
+                {
+                    result.emplace(entry.first, entry.second);
+                }
+            }
+        }
+
+        return result;
+    }
```
castler: why is this implemented in the header?

score/mw/com/impl/configuration/configuration.h
```diff
@@ -156,6 +151,110 @@ class Configuration final
...
+    ServiceTypeDeployments GetServiceTypes() const noexcept
+    {
+        auto result = ServiceTypeDeployments();
+
+        const auto current_list = std::atomic_load_explicit(&service_types_, std::memory_order_acquire);
+        if (current_list != nullptr && !current_list->empty())
+        {
+            for (auto& element : *current_list.get())
+            {
+                for (auto& entry : *element.get())
```
castler: Will these things be moved? If not, we should make it const again.

score/mw/com/impl/configuration/configuration.h
```diff
@@ -156,6 +151,110 @@ class Configuration final
     std::set<std::string_view> GetInstancesOfServiceType(std::string_view service_type) const noexcept;
 
   private:
+    /// \brief Returns a copy of the merged entries across all persisted maps of service type deployments.
+    /// \attention This returns an owned copy, so no references, pointers or string_views should be extracted and used
+    ///         beyond the copy's lifetime. For that use case, use ForEachServiceType() instead, which iterates directly
+    ///         over the persisted storage.
+    ServiceTypeDeployments GetServiceTypes() const noexcept
+    {
+        auto result = ServiceTypeDeployments();
```
castler: you are initializing here the variable no?

Then easiest would be:

ServiceTypeDeployments result{};

otherwise its not clear if you are invoking a constructor in 160 or a method or ....

score/mw/com/impl/configuration/configuration.h
```diff
@@ -156,6 +151,110 @@ class Configuration final
...
+        return result;
+    }
+
+    /// \brief Returns a copy of the merged entries across all persisted maps of service isntance deployments.
+    /// \attention This returns an owned copy, so no references, pointers or string_views should be extracted and used
+    ///         beyond the copy's lifetime. For that use case, use ForEachServiceInstance() instead, which iterates
+    ///         directly over the persisted storage.
+    ServiceInstanceDeployments GetServiceInstances() const noexcept
+    {
+        auto result = ServiceInstanceDeployments();
```
castler: Same statement as above, prefer {}-Initialization.

score/mw/com/impl/configuration/configuration.h
```diff
@@ -156,6 +151,110 @@ class Configuration final
...
+
+        const auto current_list = std::atomic_load_explicit(&service_instances_, std::memory_order_acquire);
+        if (current_list != nullptr && !current_list->empty())
+        {
+            for (auto& element : *current_list.get())
+            {
+                for (auto& entry : *element.get())
+                {
```
castler: Same question above, should those be moved - if no, make it `const`

score/mw/com/impl/configuration/configuration.h
```diff
@@ -156,6 +151,110 @@ class Configuration final
...
+            for (const auto& entry : *element)
+            {
+                callback(entry);
+            }
+        }
+    }
+
+    /// \brief Helper function to check if entry for this service_identifier is already stored in list of service type
+    /// deployments
+    static bool CheckServiceTypeExists(
```
castler, line 220: Why do we make this static and the pass the member variable? Same question for the function below

score/mw/com/impl/configuration/configuration.h
```diff
@@ -172,15 +271,32 @@ class Configuration final
...
+     * Each write update, i.e. AddServiceInstanceDeployment() and MergeServiceEntries() will create a new map
+     * that will be added to this list. The different versions are stored in a list so that the pointers to the elements
+     * in it, stay valid for the lifetime of this configuration. Stored as std::shared_ptr so that atomic updates can be
+     * done to the list of maps, while the previous generations are still accessible by readers.
+     */
+    std::shared_ptr<std::list<std::shared_ptr<ServiceInstanceDeployments>>> service_instances_;
     GlobalConfiguration global_configuration_;
     TracingConfiguration tracing_configuration_;
+
+    std::mutex merge_mutex_;
```
castler, line 278: Please give an explanation in a comment what this mutex is used for.

Only used to avoid parallel merges...or something like that. Not used in "read" path, due to atomic list construct...yada yada

castler (commented):
As discussed, we have to rethink the "add" logik.

score/mw/com/impl/instance_identifier.h
SebSparrowHawk: Removed

score/mw/com/impl/runtime.cpp
SebSparrowHawk: Yes, removed that.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: At the moment there are methods "ServiceTypeDeployments GetServiceTypes() const noexcept" and "ServiceInstanceDeployments GetServiceInstances() const noexcept" that are defined as private methods. 
Because they should only be used internally and are useful for some test checks.

score/mw/com/impl/instance_identifier.h
SebSparrowHawk: Right, removed

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: Removed it.

score/mw/com/impl/configuration/configuration.cpp
SebSparrowHawk: With the introduction of the std::mutex member variable, I removed the default move constructor.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: With the introduction of the std::mutex member variable, I removed the default move constructor.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: I missed moving the implementation to the .cpp file, fixed.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: Won't be moved, I'll make them const

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: Won't be moved, I'll make them const

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: I wanted to avoid to call std:.atomic_load_explicit(...) in a loop and therefore decided to load it just once and pass it by reference. 
If that should be avoided, I could think about restructuring the whole check logic.

score/mw/com/impl/configuration/configuration.h
SebSparrowHawk: the workflow that you depicted sounds reasonable. 
One thing: Currently validate is called within the Configuration::Parse() directly and then we would still need this parameter to decide that an add-on configuration shall not be (completely) validated, before the merge.

castler (dismissed):

castler (dismissed):

castler (approved):

Issue #852, opened 2026-08-05
Improvement: GenericProxy - add API to read out sample alignment


crimson11:
### What

Currently `GenericProxyEvent` only has an API to get the (aligned) sample size ( `GetSampleSize()` ). It misses an API to read out the alignment!
This is required: If the typoe-erased sample data gets retrieved via a GenericProxyEvent, but then needs to be stored at a (shared) memory location, where an accessor exists, which knows the underlying type and does a type-agnostic access, it needs to be stored at a correctly aligned address! Thus, `GenericProxyEvent` needs to provide an API to do that.

Options:
1. Add a new API `size_t GetSampleAlignment() const` (and keep `GetSampleSize()` )
2. Add a new API `memory::DataTypeSizeInfo GetSampleSizeInfo() const` (and mark `GetSampleSize()` as deprecated)

The 2nd version relies on memory::DataTypeSizeInfo, which contains both, alignedsize and alignment.

### How

see description.

### Estimates for realization

3 days
If solution (2) is chosen also API deprecation has to be handled.

### Category

- [ ] Affects Detailed Design

### Requirements / Architecture

- [x] Requirements / Architecture are not affected by this change?

NemanjaTrifunovicRTRK:
Hello @crimson11,

I fixed it using the first proposed solution.
PR #931 is linked to this issue.

NemanjaTrifunovicRTRK:
PR https://github.com/eclipse-score/communication/pull/931 is merged, I'm closing this issue.

PR #931, opened 2026-08-14
Event sample data type alignment fix

NemanjaTrifunovicRTRK:
Added API for getting alignment of event data type.
Issue #852 

Also fixed alignment of event data type in GatewayApplication.
Alignment was hardcoded to `0U`, now it is collected from `memory::DataTypeSizeInfo`.
Issue #387 

Documentation updated
- `score/mw/com/dependability/software_architectural_design/mw_com_dii/README.md`
- `score/mw/com/design/skeleton_proxy/generic_proxy/generic_proxy_model.puml`

Implementation extended:
- `score/mw/com/gateway/gateway_application/gateway_application.cpp`
- `score/mw/com/impl/bindings/lola/generic_proxy_event.cpp`
- `score/mw/com/impl/bindings/lola/generic_proxy_event.h`
- `score/mw/com/impl/generic_proxy_event.cpp`
- `score/mw/com/impl/generic_proxy_event.h`
- `score/mw/com/impl/generic_proxy_event_binding.h`

Tests updated:

- `score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp`
- `score/mw/com/impl/bindings/mock_binding/generic_proxy_event.h`
- `score/mw/com/impl/generic_proxy_event_test.cpp`

### DIFF
--- a/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc_generic_proxy.trlc
+++ b/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc_generic_proxy.trlc
@@ -510,6 +510,7 @@ section "com" {
                 }
 
                 /* broken_link_c/issue/14035184 */
+                /* GetSampleSize is deprecated in favor of GetDataTypeSizeInfo (see GenericProxyEventGetDataTypeSizeInfo), it will be removed in https://github.com/eclipse-score/communication/issues/975 */
                 ScoreReq.CompReq GenericProxyEventGetSampleSize {
                     description = '''The {{GenericProxyEvent}} shall provide a public method:
 
@@ -518,6 +519,19 @@ section "com" {
                     Behaviour: {{GetSampleSize}} shall adhere to the re-entrancy and thread-safety specifications described in Re-entrancy and Thread-safety.
 
                     Return value: returns the aligned size in bytes of the underlying event sample data type. I.e. in case the underlying data type is T, it shall return {{sizeof(T)}}.'''
+                    safety = ScoreReq.Asil.B
+                    derived_from = [Communication.EventType@1]
+                    version = 1
+                }
+
+                ScoreReq.CompReq GenericProxyEventGetDataTypeSizeInfo {
+                    description = '''The {{GenericProxyEvent}} shall provide a public method:
+
+                    {{{memory::DataTypeSizeInfo GetDataTypeSizeInfo() const noexcept}}}
+
+                    Behaviour: {{GetDataTypeSizeInfo}} shall adhere to the re-entrance and thread-safety specifications described in Re-entrance and Thread-safety.
+
+                    Return value: returns the data type size info (size in bites and alignment in bytes) of the underlying event sample data type.'''
                     safety = ScoreReq.Asil.B
                     derived_from = [Communication.EventType@1]
                     version = 1
--- a/score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
+++ b/score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
@@ -77,6 +77,25 @@ TEST_F(LolaGenericProxyEventFixture, GetSampleSize)
     EXPECT_EQ(generic_proxy_event_->GetSampleSize(), sizeof(SampleType));
 }
 
+TEST_F(LolaGenericProxyEventFixture, GetDataTypeSizeInfo)
+{
+    RecordProperty("lobster-tracing", "GenericProxyEventGetDataTypeSizeInfo");
+    RecordProperty(
+        "Description",
+        "Checks that GetDataTypeSizeInfo will return the data type size info of the underlying event data type.");
+    RecordProperty("TestType", "Requirements-based test");
+    RecordProperty("Priority", "1");
+    RecordProperty("DerivationTechnique", "Analysis of requirements");
+
+    // Given a valid GenericProxyEvent
+    WithAGenericProxyEvent(element_fq_id_, event_name_);
+
+    // Expect, that asking about the Sample size, we get the sizeof the underlying event data type (which is
+    // std::uint32_t in case of LolaProxyEventResources)
+    EXPECT_EQ(generic_proxy_event_->GetDataTypeSizeInfo().Alignment(), alignof(SampleType));
+    EXPECT_EQ(generic_proxy_event_->GetDataTypeSizeInfo().Size(), sizeof(SampleType));
+}
+
 TEST_F(LolaGenericProxyEventFixture, HasSerializedFormat)
 {
     RecordProperty("Verifies", "SCR-14035199");
--- a/score/mw/com/impl/generic_proxy_event.h
+++ b/score/mw/com/impl/generic_proxy_event.h
@@ -80,7 +80,11 @@ class GenericProxyEvent : public ProxyEventBase
 
     /// \brief return the (aligned) size in bytes of the underlying event sample data type.
     /// \return size in bytes.
+    [[deprecated("Use GetDataTypeSizeInfo() for size and alignment information, issue #975")]]
     std::size_t GetSampleSize() const noexcept;
+
+    /// \brief return the size and alignment information of the underlying event sample data type.
+    memory::DataTypeSizeInfo GetDataTypeSizeInfo() const;
 
     /// \brief reports, whether the event sample data the SamplePtr<void> points to is in some internal serialized
     ///        format (true) or it is the binary representation of the underlying C++ data type (false).
--- a/score/mw/com/impl/generic_proxy_event_binding.h
+++ b/score/mw/com/impl/generic_proxy_event_binding.h
@@ -52,7 +52,11 @@ class GenericProxyEventBinding : public ProxyEventBindingBase
 
     /// \brief return the (aligned) size in bytes of the underlying event sample data type.
     /// \return size in bytes.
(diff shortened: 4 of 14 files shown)


score/mw/com/impl/bindings/lola/generic_proxy_event_test.cpp
```diff
@@ -77,6 +77,24 @@ TEST_F(LolaGenericProxyEventFixture, GetSampleSize)
     EXPECT_EQ(generic_proxy_event_->GetSampleSize(), sizeof(SampleType));
 }
 
+TEST_F(LolaGenericProxyEventFixture, GetSampleAlignment)
+{
+    RecordProperty("Verifies", "SCR-14035184");
```
LittleHuba: Please drop this reference to this requirement. That requirement does not fit this API.
Instead, please create a new requirement here:
https://github.com/eclipse-score/communication/blob/main/score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc_generic_proxy.trlc#L520C17-L531C18

Same for other tests where you reference this requirement.

NemanjaTrifunovicRTRK: I fixed it.

limdor (commented):
What are all these IDs added? like 14035185 or 31545523. They seem completly random

NemanjaTrifunovicRTRK:
@limdor 
> What are all these IDs added? like 14035185 or 31545523. They seem completly random

Yes, it is random. I didn't find there is instructions for ID decision.
Inform me if these IDs should be created to have a specific value.
Thank you.

score/mw/com/impl/generic_proxy_event_binding.h
```diff
@@ -54,6 +54,10 @@ class GenericProxyEventBinding : public ProxyEventBindingBase
     /// \return size in bytes.
     virtual std::size_t GetSampleSize() const noexcept = 0;
 
+    /// \brief return the alignment in bytes of the underlying event sample data type.
+    /// \return alignment in bytes.
+    virtual std::size_t GetSampleAlignment() const noexcept = 0;
```
bemerybmw: If we need to change the public api to expose the alignment, I would rather that we deprecate GetSampleSize and add a new function GetDataTypeSizeInfo which returns https://github.com/eclipse-score/baselibs/blob/main/score/memory/data_type_size_info.h

bemerybmw (changes requested):

score/mw/com/impl/generic_proxy_event_binding.h
NemanjaTrifunovicRTRK: The current solution is proposed in issue #852 as the first option by @crimson11.

I'd rather keep this kind of solution because it doesn't require changing existing unit tests, just extending them, and it's also backward compatible.

bemerybmw: I understand that there were 2 options provided by @crimson11 but IMO, the second option is the preferred one. We have many places in our code in which we need size and alignment which we treat differently (e.g. sometimes using separate variables, sometimes using [DataTypeMetaInfo](https://github.com/eclipse-score/communication/blob/main/score/mw/com/impl/data_type_meta_info.h) and sometimes using [DataTypeSizeInfo](https://github.com/eclipse-score/baselibs/blob/main/score/memory/data_type_size_info.h). I think that we should pick one approach (also makes it easier to use these within code, because we don't need to convert from one to the other) and I think the latter makes the most sense as it enforces invariants specified in the standard (e.g. that size is a multiple of alignment and that alignment is a power of 2).

limdor: I also think option 2 would be better

NemanjaTrifunovicRTRK: I changed it to option 2.

btirunagaru (commented):
Please fix the typos.

limdor:
> @limdor
> 
> > What are all these IDs added? like 14035185 or 31545523. They seem completly random
> 
> Yes, it is random. I didn't find there is instructions for ID decision. Inform me if these IDs should be created to have a specific value. Thank you.

@NemanjaTrifunovicRTRK the ids are legacy from when we had this code internally. We will remove all these ids long term. For now, just do not add new IDs and make the linking via requirement name.
You can see an example on how it will look like in https://github.com/eclipse-score/communication/pull/952/changes

NemanjaTrifunovicRTRK:
> > @limdor
> > > What are all these IDs added? like 14035185 or 31545523. They seem completly random
> > 
> > 
> > Yes, it is random. I didn't find there is instructions for ID decision. Inform me if these IDs should be created to have a specific value. Thank you.
> 
> @NemanjaTrifunovicRTRK the ids are legacy from when we had this code internally. We will remove all these ids long term. For now, just do not add new IDs and make the linking via requirement name. You can see an example on how it will look like in https://github.com/eclipse-score/communication/pull/952/changes

@limdor, done.

NemanjaTrifunovicRTRK:
> Please fix the typos.

@btirunagaru, done.

LittleHuba (dismissed):

score/mw/com/impl/generic_proxy_event.h
```diff
@@ -80,8 +80,13 @@ class GenericProxyEvent : public ProxyEventBase
 
     /// \brief return the (aligned) size in bytes of the underlying event sample data type.
     /// \return size in bytes.
+    [[deprecated("Use GetDataTypeSizeInfo() for size and alignment information")]]
```
bemerybmw: Please also link the deprecated ticket here.

score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc_generic_proxy.trlc
bemerybmw, line 512: Can you please create a ticket to remove the deprecated function and link it here? I don't think we have a mechanism to deprecate requirements, so just adding a comment saying it's deprecated and linking the ticket would be fine.

score/mw/com/impl/generic_proxy_event_binding.h
```diff
@@ -52,8 +52,13 @@ class GenericProxyEventBinding : public ProxyEventBindingBase
 
     /// \brief return the (aligned) size in bytes of the underlying event sample data type.
     /// \return size in bytes.
+    [[deprecated("Use GetDataTypeSizeInfo() for size and alignment information")]]
```
bemerybmw: Please also link the deprecated ticket here.

score/mw/com/impl/generic_proxy_event.h
```diff
@@ -80,8 +80,13 @@ class GenericProxyEvent : public ProxyEventBase
...
     std::size_t GetSampleSize() const noexcept;
 
+    /// \brief return the size and alignment information of the underlying event sample data type.
+    /// \return DataTypeSizeInfo instance.
```
bemerybmw: In general, we only add comments when they're adding information that isn't clear from the code."/// \return DataTypeSizeInfo instance." is not adding anything so it can be removed.

score/mw/com/impl/generic_proxy_event.h
```diff
@@ -80,8 +80,13 @@ class GenericProxyEvent : public ProxyEventBase
...
+    memory::DataTypeSizeInfo GetDataTypeSizeInfo() const noexcept;
```
bemerybmw: Newly added functions should not be marked as noexcept. There was a decision made that we should only mark destructors and move operators as noexcept.

score/mw/com/impl/generic_proxy_event_binding.h
```diff
@@ -52,8 +52,13 @@ class GenericProxyEventBinding : public ProxyEventBindingBase
...
     virtual std::size_t GetSampleSize() const noexcept = 0;
 
+    /// \brief return the size and alignment information of the underlying event sample data type.
+    /// \return DataTypeSizeInfo instance.
```
bemerybmw: Can remove /// \return DataTypeSizeInfo instance.

bemerybmw (changes requested):
Please also squash the commits into a single commit with an informative message

score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc_generic_proxy.trlc
NemanjaTrifunovicRTRK: Issue #975

bemerybmw:
@NemanjaTrifunovicRTRK Please don't resolve the reviewer comments yourself. You can put a thumbs up or a comment indicating that you've addressed the comment and then the reviewer can resolve the comment once they've checked your change.

score/mw/com/dependability/requirements/component_requirements/component_requirements_ipc_generic_proxy.trlc
bemerybmw: Please add a comment here linking the ticket to remove the deprecated requirement

bemerybmw (changes requested):
You missed one comment. Apart from that, please squash the commits again. We shouldn't have merge commits. If you want to update the pr with latest master, then do a rebase instead of a merge:

Image

NemanjaTrifunovicRTRK:
> You missed one comment. Apart from that, please squash the commits again. We shouldn't have merge commits. If you want to update the pr with latest master, then do a rebase instead of a merge:
> 
> Image

@bemerybmw, thank you for the instructions.

I fixed it now, you can check.

bemerybmw (approved):

PR #934, opened 2026-08-14
Use public api instead

sergejwinterbmw:
Issue: SWP-175391

### DIFF
--- a/score/mw/com/gateway/transport_layer/sample/test/app2_main.cpp
+++ b/score/mw/com/gateway/transport_layer/sample/test/app2_main.cpp
@@ -47,6 +47,10 @@ void SendMessages(BidirectionalTransport& transport)
         score::mw::com::impl::ServiceElementType::EVENT,
         "TestEvent"};
     const auto notification_result = transport.SendNotification(notification);
+    if (!notification_result)
+    {
+        std::cerr << "Failed to send notification: " << notification_result.error() << std::endl;
+    }
 }
 
 int ExecuteWithRegularConnection()
--- a/score/mw/com/gateway/gateway_application/BUILD
+++ b/score/mw/com/gateway/gateway_application/BUILD
@@ -35,12 +35,9 @@ cc_library(
     ],
     deps = [
         ":gateway_core",
+        "//score/mw/com:types",
         "//score/mw/com/gateway/gateway_application/configuration:gateway_configuration",
         "//score/mw/com/gateway/transport_layer:transport",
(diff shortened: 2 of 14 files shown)


score/mw/com/gateway/transport_layer/sample/test/app2_main.cpp
```diff
@@ -44,9 +44,10 @@ void SendMessages(BidirectionalTransport& transport)
 
     auto notification = score::mw::com::gateway::RegisterNotificationRequest{
         score::mw::com::impl::InstanceSpecifier::Create(std::string{"TestService/Instance1"}).value(),
-        score::mw::com::impl::ServiceElementType::EVENT,
+        score::mw::com::ServiceElementType::EVENT,
         "TestEvent"};
     const auto notification_result = transport.SendNotification(notification);
+    std::ignore = notification_result;  // Ignore the result for this test
```
limdor: What is the reason to ignore it? Can we leave it out of this PR this change and discuss it in a separate one? It is quite a different topic.
To me it looks like or we should at least issue a log, or basically change the signature of the `SendMessage` and forward the result. But I also see that above there is the `send_result` that also is not handled.

score/mw/com/impl/service_element_map_view.h
```diff
@@ -44,6 +44,16 @@ class ServiceElementMapView
...
+    const_iterator begin() const noexcept
+    {
+        return cbegin();
+    }
+
+    const_iterator end() const noexcept
+    {
+        return cend();
+    }
+
```
limdor:
```suggestion
```
This is a bit strange to have `begin` and `end` to return const iterators.
What we should check is why `cbegin` and `cend` is not taken.

score/mw/com/types.h
```diff
@@ -66,6 +67,8 @@ using FindServiceHandle = ::score::mw::com::impl::FindServiceHandle;
 /// \brief Handle to a service instance.
 using HandleType = ::score::mw::com::impl::HandleType;
 
+using ServiceElementType = ::score::mw::com::impl::ServiceElementType;
```
limdor: I had a short sync with @bemerybmw regarding this change and right now this is not part of the public API. If we want to make it public, I would not change it in this PR.
I would like to see this PR to first move to the public API that is currently agreed, and leave potential extensions for later.

sergejwinterbmw: moved back to impl

limdor (approved):

PR #937, opened 2026-08-17
mw/com: expose public assumptions_of_use target for safety analysis AoUs

akritisingh1109:
Adds a public assumptions_of_use Bazel target wrapping aou.trlc under score/mw/com/dependability/safety_analysis

### DIFF
--- a/score/mw/com/dependability/safety_analysis/BUILD
+++ b/score/mw/com/dependability/safety_analysis/BUILD
@@ -13,6 +13,7 @@
 
 load(
     "@score_tooling//bazel/rules/rules_score:rules_score.bzl",
+    "assumptions_of_use",
     "fmea",
 )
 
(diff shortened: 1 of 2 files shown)


score/mw/com/dependability/safety_analysis/BUILD
```diff
@@ -36,3 +37,9 @@ fmea(
     root_causes = [":root_causes"],
     visibility = ["//score/mw/com/dependability:__subpackages__"],
 )
+
+assumptions_of_use(
+    name = "assumptions_of_use",
```
limdor, line 42: This feels wrong. With this change we have the file `aou.trlc` in two targets, `fmea` target and the `assumptions_of_use` target.
If the problem is that `aou.trlc` contains AoUs and control measures, we should just split it first.
I cannot see the AoUs linked in the fmea in https://eclipse-score.github.io/tooling/latest/user_guide/dependability_analysis.html#fmea, is this correct? Are AoUs not linked to an fmea? I would have expected that the AoUs that are the result of hte FMEA they are linked also to the fmea.

limdor (changes requested):

score/mw/com/dependability/safety_analysis/BUILD
```diff
@@ -31,8 +32,14 @@ filegroup(
...
+    controlmeasures = [":assumptions_of_use"],
     failuremodes = ["failure_modes.trlc"],
     root_causes = [":root_causes"],
     visibility = ["//score/mw/com/dependability:__subpackages__"],
 )
+
+assumptions_of_use(
+    name = "assumptions_of_use",
+    srcs = ["aou.trlc"],
+    visibility = ["//visibility:public"],
```
LittleHuba, line 44: Please reduce the visibility to within the module.
Other modules should not access this target directly but use the dependable_element instead.


```suggestion
    visibility = ["//score/mw/com/dependability:__pkg__"],
```

score/mw/com/dependability/safety_analysis/BUILD
LittleHuba: It is still necessary to split AoUs from control measures.

hoe-jo (dismissed):

PR #938, opened 2026-08-17
Enforce doc warnings

limdor:


score/mw/com/com_error_domain.h
```diff
@@ -15,7 +15,11 @@
 
 #include "score/mw/com/impl/com_error.h"
 
-/// \api
+///  \api
+namespace score::mw::com::dummy{
+    // This is just a dummy empty namespace otherwise our documentation build fails if there is no namespace tagged with api
+}
+
```
LittleHuba: Why the dummy namespace in here?

score/mw/com/impl/traits.h
```diff
@@ -456,12 +456,10 @@ template <template <class> class Interface>
 std::optional<std::unordered_map<HandleType, std::queue<Result<ProxyWrapperClass<Interface>>>>>
     ProxyWrapperClass<Interface>::creation_results_{};
 
-/// \api
 /// \brief Interpret an interface that follows our traits as proxy (see description above)
 template <template <class> class T>
 using AsProxy = ProxyWrapperClass<T>;
 
-/// \api
```
LittleHuba, line 464: Based on the commit description I would have expected that the tag here is necessary?

score/mw/com/com_error_domain.h
limdor: I explained in the commit message. The problem is or doc generation. That is what we get if no namespace is tagged
```
ERROR: /home/q463998/github.com/communication/docs/sphinx/BUILD:45:17: output 'docs/sphinx/generated/api_namespaces.rst' was not created
ERROR: /home/q463998/github.com/communication/docs/sphinx/BUILD:45:17: Generating API RST files for generate_api_rst failed: not all outputs were created or valid
Target //docs/sphinx:sphinx_doc failed to build
```

score/mw/com/impl/traits.h
limdor: This is in the `impl` namespace. There is no right answer right now without blowing up the full fix.
If we add it, then the class in the `impl` namespace appears in the documentation like a regular class. If we do not add it, then the user cannot follow the alias from the public `AsProxy` to know what it is available.

score/mw/com/com_error_domain.h
LittleHuba: But why the naemsapce in `com_error_domain.h`? Personally, I would have seen this more in a separate target.

limdor: It does not matter where, it needs to be somewhere. I personally do not care in which file I put it. Just let me know in which one you prefer I add it.

limdor: I added an additional commit with an alternative solution but it is also not super nice.

LittleHuba (approved):

limdor:
We need to go back to the old approach because of the API tests.
https://github.com/eclipse-score/communication/pull/943

PR #940, opened 2026-08-17
Enable pylint

LittleHuba:


LittleHuba:
I will create a ticket to follow up on the in-code suppressions for pylint that were necessary for introduction of the check.

LittleHuba:
We decided for ruff instead of pylint. Closing in favor of #954

PR #947, opened 2026-08-18
Fix serialization and deserialization for ProvideServiceRequest in Gateway

NemanjaTrifunovicRTRK:
The problem was with the deserialization of the field of type` score::mw::com::impl::EventInfo` and its `std::string_view` field name:

```
struct EventInfo
{
    std::string_view name;
    DataTypeMetaInfo data_type_meta_info;
};
```

std::string_view is essentially a pointer into a string, so its (de)serialization requires the string's value to be copied, not just the pointer.

Implementation extended:

- `score/mw/com/gateway/transport_layer/sample/messages/BUILD`
- `score/mw/com/gateway/transport_layer/sample/messages/serialization.h`

Unit tests extended:

- ` score/mw/com/gateway/transport_layer/sample/messages/serialization.h`

Issue #387

### DIFF
--- a/score/mw/com/gateway/transport_layer/sample/messages/serialization.h
+++ b/score/mw/com/gateway/transport_layer/sample/messages/serialization.h
@@ -20,6 +20,7 @@
 #include <cstring>
 #include <limits>
 #include <string>
+#include <string_view>
 #include <tuple>
 #include <type_traits>
 #include <vector>
@@ -49,6 +50,30 @@ struct is_std_string<std::string> : std::true_type
 {
 };
 
+// Helper to detect std::string_view
+template <typename T>
+struct is_std_string_view : std::false_type
+{
+};
+
+template <>
+struct is_std_string_view<std::string_view> : std::true_type
+{
+};
+
+// Helper to detect types that opt into member-wise (de)serialization via GetSerializeMembers(),
+// so they are not accidentally routed to a raw memcpy just because they happen to be trivially
+// copyable (e.g. a struct holding a std::string_view).
+template <typename T, typename = void>
+struct has_get_serialize_members : std::false_type
+{
+};
+
+template <typename T>
+struct has_get_serialize_members<T, std::void_t<decltype(std::declval<T&>().GetSerializeMembers())>> : std::true_type
+{
+};
+
 // Tag types for dispatch
 struct trivially_copyable_tag
 {
@@ -57,6 +82,9 @@ struct std_vector_tag
 {
 };
 struct std_string_tag
+{
+};
+struct std_string_view_tag
 {
 };
 struct non_trivial_tag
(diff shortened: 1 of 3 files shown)


score/mw/com/gateway/transport_layer/sample/messages/serialization.h
```diff
@@ -390,6 +423,34 @@ score::Result<uint32_t> Deserialize(T& non_trivial_copyable,
     return total_consumed_bytes;
 }
 
+/**
+ * @brief Deserialize implementation for score::mw::com::impl::EventInfo.
+ * @details Necessery because EventInfo has a std::string_view member, which is not trivially copyable and not a
+ * std::string.
+ */
+template <>
+inline score::Result<uint32_t> Deserialize<score::mw::com::impl::EventInfo>(
```
LittleHuba: Specializing for this high-level type is probably too coarse.
I would expect that EventInfo falls into https://github.com/NemanjaTrifunovicRTRK/communication/blob/4bad624b0978207811515b9bdf449a7de8bd6622/score/mw/com/gateway/transport_layer/sample/messages/serialization.h#L394

This would mean you only need to provide `Deserialize` specializations for the internal non-trivially-copyable types.

Same pattern applies for `Serialize()`. Can you please check whether you can trim the fix into that direction?

NemanjaTrifunovicRTRK: Deserialization of ` score::mw::com::impl::EventInfo` fell into https://github.com/NemanjaTrifunovicRTRK/communication/blob/4bad624b0978207811515b9bdf449a7de8bd6622/score/mw/com/gateway/transport_layer/sample/messages/serialization.h#L348.

LittleHuba: In that case I think we need to sharpen https://github.com/NemanjaTrifunovicRTRK/communication/blob/4bad624b0978207811515b9bdf449a7de8bd6622/score/mw/com/gateway/transport_layer/sample/messages/serialization.h#L75 to also exclude `std::string_view`s. Once that is done, you should have the expected behavior I outlined above.

NemanjaTrifunovicRTRK: Please check score/mw/com/gateway/transport_layer/sample/messages/serialization.h again.
I changed the implementation to not specialize EventInfo type.

score/mw/com/gateway/transport_layer/sample/messages/serialization.h
```diff
@@ -307,6 +371,28 @@ score::Result<uint32_t> Deserialize(T& target_string, score::cpp::span<const std
...
+ */
+template <typename T>
+score::Result<uint32_t> Deserialize(T& target_string_view,
+                                    score::cpp::span<const std::byte> source_buffer,
+                                    std_string_view_tag)
+{
+    auto string_start = reinterpret_cast<const char*>(source_buffer.data());
+    const auto string_len = strnlen(string_start, source_buffer.size());
+    const auto remaining_size = source_buffer.size();
+    if (string_len == remaining_size || string_len == 0)
```
LittleHuba: Why do you consider zero-length strings as invalid?

LittleHuba:
As discussed in the weekly, the actual fix is more complex.
I went ahead and implemented it in #1039

PR #949, opened 2026-08-18
Fix aspect lint findings

sergejwinterbmw:


score/mw/com/impl/service_element_map_view.h
```diff
@@ -44,6 +44,16 @@ class ServiceElementMapView
...
+    const_iterator begin() const noexcept
+    {
+        return cbegin();
+    }
+
+    const_iterator end() const noexcept
+    {
+        return cend();
+    }
+
```
LittleHuba: Should these not return `iterator`?

limdor: I believe it was a very intentional design that this map view does not allow you to get none const iterators.
I would make `begin() const` to return `const_iterator` even if this is not fulfilling container requirements.

We are anyway as of today not fulfilling container requirements by not providing `begin()` and `end()`

bemerybmw: The map itself should not be modified, i.e. a user cannot add or remove elements. However, the user should be able to get a non const iterator to an event so that they can for example call GetNewSamples on the retrieved event which is non const.

limdor: Ok then to me it looks good, once the formatting is fixed we can merge it.

limdor (approved):

PR #954, opened 2026-08-18
Introduce ruff

LittleHuba:


castler (approved):

quality/static_analysis/static_analysis.bazelrc
```diff
@@ -29,6 +29,20 @@ build:clang-tidy --extra_toolchains=@ferrocene_x86_64_unknown_linux_gnu_llvm//:r
 build:clippy --config=_lint
 build:clippy --aspects=@score_rust_policies//clippy:linters.bzl%clippy_strict
 
+# Ruff configuration
```
limdor, line 32: It can be done in a follow up, but it would be good to sort a bit in this file or split the bazelrc in multiple ones, one for each tool.
Right now the clang-tidy configs are before these configs but the clang-tidy-fix config are after. Thex should be group by tool or by check and fix, but right now is kind of a mix.

limdor (dismissed):
Commit 163057377e8e5e27b12239b3ff88fff0f48679a6 has a mix of automatic formatting and manual changes. Please split it otherwise review is really hard.

PR #961, opened 2026-08-19
Add clippy and ruff to CI.md

LittleHuba:
We introduced clippy and ruff in the CI.

Adds them to the documentation to keep it up to date.

### DIFF
--- a/CI.md
+++ b/CI.md
@@ -110,6 +110,11 @@ Also, it lints the files to ensure the basic Bazel best practices.
 Ensure that the modified C++ files are formatted following our decided style.
 The goal is not to ensure a specific format but to have consistency in the project.
 
+#### Formatting of Python files
+
+Ensure that the modified Python files are formatted following our decided style.
+The goal is not to ensure a specific format but to have consistency in the project.
+
 #### Build everything and run unit tests (x86-64 linux)
 
 Builds everything and runs all unit tests for all languages (C++, Rust, Python, etc.).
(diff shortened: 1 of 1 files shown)


CI.md
```diff
@@ -110,6 +110,11 @@ Also, it lints the files to ensure the basic Bazel best practices.
 Ensure that the modified C++ files are formatted following our decided style.
 The goal is not to ensure a specific format but to have consistency in the project.
 
+#### Ruff format
```
limdor:
```suggestion
#### Formatting of Pyhton files
```
I know above we explicitly call it clang format, and that might have been a mistake, but considering that for Python there are a lot of formatters available, I would just consider that we want to have formatting as part of the design and not what tool we want to use for the Python formatting.

CI.md
```diff
@@ -174,6 +179,14 @@ Runs the [clang-tidy checks](./.clang-tidy) for the target.
 
 Note: Not yet implemented, it will be done with https://github.com/eclipse-score/communication/issues/790.
 
+#### Clippy linting (x86-64 linux)
+
+Runs the [default clippy checks](https://rust-lang.github.io/rust-clippy/master/index.html) for the host.
+
+#### Ruff linting (x86-64 linux)
+
+Runs the [ruff checks](./.ruff.toml) for the host.
```
limdor:
```suggestion
#### Linting for Python (x86-64 linux)

Runs the [ruff checks](./.ruff.toml) for the host.
```
I think adding the tool name in the section was a mistake for C++. I would go with a generic headline and the tool in the description.
In the paragraph just add a link to ruff docs that people could look what ruff is and just say that we choose ruff over pylint because of the speed and it gets the job done.

CI.md
```diff
@@ -174,6 +179,14 @@ Runs the [clang-tidy checks](./.clang-tidy) for the target.
 
 Note: Not yet implemented, it will be done with https://github.com/eclipse-score/communication/issues/790.
 
+#### Clippy linting (x86-64 linux)
+
+Runs the [default clippy checks](https://rust-lang.github.io/rust-clippy/master/index.html) for the host.
```
limdor:
```suggestion
#### Linting for Rust (x86-64 linux)

Runs the [default clippy checks](https://rust-lang.github.io/rust-clippy/master/index.html) for the host.
```

CI.md
LittleHuba: Should I also adapt for C++?

limdor (approved):

PR #964, opened 2026-08-19
Warning fixes

eclipse-impl:
These small details were causing warning, that would then lead the builds to fail.

### DIFF
--- a/score/mw/com/impl/methods/skeleton_method.h
+++ b/score/mw/com/impl/methods/skeleton_method.h
@@ -286,6 +286,8 @@ Result<void> SkeletonMethod<ReturnType(ArgTypes...)>::RegisterHandlerImpl(Callab
                 }
                 else
                 {
+                    // When WithQuality != kYes it does not matter the quality type
+                    std::ignore = quality_type;
                     return stateless_type_erased_handler(
                         user_callback, std::optional<QualityType>{}, type_erased_in_args, type_erased_return);
                 }


score/mw/com/impl/methods/skeleton_method.h
```diff
@@ -286,6 +286,7 @@ Result<void> SkeletonMethod<ReturnType(ArgTypes...)>::RegisterHandlerImpl(Callab
                 }
                 else
                 {
+                    std::ignore = quality_type;
```
limdor, line 290: I would rather use `maybe_unused` in the paramter of the lambda to already show there that it is not used in all cases
https://en.cppreference.com/cpp/language/attributes/maybe_unused

LittleHuba (dismissed):

score/mw/com/impl/bindings/lola/linear_search_map.h
```diff
@@ -134,8 +134,8 @@ class LinearSearchMap
 template <typename Key, typename MappedType, typename KeyEqual, typename Allocator>
 LinearSearchMap<Key, MappedType, KeyEqual, Allocator>::LinearSearchMap(const size_type max_number_of_elements,
                                                                        memory::shared::ManagedMemoryResource& resource,
-                                                                       const KeyEqual& key_equal)
-    : storage_{max_number_of_elements, Allocator{resource}}, key_equal_{key_equal}
+                                                                       const KeyEqual& key_equal_inst)
+    : storage_{max_number_of_elements, Allocator{resource}}, key_equal_{key_equal_inst}
```
limdor: What is this shadowing? The value type?
To me the right fix would be to rename https://github.com/eclipse-impl/communication/blob/b962abc12552da8ab3e1d468965f260e3a2110b6/score/mw/com/impl/bindings/lola/linear_search_map.h#L76
I am a bit confused on why we decided to call it key_equal when in the standard it says that it should be `key_compare`. On a quick look it seems to me that it is the same thing and we just got the name wrong.
@bemerybmw do you remember if this alias name deviation was intentional?

limdor (changes requested):

score/mw/com/impl/bindings/lola/linear_search_map.h
eclipse-impl: Yes, the collision was against the type definition. I would have no problem changing it there, although a local parameter seemed to me to be a more constrained change.

score/mw/com/impl/methods/skeleton_method.h
eclipse-impl: Would also be fine for me, although it could be the case that external tooling, like static analysers or coverage instrumentation tools, would not properly understand the new construction, whereas this is known to work and also makes it clear when the variable is used and when it's not.

score/mw/com/impl/bindings/lola/linear_search_map.h
limdor: It is the most constrained change, but I would still like if we can use the oportunity to just fix it properly.

score/mw/com/impl/methods/skeleton_method.h
limdor: Also fine for me to keep it, but I do not think a tool would understand ignore but not maybe.
In that case, I would like you add at least a comment.
```suggestion
                    // When WithQuality != kYes it does not matter the quality type
                    std::ignore = quality_type;
```

score/mw/com/impl/bindings/lola/linear_search_map.h
limdor: Because the biggest issue is why we did not detect these warnings. I just created https://github.com/eclipse-score/communication/pull/966 that fixes the root problem and also fixes this warning.

score/mw/com/impl/methods/skeleton_method.h
limdor: @eclipse-impl could you give me the exact warning that you were getting? I would make sure that we are catching it in our CI and we do not reintroduce it.

score/mw/com/impl/bindings/lola/linear_search_map.h
bemerybmw: @limdor I don't think there was a reason to deviate with the naming. It should be fine to change the name to key_compare.

limdor (approved):

PR #968, opened 2026-08-20
Add more warnings

limdor:


SebSparrowHawk (approved):
Change in "score/mw/com/gateway/transport_layer/sample/test/app1_main.cpp" looks good to me. 
Actual message content is not relevant for this test.

bemerybmw (approved):

PR #973, opened 2026-08-20
POC fast_llvm toolchain

AlexanderLanin:
extraction of llvm tar:
* original toolchain 220s
* this toolchain 70s

### DIFF
--- a/fast_llvm_repo.bzl
+++ b/fast_llvm_repo.bzl
@@ -54,6 +54,9 @@ _ARCH_ALIASES = {
     "x86_64": "x86_64",
 }
 
+def _log(message):
+    print("[fast_llvm_repo] " + message)
+
 # Implementation of `fast_llvm_repo`. Repository rules run while Bazel is
 # preparing external dependencies, before analysing or building project code.
 def _fast_llvm_repo_impl(ctx):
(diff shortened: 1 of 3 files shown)


fast_llvm_repo.bzl
```diff
@@ -0,0 +1,164 @@
...
+                "-xf",
+                archive,
+                "--strip-components=1",
+                # Extract into the external repository represented by `ctx`.
+                "-C",
+                ctx.path("."),
+            ],
+            timeout = 1800,
+            quiet = False,
+        )
```
LittleHuba, line 135: Why not use https://docs.bazel.build/versions/0.29.1/repo/http.html#http_archive

AlexanderLanin: Thanks for pointing this out — it seems I forgot to include any documentation for this POC.

Functionally, `http_archive` would be a valid solution. The LLVM toolchain internally creates a repository for the LLVM distribution and uses Bazel’s standard download-and-extract mechanism for it. That `download_and_extract()` step is exactly the slow part this POC is trying to improve.

The custom rule downloads the pinned archive with `ctx.download()` and extracts it using `xz -T0 | tar`, allowing XZ decompression to run in parallel. In my measurements, this reduced the LLVM setup time from roughly 220 seconds to 70 seconds.

So the custom rule is not required for correctness. It replaces only the internally generated LLVM distribution repository, while keeping the existing toolchain configuration and BUILD layout intact. The trade-off is additional maintenance and currently limited platform coverage, which is why this is still a POC.

If the additional extraction time is acceptable, sticking with the original `toolchains_llvm` setup is indeed the simpler option.

LittleHuba: The brittleness comes from calling host tools directly in the repository-rule.
While this makes it fast, it may also make it behave differently depending on what version is installed in the system.

Not quite sure that the benefit is high enough to warant this. After all, in 99% of the cases, the toolchain is cached. Redownload really only happens if one builds without cache or invalidates it by modifying the toolchain.

@castler what are your thoughts on this?

PR #996, opened 2026-08-24
Gateway application binary

NemanjaTrifunovicRTRK:
Add gateway application binary, reusable runner library, and sample config

- Add the gateway_application_runner library (RunGatewayApplication() /
  SignalHandler()), which owns gateway config parsing plus the
  setup/start/wait-for-shutdown sequence. It has no dependency on
  main()/argv, so it can be linked into other applications or libraries
  outside this repo that want to embed a gateway process.

- Add gateway_application_bin, a thin main() that parses the two CLI
  args, initializes the mw::com runtime, installs SIGINT/SIGTERM
  handlers, and calls into the runner library — the entry point used
  within this repo to run a standalone gateway process.

- Add sample runtime configuration under gateway_application/etc/
  (mw_com config, gateway config, sample transport config, and
  logging.json) for both the source and destination gateway roles,
  with a README documenting how the files link together and how to
  run the pair locally. Interfaces in configuration linked to `score/mw/com/example/com-api-example` and `score/mw/com/doc/tutorial/chapter_1`

Issue #387

### DIFF
--- a/quality/visibility_guard/public_targets.golden
+++ b/quality/visibility_guard/public_targets.golden
@@ -54,6 +54,7 @@
 //score/mw/com/dependability:mw_com_doc_needs
 //score/mw/com/dependability:mw_com_index
 //score/mw/com/dependability:mw_com_rst
+//score/mw/com/gateway/gateway_application:gateway_application_runner
 //score/mw/com/performance_benchmarks/macro_benchmark/config_generator:config_generator_rule.bzl
 //score/mw/com/performance_benchmarks/macro_benchmark:logging.json
 //score/mw/com/performance_benchmarks/macro_benchmark:lola_benchmarking_client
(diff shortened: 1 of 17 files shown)


LittleHuba (commented):
The general idea to provide an example is very nice.
I'm currently not really happy with the place though.
I think it would make more sense to put this somewhere into the tutorial in its own chapter.
We did not provide an implementation, since the gatewaying logic for VM use cases is quite specific to the hypervisor. But a tutorial chapter on this would help people to implement this for their own hypervisor.
@crimson11 what do you think?

NemanjaTrifunovicRTRK:
> The general idea to provide an example is very nice. I'm currently not really happy with the place though. I think it would make more sense to put this somewhere into the tutorial in its own chapter. We did not provide an implementation, since the gatewaying logic for VM use cases is quite specific to the hypervisor. But a tutorial chapter on this would help people to implement this for their own hypervisor. @crimson11 what do you think?
@LittleHuba,

The idea for `gateway_application_runner` is to provide public API that can be integrated into a 3rd-party applications, so I thik that this part should stay in this place. For me this is the main part cause I plan to use it in my company's project.

For the `gateway_application_bin` it makes sense to be in the tutorial part. Also, it can be extracted to another PR if make more sense to you.

LittleHuba:
Marked as ready for review based on discussion in weekly.

PR #1003, opened 2026-08-25
fix: resolve performance-unnecessary-value-param clang-tidy findings

castler:
## Summary

Fixes all in-repo findings of the `performance-unnecessary-value-param` clang-tidy check, which flags function parameters passed by value but only ever read (never moved-from), recommending `const&` instead to avoid an unnecessary copy.

## Why this needed extensive manual work

Unlike some other checks fixed in earlier PRs, this check's FixIt is **partially** automatic and has two distinct incompleteness modes:

1. **Declaration/definition split**: the FixIt only patches the diagnostic location (usually the out-of-line `.cpp` definition), never the paired header declaration. For member functions this causes a hard compile error (signature mismatch); for free functions it can silently create a duplicate overload, surfacing as a linker error instead.
2. **No FixIt at all** for header-only inline functions, template member functions, and constructors with member-initializer lists — confirmed by inspecting the generated patch files (many were empty even at flagged locations).

This PR:
- Applies the upstream FixIt patches to affected `.cpp` files.
- Manually synchronizes every affected header declaration to match.
- Manually converts ~25 header-only inline functions, constructors, and templates that clang-tidy flags but cannot autofix.
- Removes now-redundant `std::move()` calls at call sites where the callee parameter changed from by-value to `const&`, including cascading cases uncovered once an inner callee was fixed (e.g. `ServiceDiscovery::BindingSpecificStartFindService` → `StartFindServiceImpl`, and `ServiceElementMessage`'s derived `Register/UnregisterNotificationRequest`, `UpdateNotification`).
- Updates `score/mw/com/api_surface.lock.json` to reflect the intentional public signature change of `Runtime::ResolveInstanceIDs`.

## Excluding `std::shared_ptr` and `std::weak_ptr` from this check

Added `std::shared_ptr;std::weak_ptr` to `performance-unnecessary-value-param.AllowedTypes` (note: the option matches on the type name *without* template arguments, so the value is `std::shared_ptr`/`std::weak_ptr`, not `std::shared_ptr<.*>`/`std::weak_ptr<.*>`) and reverted every `std::shared_ptr`- and `std::weak_ptr`-parameter change originally made for this check back to its original by-value form.

Rationale: copying a `shared_ptr`/`weak_ptr` is a cheap atomic refcount operation, not a heap allocation, and this check has no interprocedural view of pass-through call chains — it was flagging intermediate forwarding functions individually even though only a single copy ultimately occurs at whichever function actually owns/stores the pointer. Rather than fight the checker's local view on every future shared_ptr/weak_ptr parameter, both are now excluded from this check entirely.

## Applying the C++ Core Guidelines "consume parameter" pattern (F.18)
While auditing every `const&` conversion in this PR that removed a pre-existing `std::move()`, one genuine sink/consume parameter was found: `GatewayApplication`'s constructors used to take `GatewayConfiguration` by value and `std::move` it into the `app_configuration_` member. Naively converting that parameter to `const&` (as clang-tidy suggests) would silently turn the move into a copy.

Per [F.18](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#rf-consume) ("for 'will-move-from' parameters, pass by `X&&` and `std::move` the parameter"), this parameter is instead converted to `GatewayConfiguration&&`, preserving the move into the member. All call sites already pass a prvalue (`MakeConfig(...)`), so no call-site changes were required.

Every other `const&` conversion in this PR that also removed a `std::move()` was audited and found **not** to be a genuine sink case — in each instance the `std::move()` was feeding a parameter that was itself already `const&` (e.g. `Skeleton::Create`, `ServiceElementMessage`'s base constructor, `std::promise::set_value`), so the move was already a no-op before this PR and removing it causes no behavioral change.

## Keeping `PolymorphicOffsetPtrAllocator::deallocate` pass-by-value

Reverted `deallocate`'s pointer parameter back to by-value with a `NOLINT` + justification comment, instead of converting it to `const&`. The C++ standard's `Cpp17Allocator` named requirements (`[allocator.requirements]`) specify the expression `a.deallocate(p, n)` matching `std::allocator::deallocate`'s by-value signature, which `std::allocator_traits` and allocator-aware containers rely on. Deviating from that standard interface shape for a small, trivially-copyable pointer-like type (`OffsetPtr<T>`) has no measurable benefit.

## Keeping `PolymorphicOffsetPtrAllocator::deallocate` pass-by-value

Reverted `deallocate`'s pointer parameter back to by-value with a `NOLINT` + justification comment, instead of converting it to `const&`. The C++ standard's `Cpp17Allocator` named requirements (`[allocator.requirements]`) specify the expression `a.deallocate(p, n)` matching `std::allocator::deallocate`'s by-value signature, which `std::allocator_traits` and allocator-aware containers rely on. Deviating from that standard interface shape for a small, trivially-copyable pointer-like type (`OffsetPtr<T>`) has no measurable benefit.

## Fixing `modernize-pass-by-value` findings triggered by this check's own fixes

Some of this check's `const&` conversions newly triggered `modernize-pass-by-value`: `ServiceTypeDeployment`, `ServiceInstanceDeployment`, and `ConfigurationStore`'s constructors take a parameter by `const&` and then unconditionally copy it into a member — exactly the "pass by value and `std::move`" pattern that check flags. These constructors were converted to take the affected parameters (`BindingInformation`, `ServiceIdentifierType`, `LolaServiceTypeDeployment`) by value with `std::move` into the member, satisfying both checks simultaneously.

All other in-repo `modernize-pass-by-value` findings were confirmed (via aggregating `bazel build --config=clang-tidy` SARIF reports) to be pre-existing on `upstream/main` and unrelated to this PR, so are intentionally left untouched.

## Out of scope

Six remaining findings in `external/score_baselibs+` headers (`dynamic_array.h`, `non_relocatable_vector.h`) are in a vendored external dependency outside this repo's control, consistent with the precedent established in the exclude-external-findings PR (#995).

## Verification

- Full plain build: succeeds, no compile/link errors.
- Full `--config=clang-tidy` build (scoped to this check, plus `modernize-pass-by-value`): 0 remaining in-repo findings for either check (only the 6 external + 6 pre-existing/unrelated findings noted above remain), confirmed after the `std::shared_ptr`/`std::weak_ptr` exclusion/revert, the F.18 conversion, and the `modernize-pass-by-value` fixes.
- Full test suite: passes, excluding pre-existing Docker network-pool-exhaustion environment failures (`all predefined address pools have been fully subnetted`) unrelated to this change.

### DIFF
--- a/score/mw/com/doc/tutorial/chapter_3/consumer/consumer.cpp
+++ b/score/mw/com/doc/tutorial/chapter_3/consumer/consumer.cpp
@@ -105,6 +105,12 @@ int main()
     // This handler is invoked asynchronously by score::mw::com whenever the set of matching service instances changes.
     // It receives *all* currently matching handles. We create a proxy for every newly discovered instance.
     auto find_service_handler = [&discovered_instances_mutex, &discovered_instances, &known_handles](
+                                    // The enclosing score::mw::com::FindServiceHandler is a type-erased callback
+                                    // whose call signature takes this parameter by value; the caller already
+                                    // copies the container into that fixed signature before invoking this lambda,
+                                    // so taking it by const& here would not avoid any copy - it would only
+                                    // (misleadingly) hide the fact that one already happened.
+                                    // NOLINTNEXTLINE(performance-unnecessary-value-param)
                                     score::mw::com::ServiceHandleContainer<HelloWorldProxy::HandleType> handles,
                                     score::mw::com::FindServiceHandle) noexcept {
         std::lock_guard<std::mutex> lock{discovered_instances_mutex};
--- a/score/mw/com/gateway/transport_layer/sample/messages/gateway_messages.h
+++ b/score/mw/com/gateway/transport_layer/sample/messages/gateway_messages.h
@@ -46,7 +46,7 @@ class ProvideServiceRequest : public TransportMessage
   public:
     ProvideServiceRequest() : TransportMessage(MessageType::kProvideServiceRequest) {}
 
-    ProvideServiceRequest(score::mw::com::InstanceSpecifier service_instance_specifier,
+    ProvideServiceRequest(const score::mw::com::InstanceSpecifier& service_instance_specifier,
                           std::vector<score::mw::com::EventInfo> service_elements,
                           std::uint32_t shm_control_size = 0U,
                           std::uint32_t shm_data_size = 0U)
@@ -142,7 +142,7 @@ class OfferServiceRequest : public TransportMessage
   public:
     OfferServiceRequest() : TransportMessage(MessageType::kOfferServiceRequest) {}
 
-    explicit OfferServiceRequest(score::mw::com::InstanceSpecifier service_instance_specifier)
+    explicit OfferServiceRequest(const score::mw::com::InstanceSpecifier& service_instance_specifier)
         : TransportMessage(MessageType::kOfferServiceRequest),
           instance_specifier_(std::string{service_instance_specifier.ToString()})
     {
@@ -175,7 +175,7 @@ class StopOfferServiceRequest : public TransportMessage
   public:
     StopOfferServiceRequest() : TransportMessage(MessageType::kStopOfferServiceRequest) {}
 
-    explicit StopOfferServiceRequest(score::mw::com::InstanceSpecifier service_instance_specifier)
+    explicit StopOfferServiceRequest(const score::mw::com::InstanceSpecifier& service_instance_specifier)
         : TransportMessage(MessageType::kStopOfferServiceRequest),
           instance_specifier_(std::string{service_instance_specifier.ToString()})
     {
@@ -222,7 +222,7 @@ class ServiceElementMessage : public TransportMessage
     using TransportMessage::TransportMessage;
 
     ServiceElementMessage(MessageType message_type,
-                          score::mw::com::InstanceSpecifier service_instance_specifier,
+                          const score::mw::com::InstanceSpecifier& service_instance_specifier,
                           impl::ServiceElementType element_type,
                           std::string element_name)
         : TransportMessage(message_type),
@@ -270,11 +270,11 @@ class RegisterNotificationRequest final : public ServiceElementMessage
 {
   public:
     RegisterNotificationRequest() : ServiceElementMessage(MessageType::kRegisterNotificationRequest) {}
-    RegisterNotificationRequest(score::mw::com::InstanceSpecifier service_instance_specifier,
+    RegisterNotificationRequest(const score::mw::com::InstanceSpecifier& service_instance_specifier,
                                 impl::ServiceElementType element_type,
                                 std::string element_name)
         : ServiceElementMessage(MessageType::kRegisterNotificationRequest,
-                                std::move(service_instance_specifier),
+                                service_instance_specifier,
                                 element_type,
                                 std::move(element_name))
     {
@@ -286,11 +286,11 @@ class UnregisterNotificationRequest final : public ServiceElementMessage
 {
   public:
     UnregisterNotificationRequest() : ServiceElementMessage(MessageType::kUnregisterNotificationRequest) {}
-    UnregisterNotificationRequest(score::mw::com::InstanceSpecifier service_instance_specifier,
+    UnregisterNotificationRequest(const score::mw::com::InstanceSpecifier& service_instance_specifier,
                                   impl::ServiceElementType element_type,
                                   std::string element_name)
         : ServiceElementMessage(MessageType::kUnregisterNotificationRequest,
-                                std::move(service_instance_specifier),
+                                service_instance_specifier,
                                 element_type,
                                 std::move(element_name))
     {
@@ -303,11 +303,11 @@ class UpdateNotification final : public ServiceElementMessage
 {
   public:
     UpdateNotification() : ServiceElementMessage(MessageType::kUpdateNotification) {}
-    UpdateNotification(score::mw::com::InstanceSpecifier service_instance_specifier,
+    UpdateNotification(const score::mw::com::InstanceSpecifier& service_instance_specifier,
                        impl::ServiceElementType element_type,
                        std::string element_name)
         : ServiceElementMessage(MessageType::kUpdateNotification,
-                                std::move(service_instance_specifier),
+                                service_instance_specifier,
                                 element_type,
                                 std::move(element_name))
     {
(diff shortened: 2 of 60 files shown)


score/mw/com/gateway/transport_layer/sample/messages/gateway_messages.h
```diff
@@ -270,11 +270,11 @@ class RegisterNotificationRequest final : public ServiceElementMessage
 {
   public:
     RegisterNotificationRequest() : ServiceElementMessage(MessageType::kRegisterNotificationRequest) {}
-    RegisterNotificationRequest(score::mw::com::InstanceSpecifier service_instance_specifier,
+    RegisterNotificationRequest(const score::mw::com::InstanceSpecifier& service_instance_specifier,
                                 impl::ServiceElementType element_type,
                                 std::string element_name)
         : ServiceElementMessage(MessageType::kRegisterNotificationRequest,
-                                std::move(service_instance_specifier),
+                                service_instance_specifier,
```
castler, line 277: RegisterNotificationRequest  (line 269) publicly inherits from  ServiceElementMessage  (line 208), and its constructor's initializer list (line 276-279) directly invokes ServiceElementMessage 's constructor (line 224-233) as a base-class construction — not a separate/unrelated call. So the same  service_instance_specifier  argument flows straight through from RegisterNotificationRequest 's param, through the base-class init at line 277, into  ServiceElementMessage 's constructor where  .ToString()  is called at line 229. This confirms the parameter is read-only end-to-end and const&  (no move, no universal reference) is correct throughout.

score/mw/com/test/common_test_resources/skeleton_container.h
```diff
@@ -49,10 +49,10 @@ class SkeletonContainer
 };
 
 template <typename Skeleton>
-void SkeletonContainer<Skeleton>::CreateSkeleton(InstanceSpecifier instance_specifier,
+void SkeletonContainer<Skeleton>::CreateSkeleton(const InstanceSpecifier& instance_specifier,
                                                  const std::string& failure_message_prefix)
 {
-    auto skeleton_result = Skeleton::Create(std::move(instance_specifier));
+    auto skeleton_result = Skeleton::Create(instance_specifier);
```
castler, line 55: SkeletonContainer::CreateSkeleton(const InstanceSpecifier&)  →  Skeleton::Create(const InstanceSpecifier&)  → reads only, no ownership taken anywhere

score/mw/com/doc/tutorial/chapter_3/consumer/consumer.cpp
```diff
@@ -105,7 +105,7 @@ int main()
     // This handler is invoked asynchronously by score::mw::com whenever the set of matching service instances changes.
     // It receives *all* currently matching handles. We create a proxy for every newly discovered instance.
     auto find_service_handler = [&discovered_instances_mutex, &discovered_instances, &known_handles](
-                                    score::mw::com::ServiceHandleContainer<HelloWorldProxy::HandleType> handles,
+                                    const score::mw::com::ServiceHandleContainer<HelloWorldProxy::HandleType>& handles,
```
crimson11: see my teams comment. This makes no sense, since the caller still uses a byValue signature and does the copy! This change just "hides" the fact that we still have copy semantics! You are - with this change - now just taking a reference to the copy! this is NOT the intention of your change.

score/mw/com/impl/bindings/lola/test/skeleton_test_resources.cpp
```diff
@@ -405,7 +405,7 @@ void SkeletonMockedMemoryFixture::CleanUpSkeleton()
     // Because the skeleton will hold a raw pointer to that configuration item and on destruction of the SkeletonGuard
     // as member of this fixture, will invoke "StopOffer" which needs to access this configuration items.
     // Thus, we need to clean up earlier - which will cause now mock calls which have not been there before
-    EXPECT_CALL(shared_memory_factory_mock_, Remove(_)).WillRepeatedly([](auto) {});
+    EXPECT_CALL(shared_memory_factory_mock_, Remove(_)).WillRepeatedly([](const auto&) {});
```
crimson11: Same potential issue as I mentioned above!
Easy workaround: Replace the line with:
`EXPECT_CALL(shared_memory_factory_mock_, Remove(_)).Times(testing::AtLeast(1));`
then we don't need any suppression!

score/mw/com/impl/bindings/lola/event_meta_info.h
```diff
@@ -27,7 +27,7 @@ class EventMetaInfo
 {
   public:
     EventMetaInfo(const memory::DataTypeSizeInfo data_type_info,
-                  const memory::shared::OffsetPtr<void> event_slots_raw_array)
+                  const memory::shared::OffsetPtr<void>& event_slots_raw_array)
```
crimson11, line 30: Hm - how did the tests even work? Afaik `EventMetaInfo` is placed into shared-memory! How should a normal reference (aka pointer) work in an object placed in shm? What am I missing here?

score/mw/com/test/fields/field_initial_value/client.cpp
```diff
@@ -56,7 +56,8 @@ void run_client(const std::size_t num_retries, const std::chrono::milliseconds r
     std::promise<std::vector<TestDataProxy::HandleType>> service_discovery_promise{};
     auto service_discovery_future = service_discovery_promise.get_future();
     auto lola_proxy_handles_result = TestDataProxy::StartFindService(
-        [moved_service_discovery_promise = std::move(service_discovery_promise)](auto handles, auto handle) mutable {
+        [moved_service_discovery_promise = std::move(service_discovery_promise)](const auto& handles,
```
crimson11: see my comment above. Misleading! The copy is done anyhow as the caller sees a FindServiceHandler signature and DOES the copy for handles! Please revert this change as it is misleading.

score/mw/com/test/partial_restart/provider_restart_max_subscribers/consumer.cpp
```diff
@@ -135,7 +135,7 @@ void ConsumerActions::DoConsumerActionsBeforeRestart() noexcept
     // Step (C.1) - Start an async FindService Search
     // ********************************************************************************
     std::cout << "Consumer Step (C.1): Call StartFindService" << std::endl;
-    auto find_service_callback = [this](auto service_handle_container, auto) noexcept {
+    auto find_service_callback = [this](const auto& service_handle_container, auto) noexcept {
```
crimson11: see above. Makes no sense here.

crimson11 (changes requested):

score/mw/com/impl/bindings/lola/event_meta_info.h
crimson11: Irrelevant/Wrong review comment!

crimson11 (approved):

PR #1006, opened 2026-08-25
Add integration test for mw::com add-on configurations

SebSparrowHawk:
This integration tests ensures that after merging two configurations, an existing instance identifer is still valid and can be used for communication. Also new service instances as defined in the add-on configuration can be created.

### DIFF
--- a/score/mw/com/test/loading_add_on_configuration/add_on_loading_application.cpp
+++ b/score/mw/com/test/loading_add_on_configuration/add_on_loading_application.cpp
@@ -1,0 +1,310 @@
+/*******************************************************************************
+ * Copyright (c) 2026 Contributors to the Eclipse Foundation
+ *
+ * See the NOTICE file(s) distributed with this work for additional
+ * information regarding copyright ownership.
+ *
+ * This program and the accompanying materials are made available under the
+ * terms of the Apache License Version 2.0 which is available at
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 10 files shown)


score/mw/com/test/loading_add_on_configuration/add_on_loading_application.cpp
```diff
@@ -0,0 +1,310 @@
...
+#include <map>
+#include <utility>
+#include <vector>
+
+using namespace std::chrono_literals;
+
+namespace
+{
+
+std::string ParseServiceInstanceManifest(int argc, const char** argv)
```
crimson11: Name is imho misleading. This function doesn't really parse anything? Please call it `GetAddOnManifestPath`

score/mw/com/test/loading_add_on_configuration/add_on_loading_application.cpp
```diff
@@ -0,0 +1,310 @@
...
+                                                                        notifications.Get(kPhase1SubscribedShmPath),
+                                                                        stop_token);
+
+        if (first_send_result != EXIT_SUCCESS)
+        {
+            return first_send_result;
+        }
+
+        // Load add-on configuration
+        const auto service_instance_manifest_path = ParseServiceInstanceManifest(argc, argv);
```
crimson11: service_instance_manifest_path  -> addon_service_instance_manifest_path

SebSparrowHawk:
Closed in favour of https://github.com/eclipse-score/communication/pull/1030, which uses an easier approach for this integration test.

PR #1007, opened 2026-08-25
move skeleton service element requirements

sahithi-nukala:


LittleHuba (commented):
@crimson11 @bemerybmw is it fine if we loose some old linkage here?
E.g. we had previously 26 linked requirements and now only 7.

bemerybmw:
> @crimson11 @bemerybmw is it fine if we loose some old linkage here? E.g. we had previously 26 linked requirements and now only 7.

No, we should not lose any linkages. Only the requirements which have been moved should be changed, the old ones should stay as they are and should be updated when they're moved. I think something like this would be fine:

```
RecordProperty("Verifies", "SCR-17563743, SCR-21553554");
RecordProperty("lobster-tracing", "Communication.SkeletonFieldClassUpdate");
```

Then when SCR-17563743 and SCR-21553554 are moved to the correct location, they can be added to the lobster-tracing section.

score/mw/com/impl/skeleton_base_test.cpp
```diff
@@ -465,6 +470,7 @@ using SkeletonBaseOfferDeathTest = SkeletonBaseOfferFixture;
 TEST_F(SkeletonBaseOfferDeathTest, TerminateOnOfferWithNoBinding)
 {
     RecordProperty("Verifies", "SCR-6222081, SCR-21856131");
+    RecordProperty("lobster-tracing", "Communication.SkeletonOfferService");
```
gdadunashvili: why was this added here? There was no corresponding codebeamer link

sahithi-nukala: this mapping is based on the test actual behavior matching the requirement description

gdadunashvili: I'm not sure about this. Either way this is not part of this PR, which is supposed to refactor existing requirements. I would suggest removing this now

gdadunashvili (approved):

LittleHuba (approved):

PR #1008, opened 2026-08-25
mw/com/dependability: Split requirements into requirements for units

gdadunashvili:


score/mw/com/dependability/requirements/component_requirements/binding_independent/proxy_and_service_elements/component_requirements_ipc_proxy_event.trlc
bemerybmw, line 1: I guess that you should have split the directory binding_independent_proxy_and_service_elements to be something like binding_independent/proxy_and_service_elements. Because soon we would also have sibling folders e.g. binding_independent/skeleton_and_service_elements.

score/mw/com/dependability/requirements/component_requirements/binding_independent/proxy_and_service_elements/component_requirements_ipc_proxy_event.trlc
```diff
@@ -0,0 +1,231 @@
...
+section "com" {
+
+    section "Public_API_Specification" {
+
+        section "Typed_Proxy" {
+
+            section "Proxy_Event" {
+
+                /* broken_link_c/issue/14137268 */
+                /* ToDo: Potentially uncovered requirement! */
```
bemerybmw, line 26: The issue here is that this is not a requirement but is rather information. Apparently this wasn't taken into account when copying from code beamer to trlc. I'm not sure how exactly they can be marked as information, but these shouldn't be requirments. We should check in codebeamer all requirements that we extract into separate files to make sure that they're:
* still valid from 25-07
* Not deleted / information

score/mw/com/dependability/requirements/component_requirements/binding_independent/proxy_and_service_elements/component_requirements_ipc_proxy_event.trlc
```diff
@@ -99,6 +98,7 @@ section "com" {
                 }
 
                 /* broken_link_c/issue/18805456 */
```
bemerybmw: IMO, when they're real requirements which are missing tests (which is likely the case here since the state change handler was added recently), I'd already remove the codebeamer link.

score/mw/com/dependability/requirements/component_requirements/binding_independent/proxy_and_service_elements/component_requirements_ipc_proxy_event.trlc
```diff
@@ -115,6 +115,7 @@ section "com" {
                 }
 
                 /* broken_link_c/issue/18805552 */
```
bemerybmw: Same here.

bemerybmw (changes requested):

score/mw/com/dependability/requirements/component_requirements/binding_independent/proxy_and_service_elements/component_requirements_ipc_proxy_event.trlc
gdadunashvili: We agreed offline to leave this link here for now, and first finish splitting the requirements

bemerybmw (approved):

PR #1019, opened 2026-08-28
mw/com: Add internal interfaces to architectural design

bemerybmw:


LittleHuba (dismissed):
No git_override at this moment please. Blocking until release candidate is through or we know that we must bump.

PR #1030, opened 2026-08-31
Add integration test for mw::com add-on configurations

SebSparrowHawk:
This integration tests ensures that after merging two configurations, an existing instance identifer is still valid and can be used for communication. Also new service instances as defined in the add-on configuration can be created.


LittleHuba:
Please fix https://github.com/eclipse-score/communication/pull/1030#pullrequestreview-5066929734

score/mw/com/test/loading_add_on_configuration/BUILD
```diff
@@ -11,160 +11,142 @@
...
+cc_library(
+    name = "consumer_impl",
     srcs = ["consumer.cpp"],
     hdrs = ["consumer.h"],
     features = COMPILER_WARNING_FEATURES,
     deps = [
         ":common_resources",
+        ":test_constants",
         "//score/mw/com",
+        "//score/mw/com/impl:instance_specifier",
```
LittleHuba: Please don't depend on impl in integration tests. This adds technical debt.

score/mw/com/test/loading_add_on_configuration/BUILD
```diff
@@ -11,160 +11,142 @@
...
-    name = "field_provider",
+    name = "provider_impl",
     srcs = ["provider.cpp"],
     hdrs = ["provider.h"],
     features = COMPILER_WARNING_FEATURES,
     deps = [
         ":common_resources",
+        ":test_constants",
         "//score/mw/com",
+        "//score/mw/com/impl:instance_specifier",
```
LittleHuba: Same

score/mw/com/test/loading_add_on_configuration/main_consumer.cpp
```diff
@@ -0,0 +1,100 @@
...
+    // 1st step: Run consumer with service defined in initial mw::com configuration
+    score::mw::com::test::run_consumer(stop_source.get_token(),
+                                       score::mw::com::test::kRegularServiceInstanceSpecifier,
+                                       *process_synchronizer_result,
+                                       *provider_ready_synchronizer_result,
+                                       score::mw::com::test::kFirstServiceSamples);
+
+    // 2nd step: Load add-on configuration and merge into existing configuration
+    const auto service_instance_manifest_path = ParseServiceInstanceManifest(argc, argv);
+    const auto add_on_load_result = score::mw::com::runtime::InitializeRuntimeAddonConfiguration(
```
LittleHuba: Late to the party...
That name is a mouth full. I'd love to have something easier. E.g. `score::mw::com::runtime::Reconfigure()`

Why was that name chosen?

LittleHuba (commented):
In general this looks good.
I'm finding the test somewhat too simple.
What you currently do is just assign a different identifier to the same service interface.
But the reconfiguration is way more powerful.

What I would like to see are additional tests that:
1. Add a previously unknown service interface with deployment
2. Do something that would cause problems (like reconfiguring a previously known service instance where we already have a Skeleton/Proxy for
3. ...

score/mw/com/test/loading_add_on_configuration/main_consumer.cpp
SebSparrowHawk: I'm open to a different shorter name. But I'm not sure about Reconfigure(), because it sounds like you could change the existing configuration, while we are only extending it.
About the name's origin: That was chosen by the requesters but I think we can still change that.

score/mw/com/test/loading_add_on_configuration/BUILD
SebSparrowHawk: Will remove it

score/mw/com/test/loading_add_on_configuration/main_consumer.cpp
SebSparrowHawk: Changed to AddConfiguration(...)

score/mw/com/test/loading_add_on_configuration/main_provider.cpp
```diff
@@ -0,0 +1,228 @@
...
+    std::cout << "\nProvider - AddOn Only: Step 2 - Offer add-on service" << std::endl;
+    skeleton_container.OfferService("provider");
+
+    // Keep the service offered for longer than the consumer's service-discovery timeout (10s, see
+    // ProxyContainer::kServiceDiscoveryTimeout) so that a peer consumer without the add-on configuration merged
+    // has enough time to attempt (and fail) discovery.
+    constexpr auto kOfferDuration = std::chrono::seconds(12);
+    std::cout << "\nProvider - AddOn Only: Step 3 - Keep add-on service offered for " << kOfferDuration.count() << "s"
+              << std::endl;
+    std::this_thread::sleep_for(kOfferDuration);
```
LittleHuba: Not really nice. Better would be to use some synchronization mechanism.

SebSparrowHawk: Switched to ProcessSynchronizer to inform provider of finished service disovery

LittleHuba (dismissed):

LittleHuba (approved):

PR #1036, opened 2026-09-02
someip communication middleware bindings impl

sergejwinterbmw:


score/mw/com/impl/plumbing/skeleton_service_element_binding_factory_impl.h
```diff
@@ -113,7 +114,11 @@ inline lola::SkeletonEventProperties CreateSkeletonEventProperties(
 
 }  // namespace detail
 
-template <typename SkeletonServiceElementBinding, typename SkeletonServiceElement, ServiceElementType element_type>
+//TODO: can this be changed to not change all callsite for every added binding?
```
pawelrutkaq, line 117: Ie:
```
SampleType SampleTypeOf(const SkeletonEventBinding<SampleType>&);
```

and then in someip lambda

```
using SampleType = decltype(detail::SampleTypeOf(std::declval<const SkeletonServiceElementBinding&>()));
```

At the end You need SampleType or ?

pawelrutkaq:
`SkeletonMethodBindingFactoryImpl` along with `SkeletonMethodBindingFactory` are missing passing `T (Ret + Args)` down to the binding.
This is opposite to proxy impl and will make it hard to connect serialization for SOME/IP once it needs to know the type.

score/mw/com/impl/configuration/someip_service_instance_deployment.h
```diff
@@ -0,0 +1,55 @@
...
+
+namespace score::mw::com::impl
+{
+
+/// \brief Instance level deployment of the SOME/IP binding.
+///
+/// Holds everything that differs between two instances of the same service interface. In contrast to the LoLa binding
+/// there are no shared memory sizes or slot counts here, since SOME/IP transports serialized messages over a socket
+/// instead of shared memory.
+class SomeIpServiceInstanceDeployment
```
pawelrutkaq, line 31: This deployment would need more params in practice that can even be SOME/IP stack-dependent. 
The question is if we will provide those as Key/Value, untyped or no support at all. Some kind of support would allow no need to patch upstream shared code in which we are now

pawelrutkaq:
Additionally, there is less of a funny effect that happens. If ie example will create Event for std::array, the code will instantiate both SomeIP and Lola implementations and I guess the linker cannot remove them since it has no clue what exactly will be used. Just noting this, may be helpfull for code size investigations.

PR #1037, opened 2026-09-02
Introduce formatting infra for Rust and format Rust code

LittleHuba:


LittleHuba:
Requested feedback from @bharatGoswami8 whether rustfmt config is fine for him.

limdor:
Please rename PR title, it is misleading. It is doing more than just formatting

score/mw/com/rust/BUILD
```diff
@@ -18,7 +18,7 @@ rust_library(
     name = "score_com",
     srcs = ["score_com.rs"],
     crate_name = "score_com",
-    edition = "2024",
+    edition = "2021",
```
bharatGoswami8, line 21: will revert back this once [score_rust_policies](https://github.com/eclipse-score/score_rust_policies/tree/main) upgrade to `2024`

Ticket created for same -
https://github.com/eclipse-score/score_rust_policies/issues/13

score/mw/com/example/com-api-example/BUILD
```diff
@@ -17,7 +17,7 @@ rust_library(
     name = "com-api-example-lib",
     srcs = glob(["src/**/*.rs"]),
     crate_name = "com_api_example",
```
limdor, line 19: > Edition 2021 seems to be the agreed upon edition in S-CORE for now.

Please update the commit message to link where S-CORE says that Rust 2021 should be used.

score/mw/com/example/com-api-example/BUILD
```diff
@@ -17,7 +17,7 @@ rust_library(
...
-    edition = "2024",
+    edition = "2021",
```
limdor, line 20: Why are we specifying rust edition per target? This is like in C++ we would specify the C++ version per target. If edition is not specified, it defaults to the one used by the toolchain.
If there is no good reason to specify the edition in every single target, I would like we specify it only in the toolchain.

LittleHuba: Fair point. @bharatGoswami8 if you do not disagree, I'll do that change.

LittleHuba:
This PR is blocked by https://github.com/eclipse-score/score_rust_policies/issues/13
A downgrade to edition 2021 is impossible.
Upgrading to 2024 will lead to incompatibility with rustfmt.

score/mw/com/example/com-api-example/BUILD
bharatGoswami8: Agreed as we are going to use default so we can remove this completely - https://github.com/eclipse-score/toolchains_rust/blob/main/toolchains/x86_64-unknown-linux-gnu/BUILD.bazel#L24

bharatGoswami8:
> This PR is blocked by [eclipse-score/score_rust_policies#13](https://github.com/eclipse-score/score_rust_policies/issues/13) A downgrade to edition 2021 is impossible. Upgrading to 2024 will lead to incompatibility with rustfmt.

Created a ticket to upgrade the Rust Edition 2024: https://github.com/eclipse-score/toolchains_rust/issues/42

hoe-jo (approved):
Exceptionally well done

score/mw/com/impl/rust/com-api/com-api-runtime-lola/BUILD
```diff
@@ -21,6 +21,7 @@ rust_library(
         "producer.rs",
         "runtime.rs",
     ],
+    edition = "2021",
```
bharatGoswami8, line 24: I think this one we don't need as default edition we are considering from tool chain.

PR #1038, opened 2026-09-02
feat: add public config schema collection alias

AEK2HI:
Expose the _config.schema.json files for convenient access to the SOME/IP config schema file.

### DIFF
--- a/config/BUILD
+++ b/config/BUILD
@@ -1,0 +1,16 @@
+filegroup(
+    name = "config_schemas_files",
+    srcs = [
+        "//score/mw/com/gateway/gateway_application/configuration:mw_com_gateway_config_schema",
+        "//score/mw/com/gateway/transport_layer/sample/configuration:mw_com_gateway_sample_transport_config_schema",
+        "//score/mw/com/impl/configuration:mw_com_config_schema",
+        "//score/mw/com/impl/tracing/configuration:comtrace_filter_config_schema",
+    ],
+    visibility = ["//visibility:private"],
+)
+
+alias(
+    name = "config_schemas",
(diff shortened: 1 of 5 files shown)


config/BUILD
```diff
@@ -0,0 +1,16 @@
+filegroup(
+    name = "config_schemas_files",
+    srcs = [
+        "//score/mw/com/gateway/gateway_application/configuration:mw_com_gateway_config_schema",
+        "//score/mw/com/gateway/transport_layer/sample/configuration:mw_com_gateway_sample_transport_config_schema",
+        "//score/mw/com/impl/configuration:mw_com_config_schema",
```
limdor, line 6: Config schema is already exposed in https://github.com/eclipse-score/communication/blob/main/score/mw/com/BUILD#L125
I do not think we should expose a big bundle but just individually. Not everyonen will need the gateway config schema for example.
About location, right now we expose the target in `score/mw/com/BUILD`. A top level config package as you did does not make sense, it mixes multiple elements together (message passing, mw::com, etc).
For the gateway, we could expose them in `score/mw/com/gateway/BUILD`, and for tracing we could just have it in `score/mw/com/BUILD`.

I'm not really comfortable discussing this on the fly within the PR. Can you create a ticket explaining what you need and the motivation? Then we can have the discussion there involving also the other maintainers and once we have an agreement you can update the PR with the agreed changes.

limdor (changes requested):

AEK2HI:
@limdor We re-evaluated the topic and came to the conclusion that we do not require this extension to the readme. Hence, we close this PR

PR #1039, opened 2026-09-02
Make gateway transport layer messages own their data

LittleHuba:
Gateway messages must own their data. Only this way they will be able to reliably provide their contained information to upper layers until they finished the action.

ProvideServiceRequest used EventInfo to represent elements. This type is a view-style type.
The underlying buffer containing the actual data was the receive socket. But this buffer is not ensured to stay stable for the full lifetime of ProvideServiceRequest.
It may be overwritten whenever a new message is received.

Introduces a new type GatewayEventInfo which is a data-owning copy of EventInfo.


LittleHuba:
@NemanjaTrifunovicRTRK please verify that this fixes your problems.

NemanjaTrifunovicRTRK:
> @NemanjaTrifunovicRTRK please verify that this fixes your problems.

@LittleHuba, sorry for waiting.

I tested your solution for my case, it is working correctly.
Your solution can be used.

crimson11 (approved):

PR #1040, opened 2026-09-02
mw/com/dependability: Split requirements according to the high level architechture

gdadunashvili:


score/mw/com/impl/bindings/lola/proxy_event_common_test.cpp
```diff
@@ -182,7 +182,8 @@ TYPED_TEST(LolaProxyEventCommonFixture, CallingSetReceiveHandlerRegistersEventNo
 
 TYPED_TEST(LolaProxyEventCommonFixture, SubscriptionFailsWhenProviderRejectsSubscription)
 {
-    this->RecordProperty("Verifies", "SCR-21269964, SCR-17292398, SCR-14033248");
+    this->RecordProperty("Verifies", "SCR-21269964, SCR-14033248");
+    this->RecordProperty("lobster-tracing", "Communication.ProxyFieldSubscribe");
     this->RecordProperty("lobster-tracing", "Communication.ProxyEventSubscribe");
```
LittleHuba: Did you verify that having this property written twice is not an override but an extension?
E.g. do you see both traces in the report?

gdadunashvili: I merged this (and other similar) tags

LittleHuba (approved):

PR #1050, opened 2026-09-05
Rust::com Rust target edition removes and rustfmt fix

bharatGoswami8:
* Use the Score Rust toolchain's default edition instead of specifying
the edition on each Rust Bazel target.

### DIFF
--- a/score/mw/com/example/com-api-example/BUILD
+++ b/score/mw/com/example/com-api-example/BUILD
@@ -17,7 +17,6 @@ rust_library(
     name = "com-api-example-lib",
     srcs = glob(["src/**/*.rs"]),
     crate_name = "com_api_example",
-    edition = "2024",
     features = ["link_std_cpp_lib"],
     deps = [
         "//score/mw/com/example/com-api-example/com-api-gen",
@@ -35,7 +34,6 @@ rust_binary(
         "etc/logging.json",
         "etc/mw_com_config.json",
     ],
-    edition = "2024",
     env = {
         "MW_LOG_CONFIG_FILE": "score/mw/com/example/com-api-example/etc/logging.json",
     },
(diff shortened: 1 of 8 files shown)


limdor (changes requested):
In e30a1beff429845da8fb6aa9d5883fa24f4462a7 the commit message should be extended that explains what you did for reformatting. In this way the reviewer can rerun the command and see if there are any issues.
Also I'm a bit skeptical of formatting without enforcing it. Can you explain why you are reformatting in this PR? It would make more sense to do it in the PR that also introduce the enforcement otherwise it is a matter of time that it is wrongly formatted again
https://github.com/eclipse-score/communication/pull/1037/changes/ab898e2dc7737010beb7ce27424bffd9ab7c541c

About 3f0830a39a00147ec0c96cd166a71ff261d6ba9a you should explain in the commit message how you are fixing not just saying that it fixes an issue. It is ok to mention that this is an issue found by clippy but you as well as the reviewer (communicated via commit message) should understand why the change is done and how this is better than before.

LittleHuba (dismissed):

bharatGoswami8:
> In [e30a1be](https://github.com/eclipse-score/communication/commit/e30a1beff429845da8fb6aa9d5883fa24f4462a7) the commit message should be extended that explains what you did for reformatting. In this way the reviewer can rerun the command and see if there are any issues. Also I'm a bit skeptical of formatting without enforcing it. Can you explain why you are reformatting in this PR? It would make more sense to do it in the PR that also introduce the enforcement otherwise it is a matter of time that it is wrongly formatted again [ab898e2](https://github.com/eclipse-score/communication/commit/ab898e2dc7737010beb7ce27424bffd9ab7c541c)
> 
> About [3f0830a](https://github.com/eclipse-score/communication/commit/3f0830a39a00147ec0c96cd166a71ff261d6ba9a) you should explain in the commit message how you are fixing not just saying that it fixes an issue. It is ok to mention that this is an issue found by clippy but you as well as the reviewer (communicated via commit message) should understand why the change is done and how this is better than before.

The rustfmt fix was with configuration setup changes only but I agree that including it in the configuration PR would be more aligned. 

I have also updated the commit message for the clippy fix.

LittleHuba (approved):

limdor (approved):

PR #1052, opened 2026-09-07
Fix proxy method extract rvalue

sahithi-nukala:


Tejveerpratap2803 (approved):

score/mw/com/test/common_test_resources/proxy_container.h
```diff
@@ -39,7 +39,7 @@ class ProxyContainer
         return *proxy_;
     }
 
-    Proxy&& Extract()
+    Proxy&& Extract() &&
```
bemerybmw, line 46: Please add a comment explaining why this should only be called on a temporary.

bemerybmw (changes requested):

score/mw/com/test/common_test_resources/proxy_container.h
sahithi-nukala: done

bemerybmw (approved):

PR #1054, opened 2026-09-07
quality/api_surface: Report undocumented symbols without location

### LINKED ISSUE #1053: Bug: api_surface --check-docs crashes with KeyError: 'file' when undocumented symbols exist
### Description

`api_surface_test(check_docs = True)` crashes with `KeyError: 'file'` as soon as the target has at least one undocumented public symbol, instead of printing the "missing \api documentation" report.

`_format_undocumented_symbols()` (`quality/api_surface/diff_api.py`, line 58) reads `symbol['file']` / `symbol['line']`, but lock entries only carry `name`, `qualified_name`, `kind`, `signature` (see `to_lock_entry()` in `extract_api.py`).

`//score/mw/com:api_surface_docs_test` is tagged `manual`, so this is not caught in CI. The committed lock file currently has 106 undocumented symbols, so the test crashes deterministically.

### Error Occurrence  Rate

Reproducible

### How to reproduce

```
bazel test //score/mw/com:api_surface_docs_test --test_output=errors
```

```
  File ".../quality/api_surface/diff_api.py", line 75, in compare
    undocumented = _format_undocumented_symbols(current_data)
  File ".../quality/api_surface/diff_api.py", line 58, in _format_undocumented_symbols
    lines.append(f"    at {symbol['file']}:{symbol['line']}")
                           ~~~~~~^^^^^^^^
KeyError: 'file'
```

### Supporting Information

Reproduced on `main` @ 8a9a452 (Ubuntu 24.04, Bazel 8.7.0, hermetic toolchain). The affected code in `diff_api.py` is unchanged since v0.3.0. `//score/mw/com:api_surface_test` (`check_docs = False`) passes on the same tree.

### Classification

minor

### Affected Version

v0.3.0

### Category

- [ ] Safety Related
- [ ] Security Related

ThePo1es:
## Summary

`api_surface_test(check_docs = True)` crashes with `KeyError: 'file'` whenever the target has at least one undocumented public symbol, so the docs report is never printed.

`_format_undocumented_symbols()` reads `symbol['file']` / `symbol['line']`, but since 31aadcc4a the extractor deliberately strips both from every lock entry, including `undocumented_symbols`. The shell version of the diff tool already had this read; 36e83cc11 carried it over, and the non-empty path was never covered by a test.

## Change

- `diff_api.py`: print `at <file>:<line>` only when both fields are present. Otherwise the report lists the qualified name, which is enough to find the declaration.
- `tests/test_diff_api.py`: regression tests for the non-empty `undocumented_symbols` path, with and without location.

I did not make the extractor emit `file`/`line` again: `.update` copies the generated JSON verbatim into the committed lock file, so line numbers would churn it on every unrelated edit (the reason they were removed in 31aadcc4a). Happy to add a separate diagnostics output instead if you prefer locations in the report.

## Verification

- `bazel test //quality/api_surface/tests:test_diff_api`: fails with the `KeyError` before the fix, passes after.
- `//score/mw/com:api_surface_docs_test` now fails properly, listing the 106 undocumented symbols, instead of crashing. It is tagged `manual`, so CI is unaffected.

Closes #1053

### DIFF
--- a/quality/api_surface/diff_api.py
+++ b/quality/api_surface/diff_api.py
@@ -55,10 +55,11 @@ def _format_undocumented_symbols(data: dict) -> str:
     ]
     for symbol in undocumented:
         lines.append(f"  {symbol['qualified_name']} ({symbol['kind']})")
-        # Location is diagnostic only and not part of the lock entry format,
-        # so it may be absent.
-        if "file" in symbol:
-            lines.append(f"    at {symbol['file']}:{symbol.get('line', '?')}")
+        # Lock entries only carry the API shape (see to_lock_entry in
+        # extract_api.py), so location is normally absent. Print it only if a
+        # producer supplies both fields.
+        if "file" in symbol and "line" in symbol:
+            lines.append(f"    at {symbol['file']}:{symbol['line']}")
     lines.extend(
         [
             "",
--- a/quality/api_surface/diff_api.py
+++ b/quality/api_surface/diff_api.py
@@ -55,7 +55,10 @@ def _format_undocumented_symbols(data: dict) -> str:
     ]
     for symbol in undocumented:
         lines.append(f"  {symbol['qualified_name']} ({symbol['kind']})")
-        lines.append(f"    at {symbol['file']}:{symbol['line']}")
+        # Location is diagnostic only and not part of the lock entry format,
+        # so it may be absent.
+        if "file" in symbol:
+            lines.append(f"    at {symbol['file']}:{symbol.get('line', '?')}")
     lines.extend(
         [
             "",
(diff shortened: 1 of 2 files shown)


quality/api_surface/diff_api.py
```diff
@@ -55,7 +55,11 @@ def _format_undocumented_symbols(data: dict) -> str:
     ]
     for symbol in undocumented:
         lines.append(f"  {symbol['qualified_name']} ({symbol['kind']})")
-        lines.append(f"    at {symbol['file']}:{symbol['line']}")
+        # Lock entries only carry the API shape (see to_lock_entry in
+        # extract_api.py), so location is normally absent. Print it only if a
+        # producer supplies both fields.
+        if "file" in symbol and "line" in symbol:
+            lines.append(f"    at {symbol['file']}:{symbol['line']}")
```
LittleHuba, line 62: Can you please already print the file if it is present?
This is informational for users. But we still want to give them a hand to find the correct symbol to document.
So if we know what file it is in, why should we exclude that information?

PR #1056, opened 2026-09-08
tutorial: Fix line numbers in literalincludes

crimson11:
Later changes of the source-code hadn't been reflected in README.rst, thus tutorial pointing to wrong code locations. Fixed line numbers.

### DIFF
--- a/score/mw/com/doc/tutorial/chapter_1/README.rst
+++ b/score/mw/com/doc/tutorial/chapter_1/README.rst
@@ -185,7 +185,7 @@ The 1st important step happens in line 30:
 
 .. literalinclude:: provider.cpp
    :language: cpp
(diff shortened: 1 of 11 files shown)


score/mw/com/doc/tutorial/chapter_1/README.rst
```diff
@@ -185,7 +185,7 @@ The 1st important step happens in line 30:
 
 .. literalinclude:: provider.cpp
    :language: cpp
-   :lines: 30-30
```
bemerybmw, line 188: Would it be possible to somehow create tags which can be linked to? I guess the code will change in the future again so we will never keep these in sync

bemerybmw (approved):

PR #1058, opened 2026-09-08
Refactor OnProxyMethodUnsubscribe to remove code duplication

Tejveerpratap2803:
Refactor OnProxyMethodUnsubscribe to remove code duplication

- Merged OnProxyMethodUnsubscribe and OnProxyMethodUnsubscribeFinished into a single OnProxyMethodUnsubscribe method.
- Returns bool and assert/ignore decided at call site.
- Added doc comment and call site comments.
- Renamed test fixtures/cases to match.

### DIFF
--- a/score/mw/com/impl/bindings/lola/skeleton.cpp
+++ b/score/mw/com/impl/bindings/lola/skeleton.cpp
@@ -719,8 +719,9 @@ Result<void> Skeleton::OnServiceMethodsUnsubscribed(const ProxyInstanceIdentifie
     for (auto& [method_id, skeleton_method_ref] : skeleton_methods_)
     {
         const ProxyMethodInstanceIdentifier proxy_method_instance_identifier{proxy_instance_identifier, method_id};
-        // Loops over all skeleton methods for the departing proxy, not just the ones it actually subscribed
-        // to, so a missing entry is expected here and must not be asserted on.
+        // We are looping over all SkeletonMethods, not just the ones corresponding to the ProxyMethod which is
+        // unsubscribing. Therefore, it's expected that some of the calls to OnProxyMethodUnsubscribe will not
+        // unregister anything so we don't assert on the return.
         std::ignore = skeleton_method_ref.get().OnProxyMethodUnsubscribe(proxy_method_instance_identifier);
     }
 
@@ -787,7 +788,7 @@ void Skeleton::UnsubscribeMethods(const std::vector<UniqueMethodIdentifier>& met
     {
         auto& skeleton_method = skeleton_methods_.at(method_id);
         const ProxyMethodInstanceIdentifier proxy_method_instance_identifier{proxy_instance_identifier, method_id};
-        // This rolls back a just-inserted registration guard, where the entry is guaranteed to exist.
+        // A precondition of this function is that all provided method_ids were successfully registered.
         const bool element_erased = skeleton_method.get().OnProxyMethodUnsubscribe(proxy_method_instance_identifier);
         SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD(element_erased);
     }
--- a/score/mw/com/impl/bindings/lola/skeleton.cpp
+++ b/score/mw/com/impl/bindings/lola/skeleton.cpp
@@ -719,7 +719,9 @@ Result<void> Skeleton::OnServiceMethodsUnsubscribed(const ProxyInstanceIdentifie
     for (auto& [method_id, skeleton_method_ref] : skeleton_methods_)
     {
         const ProxyMethodInstanceIdentifier proxy_method_instance_identifier{proxy_instance_identifier, method_id};
-        skeleton_method_ref.get().OnProxyMethodUnsubscribeFinished(proxy_method_instance_identifier);
+        // Loops over all skeleton methods for the departing proxy, not just the ones it actually subscribed
+        // to, so a missing entry is expected here and must not be asserted on.
+        std::ignore = skeleton_method_ref.get().OnProxyMethodUnsubscribe(proxy_method_instance_identifier);
     }
 
     return {};
@@ -785,7 +787,9 @@ void Skeleton::UnsubscribeMethods(const std::vector<UniqueMethodIdentifier>& met
     {
         auto& skeleton_method = skeleton_methods_.at(method_id);
         const ProxyMethodInstanceIdentifier proxy_method_instance_identifier{proxy_instance_identifier, method_id};
-        skeleton_method.get().OnProxyMethodUnsubscribe(proxy_method_instance_identifier);
+        // This rolls back a just-inserted registration guard, where the entry is guaranteed to exist.
+        const bool element_erased = skeleton_method.get().OnProxyMethodUnsubscribe(proxy_method_instance_identifier);
+        SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD(element_erased);
     }
 }
 
--- a/score/mw/com/impl/bindings/lola/skeleton.h
+++ b/score/mw/com/impl/bindings/lola/skeleton.h
@@ -229,6 +229,9 @@ class Skeleton final : public SkeletonBinding
         const uid_t proxy_uid,
         const pid_t proxy_pid,
         const QualityType asil_level);
+    /// \brief Will unsubscribe each of the provided ProxyMethods from their corresponding SkeletonMethod.
+    ///
+    /// \pre All provided method_ids were successfully subscribed via SubscribeMethods.
     void UnsubscribeMethods(const std::vector<UniqueMethodIdentifier>& method_ids,
                             const ProxyInstanceIdentifier& proxy_instance_identifier);
 
--- a/score/mw/com/impl/bindings/lola/skeleton_method.cpp
+++ b/score/mw/com/impl/bindings/lola/skeleton_method.cpp
@@ -117,18 +117,11 @@ Result<void> SkeletonMethod::OnProxyMethodSubscribeFinished(
     return {};
 }
 
-void SkeletonMethod::OnProxyMethodUnsubscribe(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier)
+bool SkeletonMethod::OnProxyMethodUnsubscribe(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier)
 {
     const std::lock_guard lock{registration_guards_mutex_};
     const auto num_elements_erased = registration_guards_.erase(proxy_method_instance_identifier);
-    SCORE_LANGUAGE_FUTURECPP_PRECONDITION_PRD(num_elements_erased != 0U);
-}
-
-void SkeletonMethod::OnProxyMethodUnsubscribeFinished(
-    const ProxyMethodInstanceIdentifier proxy_method_instance_identifier)
-{
-    const std::lock_guard lock{registration_guards_mutex_};
-    std::ignore = registration_guards_.erase(proxy_method_instance_identifier);
+    return num_elements_erased != 0U;
 }
 
 void SkeletonMethod::UnregisterMethodCallHandlers()
--- a/score/mw/com/impl/bindings/lola/skeleton_method.h
+++ b/score/mw/com/impl/bindings/lola/skeleton_method.h
@@ -57,9 +57,11 @@ class SkeletonMethod : public SkeletonMethodBinding
         pid_t proxy_pid,
         const QualityType asil_level);
 
-    void OnProxyMethodUnsubscribe(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier);
-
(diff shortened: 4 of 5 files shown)


score/mw/com/impl/bindings/lola/skeleton_method.h
```diff
@@ -57,9 +57,11 @@ class SkeletonMethod : public SkeletonMethodBinding
         pid_t proxy_pid,
         const QualityType asil_level);
 
-    void OnProxyMethodUnsubscribe(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier);
-
-    void OnProxyMethodUnsubscribeFinished(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier);
+    // Returns true if a registration guard was erased, false if none existed. OnServiceMethodsUnsubscribed loops
```
bemerybmw: This should follow the doxygen style e.g. 
```
/// \brief 
///
/// More details...
///
/// \return Returns true if...
```

score/mw/com/impl/bindings/lola/skeleton_method.cpp
```diff
@@ -117,18 +117,11 @@ Result<void> SkeletonMethod::OnProxyMethodSubscribeFinished(
     return {};
 }
 
-void SkeletonMethod::OnProxyMethodUnsubscribe(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier)
+bool SkeletonMethod::OnProxyMethodUnsubscribe(const ProxyMethodInstanceIdentifier proxy_method_instance_identifier)
```
bemerybmw: I think it would be better to instead return the number of elements removed. This information might be useful in the future and is also more in line with the standard library. E.g. https://en.cppreference.com/cpp/container/unordered_map/erase

score/mw/com/impl/bindings/lola/skeleton_method.h
```diff
@@ -57,9 +57,11 @@ class SkeletonMethod : public SkeletonMethodBinding
...
+    // over all skeleton methods for a departing proxy, not just the ones it subscribed to, so a missing entry is
```
bemerybmw: The documentation for this function should just explain what the function is doing. We shouldn't copy and paste / reword the docs from the calling code.

E.g. Unregisters the method call handler corresponding to a given ProxyMethod which was registered in with MessagePassing in OnProxyMethodSubscribeFinished.

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -719,7 +719,9 @@ Result<void> Skeleton::OnServiceMethodsUnsubscribed(const ProxyInstanceIdentifie
     for (auto& [method_id, skeleton_method_ref] : skeleton_methods_)
     {
         const ProxyMethodInstanceIdentifier proxy_method_instance_identifier{proxy_instance_identifier, method_id};
-        skeleton_method_ref.get().OnProxyMethodUnsubscribeFinished(proxy_method_instance_identifier);
+        // Loops over all skeleton methods for the departing proxy, not just the ones it actually subscribed
```
bemerybmw: This is not correct. It's looping over all skeleton methods, not just the methods for the specified proxy. And it doesn't subscribe to a proxy but a proxy subscribes to a skeleton. Something like this would be clearer IMO:

"We are looping over all SkeletonMethods, not just the ones corresponding to the ProxyMethod which is unsubscribing. Therefore, it's expected that some of the calls to OnProxyMethodUnsubscribe will not unregister anything so we don't ignore the return.

bemerybmw: On second thought, shouldn't we assert that at least one method was successfully unsubscribed?

score/mw/com/impl/bindings/lola/skeleton.cpp
```diff
@@ -785,7 +787,9 @@ void Skeleton::UnsubscribeMethods(const std::vector<UniqueMethodIdentifier>& met
     {
         auto& skeleton_method = skeleton_methods_.at(method_id);
         const ProxyMethodInstanceIdentifier proxy_method_instance_identifier{proxy_instance_identifier, method_id};
-        skeleton_method.get().OnProxyMethodUnsubscribe(proxy_method_instance_identifier);
+        // This rolls back a just-inserted registration guard, where the entry is guaranteed to exist.
```
bemerybmw: Again, we shouldn't comment on where this function is called here. I would rather say something like:
"A precondition of this function is that all provided method_ids were succesfully registered."

You could even add this with a \pre tag in the header file in the function documetation.

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
```diff
@@ -432,17 +432,97 @@ TEST_F(SkeletonMethodOnProxyMethodUnsubscribedFixture, CallingWillUnregisterHand
...
 
     // When calling OnProxyMethodUnsubscribe with proxy_method_instance_identifier_ which was never subscribed
-    // Then the program terminates
-    SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED(
-        unit_->OnProxyMethodUnsubscribe(proxy_method_instance_identifier_));
+    // Then it reports that no element was erased
+    EXPECT_FALSE(unit_->OnProxyMethodUnsubscribe(proxy_method_instance_identifier_));
+}
+
+TEST_F(SkeletonMethodOnProxyMethodUnsubscribedFixture, CallingTwiceWithSameIdentifierIsANoOp)
```
bemerybmw: It's not a noop? It just returns false when called the second time?

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
```diff
@@ -432,17 +432,97 @@ TEST_F(SkeletonMethodOnProxyMethodUnsubscribedFixture, CallingWillUnregisterHand
...
+{
+    GivenASkeletonMethod().WithARegisteredCallback();
+
+    EXPECT_CALL(message_passing_mock_,
```
bemerybmw, line 454: Add an // Expecting comment

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
```diff
@@ -432,17 +432,97 @@ TEST_F(SkeletonMethodOnProxyMethodUnsubscribedFixture, CallingWillUnregisterHand
...
+{
+    // Two proxy instances from the SAME application (same application_id, different proxy_instance_counter)
+    const ProxyInstanceIdentifier proxy_instance_identifier_same_app{kDummyProxyInstanceCounter + 1U,
+                                                                     kDummyApplicationId};
+    const ProxyMethodInstanceIdentifier proxy_method_instance_identifier_same_app{proxy_instance_identifier_same_app,
+                                                                                  unique_method_identifier_};
+
+    GivenASkeletonMethod().WithARegisteredCallback();
+
+    EXPECT_CALL(message_passing_mock_, RegisterMethodCallHandler(kAsilLevel, proxy_method_instance_identifier_, _, _));
```
bemerybmw, line 495: Are these expect calls needed in this test? If they are, then add an // Expecting comment

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
```diff
@@ -432,17 +432,97 @@ TEST_F(SkeletonMethodOnProxyMethodUnsubscribedFixture, CallingWillUnregisterHand
...
+    EXPECT_CALL(message_passing_mock_,
+                RegisterMethodCallHandler(kAsilLevel, proxy_method_instance_identifier_same_app, _, _));
+
+    // The first proxy's handler is removed by OnProxyMethodUnsubscribe — must not fire on destruction
+    EXPECT_CALL(message_passing_mock_, UnregisterMethodCallHandler(kAsilLevel, proxy_method_instance_identifier_))
```
bemerybmw, line 500: These comments here about one being called on OnProxyMethodUnsubscribe and the other called on destruction are not enforced in the test. You should set a flag when they're called and check that they're called when you expect.

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/skeleton_method.h
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton_method.h
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
Tejveerpratap2803: Changed the name to CallingASecondTimeWithSameIdentifierReturnsFalse.

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
Tejveerpratap2803: Yes, they're needed. I have added the expecting comment.

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton.cpp
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton.h
```diff
@@ -229,6 +229,9 @@ class Skeleton final : public SkeletonBinding
         const uid_t proxy_uid,
         const pid_t proxy_pid,
         const QualityType asil_level);
+    /// \brief Rolls back the registration guards for a set of methods that were previously successfully subscribed.
```
bemerybmw: What does it mean to roll back a registration guard? What even is a registration guard. IMO, this comment is not really providing useful information to someone unfamiliar with the code, because they'd have to go and check the implementation to understand these things. I think something would like this would provide more information without mentioning too many implementation details: 

Will unsubscribe each of the provided ProxyMethods from their corresponding SkeletonMethod.

The precondition then conveys the info that all methods were succesfully subscribed.

score/mw/com/impl/bindings/lola/skeleton.h
```diff
@@ -229,6 +229,9 @@ class Skeleton final : public SkeletonBinding
...
+    ///
+    /// \pre All provided method_ids were successfully registered.
```
bemerybmw:
```suggestion
    /// \pre All provided method_ids were successfully subscribed via SubscribeMethods.
```

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
```diff
@@ -432,17 +432,116 @@ TEST_F(SkeletonMethodOnProxyMethodUnsubscribedFixture, CallingWillUnregisterHand
...
+
+    // Then the first proxy's handler has already been unregistered, and the second proxy's handler is still
+    // registered
+    EXPECT_TRUE(first_proxy_handler_unregistered);
+    EXPECT_FALSE(second_proxy_handler_unregistered);
+
+    // When destroying the SkeletonMethod
+    unit_.reset();
+
+    // Then the second proxy's handler is unregistered on destruction
```
bemerybmw: The fact that we have here 2 "then" sections hints that this test is doing too many things. The first Then section is testing that we only unregister the correct method when calling OnProxyMethodUnsubscribe. The second Then section is testing that methods are unregistered on destruction. These should be 2 separate tests.

bemerybmw (changes requested):

score/mw/com/impl/bindings/lola/skeleton.h
Tejveerpratap2803: Done, simplified the comment.

score/mw/com/impl/bindings/lola/skeleton.h
Tejveerpratap2803: Done

score/mw/com/impl/bindings/lola/skeleton_method_test.cpp
Tejveerpratap2803: Done

bemerybmw (approved):

PR #1063, opened 2026-09-09
Sw/main/someip scaffold

sergejwinterbmw:


pawelrutkaq:
One fast question: I see this is kind of "generic SOME/IP binding," using itransport as an abstraction for different possible transports if required. Where the serialization place is expected to happen, which shall know the `T` type. Currently, I dont see a place for choosing serilization and so on . Ie if i want to take it to internal someip stack and hook iTransport, i also need to hook serialization with settings.

score/mw/com/impl/bindings/someip/skeleton_event.cpp
```diff
@@ -0,0 +1,233 @@
...
+        return MakeUnexpected(ComErrc::kBindingFailure, "Send called with a nullptr sample");
+    }
+
+    auto allocated_slot_result = Allocate(std::move(guard));
+    if (!allocated_slot_result.has_value())
+    {
+        return MakeUnexpected(ComErrc::kSampleAllocationFailure, "Could not allocate slot");
+    }
+    auto allocated_slot = std::move(allocated_slot_result).value();
+    std::memcpy(allocated_slot.Get(), value_ptr, event_sample_size_info_.Size());
```
pawelrutkaq, line 65: Why do we memcopy into Allocate sample ? IN case of SOME/IP binding, the user Send(T data) is more optimal since You can just take data and serialize it directly into SHM. For a user doing first `Allocate` we need to do this intermediate heap buffer. But in both cases we shall it further to the code that does processing (ie serialization, maybe secoc) so it can do it most performantly, or ?

sergejwinterbmw: Hi Pawel - this is quite an early draft. Both of your remarks will be adressed once some other topics on serialization are finalized on our end.

pawelrutkaq: Is there any way to help get this progressing? 
Anyway It would be nice to just let me/us know where and how the serialization hookup is thought of. I guess a simple explanation would do for a moment to let me think about it ;)

Thanks

LittleHuba: There are several considerations for a good placement of this feature:
1. Serialization is quite a generic capability with protocol specific implementation
2. For serialization we need the type information (and bindings are type erased)
3. Binding choice is deployment information (we don't want to restrict this at compile time already through available serialization logic)
4. Type information for SOME/IP often comes from some central model description to safely share it between communication participants

So the best place we indentified is to put the general functionality into the frontend and provide serialization logic via dynamic loading. That keeps the implementation lean and allows bigger projects to provide the serialization information in a streamlined manner.

> Is there any way to help get this progressing?

Thanks for your offer. For the moment, I don't think there is anything you can do to help this progress faster. We are just now writing up the detailed design for this. Once it is ready, we will present it in the weekly and discuss it there. That would be the point in time where your feedback will be of most benefit.
While that happens, this PR here is providing the basic infrastructure to then start the implementation. In this PR there is nothing really controversial, since any implementation details of the binding will follow later.
When the design is nailed down, we are happy to get some more hands for the implementation.

sergejwinterbmw:
Superseded by #1135. This PR was based on `mf_type_erased_binding`, which was deleted when #897 merged; GitHub auto-closed this PR and will not reopen it because its base ref no longer exists. The work has been rebased onto `main` and continues in #1135.

PR #1073, opened 2026-09-13
memory/shared: Add OffsetRef<T>

crimson11:
Added now an OffsetRef<T> besides
OffsetPtr<T> to cover reference semantics
in shared-memory, which we also need in case
of score::mw::com/LoLa.


score/memory/shared/test_offset_ptr/BUILD
```diff
@@ -79,6 +79,9 @@ cc_library(
     srcs = ["bounds_check_memory_pool.cpp"],
     hdrs = ["bounds_check_memory_pool.h"],
     features = COMPILER_WARNING_FEATURES,
+    visibility = [
+        "//score/memory/shared/test_offset_ref:__pkg__",
+    ],
```
limdor, line 84: Before we add more tests, I would like we clarify our structure for test before it gets to hard to change. I would like to understand why we put them in a folder instead of following the _test pattern.
Basically now we have the follwing places with tests:
- File next to the cpp file
- These test_* folders
- The test/ folder with the dynamic_array_in_shm_test

score/memory/shared/test_offset_ref/BUILD
```diff
@@ -0,0 +1,52 @@
+# *******************************************************************************
+# Copyright (c) 2025 Contributors to the Eclipse Foundation
```
limdor, line 2:
```suggestion
# Copyright (c) 2026 Contributors to the Eclipse Foundation
```

score/memory/shared/BUILD
```diff
@@ -332,6 +332,27 @@ cc_library(
...
+        "offset_ref.cpp",
+    ],
+    hdrs = [
+        "offset_ref.h",
+    ],
+    features = COMPILER_WARNING_FEATURES,
+    tags = ["FFI"],
+    visibility = [
+        "//visibility:public",  # platform_only
+    ],
```
limdor, line 347:
```suggestion
    visibility = ["//visibility:public"],
```

score/memory/shared/test_offset_ref/BUILD
```diff
@@ -0,0 +1,52 @@
...
+
+cc_unit_test(
+    name = "offset_ref_test",
+    srcs = [
+        "offset_ref_test.cpp",
+    ],
+    features = COMPILER_WARNING_FEATURES,
+    visibility = [
+        "//score/memory:__pkg__",
+    ],
```
limdor, line 25:
```suggestion
```
This is not needed, we do not refer to this test from anywhere else.

score/memory/shared/test_offset_ref/BUILD
```diff
@@ -0,0 +1,52 @@
...
+# *******************************************************************************
+
+load("//quality/unit_testing:unit_testing.bzl", "cc_unit_test")
+load("//score/mw:common_features.bzl", "COMPILER_WARNING_FEATURES")
+
+cc_unit_test(
+    name = "offset_ref_test",
+    srcs = [
+        "offset_ref_test.cpp",
+    ],
```
limdor, line 21:
```suggestion
    srcs = ["offset_ref_test.cpp"],
```
It helps to keep it a bit more compact

score/memory/shared/test_offset_ref/BUILD
```diff
@@ -0,0 +1,52 @@
...
+    srcs = [
+        "offset_ref_bounds_check_test.cpp",
+    ],
+    features = COMPILER_WARNING_FEATURES + [
+        # This test catches exceptions instead of using gtest EXPECT_DEATH so we disable aborts_upon_exception.
+        "-aborts_upon_exception",
+    ],
+    visibility = [
+        "//score/memory:__pkg__",
+    ],
```
limdor, line 44:
```suggestion
```
Not needed

score/memory/shared/test_offset_ref/BUILD
```diff
@@ -0,0 +1,52 @@
...
+        "//score/memory/shared/fake:fake_memory_resources",
+        "@score_baselibs//score/mw/log:backend_stub_testutil",
+    ],
+)
+
+cc_unit_test(
+    name = "offset_ref_precondition_violation_test",
+    srcs = [
+        "offset_ref_bounds_check_test.cpp",
+    ],
```
limdor, line 37:
```suggestion
    srcs = ["offset_ref_bounds_check_test.cpp"],
```
It helps on readibility keeping it more compact

score/memory/shared/test_offset_ref/offset_ref_bounds_check_test.cpp
```diff
@@ -0,0 +1,69 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
limdor, line 2:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

score/memory/shared/test_offset_ref/offset_ref_test.cpp
```diff
@@ -0,0 +1,193 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
limdor, line 2:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

score/memory/shared/offset_ref.cpp
```diff
@@ -0,0 +1,13 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
limdor, line 2:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

score/memory/shared/offset_ref.h
```diff
@@ -0,0 +1,141 @@
+/********************************************************************************
+ * Copyright (c) 2025 Contributors to the Eclipse Foundation
```
limdor, line 2:
```suggestion
 * Copyright (c) 2026 Contributors to the Eclipse Foundation
```

score/memory/shared/BUILD
```diff
@@ -332,6 +332,27 @@ cc_library(
     ],
 )
 
+cc_library(
+    name = "offset_ref",
+    srcs = [
+        "offset_ref.cpp",
+    ],
```
limdor, line 339:
```suggestion
    srcs = ["offset_ref.cpp"],
```
It helps on readibility keeping it more compact

score/memory/shared/BUILD
```diff
@@ -332,6 +332,27 @@ cc_library(
...
+    hdrs = [
+        "offset_ref.h",
+    ],
```
limdor, line 342:
```suggestion
    hdrs = ["offset_ref.h"],
```
It helps on readibility keeping it more compact

score/memory/shared/test_offset_ref/BUILD
```diff
@@ -0,0 +1,52 @@
...
+    features = COMPILER_WARNING_FEATURES + [
+        # This test catches exceptions instead of using gtest EXPECT_DEATH so we disable aborts_upon_exception.
```
limdor, line 39:
```suggestion
        # This test uses SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED that relies on exceptions being thrown so we disable aborts_upon_exception.
```

score/memory/shared/offset_ref.h
```diff
@@ -0,0 +1,141 @@
...
+#include "score/memory/shared/offset_ptr.h"
+#include "score/memory/shared/pointer_arithmetic_util.h"
+
+#include <score/assert.hpp>
+
+#include <cstddef>
+
+namespace score::memory::shared
+{
+
```
limdor, line 26: We should also update:
- https://github.com/eclipse-score/communication/tree/main/score/memory/shared/design#offset-pointer
- https://github.com/eclipse-score/communication/blob/main/score/memory/shared/design/OffsetPtrDesign.md
- https://github.com/eclipse-score/communication/blob/main/score/memory/shared/design/offset_ptr_problems.md

Otherwise it feels strange that we talk about offset ptr and we do not talk about offset ref.

score/memory/shared/offset_ref.h
```diff
@@ -0,0 +1,141 @@
...
+/// \brief Custom implementation of an offset reference. It behaves similarly to a native C++ reference (T&) but, like
+/// OffsetPtr, is safe to use within shared memory that is mapped at different virtual addresses in different
+/// processes.
+///
+/// \details Internally, OffsetRef stores an offset (relative to its own address) to the referenced object, exactly like
+/// OffsetPtr does, and re-uses OffsetPtr's bounds-checking mechanism (detail_offset_ptr::AssertOffsetBoundsCheck()).
+/// In contrast to OffsetPtr, OffsetRef:
+/// - can only be constructed from a valid reference to T (never from a raw pointer and never "null").
```
limdor, line 34:
```suggestion
/// - can only be constructed from a valid reference to T (never from a raw pointer).
/// - it can never be "null".
```
I guess you mean in general, not from construction but also not afterwards. I think with a new sentence would read better. Does it say the same like your intention was?

score/memory/shared/test_offset_ptr/BUILD
limdor: This could be a proposal to have a bit more consistent:
- score/memory/shared/test/fancy_pointers/ (this would containe the common bounds_check_memory_pool)
- score/memory/shared/test/fancy_pointers/offset_ptr/
- score/memory/shared/test/fancy_pointers/offset_ref/
- score/memory/shared/test/dynamic_array_in_shm_test/ (as of today, no change)

score/memory/shared/offset_ptr.h
```diff
@@ -87,6 +87,29 @@ constexpr difference_type kNullPtrRepresentation = 1;
 
 bool IsBoundsCheckingEnabled() noexcept;
 
+/// \brief Performs the bounds-check (if enabled) for an offset-based address relationship, i.e. an address at
```
limdor, line 90: That the bounds check is performed only when enabled is something I'm not totally sure that this function is doing. Digging deeper I saw that `DoesOffsetPtrNotInSharedMemoryPassBoundsChecks` will still perform bounds checking of the stard and end of the pointed object in case that `offset_ptr_memory_bounds` has a value.

score/memory/shared/offset_ref.h
```diff
@@ -0,0 +1,141 @@
...
+/// - is neither copyable nor moveable (just like a native C++ reference cannot be re-bound after construction).
```
limdor, line 35:
```suggestion
/// - is neither copyable nor moveable.
```
I would propose to remove the part here about comparing with C++ reference and just mention it at the end.

score/memory/shared/offset_ref.h
```diff
@@ -0,0 +1,141 @@
...
+///
+/// Because of these two restrictions, OffsetRef does not need to reserve/handle a "null" offset representation, nor
+/// does it need to store bounds information for the "copied out of shared memory" scenario that OffsetPtr has to
+/// handle (that information is only needed to preserve correct bounds-checking when an OffsetPtr is copied to a
+/// location outside of the shared-memory region it originally referred to; since OffsetRef can never be copied/moved,
+/// this scenario cannot occur). Consequently, OffsetRef only stores the offset itself (i.e. it has the same size as a
+/// single pointer/std::ptrdiff_t), unlike OffsetPtr which additionally stores a MemoryRegionBounds.
```
limdor, line 42:
```suggestion
/// single pointer/std::ptrdiff_t), unlike OffsetPtr which additionally stores a MemoryRegionBounds.

/// The main difference with a native C++ reference is that this is an object and occupies memory. A native C++ reference it is just a name alias for an object that might end up occupying memory in certain scenarios due to how it is implemented.
```

score/memory/shared/test_offset_ptr/BUILD
limdor: After a second look, I'm not convience about the fancy_pointer folder for offset_ref, because it is not a fancy pointer and will not have the pointer traits. Maybe just like this:

- score/memory/shared/test/ (this would contain the common bounds_check_memory_pool)
- score/memory/shared/test/offset_ptr/
- score/memory/shared/test/offset_ref/
- score/memory/shared/test/dynamic_array_in_shm_test/ (as of today, no change)

score/memory/shared/offset_ref.h
```diff
@@ -0,0 +1,141 @@
...
+    {
+        get() = value;
+        return *this;
+    }
+
+    /// \brief Returns a reference to the referenced object (with bounds-checking).
+    T& get() const
+    {
+        // OffsetRef never needs to capture bounds for the "copied out of shared memory" scenario (since it is never
+        // copied), so we always pass a default constructed (empty) MemoryRegionBounds here.
```
limdor, line 98: This seems to imply that an OffsetRef will never be not in shared memory, however someone could have created an OffsetRef not in shared memory in the first place. What should happen in such scenario? It would make sense to mention something here.
I see the following situations:
- OffsetRef in shared memory referring to an object outside shared memory
- OffsetRef in shared memory referring to an object inside shared memory
- OffsetRef outside shared memory referring to an object outside shared memory
- OffsetRef outside shared memory referring to an object inside shared memory

And of course the situation where the OffsetRef or the pointed object might be half in half out. Maybe for the pointed object we want to check.

limdor (commented):
I reviewed everything except the tests, but I believe the only thing that it could be missing in the tests are missing test cases. Because in some of the comments I talk about different scenarios, I would just submit my comments and afterwards when I do the full review I would also look at the tests.

score/memory/shared/offset_ptr.h
crimson11: We can "globally" disable bounds-checking! This is referred to here. (https://github.com/eclipse-score/communication/blob/main/score/memory/shared/offset_ptr.cpp#L38)

score/memory/shared/offset_ref.h
crimson11: Removed it - it was anyhow "redundant".

score/memory/shared/offset_ref.h
crimson11: Nope: The thing is, the OffsetPtr only initializes its bounds-member, when it is copied out from shared-memory! Only then it is needed to store the bounds "locally" within the OffsetPtr!

All the use cases you describe above are also supported with OffsetRef! It is cared for by re-using the existent functionality for bounds-checking.

score/memory/shared/design/OffsetPtrDesign.md
```diff
@@ -20,15 +20,27 @@ This offset is the same in all processes, thus, a valid pointer can be calculate
...
 
 The available public member methods are taken over from the `boost::interprocess::offset_ptr` implementation.
 In order to reuse this pointer also with stl-based containers it shall implement the requirements stated by
 [std::pointer_traits](https://en.cppreference.com/w/cpp/memory/pointer_traits).
 
-### Bounds Checking OffsetPtr
+This design document mainly describes the `OffsetPtr`. But in addition to the `OffsetPtr`, we also have an "offset
+reference" type called `OffsetRef`. It is similar to a C++ reference in the sense, that it:
+
+- always references a valid T, because it can only be constructed from a T&.
```
limdor:
```suggestion
- always references a valid T, because it can only be constructed from a T& (still the user needs to guarantee proper lifetime).
```

score/memory/shared/test/test_offset_ref/BUILD
```diff
@@ -0,0 +1,43 @@
...
+load("//score/mw:common_features.bzl", "COMPILER_WARNING_FEATURES")
+
+cc_unit_test(
+    name = "offset_ref_test",
+    srcs = ["offset_ref_test.cpp"],
+    features = COMPILER_WARNING_FEATURES,
+    deps = [
+        "//score/memory/shared:offset_ref",
+        "//score/memory/shared/fake:fake_memory_resources",
+        "@score_baselibs//score/mw/log:backend_stub_testutil",
```
limdor:
```suggestion
```
After https://github.com/eclipse-score/communication/pull/1098 it is no longer needed to provide the stub backend.

score/memory/shared/test/test_offset_ref/BUILD
```diff
@@ -0,0 +1,43 @@
...
+        # This test uses SCORE_LANGUAGE_FUTURECPP_EXPECT_CONTRACT_VIOLATED that relies on exceptions being thrown so we
+        # disable aborts_upon_exception.
+        "-aborts_upon_exception",
+    ],
+    deps = [
+        "//score/memory/shared:offset_ref",
+        "//score/memory/shared/fake:fake_memory_resources",
+        "//score/memory/shared/test:bounds_check_memory_pool",
+        "@score_baselibs//score/language/futurecpp:futurecpp_test_support",
+        "@score_baselibs//score/mw/log:backend_stub_testutil",
```
limdor:
```suggestion
```
After https://github.com/eclipse-score/communication/pull/1098 it is no longer needed to provide the stub backend.

limdor (approved):

PR #1089, opened 2026-09-15
lola/messaging: declare missing safe_math dependency

mihajlo-k:
`message_passing_service_instance.cpp`includes `score/language/safecpp/safe_math/safe_math.h`, but the target did not declare the dependency. It resolved only transitively through `score/json:json_writer_impl` -> `score/json/internal/writer/json_serialize`.

Any change to the json writer's dependency graph therefore breaks this target with "safe_math.h: No such file or directory". All other targets in the repo that include this header already declare the dependency.


limdor:
Because it is used only in cpp, please add it to implementation deps

limdor (approved):

PR #1094, opened 2026-09-16
Depa ipc tracing feature flag usage

devendrapat:
Keeping the changes in this PR (https://github.com/eclipse-score/communication/pull/1022/changes) intact.
This PR also updates the version of score_baselibs for usage of required_typed_memory flag.


crimson11 (approved):

MODULE.bazel
```diff
@@ -30,6 +30,7 @@ bazel_dep(name = "score_qnx_unit_tests", version = "0.2.0")
 bazel_dep(name = "score_baselibs", version = "0.2.12")
 single_version_override(
     module_name = "score_baselibs",
+    commit = "4abf22407106ec2461f5f80a619cd15b5d294238",
```
limdor, line 33: This would require all our customer to pin this exact version. Is a new release of baselibs planned? What is the ticket to track that? In general we only want to use release versions of baselibs.

limdor (changes requested):

MODULE.bazel
limdor: I created https://github.com/eclipse-score/communication/pull/1102 that bumps baselibs to the new released version. The new version contain the changes you need, you should just remove this check once the mentioned PR is merged.

limdor: Currently we cannot bump because of https://github.com/eclipse-score/baselibs/issues/527#issuecomment-5730181076

PR #1099, opened 2026-09-17
AoU cleanup

sahithi-nukala:


LittleHuba (changes requested):
This looks wrong. At least some of those AoUs are still relevant to my knowledge.
@crimson11 @bemerybmw please say so if this knowledge is outdated.

I think the correct action is to extend the FTAs and FMEAs respectively to be more thorough.

score/mw/com/dependability/safety_analysis/aou.trlc
```diff
@@ -149,15 +149,6 @@ ScoreReq.AoU OneProducerOnlyOneAllocateePtr {
     mitigates = "TooFewMemoryAllocated, TooMuchMemoryAllocated, SendingEventSendsToWrongConsumer"
 }
 
-/* broken_link_c/issue/5835324 */
```
crimson11, line 152: What was your intention to delete this AoU? I guess - as of now it is needed/referenced by an FMEA ("Too few memory allocated"). If we would delete this AoU (see details below), then we would also need to delete the AoU above "OneProducerOnlyOneAllocateePtr", because they are both strongly related.

TBH: I'm also not really happy with this AoU. For several reasons:
1. Why shouldn't we allow this? The user is responsible for configuring the slot-count to his needs! If he configures it sufficient, there is no reason, why he can't hold more than one `SampleAllocateePtr` in parallel
2. We even at some point had a config-value (removed now) to explicitly allow the user to configure the number of `SampleAllocateePtr`, he wants to hold in parallel (to support him with some supervision).
3. We can discuss, whether this "supervision" in (2) makes sense or not ... but having this AoU, while we in code NOT checking this "restriction" (if we really want it) in code ALTHOUGH we easily could, makes no sense to me!

### Summary
There are imho two options:
A. We keep this AoU and the AoU above "OneProducerOnlyOneAllocateePtr" and also the FTA referencing this AoU as an "preventive measure". This is the "lazy" approach - keeps everything as it is now and it is "consistent".
B. We clean this up: We remove both AoUs and adapt the FMEA/FTA: Getting an error-result in SkeletonEvent::Allocate() or Send(SampleType&) is NOT an error in the case the user himself is "exhausting" the slots on provider side. Then I would rather re-phrase the existing "CorrectlyConfiguredMaximumNumberOfMaximumElementsPerSubscriber" AoU into "CorrectlyConfigurednumberOfSampleSlots", which simply states, that "numberOfSampleSlots" needs to be dimensioned correctly according to consumers AND producer need!

score/mw/com/dependability/safety_analysis/aou.trlc
crimson11, line 135: Should be adapted imho! The `Allocate()` call nowadays returns a `Result` -> should be checked for an error instead!

score/mw/com/dependability/safety_analysis/aou.trlc
```diff
@@ -284,29 +257,12 @@ ScoreReq.AoU NoGuaranteeOnExecutionTime {
     mitigates = "AnyFunctionBlocksLongerThanExpected"
 }
 
-/* broken_link_c/issue/7091801 */
-ScoreReq.AoU UsageOfConfigurationOversubscription {
```
crimson11, line 288: Why you deleted this? What was your intention?

score/mw/com/dependability/safety_analysis/aou.trlc
```diff
@@ -284,29 +257,12 @@ ScoreReq.AoU NoGuaranteeOnExecutionTime {
...
 /* broken_link_c/issue/21206172 */
 ScoreReq.AoU SameCompilerSettingsForProviderAndConsumerSide {
     description = "All compiler settings having influence on the binary representation of data exchanged via {{mw::com}}/{{LoLa}} (event, field, service-method payloads) have to be identical for compilation of code containing {{mw::com}} proxies and skeletons, which communicate."
     safety = ScoreReq.Asil.B
     version = 1
-    mitigates = ""
-}
-
-/* broken_link_c/issue/21209491 */
-ScoreReq.AoU EventOrFieldReceptionViaGenericProxyNeedsSpecificCare {
```
crimson11, line 305: Why you removed this? I guess it is still valid ... only the API `GetSampleSize` is deprecated and is now superseeded by `GetDataTypeSizeInfo`

score/mw/com/dependability/safety_analysis/aou.trlc
```diff
@@ -335,11 +291,3 @@ ScoreReq.AoU MethodInArgPtrMatches {
     version = 1
     mitigates = "Communication.CallBlocksOnCaller, Communication.CallBlocksOnCallee, Communication.CallBlocksOnUserHandler"
 }
-
-/* broken_link_c/issue/79416086 */
-ScoreReq.AoU NoGuaranteeOnSubscriptionStateCorrectness {
```
crimson11, line 340: Why deleted? It has been explicitly newly introduced! If it is currently not used/referenced ... we rather have to update our FTAs and reference it ;)

crimson11 (changes requested):

PR #1102, opened 2026-09-18
Upgrade baselib version

limdor:
This will be required for follow up work.
This change forces us to update also docs as code
and score rust.


limdor:
Blocked by https://github.com/eclipse-score/baselibs/issues/527#issuecomment-5730181076

